PR #12, opened 2015-12-24
Exclude parent module from maven deploy.

itohro:


junit-trait-runner/pom.xml
```diff
@@ -29,6 +29,10 @@
         <maven>3.0.5</maven>
     </prerequisites>
```
motlin: I think we should continue to deploy the junit-trait-runner jar. It can be useful in the same way as the testutils jar.

pom.xml
```diff
@@ -103,6 +103,7 @@
         <sonar.dynamicAnalysis>reuseReports</sonar.dynamicAnalysis>
         <sonar.core.codeCoveragePlugin>clover</sonar.core.codeCoveragePlugin>
         <sonar.clover.version>${clover.version}</sonar.clover.version>
```
motlin: Does this change prevent the parent pom from being deployed? It's fine, we don't need to publish the parent pom for now. It looks like there's extra indentation.

junit-trait-runner/pom.xml
```diff
@@ -29,6 +29,10 @@
...
 
+    <properties>
+        <maven.deploy.skip>true</maven.deploy.skip>
+    </properties>
+
     <dependencies>
```
motlin: I think we should continue to deploy the junit-trait-runner jar. It can be useful in the same way as the testutils jar.

pom.xml
```diff
@@ -103,6 +103,7 @@
...
+		<maven.deploy.skip>true</maven.deploy.skip>
     </properties>
```
motlin: Does this change prevent the parent pom from being deployed? It's fine, we don't need to publish the parent pom for now. It looks like there's extra indentation.

junit-trait-runner/pom.xml
itohro: Ok, let me keep junit-trait-runner deployed then.

pom.xml
itohro: Yes, I believe so. Let me fix the indentation.

PR #17, opened 2015-12-29
Changes for 7.0.1 bug fix release.

itohro:
- Update the version number from 7.0.0 to 7.0.1.
- Bug fix in UnifiedSet$ChainedBucket.removeLongChain() method.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
@@ -1719,6 +1719,12 @@ public class UnifiedSet<T>
                         bucket.three = null;
                         return;
                     default:
+                        if (bucket.three instanceof ChainedBucket)
+                        {
+                            i -= 3;
+                            oldBucket = bucket;
+                            continue;
+                        }
                         throw new AssertionError();
                 }
             }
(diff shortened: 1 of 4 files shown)


motlin:
I think the bugfix is already merged into master as part of PR https://github.com/eclipse/eclipse-collections/pull/16. I accidentally closed and reopened that one before merging it. Sorry for the confusion. Try rebasing onto the latest.

itohro:
Yes, I knew that. This is not meant to merge into master. 

For bug fix release, I created a branch 7.0.x (on the commit tagged as 7.0.0) and this pull-request is what I cherry-picked from master into 7.0.x branch.

motlin:
Ah, I see you created 7.0.x and the pull request goes into eclipse:7.0.x from itohro:7.0.x. I missed that.

Is a branch necessary for the 7.0.1 release? At the moment, there are no code changes on master since the 7.0.0 tag besides the bug-fix. The only other change are to the build and docs which I think are ok to include in 7.0.1.

On the other hand, a branch for a bug-fix release is good practice to be completely sure we're isolating the changes related to the fix. In that case, the 7.0.x branch may only need the first commit, for the bug fix. The second commit, for version number bumps, could be handled by the maven-release-plugin as described [here](http://central.sonatype.org/pages/apache-maven.html#performing-a-release-deployment-with-the-maven-release-plugin). Even if we bump the version numbers manually, as described [here](http://central.sonatype.org/pages/apache-maven.html#performing-a-release-deployment), the second commit should still look like a maven plugin created it, and not include IntelliJ project file changes.

itohro:
Oops, i didn't mean to commit IntelliJ project files. I thought I only chose pom files when committing from IntelliJ, but it seems I missed to do that correctly.

Good to know release plugin does this. Let me get rid of the version changes then. I think we need to re-configure Hudson release build to leverage maven-release-plugin. Let us discuss this offline.

itohro:
Dropped the commit for version update.

PR #30, opened 2016-01-25
Implement Verify.assertListMultimapsEqual(), Verify.assertSetMultimapsEqual(), Verify.assertBagMultimapsEqual(), Verify.assertSortedSetMultimapsEqual(), Verify.assertSortedBagMultimapsEqual().

nikhilnanivadekar:
Implement new methods that provide assertions for ListMultimaps, SetMultimaps and BagMultimaps.
This is a minor version change.

### DIFF
--- a/eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/VerifyTest.java
+++ b/eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/VerifyTest.java
@@ -401,12 +401,13 @@ public class VerifyTest
     @Test
     public void assertListMultimapsEquals()
     {
+        ListMultimap<Integer, String> nullMultimap = null;
+        Verify.assertListMultimapsEqual(null, nullMultimap);
+        ListMultimap<Integer, String> emptyMultimap = FastListMultimap.newMultimap();
+        Verify.assertListMultimapsEqual(FastListMultimap.<Integer, String>newMultimap(), emptyMultimap);
         final ListMultimap<Integer, String> multimap1 = FastListMultimap.newMultimap(Tuples.pair(1, "One"), Tuples.pair(2, "Two"), Tuples.pair(2, "TwoTwo"));
-        MutableListMultimap<Integer, String> multimap2 = null;
-        Verify.assertListMultimapsEqual(null, multimap2);
-        multimap2 = FastListMultimap.newMultimap(Tuples.pair(2, "Two"), Tuples.pair(1, "One"), Tuples.pair(2, "TwoTwo"));
+        MutableListMultimap<Integer, String> multimap2 = FastListMultimap.newMultimap(Tuples.pair(2, "Two"), Tuples.pair(1, "One"), Tuples.pair(2, "TwoTwo"));
         Verify.assertListMultimapsEqual(multimap1, multimap2);
-
         final MutableListMultimap<Integer, String> multimap3 = FastListMultimap.newMultimap(Tuples.pair(1, "One"), Tuples.pair(2, "Two"), Tuples.pair(2, "TwoTwo"));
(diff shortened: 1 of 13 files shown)


nikhilnanivadekar:
I have split the commits in two parts, one for unsorted and one for sorted multimaps.

motlin:
The new assertions assert everything they need to about the equality contract between two multimaps, but don't call the `equals()` method. It would be a little surprising to call `Verify.assert*MultimapsEqual()` and find that `equals()` doesn't get covered. I'd suggest adding a final call to `assertEquals` as a final step in each method, even though it seems redundant.

I'd also recommend adding one assertion between empty multimaps.

nikhilnanivadekar:
Fixed the review comment with commit 641c91b

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/VerifyTest.java
```diff
@@ -380,6 +399,413 @@ public void run()
...
+    public void assertListMultimapsEquals()
+    {
+        final ListMultimap<Integer, String> multimap1 = FastListMultimap.newMultimap(Tuples.pair(1, "One"), Tuples.pair(2, "Two"), Tuples.pair(2, "TwoTwo"));
+        MutableListMultimap<Integer, String> multimap2 = null;
+        Verify.assertListMultimapsEqual(null, multimap2);
+        multimap2 = FastListMultimap.newMultimap();
+        Verify.assertListMultimapsEqual(FastListMultimap.<Integer, String>newMultimap(), multimap2);
+        multimap2 = FastListMultimap.newMultimap(Tuples.pair(2, "Two"), Tuples.pair(1, "One"), Tuples.pair(2, "TwoTwo"));
+        Verify.assertListMultimapsEqual(multimap1, multimap2);
+
```
motlin: Some of the test multimaps are declared as new variables, and some reassign existing variables in scope. I'd prefer consistently declaring new variables.

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/VerifyTest.java
```diff
@@ -380,6 +399,413 @@ public void run()
...
+        Verify.assertError(AssertionError.class, new Runnable()
+        {
+            public void run()
+            {
+                Verify.assertListMultimapsEqual(multimap1, FastListMultimap.<Integer, String>newMultimap());
+            }
+        });
+        Verify.assertError(AssertionFailedError.class, new Runnable()
+        {
+            public void run()
```
motlin: Some of the calls to assertError assert that AssertionError is thrown. Some assert AssertionFailedError.

nikhilnanivadekar: The AssertionFailedError is thrown wherever junit.framework.Assert.failNotEquals() is called.
In this case wherever the call to Verify.assertListsEqual(), Verify.assertSetsEqual(), Verify.assertSortedSetsEqual(), Verify.assertSortedBagsEqual() fails, it throws AssertionFailedError. I will remove the deprecated call to junit.framework.Assert.failNotEquals() in a different pull request for a major version.

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/VerifyTest.java
nikhilnanivadekar: Fixed with commit ebf7ec3.
I have squashed the two code review commits.

motlin:
I see. The method `failNotEquals()` was moved from public to package private in the `org.junit` packages so we can't even get rid of these deprecated usages. That was the last thing I was wondering about. I'm going to go ahead and merge this pull request. Thank you very much!

PR #31, opened 2016-01-26
Modify pom files so that they can be updated via maven-versions-plugin.

itohro:
This is a pull request for branch "7.0.x".

I took couple of changes from master so that 7.0.x branch can be built with maven-versions-plugin.

For subsequent service releases (7.0.1, 7.0.2 etc.), version numbers should be updated via Hudson release build.

### DIFF
--- a/acceptance-tests/pom.xml
+++ b/acceptance-tests/pom.xml
@@ -17,7 +17,7 @@
     <parent>
         <artifactId>eclipse-collections-parent</artifactId>
         <groupId>org.eclipse.collections</groupId>
-        <version>7.0.0</version>
+        <version>7.1.0-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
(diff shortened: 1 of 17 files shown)


motlin:
This PR includes commit bd496ce to update the version to 7.1.0-SNAPSHOT and 4a95a7e to revert it back. We can drop both commits since they cancel each other out. Would you like me to submit a new pull request with db65d78 and 50b60d9 to the 7.0.x branch?

Even though we're working on release builds, I don't think we should leave the version number at 7.0.0 in the pom files. On 7.0.x, we should update it to 7.0.1-SNAPSHOT. What do you think?

itohro:
It'll be helpful if you could submit a new one with db65d78 and 50b60d9 to the 7.0.x branch.

Good point about 7.0.1-SNAPSHOT, i think it makes sense. I can submit that after you submit those 2 commits.

motlin:
Create https://github.com/eclipse/eclipse-collections/pull/33 as discussed

itohro:
Merged #33. Closing this PR.

PR #32, opened 2016-01-26
Add .maven folder to .gitignore file.

itohro:
This is added so that Hudson maven build can commit updated pom files as part of release build, without committing local maven repository under the workspace.

FYI attaching POC release build console: https://hudson.eclipse.org/collections/job/eclipse-collections-release/5/consoleFull

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 target
+.maven


motlin:
In general it's fine to add stuff to `.gitignore`. For my own knowledge, I'm curious where the string `.maven` is coming from. I don't see it in the configuration of the hudson build, and we don't define a `maven-settings.xml` file. I would have expected it to default to `.m2/repository` or something similar.

itohro:
As per the info below, it seems there is .m2/settings.xml in the Hudson node. I'd suspect that .maven thing might be specified there, but I'm not sure how I can see this file.. 

https://wiki.eclipse.org/Hudson#Hudson_configuration_and_tools

PR #37, opened 2016-02-04
Implement Iterate.toMultimap()

nikhilnanivadekar:
Adding `toMultimap()` on `RichIterable` is a major version change. It can be added in the static utility for a minor version.
There is a new [MultimapKeyValuePutProcedure](https://github.com/nikhilnanivadekar/eclipse-collections/blob/bc7c316629762b43d5d8b05e9a1e5956d7557501/eclipse-collections/src/main/java/org/eclipse/collections/impl/block/procedure/MultimapKeyValuePutProcedure.java) created, if this will be a major version change, I can inline the call for minor version change, and replace with the `Procedure` in major version.
I have only added the `targetMultimap` variant of `toMultimap()`, this is because I want users to control the output instead of deciding which type of `Multimap` is returned for a particular `Iterable`.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/Iterate.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/Iterate.java
@@ -72,6 +72,7 @@ import org.eclipse.collections.impl.block.factory.Procedures2;
 import org.eclipse.collections.impl.block.procedure.MapCollectProcedure;
 import org.eclipse.collections.impl.block.procedure.MaxComparatorProcedure;
 import org.eclipse.collections.impl.block.procedure.MinComparatorProcedure;
+import org.eclipse.collections.impl.block.procedure.MultimapKeyValuePutAllProcedure;
 import org.eclipse.collections.impl.factory.Lists;
 import org.eclipse.collections.impl.list.mutable.FastList;
 import org.eclipse.collections.impl.map.mutable.UnifiedMap;
@@ -2996,6 +2997,48 @@ public final class Iterate
     }
 
     /**
+     * Iterate over the specified collection applying the specified Functions to each element to calculate
+     * a key and values, add the results to {@code targetMultimap} and return the {@code targetMultimap}.
+     * <p>
+     * Example using a Java 8 lambda expression:
+     * <pre>
+     * MutableMultimap&lt;String, String&gt; multimap =
+     *      Iterate.<b>toMultimap</b>(integers, each -> "key:" + each, each -> Lists.mutable.of("value:" + each), FastListMultimap.newMultimap());
+     * </pre>
+     * <p>
+     * Example using an anonymous inner class:
+     * <pre>
+     * MutableMultimap&lt;String, String&gt; multimap =
+     *      Iterate.<b>groupByAndCollect</b>(integers,
+     *          new Function&lt;Integer, String&gt;()
+     *          {
+     *              public String valueOf(Integer each)
+     *              {
+     *                  return "key:" + each;
+     *              }
+     *          }, new Function&lt;Integer, Iterable&lt;String&gt;&gt;()
+     *          {
+     *              public Iterable&lt;String&gt; valueOf(Integer each)
+     *              {
+     *                  return Lists.mutable.of("value:" + each);
+     *              }
+     *          }, FastListMultimap.newMultimap());
+     * </pre>
+     *
+     * @see Iterate#groupBy(Iterable, Function) when only keys get transformed
+     * @see Iterate#groupByEach(Iterable, Function) when only keys get transformed and Function returns multiple keys
+     */
+    public static <T, K, V, R extends MutableMultimap<K, V>> R toMultimap(
+            Iterable<T> iterable,
+            Function<? super T, ? extends K> keyFunction,
+            Function<? super T, ? extends Iterable<V>> valuesFunction,
+            R targetMultimap)
+    {
+        Iterate.forEach(iterable, new MultimapKeyValuePutAllProcedure<T, K, V>(targetMultimap, keyFunction, valuesFunction));
+        return targetMultimap;
+    }
+
+    /**
      * Return the specified collection as a sorted List.
      */
     public static <T extends Comparable<? super T>> MutableList<T> toSortedList(Iterable<T> iterable)
@@ -3089,6 +3132,8 @@ public final class Iterate
 
     /**
      * @see RichIterable#groupBy(Function)
+     * @see Iterate#groupByEach(Iterable, Function) when function returns multiple keys
+     * @see Iterate#toMultimap(Iterable, Function, Function, MutableMultimap) when both keys and values get transformed
      */
     public static <T, V> MutableMultimap<V, T> groupBy(
             Iterable<T> iterable,
@@ -3119,6 +3164,8 @@ public final class Iterate
 
     /**
      * @see RichIterable#groupBy(Function, MutableMultimap)
+     * @see Iterate#groupByEach(Iterable, Function, MutableMultimap) when function returns multiple keys
+     * @see Iterate#toMultimap(Iterable, Function, Function, MutableMultimap) when both keys and values get transformed
      */
     public static <T, V, R extends MutableMultimap<V, T>> R groupBy(
             Iterable<T> iterable,
@@ -3202,6 +3249,8 @@ public final class Iterate
 
     /**
      * @see RichIterable#groupByEach(Function)
+     * @see Iterate#groupBy(Iterable, Function) when function returns single key
+     * @see Iterate#toMultimap(Iterable, Function, Function, MutableMultimap) when both keys and values get transformed
      */
     public static <T, V> MutableMultimap<V, T> groupByEach(
             Iterable<T> iterable,
@@ -3232,6 +3281,8 @@ public final class Iterate
 
     /**
      * @see RichIterable#groupByEach(Function, MutableMultimap)
+     * @see Iterate#groupBy(Iterable, Function, MutableMultimap) when function returns single key
+     * @see Iterate#toMultimap(Iterable, Function, Function, MutableMultimap) when both keys and values get transformed
      */
     public static <T, V, R extends MutableMultimap<V, T>> R groupByEach(
             Iterable<T> iterable,
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/utility/IterateTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/utility/IterateTest.java
@@ -47,7 +47,12 @@ import org.eclipse.collections.api.map.primitive.ObjectDoubleMap;
 import org.eclipse.collections.api.map.primitive.ObjectLongMap;
 import org.eclipse.collections.api.multimap.Multimap;
 import org.eclipse.collections.api.multimap.MutableMultimap;
+import org.eclipse.collections.api.multimap.bag.BagMultimap;
+import org.eclipse.collections.api.multimap.bag.MutableBagMultimap;
+import org.eclipse.collections.api.multimap.list.MutableListMultimap;
 import org.eclipse.collections.api.multimap.set.MutableSetMultimap;
+import org.eclipse.collections.api.multimap.sortedbag.MutableSortedBagMultimap;
+import org.eclipse.collections.api.multimap.sortedset.MutableSortedSetMultimap;
 import org.eclipse.collections.api.partition.PartitionIterable;
 import org.eclipse.collections.api.set.MutableSet;
 import org.eclipse.collections.api.tuple.Pair;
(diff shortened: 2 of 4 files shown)


itohro:
It looks good besides the unused import above. I believe adding a new procedure class is ok as minor changes (it doesn't force existing clients to use the new API). 

Just one thought after looking at the new API. Would it be also helpful if we have an additional API like below (we may want to have a better name)? The only difference is that value function is expected to return collection of values, and we can eventually call putAll() when putting values to a multimap.

    public static <T, K, V> MutableMultimap<K, V> toFlatValueMultimap(
            Iterable<T> iterable,
            Function<? super T, ? extends K> keyFunction,
            Function<? super T, Iterable<? extends V>> valueFunction,
            MutableMultimap<K, V> targetMultimap)

This doesn't need to be in this commit and can be taken into another issue.

motlin:
To be consistent with `groupBy()` and `groupByEach()`, I think that this functionality is the one that should be renamed and that the alternate type signature suggested by @itohro should get the name `toMultimap()`. What do you think?

itohro:
+1 on @motlin 's suggestion.

nikhilnanivadekar:
@motlin , @itohro What will be the names of two methods?

```
public static <T, K, V> MutableMultimap<K, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, ? extends K> keyFunction,
        Function<? super T, ? extends V> valueFunction,
        MutableMultimap<K, V> targetMultimap)

public static <T, K, V> MutableMultimap<K, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, ? extends K> keyFunction,
        Function<? super T, Iterable<? extends V>> valueFunction,
        MutableMultimap<K, V> targetMultimap)
```

Also, we need to add two more methods, to cover all the patterns:

```
public static <T, K, V> MutableMultimap<K, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, Iterable<? extends K>> keyFunction,
        Function<? super T, ? extends V> valueFunction,
        MutableMultimap<K, V> targetMultimap)

public static <T, K, V> MutableMultimap<K, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, Iterable<? extends K>> keyFunction,
        Function<? super T, Iterable<? extends V>> valueFunction,
        MutableMultimap<K, V> targetMultimap)
```

itohro:
Thinking further on naming consistency with groupBy(), groupByEach(), collect() and flatCollect(). How do the names below sound?

Note: I fixed generics signature for Iterable from ```Iterable<? extends V>``` to ```? extends Iterable<V>```.

```
public static <T, K, V> MutableMultimap<K, V> collectGroupBy(
        Iterable<T> iterable,
        Function<? super T, ? extends K> keyFunction,
        Function<? super T, ? extends V> valueFunction,
        MutableMultimap<K, V> targetMultimap)

public static <T, K, V> MutableMultimap<K, V> flatCollectGroupBy(
        Iterable<T> iterable,
        Function<? super T, ? extends K> keyFunction,
        Function<? super T, ? extends Iterable<V>> valueFunction,
        MutableMultimap<K, V> targetMultimap)

public static <T, K, V> MutableMultimap<K, V> collectGroupByEach(
        Iterable<T> iterable,
        Function<? super T, ? extends Iterable<V>> keyFunction,
        Function<? super T, ? extends V> valueFunction,
        MutableMultimap<K, V> targetMultimap)

public static <T, K, V> MutableMultimap<K, V> flatCollectGroupByEach(
        Iterable<T> iterable,
        Function<? super T, ? extends Iterable<V>> keyFunction,
        Function<? super T, ? extends Iterable<V>> valueFunction,
        MutableMultimap<K, V> targetMultimap)
```

itohro:
And if we go with the method names above, we might want to change the name and order of functions as collectFunction (from valueFunction) and groupByFunction (from keyFunction).

motlin:
Do you think all four variants are useful? Or is this for completeness?

nikhilnanivadekar:
@itohro I would want to have `toMultimap()` in the method names, this is because for consistency with other iteration patterns of `toList()`,` toSet()`, `toMap()`, `toBag()`.

@motlin it is more for completeness. I think the most useful one to have will be: 

```
public static <T, K, V> MutableMultimap<K, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, ? extends K> keyFunction,
        Function<? super T, ? extends V> valueFunction,
        MutableMultimap<K, V> targetMultimap)

```

itohro:
Putting aside the naming for a moment, I could think of the 3 APIs below are equally useful. Taking company kata domain as an example. 

* Get customer names per city: 
```collectGroupBy``` => from customers, get a multimap with key=city and value=names
* Get ordered items per city: 
```flatCollectGroupBy``` => from customers, get a multimap with key=city and value=items 
* Get cities per ordered item: 
```collectGroupByEach``` => from customers, get a multimap with  key=item and value=cities  

There might be some cases for flatCollectGroupByEach as well, but it might not be as common as other 3 variants.

Coming back to the naming, I agree we'd want to have an API named toMultimap() because that's where people may look up first when thinking about the use cases above. Following @motlin 's original suggestion, how about naming them as below? We could add references to other variants each other in Javadoc for better discoverability. 

```
public static <T, K, V> MutableMultimap<K, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, ? extends K> keyFunction,
        Function<? super T, ? extends Iterable<V>> valueFunction,
        MutableMultimap<K, V> targetMultimap)

public static <T, K, V> MutableMultimap<K, V> collectGroupBy(
        Iterable<T> iterable,
        Function<? super T, ? extends K> groupByFunction,
        Function<? super T, ? extends V> collectFunction,
        MutableMultimap<K, V> targetMultimap)

public static <T, K, V> MutableMultimap<K, V> collectGroupByEach(
        Iterable<T> iterable,
        Function<? super T, ? extends Iterable<V>> groupByEachFunction,
        Function<? super T, ? extends V> collectFunction,
        MutableMultimap<K, V> targetMultimap)
```

I'm not keen on adding the 4th one, but we could name it below if we want to.

```
public static <T, K, V> MutableMultimap<K, V> toMultimapByEach(
        Iterable<T> iterable,
        Function<? super T, ? extends Iterable<V>> keyFunction,
        Function<? super T, ? extends Iterable<V>> valueFunction,
        MutableMultimap<K, V> targetMultimap)
```

**Comment by motlin** [comment]
Some more thoughts on naming. Two of these patterns have workarounds already. They are the ones where you can call `groupBy()` or `groupByEach()` followed by `collectValues()`. Since they have a workaround, I'm not convinced we should add them at all. I thought about making these methods overloads since they have a different number of parameters than `groupBy()` and `groupByEach()`. But it would interfere with a form of `groupBy()` that groups multiple levels, in other words, two Functions from `T` to `K1` and `K2`. So I think the names suggested by @itohro are good, but I'd flip them so that they sort alphabetically with `groupBy()` instead of `collect()`.

```
public static <T, K, V> MutableMultimap<K, V> groupByCollect(
        Iterable<T> iterable,
        Function<? super T, ? extends K> groupByFunction,
        Function<? super T, ? extends V> collectFunction)

public static <T, K, V> MutableMultimap<K, V> groupByEachCollect(
        Iterable<T> iterable,
        Function<? super T, ? extends Iterable<V>> groupByEachFunction,
        Function<? super T, ? extends V> collectFunction)
```

But like I said, maybe it would be better not to add them. The methods I've found myself needing without great workarounds are these two.

```java
public static <T, V> MutableMultimap<T, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, ? extends Iterable<V>> valueFunction)

public static <T, K, V> MutableMultimap<K, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, ? extends K> keyFunction,
        Function<? super T, ? extends Iterable<V>> valueFunction)
```

nikhilnanivadekar:
I agree with @motlin the most common use cases I have come across were for these, also these methods are closer to the `toMap()` patterns which are present.

```
public static <T, V> MutableMultimap<T, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, ? extends Iterable<V>> valueFunction)

public static <T, K, V> MutableMultimap<K, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, ? extends K> keyFunction,
        Function<? super T, ? extends Iterable<V>> valueFunction)
```

I will add these two on `Iterate` for now, and add them on `RichIterable` for the next major release.

itohro:
The plan sounds good to me!

nikhilnanivadekar:
On second thoughts, if I add this: 

```
public static <T, K, V> MutableMultimap<K, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, ? extends K> keyFunction,
        Function<? super T, ? extends Iterable<V>> valueFunction)
```
the other variant:
```
public static <T, V> MutableMultimap<T, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, ? extends Iterable<V>> valueFunction)
```
gets a workaround, because now the `keyFunction` can just be `each -> each` or `Functions.identity()`

itohro:
Why don't we add ```groupByCollectForEach()``` as well then? As I mentioned above, the 3 APIs are equally useful and I'd add it, too, if we were to add ```groupByCollect()```.

One minor suggestion in Javadoc. Let's add @see tag to refer to other method variants each other. They get different names but they're very similar by nature. For example, users might lookup ```toMultimap()``` first when he/she need is actually ```groupByCollect()``` (as you did @nikhilnanivadekar , and I would have done it, too :) ). Having @see tag is useful for better discoverability.

nikhilnanivadekar:
In that case, I will split all the three as 3 separate commits under 1 pull request, or do you want 3 different pull requests.
@itohro, @motlin what do you think?

itohro:
I'm good with this => "3 separate commits under 1 pull request".

But let's wait for @motlin 's feedback.

nikhilnanivadekar:
Apart from add groupByEachCollect() can you please review the code for toMultimap() and groupByCollect(), if there is feedback, I can fix it in one commit itself.

motlin:
There's one reason you might want to split the pull request. I think adding `toMultimap()` won't require much discussion and can be pulled right away.

```java
public static <T, K, V> MutableMultimap<K, V> toMultimap(
        Iterable<T> iterable,
        Function<? super T, ? extends K> keyFunction,
        Function<? super T, ? extends Iterable<V>> valueFunction)
```

Maybe even the variant without a keyFunction can be pulled right away. For the other variants, we probably should discuss naming and get feedback from more people, including @donraab.

nikhilnanivadekar:
But do we require a variant without a keyFunction, will it not be using `Functions.identity()` or `each -> each`. 
Or do you want to add it for completeness?

motlin:
No it doesn't seem necessary. Seems nice to have.

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/Iterate.java
```diff
@@ -3258,6 +3301,45 @@ public static boolean contains(Iterable<?> iterable, Object value)
     }
 
     /**
+     * Iterate over the specified collection applying the specified Functions to each element to calculate
+     * a key and value, add the results to {@code targetMultimap} and return the {@code targetMultimap}.
+     * <p>
```
motlin: a key and several values (rather than one value)

itohro:
If you could move other commits (69c0a6c and onwards) to a separate branch, we can merge 7f77c47 right away in this PR. The changes in 7f77c4 looks good to me.

As @motlin suggested, let's discuss with @donraab in regards to the necessity and naming of the other variants.

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/Iterate.java
```diff
@@ -3118,7 +3163,11 @@ public static boolean contains(Iterable<?> iterable, Object value)
     }
 
     /**
+     * Also see {@link Iterate#groupByEach(Iterable, Function, MutableMultimap)}, {@link Iterate#toMultimap(Iterable, Function, Function, MutableMultimap)}.
```
motlin: This one Javadoc comment using "Also see" and Javadoc links seems inconsistent with the rest.

unit-tests/src/test/java/org/eclipse/collections/impl/utility/IterateTest.java
```diff
@@ -739,6 +752,97 @@ public void toMapSelectingKeyAndValue()
...
+                Tuples.pair(3, "3"),
+                Tuples.pair(3, "1"),
+                Tuples.pair(3, "3"),
+                Tuples.pair(3, "1"),
+                Tuples.pair(3, "3")};
+
+        Function<Integer, Integer> keyFunction = each -> each;
+        Function<Integer, Iterable<String>> valuesFunction = each -> Lists.mutable.of("1", String.valueOf(each));
+        MutableMultimap<Integer, String> actualMultimap1 = Iterate.toMultimap(list, keyFunction, valuesFunction, Multimaps.mutable.list.empty());
+        Verify.assertListMultimapsEqual(FastListMultimap.newMultimap(expectedPairs), (ListMultimap<Integer, String>) actualMultimap1);
```
motlin: You can get rid of the cast to `ListMultimap` if you change the assignment on the previous line to type `MutableListMultimap`. I think all casts in this test are avoidable.

motlin:
:+1:

PR #41, opened 2016-02-08
Fix memory leak in HashBiMap. 

bhav0904:
I added a third test to check for key preservation when we use forcePut().

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/HashBiMapTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/HashBiMapTest.java
@@ -113,13 +113,12 @@ public class HashBiMapTest extends AbstractMutableBiMapTestCase
         Key key2 = new Key("2");
 
         Key value1 = new Key("abc");
-        Key value2 = new Key("xyz");
 
         Key duplicateOfKey1 = new Key("1");
 
-        HashBiMap<Key, Key> biMap = this.newMapWithKeysValues(key1, value1, key2, value2);
+        HashBiMap<Key, Key> biMap = this.newMapWithKeysValues(key1, value1, key2, new Key("xyz"));
 
-        biMap.forcePut(duplicateOfKey1, value2);
+        biMap.forcePut(duplicateOfKey1, new Key("xyz"));
 
         Assert.assertSame(key1, Iterate.getFirst(biMap.entrySet()).getKey());
         Assert.assertSame(key1, Iterate.getFirst(biMap.inverse().entrySet()).getValue());
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/HashBiMapTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/HashBiMapTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015 Goldman Sachs.
+ * Copyright (c) 2016 Goldman Sachs.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -10,8 +10,11 @@
 
 package org.eclipse.collections.impl.bimap.mutable;
 
+import org.eclipse.collections.api.bimap.MutableBiMap;
 import org.eclipse.collections.impl.map.mutable.UnifiedMap;
 import org.eclipse.collections.impl.test.Verify;
+import org.eclipse.collections.impl.test.domain.Key;
+import org.eclipse.collections.impl.utility.Iterate;
 import org.junit.Assert;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


nikhilnanivadekar:
b842a56 does not have a sign-off by @bhav0904. That is why the ip validation is failing.

motlin:
Reopened this so that the Travis builds run. I think this branch is in good shape now.

bhav0904:
What are the next steps? Do I need to do anything else here?

unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/HashBiMapTest.java
```diff
@@ -102,4 +105,23 @@ public void withKeysValues()
...
+
+    @Test
+    public void forcePut_inverseKeyPreservation()
+    {
+        Key key1 = new Key("1");
+        Key key2 = new Key("2");
+
+        Key value1 = new Key("abc");
+        Key value2 = new Key("xyz");
+
```
motlin: There are no assertions using `value2`. Is it necessary?

bhav0904: I wanted to have "equal" values in the BiMap and in the forcePut so that containsKey and containsValue are both true. So I reused value2 in the forcePut call. I could inline value2 if it is confusing.

motlin: No need to inline it. What I mean is that the `Key` type exists to be explicit about `Strings` that are equal but not the same instance. You have code to create a duplicate key and later assert you find the same instance.

```java
Key duplicateOfKey1 = new Key("1");
...
Assert.assertSame(key1, Iterate.getFirst(biMap.entrySet()).getKey());
```

But there's no code like that for values. If it's important to assert separately, then you're missing assertions about values (or the inverse bimap's keys). If it's not important to test separately, then it's unnecessary to use `BiMap<Key, Key>` and you could simplify to `BiMap<Key, String>`.

unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/HashBiMapTest.java
```diff
@@ -102,4 +105,22 @@ public void withKeysValues()
...
+        Key duplicateOfKey1 = new Key("1");
+
+        HashBiMap<Key, Key> biMap = this.newMapWithKeysValues(key1, value1, key2, new Key("xyz"));
+
+        biMap.forcePut(duplicateOfKey1, new Key("xyz"));
+
+        Assert.assertSame(key1, Iterate.getFirst(biMap.entrySet()).getKey());
+        Assert.assertSame(key1, Iterate.getFirst(biMap.inverse().entrySet()).getValue());
+    }
```
motlin: Repeating my comment on the latest commit.

There are no assertions about `"abc"` or `"xyz"`. If it's important to make assertions about values (or the Bimap's invere's keys), then they should be added here. If it's not important to test separately, then it's unnecessary to use BiMap<Key, Key> and you could simplify to BiMap<Key, String>.

bhav0904: So, this?

`@Test
    public void forcePut_inverseKeyPreservation()
    {
        Key key1 = new Key("1");
        Key key2 = new Key("2");

        String value1 = "abc";
        String value2 = "xyz";

        Key duplicateOfKey1 = new Key("1");

        HashBiMap<Key, String> biMap = this.newMapWithKeysValues(key1, value1, key2, value2);

        biMap.forcePut(duplicateOfKey1, value2);

        Assert.assertSame(key1, Iterate.getFirst(biMap.entrySet()).getKey());
        Assert.assertSame(key1, Iterate.getFirst(biMap.inverse().entrySet()).getValue());
    }`

motlin: If you don't think assertions on the value are necessary. And in that case, you can just inline key2, value1, and value2.

bhav0904: I realized that we need to assertSame on the value as well. Sorry for the confusion. I think this should be good:

`@Test
    public void forcePut_inverseKeyAndValuePreservation()
    {
        Key key1 = new Key("1");
        Key value2 = new Key("xyz");

        HashBiMap<Key, Key> biMap = this.newMapWithKeysValues(key1, new Key("abc"), new Key("2"), value2);

        Key duplicateOfKey1 = new Key("1");
        Key duplicateOfValue2 = new Key("xyz");

        biMap.forcePut(duplicateOfKey1, duplicateOfValue2);

        Assert.assertSame(key1, Iterate.getFirst(biMap.entrySet()).getKey());
        Assert.assertSame(key1, Iterate.getFirst(biMap.inverse().entrySet()).getValue());

        Assert.assertSame(value2, Iterate.getFirst(biMap.entrySet()).getValue());
        Assert.assertSame(value2, Iterate.getFirst(biMap.inverse().entrySet()).getKey());
    }`

motlin: looks good to me :+1:

motlin:
:+1: Thanks again!

bhav0904:
Are we going to port this change to the master branch too?

motlin:
We planning on releasing a bug-fix version (7.0.2) and then merging 7.0.x into master.

PR #57, opened 2016-02-21
Fix GenerateMojo Deprecated Syntax Warnings

marschall:
The GenerateMojo uses the deprecated expression syntax which results in
deprecation warnings during the build.

Switch to the property or default value syntax that is recommended by
Maven.

Signed-off-by: Philippe Marschall <philippe.marschall@gmail.com>

### DIFF
--- a/eclipse-collections-code-generator-maven-plugin/src/main/java/org/eclipse/collections/codegenerator/maven/GenerateMojo.java
+++ b/eclipse-collections-code-generator-maven-plugin/src/main/java/org/eclipse/collections/codegenerator/maven/GenerateMojo.java
@@ -31,12 +31,12 @@ public class GenerateMojo extends AbstractMojo
     /**
      * Skips code generation if true.
      *
-     * @parameter expression="${skipCodeGen}"
+     * @parameter property="skipCodeGen"
      */
     private boolean skipCodeGen;
 
     /**
-     * @parameter expression="${project.build.directory}/generated-sources"
+     * @parameter default-value="${project.build.directory}/generated-sources"
      * @required
      */
     private String templateDirectory;
@@ -44,7 +44,7 @@ public class GenerateMojo extends AbstractMojo
     /**
      * The Maven project to act upon.
      *
-     * @parameter expression="${project}"
+     * @parameter property="project"
      * @required
      */
     private MavenProject project;


marschall:
The CLA check doesn't pass because I did the commit with my secondary email address. This email address is registered and validated with git.eclipse.org. Is there somehow a way to sync dev.eclipse.org and git.eclipse.org?

motlin:
Thanks! What's the difference between `property` and `default-value`?

eclipse-collections-code-generator-maven-plugin/src/main/java/org/eclipse/collections/codegenerator/maven/GenerateMojo.java
```diff
@@ -31,12 +31,12 @@
...
      * Skips code generation if true.
      *
-     * @parameter expression="${skipCodeGen}"
+     * @parameter property="skipCodeGen"
      */
     private boolean skipCodeGen;
 
     /**
-     * @parameter expression="${project.build.directory}/generated-sources"
+     * @parameter default-value="${project.build.directory}/generated-sources"
```
itohro: Per [this API doc](https://maven.apache.org/developers/mojo-api-specification.html), think this should look like below? Guess original `expression` should have been `default-value` to begin with. CC @motlin for confirmation. 

```
@parameter property="templateDirectory" default-value="${project.build.directory}/generated-sources"
```

motlin: I'm not sure. The build passes either way?

marschall: I can check that.

marschall: "default-value" is an expression that is used to calculate the value. "property" references a system property that can be overridden with 
```
mvn -DtemplateDirectory=someDirectory
```

I don't think it makes sense to be able to dynamically change the template directory from the command line. But you know best, I can add the property it if you want it. The build works either way.

Being able to skip the code generation from the command line with
```
mvn -DskipCodeGen=true
```
on the other hand makes sense to me.

motlin: Thanks for the explanation. Makes sense to me. Do you mind rebasing on top of master?

marschall: Rebased.

motlin:
I see the ip-validation works now too. Looks good. Thank you very much!

PR #60, opened 2016-02-23
Add release-artifacts profile to the build in the junit-trait-runner maven module.

motlin:
Alternatively, we could set maven.deploy.skip to true for these two modules.


motlin:
@itohro if this is ok with you, I'd manually move the 7.0.2 tag to this commit. If not, we'd have to start 7.0.3.

itohro:
Plaese go ahead with manually moving the 7.0.2 tag to this commit.

PR #63, opened 2016-02-26
Fix bug in ArrayIterate.chunk().

motlin:
With the bug, splitting `{1, 2, 3, 4, 5, 6, 7}` into chunks of size 2 was giving `[[1, 2], [1, 2], [1, 2], [1]]`.


itohro:
LGTM.

PR #66, opened 2016-03-10
Update README.md links to point to 7.0.2.

itohro:


motlin:
After this one, maybe we should move this information into the wiki?

itohro:
I'd keep the maven/gradle/ivy information in README (i.e. top page) where visitors can find at first glance. That's the very first information users would look for.

We have the official webpage that also has the same information, but we sill have number of visitors to GitHub. We should leave the information where it's very easy to find (without any clicks).

motlin:
Fair enough. I'd like to automate it. Ideally the change would happen in the same commit as upgrading the version number in pom files during a release.

itohro:
Agree and I'd want to make all version changes at once including the official web landing page. We'll see how we can achieve that.

PR #67, opened 2016-03-10
Replace forkMode with forkCount

motlin:
forkMode has been deprecated since surefire version 2.14 and been
replaced with forkCount and reuseForks.

Signed-off-by: Philippe Marschall <philippe.marschall@gmail.com>


itohro:
Travis is still running but merging now as the change is already validated in another PR.

PR #82, opened 2016-04-10
Implement MutableMultimap.putAllPairs(Iterable<Pair<K, V>> keyValuePairs).

### LINKED ISSUE #43: Implement MutableMultimap.putAllPairs(Iterable<Pair<K, V>> keyValuePair).
Currently MutableMultimap has these two bulk operations: 
`putAllPairs(Pair<K, V>... pairs) `
`putAll(K key, Iterable<? extends V> values)`

it does not have `putAllPairs(Iterable<Pair<K, V>> keyValuePairs)`

the implementation would be straightforward which is the current workaround:

`Iterate.forEach(keyValuePairs, each -> multimap.add(each));`

nikhilnanivadekar:
Fixes #43.

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/multimap/AbstractMutableMultimapTestCase.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/multimap/AbstractMutableMultimapTestCase.java
@@ -12,6 +12,8 @@ package org.eclipse.collections.impl.multimap;
 
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.bag.MutableBag;
+import org.eclipse.collections.api.list.ImmutableList;
+import org.eclipse.collections.api.list.MutableList;
 import org.eclipse.collections.api.map.MutableMap;
 import org.eclipse.collections.api.multimap.ImmutableMultimap;
 import org.eclipse.collections.api.multimap.Multimap;
@@ -22,6 +24,7 @@ import org.eclipse.collections.impl.bag.mutable.HashBag;
 import org.eclipse.collections.impl.block.procedure.CollectionAddProcedure;
 import org.eclipse.collections.impl.factory.Bags;
 import org.eclipse.collections.impl.factory.Lists;
+import org.eclipse.collections.impl.factory.Sets;
 import org.eclipse.collections.impl.map.mutable.UnifiedMap;
 import org.eclipse.collections.impl.set.mutable.UnifiedSet;
 import org.eclipse.collections.impl.test.Verify;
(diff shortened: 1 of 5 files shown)


unit-tests/src/test/java/org/eclipse/collections/impl/multimap/AbstractMutableMultimapTestCase.java
```diff
@@ -132,6 +135,33 @@ public void testPutAll()
...
 
     @Test
+    public void testPutAllPairs()
+    {
+        MutableMultimap<Integer, String> multimap1 = this.newMultimapWithKeysValues(1, "One", 2, "2");
+        MutableList<Pair<Integer, String>> pairs1 = Lists.mutable.of(Tuples.pair(1, "One"), Tuples.pair(2, "Two"), Tuples.pair(3, "Three"));
+        Assert.assertTrue(multimap1.putAllPairs(pairs1));
+        MutableMultimap<Integer, String> expected1 = this.newMultimapWithKeysValues(1, "One", 1, "One", 2, "2", 2, "Two");
+        expected1.put(3, "Three");
+        Assert.assertEquals(expected1, multimap1);
```
motlin: The first three lines make a good test of `putAllPairs()`. I find lines 4 and 5 confusing, which mutate the expected Multimap. Is it just because the factories only have overloads going up to 4 keys? If so, I'd prefer to work around that limitation in a more consistent way.

nikhilnanivadekar: Yes that's right, it is because of the overloads going only up to 4 keys. I can make it consistent by creating an empty Multimap and then go on adding a key and value. Another option is to only use 4 keys, but I think there should be a test with duplicates.
Which one would you prefer? Or is there an option better than the two I mentioned?

unit-tests/src/test/java/org/eclipse/collections/impl/multimap/AbstractMutableMultimapTestCase.java
```diff
@@ -132,6 +135,43 @@ public void testPutAll()
...
+        Assert.assertTrue(multimap3.putAllPairs(pairs3));
+        MutableMultimap<String, Integer> expected3 = this.newMultimap();
+        expected3.put("One", 1);
+        expected3.put("One", 1);
+        expected3.put("Two", 2);
+        expected3.put("Two", 2);
+        expected3.put("Three", 3);
+        Assert.assertEquals(expected3, multimap3);
+    }
+
```
itohro: We may want to add a case when putAllPairs() returns false.

nikhilnanivadekar: Added it in AbstractMutableSetMultimapTestCase.

PR #83, opened 2016-04-10
Implement keySet() on Multimap.

### LINKED ISSUE #45: Implement keySet() on Multimap.
`Multimap` has `keyBag()` however, it is an O(n) operation which loops through the elements and creates a new Bag. Since a `Multimap` is defined as a `Map` of key and value `Collection`, there should be a method `keySet()` similar to what is present on a `Map`.

The `keySet()` implementation will go on [AbstractMutableMultimap](https://github.com/eclipse/eclipse-collections/blob/8d45bc751e796c1e0f8cc21a26d4f96b060392c2/eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractMutableMultimap.java) and [AbstractImmutableMultimap](https://github.com/eclipse/eclipse-collections/blob/8d45bc751e796c1e0f8cc21a26d4f96b060392c2/eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractImmutableMultimap.java).

On `AbstractMutableMultimap`, the implementation will be:

```
public Set<K> keySet()
{
  return this.map.keySet();
}
```

On `AbstractImmutableMultimap`, the implementation will be:

```
public Set<K> keySet()
{
  return this.map.castToMap().keySet();
}
```

Note: 
1) This cannot be added on `AbstractMultimap` as the `getMap()` method on it returns a `MapIterable` and does not have a `keySet()` method as yet.
2) Current workaround is `keysView()` which is a lazy view of the unique keys.

nikhilnanivadekar:
Fixes #45.

### DIFF
--- a/RELEASE_NOTE_DRAFT.md
+++ b/RELEASE_NOTE_DRAFT.md
@@ -1,20 +1,20 @@
-8.0.0 
+8.0.0
 =====
 
 New Functionality
 -----------------
 
-* A Placeholder 
+* Implemented Multimap.keySet() to return an unmodifiable SetIterable of keys.
 
 Optimizations
 -------------
 
-* A Placeholder 
+* A Placeholder
 
 Bug fixes
 ---------
 
-* A Placeholder 
+* A Placeholder
 
 Acquiring Eclipse Collections
 -----------------------------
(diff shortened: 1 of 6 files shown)


itohro:
This should be ready to be merged as soon as a draft release note is added.

nikhilnanivadekar:
Done! Please check.

PR #84, opened 2016-04-11
Optimize MutableList.chunk() for RandomAccess lists

nikhilnanivadekar:
Current implementation of `MutableList.chunk()` uses an iterator. Optimize to use the backing array when the list is RandomAccess. 
Currently all implementations which extend `AbstractMutableList` are RandomAccess.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
@@ -22,6 +22,7 @@ import java.util.Random;
 import java.util.RandomAccess;
 import java.util.concurrent.ExecutorService;
 
+import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.block.HashingStrategy;
 import org.eclipse.collections.api.block.function.Function;
 import org.eclipse.collections.api.block.function.Function0;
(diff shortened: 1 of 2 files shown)


motlin:
Could you split the pull request? One for optimizations and one for API changes. There's no dependency between them, right?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
```diff
@@ -1116,6 +1117,29 @@ public int binarySearch(T key)
...
+
+        MutableList<RichIterable<T>> result = Lists.mutable.empty();
+        int i = 0;
+        while (i < this.size())
+        {
+            MutableList<T> batch = new FastList<T>(Math.min(size, this.size() - i));
+
+            for (int j = 0; j < size && i < this.size(); j++)
+            {
+                batch.add(this.get(i));
```
motlin: This assumes that all subclasses of AbstractMutableList implement RandomAccess. I think we're inconsistent about this assumption today and ought to split AbstractMutableList into two classes.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/ImmutableSetFactoryImpl.java
```diff
@@ -155,7 +157,28 @@
...
         {
-            return this.with();
+            return this.empty();
+        }
+
+        if (items instanceof Set)
+        {
+            if (((Set) items).size() > 4)
+            {
+                return (ImmutableSet<T>) ImmutableUnifiedSet.newSet(items);
```
motlin: I don't think this is something we should do. Generally, we want to call toArray() before constructing an immutable collection because toArray() is atomic.

nikhilnanivadekar: Ok that makes sense. I will drop this commit.
Thanks.

PR #85, opened 2016-04-11
Pull up into() from LazyIterable to RichIterable.

donraab:


itohro:
There are test failures. Could you please take a look?
https://travis-ci.org/eclipse/eclipse-collections/jobs/122367960

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -1332,6 +1332,13 @@
     double injectInto(double injectedValue, DoubleObjectToDoubleFunction<? super T> function);
 
     /**
+     * Adds all the elements in this iterable to the specific target Collection.
+     *
+     * @since 8.0
+     */
+    <R extends Collection<T>> R into(R target);
```
motlin: I think a more flexible signature would be:

```java
<R extends Collection<? super T>> R into(R target);
```

This would allow collecting into a `MutableCollection<Object>` for example.

donraab: Good idea.  I will do that in a separate commit.

itohro:
LGTM.

PR #75, opened 2016-03-23
Fix ecj compilation issues.

itohro:
FYI, this is meant to be merged into 'eclipse-compile' branch.

Which do you prefer, merging this pull-request, or directly pushing this commit to `eclipse-compile` branch in eclipse repo? I'm ok either way, but leaning towards direct push as long as this PR is reviewed. I think we should keep rebasing this branch directly in eclipse repo anyways. Let me know.

FYI I've put together a draft instruction for Eclipse users here. We can link to it from CONTRIBUTING.md once we're good with it.

https://github.com/itohro/eclipse-collections/wiki/How-to-build-Eclipse-Collections-in-Eclipse-IDE

If we were to merge pull-request, I'll need to update an [instruction of rebase](git rebase --onto master origin/eclipse-compile^ origin/eclipse-compile) from

```
git rebase --onto master origin/eclipse-compile^ origin/eclipse-compile
```

to

```
git rebase --onto master origin/eclipse-compile^^ origin/eclipse-compile
```

PR #87, opened 2016-04-13
Fix ecj compilation issues.

itohro:
Renamed branch name from #75 .

Can you merge this branch first, so that we can start following up with relevant compiler developer group.

In parallel, we can continue reviewing the instruction for Eclipse IDE users in my fork (I updated).
https://github.com/itohro/eclipse-collections/wiki/How-to-build-Eclipse-Collections-in-Eclipse-IDE


motlin:
This pull request is to merge from itohro:ecj-workarounds into eclipse:master. I was expecting into eclipse:ecj-workarounds.

itohro:
Good cathch! My bad, let me re-submit..

PR #92, opened 2016-04-18
Implement synchronized BiMap. Closes #25.

### LINKED ISSUE #25: Implement synchronized BiMaps.
The API `MutableBiMap<K, V>.asSynchronized()` already exists, but [the implementation](https://github.com/eclipse/eclipse-collections/blob/7.0.0/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java#L385) in `AbstractMutableBiMap` is unsupported.

bhav0904:


eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
```diff
@@ -543,6 +545,10 @@ public MutableShortBag collectShort(ShortFunction<? super V> shortFunction)
         return this.delegate.collectIf(predicate, function);
     }
 
+    /**
+     * @deprecated in 7.1. Use {@link OrderedIterable#zip(Iterable)} instead.
+     */
+    @Deprecated
```
itohro: It looks odd to be included in the same commit. Could this be in a separate commit? Version number should be 8.0 now that 7.1 is already released. Also the link should be `zipWithIndex(Iterable)` not `zip(Iterable)` I think.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
```diff
@@ -558,6 +564,10 @@ public MutableShortBag collectShort(ShortFunction<? super V> shortFunction)
         return this.delegate.groupByEach(function, new UnifiedSetMultimap<VV, V>());
     }
 
+    /**
+     * @deprecated in 7.1. Use {@link OrderedIterable#zip(Iterable)} instead.
+     */
+    @Deprecated
```
itohro: Same as above, let's separate this to another commit. Update version# to 8.0, too.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
```diff
@@ -0,0 +1,482 @@
...
+    }
+
+    protected SynchronizedMutableBiMap(MutableBiMap<K, V> delegate, Object lock)
+    {
+        super(delegate, lock);
+    }
+
+    /**
+     * This method will take a MutableBiMap and wrap it directly in a SynchronizedMutableMap.  It will
+     * take any other non-GS-map and first adapt it will a MapAdapter, and then return a
```
itohro: Let's say "non-EC-map" instead of "non-GS-map".

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
```diff
@@ -0,0 +1,482 @@
...
+    public ImmutableBiMap<K, V> toImmutable()
+    {
+        synchronized (this.lock)
+        {
+            return BiMaps.immutable.withAll(this);
+        }
+    }
+
+    /**
+     * @deprecated in 7.1. Use {@link OrderedIterable#zip(Iterable)} instead.
```
itohro: Change version# to 8.0, and link to `zipWIthIndex()`.

itohro:
Similar to previous change, you can put your name in copyright section in the files you created. Also modify other files to "Goldman Sachs and others.".

bhav0904:
I worked on all your review comments, Hiroshi. But the ip validaton fails because it includes Craig's commit. We have to wait for @motlin to either cherry-pick my commit and push it along with his changes, or create a separate pull request for his changes.

bhav0904:
I'm going to include another commit in the same pull request that deprecates all the zip/zipWithIndex methods in the BiMap hierarchy. Will commit tomorrow.

itohro:
Thx, the latest changes look good to me.

If @motlin is also fine, we can merge it. We can manually validate your CLA so it is fine.

motlin:
I included the cherry-picked commit in #96. That's a big pull request but if it's possible to get that merged in the near future then we could rebase this on top, have a simple history, and no CLA issue. I think it's fine to merge in either order though.

itohro:
#96 has been merged. @bhav0904 , could you please rebase your 2 commits on top, then hopefully it resolves ip validation.

bhav0904:
The build failed due to a timeout issue. Is there a way we can re-run the build without pushing any code?

itohro:
Yes, there is a re-run button (a round arrow button) on the right top side of your build.

bhav0904:
I don't have the option, @itohro. Could you re-run it for me please?

itohro:
Ah ok, I didn't know it's only visible to committers. I've kicked off the build.

bhav0904:
Thanks. Looks like it has gone through!

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
```diff
@@ -0,0 +1,478 @@
...
+import org.eclipse.collections.impl.collection.mutable.SynchronizedMutableCollection;
+import org.eclipse.collections.impl.factory.BiMaps;
+import org.eclipse.collections.impl.list.fixed.ArrayAdapter;
+import org.eclipse.collections.impl.map.AbstractSynchronizedMapIterable;
+import org.eclipse.collections.impl.map.mutable.SynchronizedBiMapSerializationProxy;
+import org.eclipse.collections.impl.map.mutable.SynchronizedMapSerializationProxy;
+import org.eclipse.collections.impl.set.mutable.SynchronizedMutableSet;
+import org.eclipse.collections.impl.utility.LazyIterate;
+
+public class SynchronizedMutableBiMap<K, V> extends AbstractSynchronizedMapIterable<K, V> implements MutableBiMap<K, V>, Serializable
```
motlin: I think this can just be named SynchronizedBiMap. The fact that it is mutable is implied. The reason we don't follow that naming convention for SynchronizedMutableList, etc. is that there are already classes named SynchronizedList built into the JDK. This would follow the naming convention of SynchronizedBag for example.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
```diff
@@ -0,0 +1,478 @@
...
+    }
+
+    protected SynchronizedMutableBiMap(MutableBiMap<K, V> delegate, Object lock)
+    {
+        super(delegate, lock);
+    }
+
+    /**
+     * This method will take a MutableBiMap and wrap it directly in a SynchronizedMutableMap.  It will
+     * take any other non-EC-map and first adapt it will a MapAdapter, and then return a
```
motlin: This Javadoc is copied from elsewhere and not accurate for BiMaps.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
```diff
@@ -0,0 +1,478 @@
...
+            return SynchronizedMutableBiMap.of(this.getDelegate().clone());
+        }
+    }
+
+    public MutableBiMap<K, V> tap(Procedure<? super V> procedure)
+    {
+        synchronized (this.lock)
+        {
+            this.forEach(procedure);
+            return this;
```
motlin: I think every method should delegate if possible.

```
return this.getDelegate().tap(procedure)
```
This class should be largely auto-generated by code --> generate --> delegate methods --> getDelegate().

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
```diff
@@ -0,0 +1,478 @@
...
+        {
+            return this.getDelegate().newEmpty();
+        }
+    }
+
+    public MutableBiMap<V, K> inverse()
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().inverse();
```
motlin: This can't be safe. You can mutate this.delegate by just calling inverse() and mutating the inverse, no?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
```diff
@@ -389,7 +393,7 @@ public ShortIterable collectShort(ShortFunction<? super V> shortFunction)
     }
 
     /**
-     * @deprecated in 7.1. Use {@link OrderedIterable#zip(Iterable)} instead.
+     * @deprecated in 8.0. Use {@link OrderedIterable#zip(Iterable)} instead.
```
motlin: Don't understand this part. Wasn't the deprecation in 7.1?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
bhav0904: This method doesn't delegate in the way you coded it is because it has to return a MutableBiMap. If we return this.getDelegate().tap(procedure), it will return a non-synchronized copy of the map - isn't that unacceptable? I delegated all methods with the exception of the ones that return "this". For example, withKeyValue(). 

Still, we can change this to:

this.getDelegate().tap(procedure);
return this;

Do you agree?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
bhav0904: Agreed. I am going to change it to:

`return new SynchronizedBiMap<V, K>(this.getDelegate().inverse());`

Is that ok?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
bhav0904: I coded this back when 7.1 wasn;t released, and I committed it that way. Later I changed everything to be 8.0. SynchronizedBiMap didn't exist in 7.1, and neither did the deprecation of these methods in the BiMap heirarchy.

bhav0904:
I have the changes ready, will push them once I get the OK from @motlin on each of my comments. Thanks!

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
motlin: Right, that makes sense.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
motlin: I think it has to share the lock too.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedMutableBiMap.java
motlin: Ok I understand now. The problem is that you have two commits in this pull request which aren't independent. The first one sets the deprecation to 7.1, which is where @itohro commented here https://github.com/eclipse/eclipse-collections/pull/92/commits/f1d2938ed7946f72bb35e3a120407b5252fc25fc#diff-1ae117ff066180d4b3d50fbfc12db191R392. The second commit changes it to 8.0. For a clean history, 7.1 should never appear in any commit.

bhav0904:
I've made the changes that @motlin suggested.

itohro:
Confirmed all changes are reflected.

PR #93, opened 2016-04-20
Implement synchronized Multimaps. Closes #22.

### LINKED ISSUE #21: Implement unmodifiable Multimaps.
Implement unmodifiable Multimap wrappers like `UnmodifiableMutableListMultimap`, `UnmodifiableMutableSetMultimap`, etc.

Implement `MutableMultimap<K, V> asUnmodifiable();` in `MutableMultimap<K, V>` and covariant overrides like `MutableListMultimap<K, V> asUnmodifiable();` in `MutableListMultimap`.

All interfaces that extend `MutableMultimap` should have an override, including
- `MutableBagIterableMultimap`
- `MutableBagMultimap`
- `MutableListMultimap`
- `MutableSetIterableMultimap`
- `MutableSetMultimap`
- `MutableSortedBagMultimap`
- `MutableSortedSetMultimap`

The overall design and inheritance hierarchy of unmodifiable `MutableMultimaps` should be similar to the unmodifiable `MutableCollection` hierarchy.

### LINKED ISSUE #22: Implement synchronized Multimaps.
Implement synchronized Multimap wrappers like `SynchronizedMutableListMultimap`, `SynchronizedMutableSetMultimap`, etc.

Implement `MutableMultimap<K, V> asSynchronized();` in `MutableMultimap<K, V>` and covariant overrides like `MutableListMultimap<K, V> asSynchronized();` in `MutableListMultimap`.

All interfaces that extend `MutableMultimap` should have an override, including
- `MutableBagIterableMultimap`
- `MutableBagMultimap`
- `MutableListMultimap`
- `MutableSetIterableMultimap`
- `MutableSetMultimap`
- `MutableSortedBagMultimap`
- `MutableSortedSetMultimap`

The overall design and inheritance hierarchy of synchronized `MutableMultimaps` should be similar to the synchronized `MutableCollection` hierarchy.

g-votte:
Implement synchronized Multimaps. Closes #22 .

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/multimap/MutableMultimap.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/multimap/MutableMultimap.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Goldman Sachs.
+ * Copyright (c) 2016 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -99,4 +99,20 @@ public interface MutableMultimap<K, V>
 
     @Override
     <V2> MutableMultimap<K, V2> collectValues(Function<? super V, ? extends V2> function);
+
+    /**
+     * Returns a synchronized (thread-safe) multimap backed by this multimap.  In order to guarantee serial access,
+     * it is critical that <strong>all</strong> access to the backing multimap is accomplished through the returned
+     * multimap.
+     * <p>
+     * The returned multimap does <i>not</i> pass the <tt>hashCode</tt> and <tt>equals</tt> operations through to the
+     * backing multimap, but relies on <tt>Object</tt>'s equals and hashCode methods.  This is necessary to preserve
+     * the contracts of these operations.
+     * <p>
+     * The returned multimap will be serializable if this multimap is serializable.
+     *
+     * @return a synchronized view of this multimap.
+     * @since 8.0
+     */
+    MutableMultimap<K, V> asSynchronized();
 }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Multimaps.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Multimaps.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Goldman Sachs.
+ * Copyright (c) 2016 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -12,16 +12,19 @@ package org.eclipse.collections.impl.factory;
 
 import java.util.Comparator;
 
+import org.eclipse.collections.api.multimap.Multimap;
 import org.eclipse.collections.api.multimap.bag.ImmutableBagMultimap;
 import org.eclipse.collections.api.multimap.bag.MutableBagMultimap;
 import org.eclipse.collections.api.multimap.list.ImmutableListMultimap;
 import org.eclipse.collections.api.multimap.list.MutableListMultimap;
 import org.eclipse.collections.api.multimap.set.ImmutableSetMultimap;
 import org.eclipse.collections.api.multimap.set.MutableSetMultimap;
+import org.eclipse.collections.api.multimap.sortedbag.MutableSortedBagMultimap;
 import org.eclipse.collections.api.multimap.sortedset.ImmutableSortedSetMultimap;
 import org.eclipse.collections.api.multimap.sortedset.MutableSortedSetMultimap;
 import org.eclipse.collections.impl.multimap.bag.HashBagMultimap;
 import org.eclipse.collections.impl.multimap.bag.ImmutableBagMultimapImpl;
+import org.eclipse.collections.impl.multimap.bag.sorted.mutable.TreeBagMultimap;
 import org.eclipse.collections.impl.multimap.list.FastListMultimap;
 import org.eclipse.collections.impl.multimap.list.ImmutableListMultimapImpl;
 import org.eclipse.collections.impl.multimap.set.ImmutableSetMultimapImpl;
@@ -297,6 +300,7 @@ public final class Multimaps
         public final MutableSetMultimapFactory set = new MutableSetMultimapFactory();
         public final MutableSortedSetMultimapFactory sortedSet = new MutableSortedSetMultimapFactory();
         public final MutableBagMultimapFactory bag = new MutableBagMultimapFactory();
+        public final MutableSortedBagMultimapFactory sortedBag = new MutableSortedBagMultimapFactory();
 
         private MutableMultimaps()
         {
@@ -361,6 +365,11 @@ public final class Multimaps
                 fastListMultimap.put(key3, value3);
                 return fastListMultimap;
             }
+
+            public <K, V> MutableListMultimap<K, V> withAll(Multimap<? extends K, ? extends V> multimap)
+            {
+                return new FastListMultimap<K, V>(multimap);
+            }
         }
 
         public static final class MutableSetMultimapFactory
@@ -422,6 +431,11 @@ public final class Multimaps
                 unifiedSetMultimap.put(key3, value3);
                 return unifiedSetMultimap;
             }
+
+            public <K, V> MutableSetMultimap<K, V> withAll(Multimap<? extends K, ? extends V> multimap)
+            {
+                return new UnifiedSetMultimap<K, V>(multimap);
+            }
         }
 
         public static final class MutableSortedSetMultimapFactory
@@ -477,6 +491,11 @@ public final class Multimaps
                 treeSortedSetMultimap.put(key2, value2);
                 treeSortedSetMultimap.put(key3, value3);
                 return treeSortedSetMultimap;
+            }
+
+            public <K, V> MutableSortedSetMultimap<K, V> withAll(Multimap<? extends K, ? extends V> multimap)
+            {
+                return new TreeSortedSetMultimap<K, V>(multimap);
             }
         }
 
@@ -538,6 +557,57 @@ public final class Multimaps
                 hashBagMultimap.put(key2, value2);
                 hashBagMultimap.put(key3, value3);
                 return hashBagMultimap;
+            }
+
+            public <K, V> MutableBagMultimap<K, V> withAll(Multimap<? extends K, ? extends V> multimap)
+            {
+                return new HashBagMultimap<K, V>(multimap);
+            }
+        }
+
+        public static final class MutableSortedBagMultimapFactory
+        {
+            private MutableSortedBagMultimapFactory()
+            {
+            }
+
+            public <K, V> MutableSortedBagMultimap<K, V> empty()
+            {
+                return TreeBagMultimap.newMultimap();
+            }
+
+            public <K, V> MutableSortedBagMultimap<K, V> with()
+            {
+                return this.empty();
+            }
+
+            public <K, V> MutableSortedBagMultimap<K, V> with(K key, V value)
+            {
+                TreeBagMultimap<K, V> treeBagMultimap = TreeBagMultimap.newMultimap();
+                treeBagMultimap.put(key, value);
+                return treeBagMultimap;
+            }
+
+            public <K, V> MutableSortedBagMultimap<K, V> with(K key1, V value1, K key2, V value2)
+            {
+                TreeBagMultimap<K, V> treeBagMultimap = TreeBagMultimap.newMultimap();
+                treeBagMultimap.put(key1, value1);
+                treeBagMultimap.put(key2, value2);
+                return treeBagMultimap;
+            }
+
+            public <K, V> MutableSortedBagMultimap<K, V> with(K key1, V value1, K key2, V value2, K key3, V value3)
+            {
+                TreeBagMultimap<K, V> treeBagMultimap = TreeBagMultimap.newMultimap();
+                treeBagMultimap.put(key1, value1);
+                treeBagMultimap.put(key2, value2);
+                treeBagMultimap.put(key3, value3);
+                return treeBagMultimap;
+            }
+
+            public <K, V> MutableSortedBagMultimap<K, V> withAll(Multimap<? extends K, ? extends V> multimap)
+            {
+                return new TreeBagMultimap<K, V>(multimap);
             }
         }
     }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
@@ -1,0 +1,402 @@
+/*
+ * Copyright (c) 2016 Shotaro Sano.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompany this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.impl.multimap;
+
+import java.util.Collection;
+import java.util.Set;
+
+import net.jcip.annotations.GuardedBy;
+import org.eclipse.collections.api.RichIterable;
+import org.eclipse.collections.api.bag.Bag;
+import org.eclipse.collections.api.block.function.Function;
+import org.eclipse.collections.api.block.function.Function0;
+import org.eclipse.collections.api.block.function.Function2;
+import org.eclipse.collections.api.block.predicate.Predicate2;
+import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.block.procedure.Procedure2;
+import org.eclipse.collections.api.map.MutableMap;
+import org.eclipse.collections.api.multimap.Multimap;
+import org.eclipse.collections.api.multimap.MutableMultimap;
+import org.eclipse.collections.api.set.SetIterable;
+import org.eclipse.collections.api.tuple.Pair;
+import org.eclipse.collections.impl.bag.mutable.SynchronizedBag;
+import org.eclipse.collections.impl.map.mutable.SynchronizedMutableMap;
+import org.eclipse.collections.impl.set.mutable.SynchronizedMutableSet;
+import org.eclipse.collections.impl.set.mutable.UnmodifiableMutableSet;
+import org.eclipse.collections.impl.utility.LazyIterate;
+
+public abstract class AbstractSynchronizedMultimap<K, V> implements MutableMultimap<K, V>
+{
+    @GuardedBy("this.lock")
+    protected final MutableMultimap<K, V> delegate;
+    private final Object lock;
+
+    protected AbstractSynchronizedMultimap(MutableMultimap<K, V> multimap, Object newLock)
+    {
+        if (multimap == null)
+        {
+            throw new IllegalArgumentException("Cannot create a AbstractSynchronizedMultimap on a null multimap");
+        }
+
+        this.delegate = multimap;
+        this.lock = newLock == null ? this : newLock;
+    }
+
+    protected AbstractSynchronizedMultimap(MutableMultimap<K, V> multimap)
+    {
+        this(multimap, null);
+    }
+
+    protected MutableMultimap<K, V> getDelegate()
+    {
+        return this.delegate;
+    }
+
+    protected Object getLock()
+    {
+        return this.lock;
+    }
+
+    protected Object writeReplace()
+    {
+        return new SynchronizedMultimapSerializationProxy<>(this.getDelegate());
+    }
+
+    @Override
+    public boolean equals(Object obj)
+    {
+        synchronized (this.lock)
+        {
+            return this.delegate.equals(obj);
+        }
+    }
+
+    @Override
+    public int hashCode()
+    {
+        synchronized (this.lock)
+        {
+            return this.delegate.hashCode();
+        }
+    }
+
+    @Override
+    public String toString()
+    {
+        synchronized (this.lock)
+        {
+            return this.delegate.toString();
+        }
+    }
+
+    @Override
+    public boolean put(K key, V value)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().put(key, value);
+        }
+    }
+
+    @Override
+    public boolean add(Pair<K, V> keyValuePair)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().add(keyValuePair);
+        }
+    }
+
+    @Override
+    public boolean remove(Object key, Object value)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().remove(key, value);
+        }
+    }
+
+    @Override
+    public boolean putAllPairs(Pair<K, V>... pairs)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().putAllPairs(pairs);
+        }
+    }
+
+    @Override
+    public boolean putAllPairs(Iterable<Pair<K, V>> pairs)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().putAllPairs(pairs);
+        }
+    }
+
+    @Override
+    public boolean putAll(K key, Iterable<? extends V> values)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().putAll(key, values);
+        }
+    }
+
+    @Override
+    public <KK extends K, VV extends V> boolean putAll(Multimap<KK, VV> multimap)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().putAll(multimap);
+        }
+    }
+
+    @Override
+    public void clear()
+    {
+        synchronized (this.lock)
+        {
+            this.getDelegate().clear();
+        }
+    }
+
+    @Override
+    public boolean isEmpty()
+    {
+        synchronized (this.lock)
+        {
+            return this.delegate.isEmpty();
+        }
+    }
+
+    @Override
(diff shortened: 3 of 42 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Multimaps.java
```diff
@@ -265,6 +269,11 @@ private MutableListMultimapFactory()
                 fastListMultimap.put(key3, value3);
                 return fastListMultimap;
             }
+
+            public <K, V> MutableListMultimap<K, V> withAll(Multimap<K, V> multimap)
+            {
+                return new FastListMultimap(multimap);
```
itohro: Think this should be `return new FastListMultimap<K, V>(multimap);`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
```diff
@@ -0,0 +1,239 @@
...
+
+    public <R extends Collection<V>> MutableMap<K, R> toMap(Function0<R> collectionFactory)
+    {
+        synchronized (this.lock)
+        {
+            return this.delegate.toMap(collectionFactory);
+        }
+    }
+
+    protected Multimap getDelegate()
```
itohro: Return type should be `Multimap<K, V>`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Multimaps.java
motlin: In addition, I think the generics of the parameter should be `Multimap<? extends K, ? extends V> multimap` to match the generics on the constructor that the method delegates to.

eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Multimaps.java
```diff
@@ -383,6 +402,57 @@ private MutableBagMultimapFactory()
...
+            }
+        }
+
+        public static final class MutableSortedBagMultimapFactory
+        {
+            private MutableSortedBagMultimapFactory()
+            {
+            }
+
+            public <K, V> MutableSortedBagMultimap<K, V> empty()
```
motlin: Doesn't necessarily need to be in this PR, but this `empty()` method reminded me that for sorted structures, it's common to have an overload that takes a `Comparator`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
```diff
@@ -0,0 +1,239 @@
...
+import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.block.procedure.Procedure2;
+import org.eclipse.collections.api.map.MutableMap;
+import org.eclipse.collections.api.multimap.Multimap;
+import org.eclipse.collections.api.set.SetIterable;
+import org.eclipse.collections.api.tuple.Pair;
+
+public abstract class AbstractSynchronizedMultimap<K, V> implements Multimap<K, V>
+{
+    protected final Object lock;
```
itohro: If you look at the 6 children of this class, they are directly accessing this.lock. Think they should be accessing via method this.getLock() as similar to other synchronized classes. 

I suspect the declaration of lock object should be private to begin with. However other synchronized class parents are also declaring lock object as protected, while their children are accessing lock object via getLock() method. I think this was some historical defect which has been carried with copy/paste.

Maybe @motlin can confirm if my understanding is correct.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/bag/SynchronizedPutHashBagMultimap.java
```diff
@@ -161,6 +161,11 @@ public void value(K key, MutableBag<V> bag)
         return this.collectValues(function, HashBagMultimap.<K, V2>newMultimap());
     }
 
+    public MutableBagMultimap<K, V> asSynchronized()
+    {
+        return SynchronizedMutableBagMultimap.of(this);
```
motlin: This is an interesting case. `SynchronizedPutHashBagMultimap.asSynchronized()` returning a wrapper with a second lock.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/bag/sorted/mutable/SynchronizedMutableSortedBagMultimap.java
```diff
@@ -0,0 +1,210 @@
...
+
+    public MutableSortedBag<V> removeAll(Object key)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().removeAll(key);
+        }
+    }
+
+    public void clear()
```
motlin: I think a method like `clear()` which is on `MutableMultimap` could be pulled up higher in the inheritance hierarchy.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
motlin: There are two protected fields, the lock and the delegate. We access the delegate through methods so that we can cast it to a more specific type. It may be possible to use generics instead, and make the type of `delegate` some `T extends MutableMultimap<K,V>`. I'm not sure if that's possible, or desirable since it increases complexity. But if possible, I think it would be fine to get rid of the getters completely.

itohro:
There have been number of changes merged into master since last time. Please rebase and squash your commits.

Also please add a draft release note in `RELEASE_NOTE_DRAFT.md` file.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/list/SynchronizedPutFastListMultimap.java
```diff
@@ -170,4 +170,9 @@ public void value(K key, MutableList<V> list)
     {
         return this.collectValues(function, FastListMultimap.<K, V2>newMultimap());
     }
+
+    public MutableListMultimap<K, V> asSynchronized()
+    {
+        return SynchronizedMutableListMultimap.of(FastListMultimap.newMultimap(this));
```
itohro: Guess this line was changed after @motlin 's comment? Actually if you look inside `SynchronizedMutableListMultimap.of()`, this is already converted to non-synchronized list multimap.

```java
return new SynchronizedMutableListMultimap<K, V>(Multimaps.mutable.list.withAll(multimap));
```

Not sure if it was already like this when @motlin commented or it got changed afterwards. Anyways I suppose this line doesn't need to to wrap with `FastListMultimap.newMultimap()` anymore.

Same cases for other SynchronizedPut* classes.

itohro:
Again, please add a draft release note in `RELEASE_NOTE_DRAFT.md` file.

It looks good to me overall besides the parts I commented.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/list/SynchronizedPutFastListMultimap.java
motlin: I'm still not sure about this. Calling SynchronizedPutFastListMultimap.asSynchronized() is weird, and we need to think about it carefully. Usually you'd have one because you just called iterable.asParallel(...).groupBy(...). If you call asSynchronized() and it takes O(n) time, it defeats the purpose of whatever came earlier.  Maybe the best solution is to throw Unsupported. Not sure yet.

itohro:
Thanks for the update. I think throwing Unsupported makes sense for now.

Can you handle 2 minor changes? 

- Get rid of an extra space in `RELEASE_NOTE_DRAFT.md`
- Now that throwing Unsupported,  `asSynchronized()` in SynchronizedPut*Multimap can be pulled up to `SynchronizedPutUnifiedSetMultimap`

One these are done, I think this is good to be merged unless @motlin has any other concerns.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/multimap/MutableMultimap.java
```diff
@@ -80,4 +80,20 @@
 
     @Override
     <V2> MutableMultimap<K, V2> collectValues(Function<? super V, ? extends V2> function);
+
+    /**
+     * Returns a synchronized (thread-safe) collection backed by this collection.  In order to guarantee serial access,
+     * it is critical that <strong>all</strong> access to the backing collection is accomplished through the returned
```
motlin: Use the word `multimap` instead of `collection` throughout. These types don't extend `java.util.Collection`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Multimaps.java
```diff
@@ -217,6 +220,7 @@ private ImmutableBagMultimapFactory()
         public final MutableSetMultimapFactory set = new MutableSetMultimapFactory();
         public final MutableSortedSetMultimapFactory sortedSet = new MutableSortedSetMultimapFactory();
         public final MutableBagMultimapFactory bag = new MutableBagMultimapFactory();
+        public final MutableSortedBagMultimapFactory sortedBag = new MutableSortedBagMultimapFactory();
```
motlin: This change is great, but isn't described by the commit message. I'd recommend splitting the commit into two, and covering both in the release notes draft.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
```diff
@@ -0,0 +1,239 @@
...
+import org.eclipse.collections.api.bag.Bag;
+import org.eclipse.collections.api.block.function.Function0;
+import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.block.procedure.Procedure2;
+import org.eclipse.collections.api.map.MutableMap;
+import org.eclipse.collections.api.multimap.Multimap;
+import org.eclipse.collections.api.set.SetIterable;
+import org.eclipse.collections.api.tuple.Pair;
+
+public abstract class AbstractSynchronizedMultimap<K, V> implements Multimap<K, V>
```
motlin: I think this should implement `MutableMultimap<K, V>`. There's no such thing as a synchronized wrapper for a non-mutable container.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMutableMultimap.java
```diff
@@ -0,0 +1,151 @@
...
+package org.eclipse.collections.impl.multimap;
+
+import org.eclipse.collections.api.block.function.Function;
+import org.eclipse.collections.api.block.function.Function2;
+import org.eclipse.collections.api.block.predicate.Predicate2;
+import org.eclipse.collections.api.multimap.Multimap;
+import org.eclipse.collections.api.multimap.MutableMultimap;
+import org.eclipse.collections.api.tuple.Pair;
+
+public abstract class AbstractSynchronizedMutableMultimap<K, V>
```
motlin: The code here should just be merged into the superclass.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/bag/sorted/mutable/SynchronizedMutableSortedBagMultimap.java
```diff
@@ -0,0 +1,202 @@
...
+    {
+        super(newMultiMap);
+    }
+
+    public SynchronizedMutableSortedBagMultimap(MutableSortedBagMultimap<K, V> newMultimap, Object newLock)
+    {
+        super(newMultimap, newLock);
+    }
+
+    protected Object writeReplace()
```
motlin: All the `writeReplace()` methods are identical. It could be pulled up to a super class.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/bag/sorted/mutable/SynchronizedMutableSortedBagMultimap.java
```diff
@@ -0,0 +1,202 @@
...
+
+public class SynchronizedMutableSortedBagMultimap<K, V>
+        extends AbstractSynchronizedMutableMultimap<K, V>
+        implements MutableSortedBagMultimap<K, V>, Serializable
+{
+    private static final long serialVersionUID = 1L;
+
+    public SynchronizedMutableSortedBagMultimap(MutableSortedBagMultimap<K, V> newMultiMap)
+    {
+        super(newMultiMap);
```
motlin: Multimap instead of MultiMap.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/multimap/MutableMultimap.java
```diff
@@ -80,4 +80,20 @@
...
+
+    /**
+     * Returns a synchronized (thread-safe) multimap backed by this multimap.  In order to guarantee serial access,
+     * it is critical that <strong>all</strong> access to the backing multimap is accomplished through the returned
+     * multimap.
+     * <p>
+     * The returned multimap does <i>not</i> pass the <tt>hashCode</tt> and <tt>equals</tt> operations through to the
+     * backing multimap, but relies on <tt>Object</tt>'s equals and hashCode methods.  This is necessary to preserve
+     * the contracts of these operations in the case that the backing multimap is a set or a list.
+     * <p>
```
motlin: This bit of Javadoc is confusing "backing multimap is a set or list"

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
```diff
@@ -0,0 +1,399 @@
...
+        }
+    }
+
+    @Override
+    public SetIterable<K> keySet()
+    {
+        synchronized (this.lock)
+        {
+            return this.delegate.keySet();
+        }
```
motlin: All of the various views need to be synchronized. Take a look at `SynchronizedMutableMap.keySet()` as an example.

nikhilnanivadekar:
Serialization tests are missing for the synchronized multimaps.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/bag/AbstractMutableBagMultimap.java
```diff
@@ -129,4 +129,10 @@ public void putOccurrences(K key, V value, int occurrences)
             this.addToTotalSize(occurrences);
         }
     }
+
+    @Override
+    public MutableBagMultimap<K, V> asSynchronized()
+    {
+        return SynchronizedBagMultimap.of(this);
```
nikhilnanivadekar: This is O(n) operation, the convention is that if a method is as<Something>() it should be O(1).

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/bag/SynchronizedBagMultimap.java
```diff
@@ -0,0 +1,208 @@
...
+     * This method will take a Multimap and wrap it directly in a SynchronizedMutableMultimap.
+     */
+    public static <K, V, M extends Multimap<K, V>> SynchronizedBagMultimap<K, V> of(M multimap)
+    {
+        if (multimap == null)
+        {
+            throw new IllegalArgumentException("cannot create a SynchronizedMutableMultimap for null");
+        }
+
+        return new SynchronizedBagMultimap<K, V>(Multimaps.mutable.bag.withAll(multimap));
```
nikhilnanivadekar: You can use diamond operator instead of `new SynchronizedBagMultimap<K, V>`
This can be changed to `new SynchronizedBagMultimap<>`

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/bag/SynchronizedPutHashBagMultimap.java
```diff
@@ -175,4 +175,10 @@ public void putOccurrences(K key, V value, int occurrences)
             }
         }
     }
+
+    @Override
+    public MutableBagMultimap<K, V> asSynchronized()
+    {
+        throw new UnsupportedOperationException("Cannot call asSynchronized() on SynchronizedPutMultimap");
```
nikhilnanivadekar: To keep the exception consistent it should be:
`throw new UnsupportedOperationException("Cannot call asSynchronized() on " + this.getClass().getSimpleName())`

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/bag/sorted/mutable/SynchronizedSortedBagMultimap.java
```diff
@@ -0,0 +1,211 @@
...
+        super(newMultimap);
+    }
+
+    public SynchronizedSortedBagMultimap(MutableSortedBagMultimap<K, V> newMultimap, Object newLock)
+    {
+        super(newMultimap, newLock);
+    }
+
+    /**
+     * This method will take a Multimap and wrap it directly in a SynchronizedMutableMultimap.
```
nikhilnanivadekar: I think we should not mention _wrap_ here as a new MutableMultimap will be created.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/bag/sorted/mutable/SynchronizedSortedBagMultimap.java
```diff
@@ -0,0 +1,211 @@
...
+import org.eclipse.collections.impl.factory.Multimaps;
+import org.eclipse.collections.impl.multimap.AbstractSynchronizedMultimap;
+
+public class SynchronizedSortedBagMultimap<K, V>
+        extends AbstractSynchronizedMultimap<K, V>
+        implements MutableSortedBagMultimap<K, V>, Serializable
+{
+    private static final long serialVersionUID = 1L;
+
+    public SynchronizedSortedBagMultimap(MutableSortedBagMultimap<K, V> newMultimap)
```
nikhilnanivadekar: The input parameter should be called `multimap`

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
```diff
@@ -0,0 +1,399 @@
...
+
+    protected Object writeReplace()
+    {
+        return new SynchronizedMultimapSerializationProxy<K, V>(this.getDelegate());
+    }
+
+    @Override
+    public boolean equals(Object obj)
+    {
+        synchronized (this.lock)
```
nikhilnanivadekar: To keep usage consistent `this.lock` should be replaced by `this.getLock()`. Same comment for `this.delegate`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
```diff
@@ -0,0 +1,399 @@
...
+    @Override
+    public <V2, R extends MutableMultimap<K, V2>> R collectValues(Function<? super V, ? extends V2> function, R target)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().collectValues(function, target);
+        }
+    }
+
+    protected MutableMultimap<K, V> getDelegate()
```
nikhilnanivadekar: Please move this to the top. 
Above `getLock()` around line 51 will be an appropriate location. This is to maintain the order of methods.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
itohro: This is already somewhat discussed earlier in this thread (though not clearly concluded). If we were to clean-up, it's rather to get rid of getter method from relevant classes. It might be possible but it's a bit too widespread change for this PR so I think we can leave it and it can be addressed separately.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/bag/SynchronizedBagMultimap.java
motlin: To clarify Nikhil's previous comment, the implementations of `asSynchronized()` are fine, but the implementation of `of()` is not because it's `of()` which is written in an O(n) way. We should not create a copy of the delegate multimap, and we don't need flexible generics here (`M extends Multimap`). A SynchronizedBagMultimap should wrap a `MutableBagMultimap` only. If a user wants to convert from a list multimap to a bag multimap they already can, and it's unrelated to synchronized wrappers.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/bag/SynchronizedBagMultimap.java
```diff
@@ -0,0 +1,208 @@
...
+     * Additionally, a developer specifies which lock to use with the collection.
+     */
+    public static <K, V, M extends Multimap<K, V>> SynchronizedBagMultimap<K, V> of(M multimap, Object lock)
+    {
+        if (multimap == null)
+        {
+            throw new IllegalArgumentException("cannot create a SynchronizedMutableMultimap for null");
+        }
+
+        return new SynchronizedBagMultimap<K, V>(Multimaps.mutable.bag.withAll(multimap), lock);
```
motlin: Same here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
g-votte: @motlin @nikhilnanivadekar @itohro 

Thanks a lot for your comments!
I actually have 2 questions. Would you mind advising me on them?

(1) Whether should #keySet return **_SynchronizedMutableSet_** or **_UnmodifiableMutableSet_**?
The former might be better according to the comment by @motlin , whereas **_Multimap_**’s javadoc mentions the result should be an unmodifiable one. As well, returning a synchronized one breaks **_AbstractMultimapTestCase#keySet_**, as we can see in the current Travis status (build #422.4).

(2) At the same lines as (1), is there any way to assume O(1) calculation? The problem is: **_Multimap#keySet_** returns a **_SetIterable_** while we’d need to pass a **_Set_** to get a **_SynchronizedMutableSet_**. I think this gap causes us a detour casting like **_SynchronizedMutableSet.of(this.getDelegate().keySet().toSet())_**, which does NOT assume O(1).

itohro: Thanks for making a good progress on this. 

As for (1), I suggest we modify `AbstractMultimapTestCase#keySet` to check it throw UnsupportedOperationException when being mutated, instead of asserting the keySet being instance of Unmodifiable. I think you can just make the delegate's keySet unmodifiable to make it pass. 

(2) looks to be a bit too much to address in this change... I suggest we compromise O(1) requirement for synchronized multimaps keySet and make them O(N) for now (with documenting that in Javadoc). I'd suggest to raise another issue to improve this performance, that requires a design discussion and potentially a good amount of refactoring. 

@motlin , @nikhilnanivadekar , what do you think?

motlin: keySet() must not take linear time. Why not just return the delegate keySet
wrapped in an unmodifiable wrapper and rewrapped in a synchronized wrapper
that shares a lock with the multimap?

itohro: The tricky part here is that the return type of `Multimap.keySet()` is supposed to be `SetIterable` (refer to issue #45). There is no unmodifiable wrapper or synchronized wrapper that takes `SetIterable` as an input parameter, so we'd need to implement ones in order to do so. I guess we can address this in a separate issue before 8.0 official release. It's just that I think it's too much for @g-votte to take up in this same issue.

motlin: Can you cast?

itohro: Good point and I think it'd work:) Better workaround than O(N).

@g-votte can you try and update?

g-votte: @motlin @itohro

Thanks for your input!

Casted #keySet result and wrap it in synchronized wrapper. The set also should be wrapped in unmodifiable one implicitly because of Multimap#keymap behavior. 

```java
@Override
public SetIterable<K> keySet()
{
    synchronized (this.lock)
    {
        return SynchronizedMutableSet.of((Set<K>) this.getDelegate().keySet(), this.lock);
    }
}
```

Replaced a line in AbstractMultimapTestCase#keySet so that it checks UnsupportedOperationException for mutation, instead of asserting keySet’s class.

```java
@Test
public void keySet()
{
    Multimap<String, Integer> multimap =
            this.newMultimapWithKeysValues("One", 1, "One", 1, "Two", 2, "Three", 3);
    Set<String> keySet = (Set<String>) multimap.keySet();
    Verify.assertThrows(UnsupportedOperationException.class, () -> keySet.add("Four"));
    Assert.assertEquals(Sets.mutable.of("One", "Two", "Three"), keySet);
}
```

motlin: Could you add assertions for removing from the keySet(), in addition to
adding to it?

g-votte: Added a line to assert removing.

```java
Verify.assertThrows(UnsupportedOperationException.class, () -> keySet.remove("Four"));
```

itohro:
Looks good to me overall.

Please add 2 changes to `RELEASE_NOTE_DRAFT.md`, then I think this is good to be merged.

itohro:
Merged the PR. Thanks @g-votte for your contribution!

PR #96, opened 2016-04-22
Fix several type signature problems due to missing overrides, incorrect return types, and generics problems.

motlin:
- Remove unnecessary overrides of anySatisfyWith, allSatisfyWith, and noneSatisfyWith from AbstractRichIterable.
- Remove unnecessary overrides of ParallelIterable.flatCollect().
- Change AbstractSynchronizedRichIterable.groupByUniqueKey() to return MapIterable instead of MutableMap.
- Change AbstractMutableMapIterable.groupByUniqueKey() to return MutableMapIterable instead of MutableMap.
- Change MutableMapIterable.aggregateBy() to return MutableMap instead of MutableMapIterable.
- Fix return types on MapIterable.collect(Function2) overrides.
- Fix return type of UnifiedSetWithHashingStrategy.groupByEach() method to preserve hashing strategy.
- Fix generics on Multimap.forEachKeyValue().
- Fix generics on Multimap.forEachKeyMultiValues().
- Add missing overrides for collect() in OrderedMaps.
- Add missing override UnsortedMapIterable.partitionWith().
- Add missing overrides of RichIterable.tap().
- Make StackIterable implement OrderedIterable.
- Make primitive-object maps more bag-like. Change the filtering and transformation methods to return bags since they contain duplicates and no meaningful order.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
@@ -267,7 +267,7 @@ public interface MutableMapIterable<K, V> extends MapIterable<K, V>, Map<K, V>
 
     MutableCollection<Pair<V, Integer>> zipWithIndex();
 
-    <KK, VV> MutableMapIterable<KK, VV> aggregateInPlaceBy(Function<? super V, ? extends KK> groupBy, Function0<? extends VV> zeroValueFactory, Procedure2<? super VV, ? super V> mutatingAggregator);
+    <KK, VV> MutableMap<KK, VV> aggregateInPlaceBy(Function<? super V, ? extends KK> groupBy, Function0<? extends VV> zeroValueFactory, Procedure2<? super VV, ? super V> mutatingAggregator);
 
-    <KK, VV> MutableMapIterable<KK, VV> aggregateBy(Function<? super V, ? extends KK> groupBy, Function0<? extends VV> zeroValueFactory, Function2<? super VV, ? super V, ? extends VV> nonMutatingAggregator);
+    <KK, VV> MutableMap<KK, VV> aggregateBy(Function<? super V, ? extends KK> groupBy, Function0<? extends VV> zeroValueFactory, Function2<? super VV, ? super V, ? extends VV> nonMutatingAggregator);
 }
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/primitive/ImmutablePrimitiveObjectMap.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/primitive/ImmutablePrimitiveObjectMap.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2016 Goldman Sachs.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompany this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.api.map.primitive;
+
+import org.eclipse.collections.api.bag.ImmutableBag;
+import org.eclipse.collections.api.bag.primitive.ImmutableBooleanBag;
+import org.eclipse.collections.api.bag.primitive.ImmutableByteBag;
+import org.eclipse.collections.api.bag.primitive.ImmutableCharBag;
+import org.eclipse.collections.api.bag.primitive.ImmutableDoubleBag;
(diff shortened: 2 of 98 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/primitive/PrimitiveObjectMap.java
```diff
@@ -0,0 +1,135 @@
...
+import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.map.UnsortedMapIterable;
+import org.eclipse.collections.api.multimap.bag.BagMultimap;
+import org.eclipse.collections.api.ordered.OrderedIterable;
+import org.eclipse.collections.api.partition.bag.PartitionBag;
+import org.eclipse.collections.api.set.UnsortedSetIterable;
+import org.eclipse.collections.api.tuple.Pair;
+
+/**
+ * @since 8.0
```
itohro: Missing a dot here.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/primitive/ImmutablePrimitiveObjectMap.java
```diff
@@ -0,0 +1,104 @@
...
+import org.eclipse.collections.api.block.predicate.Predicate;
+import org.eclipse.collections.api.block.predicate.Predicate2;
+import org.eclipse.collections.api.block.procedure.Procedure2;
+import org.eclipse.collections.api.map.ImmutableMap;
+import org.eclipse.collections.api.multimap.bag.ImmutableBagMultimap;
+import org.eclipse.collections.api.ordered.OrderedIterable;
+import org.eclipse.collections.api.partition.bag.PartitionImmutableBag;
+import org.eclipse.collections.api.set.ImmutableSet;
+import org.eclipse.collections.api.tuple.Pair;
+
```
itohro: Maybe add `@since 8.0` javadoc here, too?

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/parallel/list/ListIterableParallelIterable.java
```diff
@@ -85,12 +85,6 @@ public T detect(Predicate<? super T> predicate)
     }
 
     @Override
-    public <V> ParallelListIterable<V> flatCollect(Function<? super T, ? extends Iterable<V>> function)
-    {
-        return new ParallelFlatCollectListIterable<T, V>(this, function);
-    }
-
```
itohro: This method used to instantiate `ParallelFlatCollectListIterable` while its super class is instantiating `ParallelFlatCollectIterable `. Is this intentional to change the behaviour?

itohro: Sorry, please ignore above. I was looking at the incorrect class. 

It's not changing the behaviour so this is ok.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -267,7 +267,7 @@
 
     MutableCollection<Pair<V, Integer>> zipWithIndex();
 
-    <KK, VV> MutableMapIterable<KK, VV> aggregateInPlaceBy(Function<? super V, ? extends KK> groupBy, Function0<? extends VV> zeroValueFactory, Procedure2<? super VV, ? super V> mutatingAggregator);
+    <KK, VV> MutableMap<KK, VV> aggregateInPlaceBy(Function<? super V, ? extends KK> groupBy, Function0<? extends VV> zeroValueFactory, Procedure2<? super VV, ? super V> mutatingAggregator);
 
-    <KK, VV> MutableMapIterable<KK, VV> aggregateBy(Function<? super V, ? extends KK> groupBy, Function0<? extends VV> zeroValueFactory, Function2<? super VV, ? super V, ? extends VV> nonMutatingAggregator);
+    <KK, VV> MutableMap<KK, VV> aggregateBy(Function<? super V, ? extends KK> groupBy, Function0<? extends VV> zeroValueFactory, Function2<? super VV, ? super V, ? extends VV> nonMutatingAggregator);
```
itohro: These methods are now referencing child interface `MutableMap` in return types. I guess this is a compromise we deliberately make?

itohro:
I think the changes with commit comment "Change ..." would be worth adding in draft release note.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
motlin: Yes, it's deliberate, and somewhat common already. For example, all the RichIterable.to*() methods like RichIterable.toList() are referencing a subtype. In this case, it's acknowledging that after collapsing map keys, any original ordering is lost.

motlin:
Fixed the Javadoc problems in 71c787c and added draft release notes in fab7822. Thanks @itohro!

motlin:
😄  thanks!

PR #97, opened 2016-04-25
Implement LazyIterable.takeWhile(Predicate) and LazyIterable.dropWhile(Predicate).

nikhilnanivadekar:
1) Implement LazyIterable.takeWhile(Predicate) and LazyIterable.dropWhile(Predicate).
2) Bug fix for Interval.take(0).
3) Fix warnings for Javadoc in LazyIterate where period was missing.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/LazyIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/LazyIterable.java
@@ -25,6 +25,7 @@ import org.eclipse.collections.api.block.function.primitive.ShortFunction;
 import org.eclipse.collections.api.block.predicate.Predicate;
 import org.eclipse.collections.api.block.predicate.Predicate2;
 import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.ordered.OrderedIterable;
 import org.eclipse.collections.api.tuple.Pair;
 
 /**
@@ -78,6 +79,18 @@ public interface LazyIterable<T>
      * Creates a deferred drop iterable for the current iterable using the specified count as the limit.
      */
     LazyIterable<T> drop(int count);
+
+    /**
+     * @see OrderedIterable#takeWhile(Predicate)
+     * @since 8.0
+     */
+    LazyIterable<T> takeWhile(Predicate<? super T> predicate);
+
+    /**
+     * @see OrderedIterable#dropWhile(Predicate)
+     * @since 8.0
+     */
+    LazyIterable<T> dropWhile(Predicate<? super T> predicate);
 
     /**
      * Creates a deferred distinct iterable to get distinct elements from the current iterable.
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/block/predicate/DropWhileIterablePredicate.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/block/predicate/DropWhileIterablePredicate.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2016 Goldman Sachs.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompany this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.impl.block.predicate;
+
+import org.eclipse.collections.api.block.predicate.Predicate;
+
+/**
+ * @since 8.0
+ */
+public class DropWhileIterablePredicate<T> implements Predicate<T>
+{
+    private static final long serialVersionUID = 1L;
+
+    private final Predicate<? super T> predicate;
+    private boolean doneDroppingElements;
+
+    public DropWhileIterablePredicate(Predicate<? super T> predicate)
+    {
+        this.predicate = predicate;
+    }
+
+    public boolean accept(T each)
+    {
+        if (!this.doneDroppingElements && !this.predicate.accept(each))
+        {
+            this.doneDroppingElements = true;
+        }
+        return this.doneDroppingElements;
+    }
+}
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/TakeWhileIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/TakeWhileIterable.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright (c) 2016 Goldman Sachs.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompany this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.impl.lazy;
+
+import java.util.Iterator;
+
+import org.eclipse.collections.api.block.predicate.Predicate;
+import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.block.procedure.Procedure2;
+import org.eclipse.collections.api.block.procedure.primitive.ObjectIntProcedure;
+import org.eclipse.collections.impl.lazy.iterator.TakeWhileIterator;
+
+/**
+ * Iterates over the elements of the adapted Iterable until the predicate returns false.
+ *
+ * @since 8.0
+ */
+public class TakeWhileIterable<T> extends AbstractLazyIterable<T>
+{
+    private final Iterable<T> adapted;
+    private final Predicate<? super T> predicate;
+
+    public TakeWhileIterable(Iterable<T> newAdapted, Predicate<? super T> predicate)
+    {
+        if (predicate == null)
+        {
+            throw new IllegalStateException("Predicate cannot be null");
+        }
+        this.adapted = newAdapted;
+        this.predicate = predicate;
+    }
+
+    public void each(Procedure<? super T> procedure)
+    {
+        for (T each : this.adapted)
+        {
+            if (!this.predicate.accept(each))
+            {
+                return;
+            }
+            procedure.value(each);
+        }
+    }
+
+    @Override
+    public void forEachWithIndex(ObjectIntProcedure<? super T> procedure)
+    {
+        int i = 0;
+        for (T each : this.adapted)
+        {
+            if (!this.predicate.accept(each))
+            {
+                return;
+            }
+            procedure.value(each, i);
+            i++;
+        }
+    }
+
+    @Override
+    public <P> void forEachWith(Procedure2<? super T, ? super P> procedure, P parameter)
+    {
+        for (T each : this.adapted)
+        {
+            if (!this.predicate.accept(each))
+            {
+                return;
+            }
+            procedure.value(each, parameter);
+        }
+    }
+
+    public Iterator<T> iterator()
+    {
+        return new TakeWhileIterator<T>(this.adapted, this.predicate);
+    }
+}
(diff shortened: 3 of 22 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/list/Interval.java
```diff
@@ -1063,7 +1063,7 @@ public Interval take(int count)
         {
             return Interval.fromToBy(this.from, this.locationAfterN(count - 1), this.step);
         }
-        return Interval.fromToBy(this.from, this.from, this.step);
```
motlin: What was the old behavior?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/LazyIterable.java
```diff
@@ -80,6 +80,22 @@
...
+     */
+    LazyIterable<T> takeWhile(Predicate<? super T> predicate);
+
+    /**
+     * Creates a deferred dropWhile iterable for the current iterable using the specified predicate.
+     * Short circuits at the first element which satisfies the Predicate.
+     *
+     * @since 8.0
+     */
+    LazyIterable<T> dropWhile(Predicate<? super T> predicate);
```
motlin: See OrderedIterable for example Javadoc.

eclipse-collections/src/main/java/org/eclipse/collections/impl/block/predicate/DropWhileIterablePredicate.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public boolean accept(T each)
+    {
+        if (!this.evaluate && !this.predicate.accept(each))
+        {
+            this.evaluate = true;
+        }
+        return this.evaluate;
+    }
```
motlin: This pattern of a cached boolean is only useful when *not* short circuiting.

eclipse-collections/src/main/java/org/eclipse/collections/impl/block/predicate/DropWhileIterablePredicate.java
```diff
@@ -0,0 +1,35 @@
...
+        this.predicate = predicate;
+    }
+
+    public boolean accept(T each)
+    {
+        if (!this.evaluate && !this.predicate.accept(each))
+        {
+            this.evaluate = true;
+        }
+        return this.evaluate;
```
motlin: Let's find a more meaningful name than `evaluate`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/TakeWhileIterable.java
```diff
@@ -0,0 +1,133 @@
...
+        while (accept && iterator.hasNext())
+        {
+            T next = iterator.next();
+            if (this.predicate.accept(next))
+            {
+                procedure.value(next);
+            }
+            else
+            {
+                accept = false;
```
motlin: You don't need a `while` loop or the accept variable. Just return here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/TakeWhileIterable.java
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+    }
+
+    @Override
+    public Object[] toArray()
+    {
+        Object[] temp = new Object[this.size()];
```
motlin: I think you should delete the override of toArray(). The class `TakeIterable` has an override of toArray() as an optimization, because we know that the size is likely to be `count`. In `TakeWhileIterable` we have no guess what the size will be, and we must always avoid calling `this.size()` on LazyIterables.

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/iterator/DropWhileIterator.java
```diff
@@ -0,0 +1,51 @@
...
+    private final Iterator<T> delegateIterator;
+
+    public DropWhileIterator(Iterable<T> iterable, Predicate<? super T> predicate)
+    {
+        this(iterable.iterator(), predicate);
+    }
+
+    public DropWhileIterator(Iterator<T> iterator, Predicate<? super T> predicate)
+    {
+        this.delegateIterator = new SelectIterator<T>(iterator, new DropWhileIterablePredicate<T>(predicate));
```
motlin: Since all methods delegate, why not inline this code into `DropWhileIterable.iterator()` and remove this class?

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/iterator/TakeWhileIterator.java
```diff
@@ -0,0 +1,77 @@
...
+
+package org.eclipse.collections.impl.lazy.iterator;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+import org.eclipse.collections.api.block.predicate.Predicate;
+
+/**
+ * Iterates over the first count elements of the iterator until predicate returns false.
```
motlin: Javadoc is pasted from another class and isn't accurate here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/iterator/TakeWhileIterator.java
```diff
@@ -0,0 +1,77 @@
...
+
+            this.terminate = true;
+            this.next = NULL;
+        }
+        return false;
+    }
+
+    public T next()
+    {
+        if (this.next != NULL || this.hasNext())
```
motlin: I think `this.next != NULL` is redundant here.

unit-tests-java8/src/test/java/org/eclipse/collections/test/lazy/DropWhileIterableTestNoIteratorTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.eclipse.collections.impl.block.factory.Predicates;
+import org.eclipse.collections.impl.lazy.DropWhileIterable;
+import org.eclipse.collections.impl.test.junit.Java8Runner;
+import org.eclipse.collections.test.IterableTestCase;
+import org.eclipse.collections.test.LazyNoIteratorTestCase;
+import org.eclipse.collections.test.list.mutable.FastListNoIterator;
+import org.junit.runner.RunWith;
+
+@RunWith(Java8Runner.class)
+public class DropWhileIterableTestNoIteratorTest implements LazyNoIteratorTestCase
```
motlin: The word Test appears two times.

unit-tests-java8/src/test/java/org/eclipse/collections/test/lazy/DropWhileIterableTestNoIteratorTest.java
```diff
@@ -0,0 +1,33 @@
...
+{
+    @Override
+    public <T> LazyIterable<T> newWith(T... elements)
+    {
+        MutableList<T> list = new FastListNoIterator<>();
+        IterableTestCase.addAllTo(elements, list);
+        return new DropWhileIterable<>(list, Predicates.alwaysFalse());
```
motlin: This follows the pattern of the DropIterableNoIteratorTest, so it's already ok. I think both could be improved by appending a copy of new Object() to the front of the list and then dropping it with a count of 1 or a predicate that compares the instance is the same.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/Interval.java
nikhilnanivadekar:
```
Interval interval = Interval.fromTo(1, 4);
Assert.assertEquals(Lists.immutable.empty(), interval.take(0));

```
Result:
`java.lang.AssertionError: `
Expected :[]
Actual   :Interval from: 1 to: 1 step: 1 size: 1

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/iterator/DropWhileIterator.java
nikhilnanivadekar: I can do that. 
Since it is a major version, can I deprecate `DropIterator`?

motlin: Sure. It's also inlineable?

unit-tests-java8/src/test/java/org/eclipse/collections/test/lazy/DropWhileIterableTestNoIteratorTest.java
nikhilnanivadekar: This follows the same convention as `DropIterableTestNoIteratorTest`. 
I will rename it to `DropIterableNoIteratorTest`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/iterator/DropWhileIterator.java
nikhilnanivadekar: Yes. `DropWhileIterator` follows the exact same pattern as `DropIterator`. The deprecation would just re-direct to use:

`new SelectIterator<T>(iterable.iterator(), new DropIterablePredicate<T>(count))`

eclipse-collections/src/main/java/org/eclipse/collections/impl/block/predicate/DropWhileIterablePredicate.java
nikhilnanivadekar: How about `accept` or `result`?

motlin: It's used to determine whether we're done dropping elements and should *not* evaluate the predicate. We could keep the name evaluate if we negated every usage. How about doneDroppingElements?

nikhilnanivadekar: Sounds good.

nikhilnanivadekar:
I have fixed all the review comments. All the builds are passing.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/Interval.java
motlin: I think you should clarify in the release notes that the old behavior was to return an interval of size one. And move the note to the top because it's the first behavior change.

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/TakeWhileIterable.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Iterator;
+
+import org.eclipse.collections.api.block.predicate.Predicate;
+import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.block.procedure.Procedure2;
+import org.eclipse.collections.api.block.procedure.primitive.ObjectIntProcedure;
+import org.eclipse.collections.impl.lazy.iterator.TakeWhileIterator;
+
+/**
+ * Iterates over the elements of the adapted Iterable until the predicate returns false.
```
motlin: Completely new classes should get a Javadoc @since tag.

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/TakeWhileIterable.java
```diff
@@ -0,0 +1,92 @@
...
+        {
+            if (this.predicate.accept(each))
+            {
+                procedure.value(each);
+            }
+            else
+            {
+                return;
+            }
+        }
```
motlin: My style preference would be

```java
if (!this.predicate.accept(each))
{
    return;
}

procedure.value(each);
```

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/Interval.java
nikhilnanivadekar: Done. Also added a note saying that it is a behavior breaking change.

nikhilnanivadekar:
Fixed the new review comments.

unit-tests-java8/src/test/java/org/eclipse/collections/test/lazy/DropWhileIterableNoIteratorTest.java
```diff
@@ -0,0 +1,35 @@
...
+public class DropWhileIterableNoIteratorTest implements LazyNoIteratorTestCase
+{
+    @Override
+    public <T> LazyIterable<T> newWith(T... elements)
+    {
+        MutableList<T> list = new FastListNoIterator<>();
+        T object = (T) new Object();
+        list.add(object);
+        IterableTestCase.addAllTo(elements, list);
+        return new DropWhileIterable<>(list, each -> each == object);
```
motlin: Perfect. I feel like this gives us good coverage.

nikhilnanivadekar:
Rebased.

PR #98, opened 2016-04-25
Deprecate or hide collection factory constructors. Closes #26.

### LINKED ISSUE #26: Deprecate or hide collection factory constructors.
Clarify that classes like collection factory classes are not meant to be instantiated more than once. The current API allows you to get an empty immutable list like this.

``` java
ImmutableList<Object> list1 = new ImmutableListFactoryImpl().empty();
```

The better way to get an empty immutable list is like this.

``` java
ImmutableList<Object> list2 = Lists.immutable.empty();
```

Both examples return the singleton instance of the empty immutable list. In other words, `list1` and `list2` are the same instance. The second example creates no garbage, making the first example seem especially wasteful. `Lists.immutable` is intended to hold the only instance of `ImmutableListFactoryImpl` in the JVM.

For all collection factory classes, change them to implement the singleton pattern, or clarify that they aren't meant to be instantiated by deprecating their constructors and adding Javadoc links to the classes like `Lists` that should be used instead.

If this is handled through deprecation, it can be done in a minor release. Changing the collection factories to implement the singleton pattern would be a [compatibility breaking change](https://github.com/eclipse/eclipse-collections/blob/master/CONTRIBUTING.md#semantic-versioning).

bhav0904:


motlin:
Please rebase onto master.

bhav0904:
Done.

motlin:
I slightly prefer the deprecation approach to the hiding approach. But if we're hiding, why not use enums?

```
public enum MutableListFactoryImpl implements MutableListFactory
{
    INSTANCE;

...
}
```

bhav0904:
Since it was unlikely that any user used the constructors directly, I thought that it was cleaner to just hide them. I agree that the enum approach is better than explicitly making the constructors private. Will change it to enums and re-submit.

bhav0904:
Made them enums.

motlin:
The code looks good to me. Could you add a note to the draft release notes markdown file? It's technically not backwards compatible so it deserves a short explanation.

motlin:
There are a few errors in the CheckStyle build that look easy to fix - just trailing whitespace.

nikhilnanivadekar:
@bhav0904 can you please update, rebase and push? I would like to review and close this out for 9.0.0 release. Thanks!

bhav0904:
Ok, I will do it tonight.

bhav0904:
Sorry for the delay, I rebased onto master. Let me know if I need to do anything else

nikhilnanivadekar:
@bhav0904 can you please squash the 2 commits. One of them is purely whitespace changes, so dropping that would be fine.
I would expect the *Impls to be package private and there by hidden. Am I missing something?

bhav0904:
I squashed the commits. This was a long time ago and I don't remember all the details, but @motlin and I had a discussion whether to make the constructors private, or just make the classes enums. We didn't talk about making it package private.
Since the only usage is now within the same package, I suppose they could be made package private. I will do it as soon as I can. When is 9.0 scheduled?

donraab:
September 6th we are aiming to be feature complete.

https://github.com/eclipse/eclipse-collections/wiki/Roadmap

nikhilnanivadekar:
In this PR (it can be different commit) I would want to make the implementations package private. Idea is that users should not be able to do `new MutableListFactoryImpl`
By hiding the constructors this is achieved, however, for completeness I would want it to be made package private. It is similar to our Immutable* implementations, where the users should not care/know about the concrete implementation.

nikhilnanivadekar:
@bhav0904 if you can resolve the merge conflicts and push, I will merge it. We can hide the factory constructors as a different PR. It would be good to have this for 9.0.0.

nikhilnanivadekar:
@bhav0904 can you please allow updates from maintainers on this PR? I have resolved the merge conflicts.
Refer to last few points on this page: https://help.github.com/articles/creating-a-pull-request-from-a-fork/
It would be good to have this for Milestone 4. Thanks!

nikhilnanivadekar:
Will be fixed via #336.

PR #99, opened 2016-04-25
Upgrade compilation source/target to JDK 8.

itohro:
- Update pom files, ant files and Eclipse/IntelliJ project files
- Remove version specific logic from `ConcurrentMutableHashMap`
- Fix findbugs error in `Comparators` after the upgrade


acceptance-tests/build.xml
```diff
@@ -13,8 +13,8 @@
 <project name="acceptance-tests" default="test" basedir="." xmlns:ivy="antlib:org.apache.ivy.ant">
     <property name="src.dir" location="src/main/java" />
     <property name="testsrc.dir" location="src/test/java" />
-    <property name="source.level" value="1.5" />
-    <property name="target.level" value="1.5" />
+    <property name="source.level" value="1.8" />
```
motlin: Properties like this no longer vary per module, so ideally they'd be defined only once (once each for ant and maven).

eclipse-collections-code-generator-ant/eclipse-collections-code-generator-ant.iml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module org.jetbrains.idea.maven.project.MavenProjectsManager.isMavenModule="true" type="JAVA_MODULE" version="4">
-  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_5" inherit-compiler-output="false">
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_8" inherit-compiler-output="false">
```
motlin: Similarly, in IntelliJ, we should now configure the language level to be 8 at the project level, and define each module to use the project language level.

eclipse-collections/src/main/java/org/eclipse/collections/impl/block/factory/Comparators.java
```diff
@@ -268,7 +268,7 @@ public int compare(T value1, T value2)
 
         public int compare(SortedSetIterable<T> setA, SortedSetIterable<T> setB)
         {
-            int compareTo = Integer.valueOf(setA.size()).compareTo(setB.size());
+            int compareTo = Integer.compare(setA.size(), setB.size());
```
motlin: This change is cool. Did a static analysis tool catch it? Was it an Inspection?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentMutableHashMap.java
```diff
@@ -176,10 +175,6 @@ public V remove(Object key)
 
     public Set<Entry<K, V>> entrySet()
     {
-        if ("1.5".equals(JAVA_SPECIFICATION_VERSION))
-        {
-            return new SafeEntrySetAdapter<K, V>(this.delegate.entrySet());
```
motlin: Could you delete `SafeEntrySetAdapter` after removing this usage?

motlin:
Made a few suggestions but overall this is an exciting PR that I've been looking forward to for a long time.

 😄

eclipse-collections/src/main/java/org/eclipse/collections/impl/block/factory/Comparators.java
itohro: Yes, it was findbugs actually.

itohro:
Thanks @motlin for the suggestions! I've reflected them.

PR #103, opened 2016-05-07
Enforce maven version to 3.1.0+ and add supplimentary information to CONTRIBUTING.md.

itohro:
Reflect couple of feedback from contributors.
- It is reported that build fails with maven 3.0.x. Explicitly mention minimum working version 3.1.0.
- It is reported that findbugs build fails following the instruction. Made it sync with Travis CI findbugs build option.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,10 @@ Contact us
 
 Build
 -----
+The Eclipse Collections build requires below as dependencies.
+
+- Java 1.8
+- Maven 3.1.0+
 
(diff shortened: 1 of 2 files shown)


motlin:
I think that #57 broke compatibility with older versions of maven. One option is to revert it, and then maven 3.0 will just work again. Another option is to use [maven prerequisites](https://maven.apache.org/pom.html#Prerequisites) to enforce 3.1.0 or newer.

itohro:
Thanks for the suggestion.

I tested maven prerequisites but somehow it didn't fail properly. I used maven enforcer plugin instead, that worked perfectly.

Issue #88, opened 2016-04-13
Implement RichIterable.getOnly().


motlin:
The implementation should be similar to `IterableIterate.getOnly()` but be available on the interface `RichIterable.` If the iterable has exactly one item, return it. Otherwise, throw `IllegalArgumentException` with a short explanation.

itohro:
@opengl-8080 showed an interest to contribute to this issue, so I'd reserve this issue for him. 

@opengl-8080, feel free to post questions in Japanese. I can translate here.

opengl-8080:
Thank you for providing opportunity of contribution.

I have a question about work of this issue.

My understanding is following.

- Add `getOnly()` method into `RichIterable` interface.
- Implement `getOnly()` method at some sub classes(`UnmodifiableRichIterable`, `SynchronizedRichIterable`, `AbstractRichIterable`, and more...).
    - I can't implement `getOnly()` method as default method in `RichIterable`. Because language level of `eclipse-collections` module is 1.5.
- Add test method for `getOnly()` method into `RichITerableTestCase`.

Is this understanding correct?

Thank you.

donraab:
Yes, you are correct.  We need to upgrade the language level of all of the modules to 1.8.  I had a similar problem this evening when I tried implementing summarizingInt/Long/Float/Double on RichIterable using default methods.

itohro:
Yes, we'll upgrade language level to 1.8 shortly.

Afterwards you could use default method in RichIterable. 

@opengl-8080, your basic understanding is correct. Along the way you'd most likely need to modify more files (e.g. stg files for primitive collections, other test files for test variants etc.), but you'll find those out while you build your project and run tests.

opengl-8080:
@donraab, @itohro
Thank you for your reply!


@itohro
> e.g. stg files for primitive collections, other test files for test variants etc.

Does it means following?

For Example, I add `getOnly()` method into `RichIterable`. Then `ImmutableByteObjectEmptyMap` become compile error.
Therefore I need modify 2 files (`immutablePrimitiveObjectEmptyMap.stg`, `immutablePrimitiveObjectEmptyMapTest.stg`).

itohro:
Yes, that's right.

motlin:
[...]

PR #104, opened 2016-05-07
Implement RichIterable.getOnly(). #88

opengl-8080:
Signed-off-by: opengl-8080 <tomcat.port.8080+eclipse-community@gmail.com>

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
@@ -1652,6 +1652,35 @@ public class <name>ObjectHashMap\<V> implements Mutable<name>ObjectMap\<V>, Exte
         return null;
     }
 
+    public V getOnly()
+    {
+        if (this.size() != 1)
+        {
+            throw new IllegalStateException("Size must be 1 but was " + this.size());
+        }
+
+        if (this.sentinelValues != null)
+        {
+            if (this.sentinelValues.containsZeroKey)
+            {
+                return this.sentinelValues.zeroValue;
+            }
+            if (this.sentinelValues.containsOneKey)
+            {
+                return this.sentinelValues.oneValue;
+            }
+        }
+        for (int i = 0; i \< this.keys.length; i++)
+        {
+            if (isNonSentinel(this.keys[i]))
+            {
+                return this.values[i];
+            }
+        }
+
+        return null;
+    }
+
     public MutableList\<V> toList()
     {
         MutableList\<V> list = Lists.mutable.of();
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/AbstractRichIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/AbstractRichIterable.java
@@ -763,4 +763,15 @@ public abstract class AbstractRichIterable<T> implements RichIterable<T>
         this.forEach(new GroupByUniqueKeyProcedure<>(target, function));
         return target;
     }
+
+    @Override
+    public T getOnly()
+    {
+        if (this.size() == 1)
+        {
+            return this.getFirst();
+        }
+
+        throw new IllegalStateException("Size must be 1 but was " + this.size());
+    }
 }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableArrayBag.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableArrayBag.java
@@ -246,6 +246,22 @@ public class ImmutableArrayBag<T>
     }
 
     @Override
+    public T getOnly()
+    {
+        if (this.counts.length == 0)
+        {
+            throw new IllegalStateException("Size must be 1 but was 0");
+        }
+
+        if (this.counts.length > 1 || this.counts[0] > 1)
+        {
+            throw new IllegalStateException("Size must be 1 but was greater than 1");
+        }
+
+        return this.getFirst();
+    }
+
+    @Override
     public ImmutableBag<T> select(Predicate<? super T> predicate)
     {
         return this.select(predicate, HashBag.newBag()).toImmutable();
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/AbstractLazyIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/AbstractLazyIterable.java
@@ -11,6 +11,7 @@
 package org.eclipse.collections.impl.lazy;
 
 import java.util.Collection;
+import java.util.Iterator;
 
 import net.jcip.annotations.Immutable;
 import org.eclipse.collections.api.LazyBooleanIterable;
@@ -124,6 +125,25 @@ public abstract class AbstractLazyIterable<T>
         T[] result = (T[]) new Object[1];
         this.each(each -> result[0] = each);
         return result[0];
+    }
+
+    @Override
+    public T getOnly()
+    {
+        Iterator<T> iterator = this.iterator();
+
+        if (!iterator.hasNext())
+        {
+            throw new IllegalStateException("Size must be 1 but was 0");
+        }
+
+        T result = iterator.next();
+        if (iterator.hasNext())
+        {
+            throw new IllegalStateException("Size must be 1 but was greater than 1");
+        }
+
+        return result;
     }
 
     @Override
--- a/unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
+++ b/unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
@@ -272,6 +272,26 @@ public interface RichIterableTestCase extends IterableTestCase
     }
 
     @Test
+    default void RichIterable_getOnly()
+    {
+        RichIterable<Integer> iterable = this.newWith(3);
+        Integer only = iterable.getOnly();
+        assertThat(only, is(3));
+
+        Iterator<Integer> iterator = iterable.iterator();
+        assertThat(iterator.next(), is(only));
+        assertThat(iterator.hasNext(), is(false));
+
+        assertThrows(IllegalStateException.class, () -> this.newWith().getOnly());
+        assertThrows(IllegalStateException.class, () -> this.newWith(1, 2).getOnly());
+
+        if (this.allowsDuplicates())
+        {
+            assertThrows(IllegalStateException.class, () -> this.newWith(1, 1).getOnly());
+        }
+    }
+
+    @Test
     default void RichIterable_getFirst_and_getLast()
     {
         RichIterable<Integer> iterable = this.newWith(3, 2, 1);
--- a/RELEASE_NOTE_DRAFT.md
+++ b/RELEASE_NOTE_DRAFT.md
@@ -10,6 +10,7 @@ New Functionality
 * Implemented MutableMultimap.putAllPairs(Iterable<Pair<K, V>> keyValuePairs).
 * Implemented LazyIterable.takeWhile(Predicate<? super T> predicate) and LazyIterable.dropWhile(Predicate<? super T> predicate).
 * Implemented unmodifiable BiMaps.
+* Implemented RichIterable.getOnly().
 * Pull up into() from LazyIterable to RichIterable.
 * Made StackIterable implement OrderedIterable.
 * Implemented summarizeInt/Float/Long/Double({Int/Float/Long/Double}Function) on RichIterable
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
@@ -129,6 +129,15 @@ public interface RichIterable<T>
     T getLast();
 
     /**
+     * Returns the element if the iterable has exactly one element. Otherwise, throw {@link IllegalStateException}.
+     *
+     * @return an element of an iterable.
+     * @throws IllegalStateException if iterable is empty or has multiple elements.
+     * @since 8.0
+     */
+    T getOnly();
+
+    /**
      * Returns true if the iterable has an element which responds true to element.equals(object).
      *
      * @since 1.0
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutablePrimitiveObjectEmptyMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutablePrimitiveObjectEmptyMap.stg
@@ -228,6 +228,11 @@ final class Immutable<name>ObjectEmptyMap\<V> implements Immutable<name>ObjectMa
         return null;
     }
 
+    public V getOnly()
+    {
+        throw new IllegalStateException("Size must be 1 but was " + this.size());
+    }
+
     public boolean contains(Object object)
     {
         return false;
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutablePrimitiveObjectHashMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutablePrimitiveObjectHashMap.stg
@@ -203,6 +203,11 @@ final class Immutable<name>ObjectHashMap\<V> extends AbstractImmutable<name>Obje
         return this.delegate.getLast();
     }
 
+    public V getOnly()
+    {
+        return this.delegate.getOnly();
+    }
+
     public boolean contains(Object object)
     {
         return this.delegate.contains(object);
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutablePrimitiveObjectSingletonMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutablePrimitiveObjectSingletonMap.stg
@@ -308,6 +308,11 @@ final class Immutable<name>ObjectSingletonMap\<V> extends AbstractImmutable<name
         return this.value1;
     }
 
+    public V getOnly()
+    {
+        return this.value1;
+    }
+
     public boolean contains(Object object)
     {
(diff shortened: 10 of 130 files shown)


itohro:
Thanks for your submission! It is a pretty great first step.

Actually there are still number of places that can be optimised. For example, fixed size collections (e.g. `SingletonList`, `DoubletonList` etc.) could return `this.element1` and so on.

The easiest step to follow is to find the string "T getFirst()" with global search (Ctrl + Shift + F in IntelliJ), then find out the files that are not changed yet (e.g. not blue colored).  That's where `getOnly()` is also most likely added.

Also search "void getFirst()" to find out the potential places to change in unit tests, too.

One minor suggestion is to change commit comment to "Implement RichIterable.getOnly(). Closes #88" so that it closes the issue.

opengl-8080:
Thank you for your review!

I will try to do your advice.
And, I will change commit comment.

opengl-8080:
I tried to search `T getFirst()`. Then, I found classes to need to modify.

I modified fixed size collections.
At singleton collections, they always return singleton value.
At empty or fixed multiple values collections, they always throw exception.

However, I found other patterns like followings.
Should I modify these classes too?

### 1 - Random access
```java
    @Override
    public T getFirst()
    {
        return this.isEmpty() ? null : this.items[0];
    }
```

- `AbstractArrayAdapter`
- `AbstractMutableList.SubList`
- `FastList`
- `ImmutableTreeSet`

Is it faster than to use Iterator?

### 2 - For loop
```java
    @Override
    public T getFirst()
    {
        for (int i = 0; i < this.table.length; i++)
        {
            Object cur = this.table[i];
            if (cur instanceof ChainedBucket)
            {
                return this.nonSentinel(((ChainedBucket) cur).zero);
            }
            if (cur != null)
            {
                return this.nonSentinel(cur);
            }
        }
        return null;
    }
```

- `UnifiedSet`
- `UnifiedSetWithHashingStrategy`
- `UnifiedMap`
- `UnifiedMapWithHashingStrategy`
- `CodePointAdapter`

Is it faster than to use Iterator?

### 3 - Use util class
```java
    @Override
    public T getFirst()
    {
        return ListIterate.getFirst(this);
    }
```

- `AbstractMutableList`
- `ImmutableArrayBag`
- `LazyIterableAdapter`
- `ImmutableSortedBagImpl`
- `CollectIterable`

It seems that using util class is difficult.
Because, util class (`IterableIterate`) throws `IllegalArgumentException` when iterable is empty or it has multiple values.
However, `getOnly()` method must throw `IllegalStateException`.

So, I guess these classes need not to be modified.
Because, these `getOnly()` methods will become to same as `AbstractRichIterable.getOnly()`.


### 4 - Delegate to keysView
```java
    @Override
    public T getFirst()
    {
        return this.getKeysView().getFirst();
    }
```

- `AbstractMapIterable`
- `AbstractMutableBagIterable`

Is it faster than to use Iterator?


### 5 - Iterate.detect()
```java
    @Override
    public T getFirst()
    {
        return Iterate.detect(this.adapted, this.predicate);
    }
```

- `SelectInstancesOfIterable`
- `AbstractLazyIterable`
- `RejectIterable`
- `SelectIterable`
- `TapIterable`

Is it faster than to use Iterator?

### 6 - Using implementation of sub class
```java
    @Override
    public T getFirst()
    {
        return this.first();
    }
```

- `AbstractImmutableSortedSet`

I think this class need not be modified.

`this.first()` is an abstract method.
It is defined at `java.util.SortedSet`.

Sub classes of `AbstractImmutableSortedSet` are `ImmutableTreeSet` and `ImmutableEmptySortedSet`.

`ImmutableTreeSet` is a pattern of "1 - Random access".
`ImmutableEmptySortedSet` is fixed size collection, so I modified it to always throwing exception.


### 7 - Delegate
```java
    public T getFirst()
    {
        return this.delegate.getFirst();
    }
```

- `CodePointList`
- `ImmutableHashBag`
- `SortedSetAdapter`
- `ImmutableUnifiedSet`
- `ImmutableUnifiedSetWithHashingStrategy`
- `ReverseIterable`
- `TreeSortedSet`

I think these classes should be modified.
Because, if delegate is fixed size classes, it will be faster than using iterator.


Thank you.

itohro:
Thanks for your thorough analysis!

Please see my comments.

1. It's faster to use index access for RandomAccess cases. Refer to [RandomAccess JavaDoc](http://docs.oracle.com/javase/7/docs/api/java/util/RandomAccess.html)
2. It's not easy to tell if these are faster than using iterator, but I suppose they are. One minor difference is that iterator would create a garbage. (If you're interested you could quickly try experimenting with adding tests under JMH module) 
3. Agree `AbstractMutableList` may not need to be changed. But some implementation that uses `ArrayIterate.getFirst()` are more optimal than iterator.
4. Did you add test to `MapIterableTestCase`? I think `AbstractMapIterable` needs to be changed at least. I cannot easily tell `AbstractMutableBagIterable`. 
5. It is more optimal to use Iterate.detect() in case backing iterable is RandomAccess data structure. 
6. Agree for `ImmutableEmptySortedSet`, but `ImmutableTreeSet` needs to be changed for the same RandomAccess reason.
7. Yes.

Considering above, `getOnly()` should pretty much follow what `getFirst()` has been doing that are already correct and optimal. So instead of duplicating the work in the classes above, it could be as simple as delegating to `getFirst()` in `AbstractRichIterable`. You can check if `size()` is 1 and return `getFirst()`. Otherwise throw `IllegalStateException`. This way the amount of change could be significantly reduced, and implementation is even optimal.

motlin:
You should never call the size() method on LazyIterable where it's O(n) so
I don't like to call size() in AbstractRichIterable if we can avoid it.

opengl-8080:
@itohro
Thank you for your answer.
I'm sorry for the delay in my response.

> So instead of duplicating the work in the classes above, it could be as simple as delegating to `getFirst()` in `AbstractRichIterable`.

In JJUG CCC, I suggested you that I implement `getOnly()` at `AbstractRichIterable` with `size()`, `getFirst()` and I override `getOnly()` at `AbstractLazyIterable` without `size()`.
However, I realized that it is not fine. Because `getFirst()` method is deprecated in `AbstractRichIterable`.
(May I use `getFirst()` in `AbstractRichIterable`?)


So I guess that `getOnly()` should be implemented in each sub classes.

In class that has fast `size()` method, I implement `getOnly()` method with `size()` and `getFirst()` method.

> `AbstractArrayAdapter`, `AbstractMutableList.SubList`, `FastList`, `ImmutableTreeSet`, `UnifiedSet`, `UnifiedSetWithHashingStrategy`, `UnifiedMap`, `UnifiedMapWithHashingStrategy`, `ImmutableSortedBagImpl`


In class that has slow `size()` method, I will not implement `getOnly()` (Using `AbstractRichIterable.getOnly()`. It is implemented by `iterator()`).

> `AbstractMutableList`, `ImmutableArrayBag`, `LazyIterableAdapter`, `ImmutableSortedBagImpl`, `CollectIterable`, `AbstractMapIterable`, `AbstractMutableBagIterable`, `SelectInstancesOfIterable`, `AbstractLazyIterable`, `RejectIterable`, `SelectIterable`, `TapIterable`


But, if the class is randam accessable I override `getOnly()` with for loop.

> `CodePointAdapter`, `ImmutableArrayBag`

motlin:
getFirst() is deprecated at the top level because not all subclasses are
ordered, so there may not be a meaningful first element. If the size is
one, then order isn't relevant. The first, last, and only element are all
the same.

Even though getFirst() is deprecated in places, we have no plans to delete
it. It's ok to use.

opengl-8080:
@motlin
Thank you!
I accept it.

I try to implement `getOnly()` with `size()` and `getFirst()` in `AbstractRichIterable`.
Then I override `getOnly()` in sub classes that have slow `size()` method (`LazyIterableAdapter`, etc...).

opengl-8080:
I modified pull request.
Please review my code again.

eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
```diff
@@ -1652,6 +1652,35 @@ public class <name>ObjectHashMap\<V> implements Mutable<name>ObjectMap\<V>, Exte
...
+        for (int i = 0; i \< this.keys.length; i++)
+        {
+            if (isNonSentinel(this.keys[i]))
+            {
+                return this.values[i];
+            }
+        }
+
+        throw new IllegalStateException("Iterable is empty.");
+    }
```
motlin: If throwing different exceptions for the cases of empty and multiple values, then both checks should be at the beginning of the method. Otherwise we can spend O(n) time checking for the empty case.

eclipse-collections/src/main/java/org/eclipse/collections/impl/AbstractRichIterable.java
```diff
@@ -763,4 +763,22 @@ public boolean containsAll(Collection<?> collection)
...
+        {
+            throw new IllegalStateException("Iterable is empty.");
+        }
+
+        if (size > 1)
+        {
+            throw new IllegalStateException("Iterable has multiple values.");
+        }
+
+        return this.getFirst();
```
motlin: An exception like this would require fewer lines and would provide more information - the size.

```
        if (this.size() == 1)
        {
            return this.getFirst();
        }
        throw new IllegalStateException("Size must be 1 but was " + this.size());
```

Another advantage is that we wouldn't be using the word iterable to refer to specific subclasses.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableArrayBag.java
```diff
@@ -246,6 +246,25 @@ public T getLast()
     }
 
     @Override
+    public T getOnly()
+    {
+        Iterator<T> iterator = this.iterator();
+
```
motlin: ImmutableArrayBag extends AbstractRichIterable, so it already has an implementation of getOnly(), as most collections do. If we're going to override that implementation, it should be to avoid the use of an iterator. ImmutableArrayBag.getFirst() doesn't use an iterator, so we can probably avoid it here too.

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/AbstractLazyIterable.java
```diff
@@ -127,6 +128,25 @@ public T getLast()
...
+            throw new IllegalStateException("Iterable is empty.");
+        }
+
+        T result = iterator.next();
+        if (iterator.hasNext())
+        {
+            throw new IllegalStateException("Iterable has multiple values.");
+        }
+
+        return result;
```
motlin: This is the correct implementation for AbstractLazyIterable, where we don't want to call size(). Besides here, we should try to eliminate all usages of iterators.

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
```diff
@@ -272,6 +272,32 @@ default void RichIterable_getLast()
...
+        RichIterable<Object> emptyIterable = this.newWith();
+        assertThrows(IllegalStateException.class, () -> emptyIterable.getOnly());
+    }
+
+    @Test
+    default void RichIterable_getOnly_exception_when_two_items()
+    {
+        RichIterable<Object> iterable = this.newWith(1, 2);
+        assertThrows(IllegalStateException.class, () -> iterable.getOnly());
+    }
```
motlin: These tests look good. I think all three could be combined into a single method, RichIterable_getOnly(). Are the assertions ever overridden separately?

It would be good to include an assertion to show that this.newWith(1, 1).getOnly() throws, with two copies of the same element for collections that support it.

eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
opengl-8080: For example, like following?

```java
public V getOnly()
{
    if (this.size() != 1)
    {
        throw new IllegalStateException("Size must be 1 but was " + this.size());
    }

    ...
```

eclipse-collections/src/main/java/org/eclipse/collections/impl/AbstractRichIterable.java
opengl-8080: I see. I modify all codes like this.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableArrayBag.java
opengl-8080: The reason why I override getOnly() at `ImmutableArrayBag`, I thought `size()` is O(n).

```java
    @Override
    public int size()
    {
        int sum = 0;
        for (int value : this.counts)
        {
            sum += value;
        }
        return sum;
    }
```

So I thought this class should not use `size()` method. Like `AbstractLazyIterable`.

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
opengl-8080: I'm sorry. I don't understand well...

> Are the assertions ever overridden separately?

What is the meaning of 'overriden separately'?
When I added test method in some test classes, I separated exception test from nromal test.


> It would be good to include an assertion to show that this.newWith(1, 1).getOnly() throws, with two copies of the same element for collections that support it.

Does this test case mean to avoid bug that is caused by like following implementation?
Call getFirst() and getLast(). Then compare these values. If they are same value return first value.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableArrayBag.java
itohro: That is correct, and @motlin 's suggestion for this particular class is to avoid using iterator as similar to how `ImmutableArrayBag.getFirst()` does.

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
itohro: The suggestion is to combine the 3 methods into a single method, unless the separated methods are overridden by subclass tests separately (supplementing this line with Japanese below to make it hopefully clearer).

もしこの例外系のテストメソッドが他のテストで個別に（= separately）オーバーライドされているのなら分けてもいいけど、そうでなければ三つをまとめて一つのメソッドでもいいんじゃないかという提案です。

As for suggestion to add `this.newWith(1, 1).getOnly()`, it is to test the different behaviour in collections, for example, FastList should throw for this case while UnifiedSet shouldn't throw.

eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
itohro: This looks good to me.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableArrayBag.java
opengl-8080: If each counts of `this.counts` are always greater than `0`, my idea is following.

```java
        if (this.counts.length == 1 && this.counts[0] == 1)
        {
            return this.getFirst();
        }

        throw new IllegalStateException("Size was not 1");
```

If `this.counts` sometimes have `0`, my idea is folloing.

```java
        int size = 0;

        for (int count : this.counts)
        {
            size += count;
            if (size > 1)
            {
                throw new IllegalStateException("Size must be 1 but was greater than 1");
            }
        }

        if (size == 0)
        {
            throw new IllegalStateException("Size must be 1 but was 0");
        }

        return this.getFirst();
```

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
opengl-8080: Thank you for your translation!

> The suggestion is to combine the 3 methods into a single method

I did not override each methods. So I combine 3 methods into a single method.

> add `this.newWith(1, 1).getOnly()`

Should I implement the test method like following?

1. I add a test method (named like `getOnly_when_multiple_same_values()`) at `RichIterableTestCase` and `AbstractRichIterableTestCase`.  
This test method verifies that  `getOnly()` throws `IllegalStateException`.
1. Run each tests then some classes will be failed (`UnifiedSetTest`, `HashBagTest` etc...).
1. I override each failed test methods. Then overriden methods verify that `getOnly()` must not throw `IllegalStateException`.

motlin: The reason to include a test like this.newWith(1, 1).getOnly() throws is for Bags, where a naive implementation like this.keysView().getOnly() would not throw.

You can do everything in a single test if you surround this particular assertion with if(this.allowsDuplicates()).

opengl-8080:
I modified code again.
Please review my code.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/mutable/TreeSortedSet.java
```diff
@@ -679,6 +679,25 @@ public T getLast()
...
+
+        T result = iterator.next();
+        if (iterator.hasNext())
+        {
+            throw new IllegalStateException("Size must be 1 but was greater than 1");
+        }
+
+        return result;
+    }
+
```
itohro: I think it's ok to call size() method in `TreeSortedSet` and this class wouldn't need to override the method.

itohro:
Thanks @opengl-8080 for your effort on this! 

Besides `TreeSortedSet` that I commented, overall the changes look good me and hopefully it is the last feedback unless @motlin has any other comment.

opengl-8080:
Thank you!
I removed `getOnly()` from `TreeSortedSet`.

itohro:
Merged the changes. Thank you @opengl-8080 for your contribution!!

PR #105, opened 2016-05-07
Add static method BooleanArrayList.newWithNValues

guoci:
Following code in eclipse-collections-code-generator/src/main/resources/impl/list/mutable/primitiveArrayList.stg

Signed-off-by: guoci <zguoci@gmail.com>

### DIFF
--- a/RELEASE_NOTE_DRAFT.md
+++ b/RELEASE_NOTE_DRAFT.md
@@ -3,7 +3,7 @@
 
 New Functionality
 -----------------
-
+* Added static method BooleanArrayList.newWithNValues, following other primitive ArrayLists.
 * Changed MutableBagIterable.addOccurrences(T item, int occurrences) to return the updated number of occurrences instead of void.
 * Implemented Bags.mutable.ofAll(Iterable<? extends T> items) and Bags.mutable.withAll(Iterable<? extends T> items).
 * Implemented Multimap.keySet() to return an unmodifiable SetIterable of keys.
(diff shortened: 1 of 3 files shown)


motlin:
Thanks for the contribution! A few comments and questions.

* I like the optimization so that it doesn't loop. I wonder if it's possible to implement the same optimization in the other primitive lists, with System.arrayCopy or something.
* I was wondering about the null check for the backing BitSet, and then I saw that we don't initialize the backing BitSet for size 0, which is consistent with other primitive lists.
* Other primitive lists throw when you call them with negative initial capacities, and I think BooleanArrayList has always been inconsistent that way. The result is that newWithNCopies with a negative number doesn't throw.
* You could add a Javadoc @since 8.0 tag, and you could add a short description in [RELEASE_NOTE_DRAFT.md](https://github.com/eclipse/eclipse-collections/blob/master/RELEASE_NOTE_DRAFT.md).

guoci:
> I like the optimization so that it doesn't loop. I wonder if it's possible to implement the same optimization in the other primitive lists, with System.arrayCopy or something.

seems possible, but should it be in another PR?
> I was wondering about the null check for the backing BitSet, and then I saw that we don't initialize the backing BitSet for size 0, which is consistent with other primitive lists.

Other primitive lists of size 0 point to singletons arrays of length 0, so they skip null checks.
> Other primitive lists throw when you call them with negative initial capacities, and I think BooleanArrayList has always been inconsistent that way. The result is that newWithNCopies with a negative number doesn't throw.

newWithNCopies does throw. Added a test for this.

> You could add a Javadoc @since 8.0 tag, and you could add a short description in RELEASE_NOTE_DRAFT.md.

Added in this PR.

motlin:
Excellent, thank you. Will wait for the build to finish and then will pull.

PR #108, opened 2016-05-13
Add summarizeInt/Float/Long/Double, reduce and reduceInPlace as default methods on RichIterable and create Collectors2 with Collector implementations for Eclipse Collections.

donraab:


eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -1564,6 +1574,88 @@
...
+    default Optional<T> reduce(BinaryOperator<T> accumulator)
+    {
+        if (this.isEmpty())
+        {
+            return Optional.empty();
+        }
+        T result = this.injectInto(null, (result1, each) ->
+                result1 == null ? each : accumulator.apply(result1, each));
+        return Optional.of(result);
+    }
```
motlin: The method `isEmpty()` can be O(n) on `LazyIterable`s so I think that `reduce()` should be rewritten to only call `each()` or should be left abstract and implemented in subclasses.

motlin:
Could you add information to [RELEASE_NOTE_DRAFT.md](https://github.com/eclipse/eclipse-collections/blob/master/RELEASE_NOTE_DRAFT.md) about the new features?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/AbstractBiMap.java
```diff
@@ -371,7 +371,7 @@ public void forEachKeyValue(Procedure2<? super K, ? super V> procedure)
     @Override
     public void each(Procedure<? super V> procedure)
     {
-        this.getInverse().forEachKey(procedure);
+        this.getDelegate().forEachValue(procedure);
     }
```
motlin: I thought this was handled separately - I don't remember where that commit lives. There's a second method that needs this change, `each()` and it's a bug-fix so I think it makes sense to fix in an earlier commit with dedicated tests.

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/CollectorJoiningTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+
+@State(Scope.Thread)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+public class CollectorJoiningTest extends AbstractJMHTestRunner
```
motlin: This test is fine though it seems unrelated to the commit message. Maybe it belongs in a separate commit?

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
```diff
@@ -885,12 +891,40 @@ default void RichIterable_iterationOrder()
...
+        });
+        assertEquals(expectedIterationOrder, summarizeLongOrder);
+
+        MutableCollection<Integer> summarizeDoubleOrder = this.newMutableForFilter();
+        this.getInstanceUnderTest().summarizeDouble(each -> {
+            summarizeDoubleOrder.add(each);
+            return 0;
+        });
+        assertEquals(expectedIterationOrder, summarizeDoubleOrder);
     }
```
motlin: The other methods (reduceInPlace) should be covered here too.

donraab: Yes, the other methods are covered in this test case as well.

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/CollectorJoiningTest.java
donraab: I added this test to benchmark any performance difference between reduceInPlace(Collector) and stream().collect(Collector) since the implementation is significantly simpler in the serial reduceInPlace case.  I may expand this benchmark if we implement reduceInPlace on ParallelIterable.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/AbstractBiMap.java
donraab: Yes, I believe you have the code and tests implemented for this.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
```diff
@@ -0,0 +1,70 @@
...
+                StringJoiner::toString,
+                EMPTY_CHARACTERISTICS);
+    }
+
+    public static <T> Collector<T, ?, MutableStack<T>> toStack()
+    {
+        return Collector.of(
+                Stacks.mutable::empty,
+                MutableStack::push,
+                (r1, r2) -> {r2.each(r1::push); return r1;},
```
motlin: Are the combiner portions of each Collector covered by tests? Most of them are obviously correct, but Stack is tricky so I want to be sure it's covered.

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
```diff
@@ -1590,13 +1591,25 @@ default void RichIterable_makeString_appendString()
...
+
+        assertEquals(
                 "[4/4/4/4/3/3/3/2/2/1]",
                 iterable.makeString("[", "/", "]"));
 
+        assertEquals(
+                iterable.makeString("[", "/", "]"),
+                iterable.reduceInPlace(makeString("[", "/", "]")));
+
         StringBuilder builder1 = new StringBuilder();
```
motlin: RichIterable_makeString_appendString is overridden in subclasses of RichIterableTestCase. I'd prefer if this were in a new test method, but it could stay in the same test as long as you also modify the overrides.

motlin: RichIterable_makeString_appendString is overridden in subclasses of RichIterableTestCase. I'd prefer if this were in a new test method, but it could stay in the same test as long as you also modify the overrides.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
```diff
@@ -50,12 +62,40 @@ private Collectors2()
...
+    }
+
+    public static <T> Collector<T, MutableList<T>, MutableStack<T>> toStack()
+    {
+        return Collector.of(
+                Lists.mutable::empty,
+                MutableList::add,
+                MutableList::withAll,
+                Stacks.mutable::ofAll,
                 EMPTY_CHARACTERISTICS);
```
motlin: I see you edited toStack() further and my previous comment doesn't apply anymore, but I'm still curious about test coverage.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
donraab: No, great catch.  I need to add some tests using parallelStream().collect() for each Collector.

PR #109, opened 2016-05-20
Fix the iteration order of BiMap.each() and BiMap.forEachValue().

motlin:


motlin:
@itohro could you take a look when you have a chance? #108 should be rebased on top of this because https://github.com/eclipse/eclipse-collections/pull/108/files#diff-b51209d7cd677a709d067dc038edab90L374 builds on the changes.

itohro:
LGTM.

PR #112, opened 2016-05-27
Change our functional interfaces to extend the JDK functional interfaces.

motlin:
The first commit changes the Object functional interfaces like Procedure to extend the JDK functional interfaces like Consumer.

The second commit handles primitive predicates. The work is not complete until we do all primitive functional interfaces but I'm hoping to get help with that effort.

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/block/predicate/primitivePredicate.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/block/predicate/primitivePredicate.stg
@@ -22,10 +22,19 @@ import java.io.Serializable;
  *
  * @since 3.0.
  */
+<if(primitive.specializedStream)>@FunctionalInterface<endif>
 public interface <name>Predicate
-        extends Serializable
+        extends <if(primitive.specializedStream)>java.util.function.<name>Predicate, <endif>Serializable
 {
     boolean accept(<type> value);
+<if(primitive.specializedStream)>
+
+    @Override
+    default boolean test(<type> value)
+    {
+        return this.accept(value);
+    }
+<endif>
 }
 
 >>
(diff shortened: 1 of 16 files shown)


findbugs-exclude.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <!--
-  ~ Copyright (c) 2016 Goldman Sachs and others.
+  ~ Copyright (c) 2016 Goldman Sachs.
```
itohro: Please keep " and others" here.

eclipse-collections-code-generator/src/main/resources/api/block/predicate/primitivePredicate.stg
```diff
@@ -23,8 +23,15 @@ import java.io.Serializable;
...
-        extends Serializable
+        extends <if(primitive.specializedStream)>java.util.function.<name>Predicate, <endif>Serializable
 {
+<if(primitive.specializedStream)>
+    default boolean test(<type> value)
+    {
+        return this.accept(value);
+    }
+
+<endif>
```
itohro: Very minor point, but non-primitive functional interfaces have the new default method after original abstract method. Maybe we want to make the method order consistent?

findbugs-exclude.xml
motlin: Thanks, fixing this manually. My IDE made this change based on the configuration in `.idea/copyright/Eclipse.xml` so we'll need to figure out how to prevent this from happening automatically.

eclipse-collections-code-generator/src/main/resources/api/block/predicate/primitivePredicate.stg
motlin: Good point. For consistency, I should also have the @FunctionalInterface and @Override annotations.

PR #113, opened 2016-05-31
Extract MutableIterable and ImmutableIterable.

motlin:
We already have `MutableCollection` and `ImmutableCollection`, but lots of types with mutable and immutable subclasses are not Collections. Some examples are Stacks, MapIterables, and BiMaps. This change extracts some of the common API between all mutable and immutable types.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/ImmutableIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/ImmutableIterable.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright (c) 2016 Goldman Sachs.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompany this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.api;
+
+import org.eclipse.collections.api.block.function.Function;
+import org.eclipse.collections.api.block.function.Function0;
+import org.eclipse.collections.api.block.function.Function2;
+import org.eclipse.collections.api.block.function.primitive.DoubleFunction;
+import org.eclipse.collections.api.block.function.primitive.FloatFunction;
+import org.eclipse.collections.api.block.function.primitive.IntFunction;
+import org.eclipse.collections.api.block.function.primitive.LongFunction;
+import org.eclipse.collections.api.block.predicate.Predicate;
+import org.eclipse.collections.api.block.predicate.Predicate2;
+import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.block.procedure.Procedure2;
+import org.eclipse.collections.api.map.ImmutableMap;
+import org.eclipse.collections.api.map.ImmutableMapIterable;
+import org.eclipse.collections.api.map.primitive.ImmutableObjectDoubleMap;
+import org.eclipse.collections.api.map.primitive.ImmutableObjectLongMap;
+import org.eclipse.collections.api.multimap.ImmutableMultimap;
+import org.eclipse.collections.api.partition.PartitionImmutableIterable;
+import org.eclipse.collections.api.tuple.Pair;
+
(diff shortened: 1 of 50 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/MutableIterable.java
```diff
@@ -0,0 +1,222 @@
...
+
+    @Override
+    MutableIntIterable collectInt(IntFunction<? super T> intFunction);
+
+    @Override
+    MutableLongIterable collectLong(LongFunction<? super T> longFunction);
+
+    @Override
+    MutableShortIterable collectShort(ShortFunction<? super T> shortFunction);
+    */
```
itohro: Are these commented methods meant to be todo for future work?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ImmutableOrderedMap.java
```diff
@@ -159,4 +158,5 @@
 
     @Override
     <KK, VV> ImmutableOrderedMap<KK, VV> aggregateBy(Function<? super V, ? extends KK> groupBy, Function0<? extends VV> zeroValueFactory, Function2<? super VV, ? super V, ? extends VV> nonMutatingAggregator);
+    */
```
itohro: Similarly are these commented methods also todo?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/ImmutableIterable.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Override
+    ImmutableIntIterable collectInt(IntFunction<? super T> intFunction);
+
+    @Override
+    ImmutableLongIterable collectLong(LongFunction<? super T> longFunction);
+
+    @Override
+    ImmutableShortIterable collectShort(ShortFunction<? super T> shortFunction);
+    */
```
itohro: Similarly are these commented methods also todo?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/MutableIterable.java
motlin: Yes, I think the same thing makes sense to do for primitive iterables. I didn't have a chance to do it, but I could if you think it's important to be in the same commit.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ImmutableOrderedMap.java
motlin: Thanks for pointing this out. The answer is a little different here. ImmutableCollection defined these methods as returning ImmutableMap. The backwards compatible thing to do is say ImmutableIterable's aggregation methods return ImmutableMapIterable and ImmutableCollection overrides that to ImmutableMap. But even that isn't right, because ImmutableSortedSet should return an Ordered map. But Ordered maps don't exist yet, so there's probably no point in breaking backwards compatibility yet.

So yes, this is a TODO for when concrete OrderedMaps exist. However, I should also probably put the overrides back into MutableCollection and ImmutableCollection in this pull request. I'll add a second commit and if it looks good to you, squash it.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/ImmutableIterable.java
motlin: Yes.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/MutableIterable.java
itohro: Thanks for clarifying. I don't think they need to be in the same commit. 

For these comments, maybe just put TODO so that these can be easily found later on?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ImmutableOrderedMap.java
itohro: Makes sense. Thanks for the explanation!

motlin:
Ok, restored the specific overrides on MutableCollection an ImmutableCollection.

itohro:
There seems to be compilation errors in build. Could you please take a look?

motlin:
Closing this - I'm not convinced it's a good idea anymore, at least in its current form. We can revisit when we use default methods for a lot more.

PR #115, opened 2016-06-03
Implement detectOptional() and detectWithOptional() on RichIterable.

itohro:
1st commit is to move the location of detectXXX methods to put them all together. 

I split the commits just to make it easier to review. I could squash them into single commit if that's better.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MapIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MapIterable.java
@@ -11,6 +11,7 @@
 package org.eclipse.collections.api.map;
 
 import java.util.Map;
+import java.util.Optional;
 
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.block.function.Function;
@@ -249,6 +250,19 @@ public interface MapIterable<K, V> extends RichIterable<V>
      * </pre>
      */
     Pair<K, V> detect(Predicate2<? super K, ? super V> predicate);
+
+    /**
+     * Return the first key and value of the map as an Optional for which the predicate evaluates to true when
+     * they are given as arguments. The predicate will only be evaluated until such pair is found or until all
+     * of the keys and values of the map have been used as arguments. That is, there may be keys and values of
+     * the map that are never used as arguments to the predicate.
+     * <p>
+     * <pre>e.g.
+     * peopleByCity.detectOptional((city, person)
+     *          -> city.getName().equals("Anytown") && person.getLastName().equals("Smith"));
+     * </pre>
+     */
+    Optional<Pair<K, V>> detectOptional(Predicate2<? super K, ? super V> predicate);
 
     /**
      * Follows the same general contract as {@link Map#equals(Object)}.
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
@@ -42,6 +42,7 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
+import java.util.Optional;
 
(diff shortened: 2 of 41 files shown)


nikhilnanivadekar:
I see you have moved `min()` and `max()` as well to group all `min()` methods together and likewise for `max()`. I agree with the changes for `detect()` but changes for `min()` and `max()` seem unnecessary.

itohro:
No I haven't touched min/max at all. It just coincidentally looks like that as a result of the detect methods move :)

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MapIterable.java
```diff
@@ -251,6 +252,19 @@
...
+     * they are given as arguments. The predicate will only be evaluated until such pair is found or until all
+     * of the keys and values of the map have been used as arguments. That is, there may be keys and values of
+     * the map that are never used as arguments to the predicate.
+     * <p>
+     * <pre>e.g.
+     * peopleByCity.detectOptional((city, person)
+     *          -> city.getName().equals("Anytown") && person.getLastName().equals("Smith"));
+     * </pre>
+     */
+    Optional<Pair<K, V>> detectOptional(Predicate2<? super K, ? super V> predicate);
```
motlin: The detect() methods have always been at the top of the hierarchy, on RichIterable, and they predate OrderedIterable. With new methods, we have the opportunity to put these methods on OrderedIterable and OrderedMap. What do you think?

eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
```diff
@@ -1071,6 +1072,16 @@ public class <name>ObjectHashMap\<V> implements Mutable<name>ObjectMap\<V>, Exte
         return null;
     }
 
+    public Optional\<V> detectOptional(Predicate\<? super V> predicate)
+    {
+        return Optional.ofNullable(this.detect(predicate));
```
motlin: There's one weird case where it's not ok to delegate to ofNullable(). The map, or iterable, could contain a null value and the predicate could be each -> each == null. There's no great way to handle that case, but it's not great to return empty() when something was actually found. The alternative would be calling Optional.of(null) which throws, but is arguably more correct.

Eventually I'd like to reverse the delegation, detect() should delegate to detectOptional() to better differentiate between finding null and not finding anything.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MapIterable.java
itohro: It's an interesting point. I'd still prefer to be able to call detect*() methods on UnifiedSet for example. Do you think it doesn't make much sense to call detect*() methods on non-OrderedIterable or non-OrderedMap?

eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
itohro: Very interesting point again (just remembered you mentioned this earlier..). 

> Eventually I'd like to reverse the delegation, detect() should delegate to detectOptional() to better differentiate between finding null and not finding anything. 

How do you envision this reverse delegation work? Is detectOptional() going to return null in case of finding null?

motlin: It's a good point, Optional.of(null) throws, unlike Scala's Option, so it cannot represent finding and returning null, or calling Map.get() and finding that the key was present but the value was null. So for backwards compatibility I guess I'm wrong, and we cannot delegate from detect to detectOptional.

motlin: I'm going to pull this as is, it's strictly an improvement. Let's continue to think about null handling and the how methods like detectIfNone should be implemented in the future. We may also considering adding a note about null handing to the Javadoc of these methods, explaining that they return Optional.empty() rather than throwing.

PR #118, opened 2016-06-13
improve speed of chunk()-method of List implementations by using subl…

### LINKED ISSUE #116: improve speed of chunk() of List implementations
We've noticed that the chunk() implementation of FastList (actually AbstractMutableCollection) iterates over the whole list and creates a new list containing chunked lists. We're using Guavas Lists.partition() method (which uses sublists instead) on top of FastList to get performance speed up.
Classes implementing List like FastList should implement chunk() by using sublists.

Here our benchmark:

```
public class FastListBenchmark {
    @State(Scope.Thread)
    public static class MyState {
        FastList<Integer> list;

        @Setup(Level.Trial)
        public void doSetup() {
            list = new FastList<>();
            for (int i = 0; i < 10000000; i++) {
                list.add(i);
            }
        }
    }

    @Benchmark
    @BenchmarkMode(Mode.All)
    public void parallelIterationViaChunk(MyState myState) throws ExecutionException, InterruptedException {
        RichIterable<RichIterable<Integer>> partitionedValues = myState.list
                .chunk(myState.list.size() / (Executor.NUMBEROFTHREADS * 8));
        iterate(partitionedValues);
    }

    @Benchmark
    @BenchmarkMode(Mode.All)
    public void parallelIterationViaPartition(MyState myState) throws ExecutionException, InterruptedException {
        List<List<Integer>> partitionedValues = Lists
                .partition(myState.list, myState.list.size() / (Executor.NUMBEROFTHREADS * 8));
        iterate(partitionedValues);
    }

    private void iterate(Iterable<? extends Iterable<?>> partitionedValues)
            throws ExecutionException, InterruptedException {
        Collection<Runnable> jobs = new ArrayList<>();
        for (Iterable<?> values : partitionedValues) {
            jobs.add(() -> {
                for (Object value : values) {
                    foobar();
                }
            });
        }
        new Executor().execute(jobs);
    }
}

public class Executor {

    public static final int NUMBEROFTHREADS = Runtime.getRuntime().availableProcessors();

    public final ExecutorService executor = Executors.newFixedThreadPool(NUMBEROFTHREADS);

    public void execute(Collection<Runnable> jobs) throws ExecutionException, InterruptedException {
        for (Runnable job : jobs) {
            executor.execute(job);
        }
        executor.shutdown(); // !!! not shutdownNow
        executor.awaitTermination(1, TimeUnit.HOURS);
    }
}
```

**Benchmark                                          Mode    Cnt   Score    Error  Units**
FastListBenchmark.parallelIterationViaChunk       thrpt    200  13.286 ±  0.042  ops/s
FastListBenchmark.parallelIterationViaPartition   thrpt    200  74.269 ±  0.445  ops/s
FastListBenchmark.parallelIterationViaChunk        avgt    200   0.075 ±  0.001   s/op
FastListBenchmark.parallelIterationViaPartition    avgt    200   0.013 ±  0.001   s/op
FastListBenchmark.parallelIterationViaChunk      sample   2689   0.077 ±  0.001   s/op
FastListBenchmark.parallelIterationViaPartition  sample  14114   0.014 ±  0.001   s/op
FastListBenchmark.parallelIterationViaChunk          ss     10   1.668 ±  0.043   s/op
FastListBenchmark.parallelIterationViaPartition      ss     10   1.472 ±  0.056   s/op

jackhammer2k:
improve speed of chunk()-method of List implementations by using sublists [Fixes #116]


motlin:
For immutable lists, this is a clear improvement and something we should definitely merge in asap. For mutable lists, it's less clear how we should proceed. As written, it's a behavior change. Before, calling ((List) list.chunk().getFirst()).remove(0) would mutate a new chunk and *not* mutate the original list. We could return unmodifiable lists, but that's also technically behavior breaking.

jackhammer2k:
Oh, indeed then it could potentially break old code. I agree that needs to be discussed.

I could not find any Javadoc or other hints in API that specify that the backing implementation must not be modified if returned RichIterable is modified (which is impossible without a cast anyway).
Thus it looks like it does not break the specification.
In addition it does only break code which relies on casting returned RichIterable, which is dangerous anyway. Code of that kind has assumptions about the returned values that are not guaranteed (instance implementing list; implementation not removing from backed implementation).

itohro:
I've been thinking about this. In addition to the breaking behaviour, either approach would make the behaviour of chunk() method for RandomAccess mutable lists inconsistent with other collections that is a bit odd.

Given the facts,
- This improvement can be only made with utilising SubList
- Using SubList would change the behaviour in mutable lists regardless making return value unmodifiable or not
- The change is anyways only applicable to RandomAccess mutable lists

How about leaving the current mutable lists `chunk()` implementation as is, and introducing a new method `chunkUnmodifiable()` only for RandomAccess mutable lists? We could document that this method performs better than chunk() and returns unmodifiable view of lists.

This approach has a disadvantage on discoverability as the method won't be found in RichIterable..

itohro:
The suggestion above may require a little more thoughts and we can continue discussing. 

In the meantime, if you limit the change to `AbstractImmutableList` in this PR (change commit message, too) and add a note in [RELEASE_NOTE_DRAFT.md](https://github.com/eclipse/eclipse-collections/blob/master/RELEASE_NOTE_DRAFT.md), then we could pull it.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
```diff
@@ -1204,7 +1204,7 @@ public int binarySearch(T key)
                 MutableList<T> batch = new FastList<>(Math.min(size, this.size() - i));
 
                 for (int j = 0; j < size && i < this.size(); j++)
-                {
+            {
```
itohro: Thanks for making the changes!

Actually this remaining change is breaking the format of curly braces. It would be great if you could squash the 2 commits into one and also completely revert the changes in this file.

PR #128, opened 2016-07-05
Applied skeletal implementation to interface migration automated refactoring

khatchad:
This is a semantics-preserving refactoring that migrates existing method implementations in classes to the corresponding implemented interface methods as default methods. The tool **does not** add new code; it only rearranges *existing* code.

- We are evaluating a research prototype automated refactoring Eclipse plug-in called [Migrate Skeletal Implementation to Interface](https://github.com/khatchad/Migrate-Skeletal-Implementation-to-Interface-Refactoring). We have applied the tool to your project in the hopes of receiving feedback.
- The approach is very conservative. That may mean that not all changes that can be made were made. Please feel free to continue the refactoring manually if you wish.
- We only migrated methods declared in abstract classes with the hopes of such methods being suitable default methods in corresponding interfaces.
- The source code should be semantically equivalent to the original.
- We have run tests prior to applying to the tool and following the application. All tests pass.

Thank you for your help in this evaluation! Any feedback you can provide would be very helpful. In particular, we are interested if each of the proposed changes are helpful or not.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/ParallelIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/ParallelIterable.java
@@ -10,9 +10,7 @@
 
 package org.eclipse.collections.api;
 
-import java.util.Collections;
 import java.util.Comparator;
-import java.util.Set;
 
 import org.eclipse.collections.api.annotation.Beta;
 import org.eclipse.collections.api.bag.MutableBag;
@@ -35,14 +33,6 @@ import org.eclipse.collections.api.map.sorted.MutableSortedMap;
 import org.eclipse.collections.api.multimap.Multimap;
 import org.eclipse.collections.api.set.MutableSet;
 import org.eclipse.collections.api.set.sorted.MutableSortedSet;
-import org.eclipse.collections.impl.bag.mutable.HashBag;
-import org.eclipse.collections.impl.bag.sorted.mutable.TreeBag;
-import org.eclipse.collections.impl.block.factory.Comparators;
-import org.eclipse.collections.impl.block.factory.Predicates;
-import org.eclipse.collections.impl.block.procedure.CollectionAddProcedure;
-import org.eclipse.collections.impl.map.mutable.ConcurrentHashMapUnsafe;
-import org.eclipse.collections.impl.set.mutable.SetAdapter;
-import org.eclipse.collections.impl.set.sorted.mutable.TreeSortedSet;
 
 /**
  * A ParallelIterable is RichIterable which will defer evaluation for certain methods like select, reject, collect, etc.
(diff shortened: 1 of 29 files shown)


motlin:
This is an interesting tool, and refactoring. In Eclipse Collections, api and impl are split across modules and live in separate jars. This refactoring breaks the maven build because it makes code in api depend on impl, a circular dependency which anyway isn't declared in the maven build. The tool adds the correct imports, they are simply in a jar that's not on the classpath.

Here's an example of a refactoring that won't compile.
https://github.com/eclipse/eclipse-collections/pull/128/files#diff-c76f689aba0f8f5e0409cd07d558b18cR66

It tries to import org.eclipse.collections.impl.block.factory.Predicates, a helper class in impl which defines many common predicates.

Other refactorings are fine, and we were considering doing by hand. For example, this one just delegates to other methods in the same interface.
https://github.com/eclipse/eclipse-collections/pull/128/files#diff-c76f689aba0f8f5e0409cd07d558b18cR164

khatchad:
@motlin Thank you for the very helpful feedback. Also, I apologize for the failed build. I thought I had checked it via maven prior to the PR issue. Indeed, the module boundaries are something we've been hitting a few times during the evaluation. Sometimes, we hit an issue where the origin module uses Java 8 but the destination module using an earlier version of Java.

khatchad:
@motlin Please do let us know if you would like us to modify this PR in any way.

motlin:
@khatchad It's up to you whether you'd like to split the changes which do and don't pass the build as it's currently configured. If you create a commit with just changes that do compile, like the second example I listed, then we could pull that change.

khatchad:
@motlin I think this is all ready. Please let me know if there is anything else required.

As for the single commit, would it be possible to squash these commits when you do the merge? Thanks!

itohro:
Thanks @khatchad for the update. I have 2 comments:

1. It seems the commits are not signed hence Eclipse ip-validation is failing. Please sign each commit as described in [CONTRIBUTING.md](https://github.com/eclipse/eclipse-collections/blob/master/CONTRIBUTING.md)
2. For the changes in `PrimitiveIterable.java`, it seems the corresponding changes (deletion of the methods in child classes) are missing. I guess it is because the refactoring was made against generated sources, not the original template files (*.stg files). You could either make the changes manually in .stg files, or just revert the changes in `PrimitiveIterable.java`.

khatchad:
@itohro, thank you for the helpful comments, please see below:

1. Done.
1. Good catch! I actually caught this bug in our tool last week and just filed a bug for it (see https://github.com/khatchad/Migrate-Skeletal-Implementation-to-Interface-Refactoring/issues/164). Currently, our tool cannot read or manipulate such template files (only java source files). I had a look at the `*.stg` files and it would seem to me that the only one that would require deletions is `eclipse-collections-code-generator/src/main/resources/impl/abstractPrimitiveIterable.stg`. Do you agree?

itohro:
Yes, i think so for #2.

khatchad:
@itohro Completed via 95dc8cd71a7b9dde930bd633bdce7cd0c57547fa. Please let me know if there is anything else required and thank you for your feedback!

itohro:
Thanks for your contribution! Squashed and merged the changes.

khatchad:
Thank you for your feedback!

On Tue, Jul 12, 2016 at 10:26:20PM -0700, Hiroshi Ito wrote:
> Thanks for your contribution! Squashed and merged the changes.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.*
> 

-- 
Raffi Khatchadourian, PhD
Assistant Professor and Open Educational Resources (OER) Fellow
Computer Systems Technology
New York City College of Technology
City University of New York
300 Jay St., Room N913
Brooklyn, NY 11201-1909
United States
+1 (718) 260-5325
rkhatchadourian@citytech.cuny.edu
http://cuny.is/khatchad

PR #136, opened 2016-08-19
Make detectOptional methods throw NullPointerException when detecting null value.

itohro:
Revisiting the comment below from #115.

>There's one weird case where it's not ok to delegate to ofNullable(). The map, or iterable, could contain a null value and the predicate could be each -> each == null. There's no great way to handle that case, but it's not great to return empty() when something was actually found. The alternative would be calling Optional.of(null) which throws, but is arguably more correct.

Following what's suggested in the last line above and also what `Streams#findFirst()` does, I made detectOptional methods throw NullPointerException when null value is detected.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/CompositeIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/CompositeIterable.java
@@ -12,6 +12,7 @@ package org.eclipse.collections.impl.lazy;
 
 import java.util.Iterator;
 import java.util.NoSuchElementException;
+import java.util.Optional;
 
 import net.jcip.annotations.Immutable;
 import org.eclipse.collections.api.block.predicate.Predicate;
@@ -98,6 +99,21 @@ public final class CompositeIterable<E>
             }
         }
         return null;
+    }
+
+    @Override
+    public Optional<E> detectOptional(Predicate<? super E> predicate)
+    {
+        for (int i = 0; i < this.iterables.size(); i++)
+        {
+            Iterable<E> eachIterable = this.iterables.get(i);
+            Optional<E> result = Iterate.detectOptional(eachIterable, predicate);
+            if (result.isPresent())
+            {
+                return result;
+            }
+        }
+        return Optional.empty();
     }
 
     public void add(Iterable<E> iterable)
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/FlatCollectIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/FlatCollectIterable.java
@@ -11,6 +11,7 @@
 package org.eclipse.collections.impl.lazy;
 
 import java.util.Iterator;
+import java.util.Optional;
 
 import net.jcip.annotations.Immutable;
 import org.eclipse.collections.api.block.function.Function;
@@ -63,7 +64,7 @@ public class FlatCollectIterable<T, V>
     public V detect(Predicate<? super V> predicate)
     {
         V[] result = (V[]) new Object[1];
-        Iterate.detect(this.adapted, each -> {
+        Iterate.anySatisfy(this.adapted, each -> {
             Iterable<V> iterable = this.function.valueOf(each);
             return Iterate.anySatisfy(iterable, each1 -> {
                 if (predicate.accept(each1))
@@ -75,6 +76,36 @@ public class FlatCollectIterable<T, V>
             });
         });
         return result[0];
+    }
+
+    @Override
+    public Optional<V> detectOptional(Predicate<? super V> predicate)
+    {
+        V[] result = (V[]) new Object[1];
+        Iterate.anySatisfy(this.adapted, each -> {
+            if (each == null)
+            {
+                throw new NullPointerException();
+            }
+            Iterable<V> iterable = this.function.valueOf(each);
+            if (iterable == null)
+            {
+                throw new NullPointerException();
+            }
+            return Iterate.anySatisfy(iterable, each1 -> {
+                if (predicate.accept(each1))
+                {
+                    if (each1 == null)
+                    {
+                        throw new NullPointerException();
+                    }
+                    result[0] = each1;
+                    return true;
+                }
+                return false;
+            });
+        });
+        return Optional.ofNullable(result[0]);
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/AbstractArrayAdapter.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/AbstractArrayAdapter.java
@@ -16,6 +16,7 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
+import java.util.Optional;
 import java.util.RandomAccess;
 
 import org.eclipse.collections.api.block.HashingStrategy;
@@ -125,10 +126,21 @@ public abstract class AbstractArrayAdapter<T>
     }
 
     @Override
-    public T detectIfNone(Predicate<? super T> predicate, Function0<? extends T> function)
+    public <P> T detectWith(Predicate2<? super T, ? super P> predicate, P parameter)
     {
-        T result = this.detect(predicate);
-        return result == null ? function.value() : result;
+        return InternalArrayIterate.detectWith(this.items, this.items.length, predicate, parameter);
+    }
+
+    @Override
+    public Optional<T> detectOptional(Predicate<? super T> predicate)
+    {
+        return InternalArrayIterate.detectOptional(this.items, this.items.length, predicate);
+    }
+
+    @Override
+    public <P> Optional<T> detectWithOptional(Predicate2<? super T, ? super P> predicate, P parameter)
+    {
+        return InternalArrayIterate.detectWithOptional(this.items, this.items.length, predicate, parameter);
     }
 
     @Override
@@ -498,22 +510,6 @@ public abstract class AbstractArrayAdapter<T>
     {
         ListIterate.rangeCheck(fromIndex, toIndex, this.items.length);
         InternalArrayIterate.forEachWithoutChecks(this.items, fromIndex, toIndex, procedure);
-    }
-
-    @Override
-    public <P> T detectWith(Predicate2<? super T, ? super P> predicate, P parameter)
-    {
-        return InternalArrayIterate.detectWith(this.items, this.items.length, predicate, parameter);
-    }
-
-    @Override
-    public <P> T detectWithIfNone(
-            Predicate2<? super T, ? super P> predicate,
-            P parameter,
-            Function0<? extends T> function)
-    {
-        T result = this.detectWith(predicate, parameter);
-        return result == null ? function.value() : result;
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayList.java
@@ -282,10 +282,9 @@ final class ImmutableArrayList<T>
     }
 
     @Override
-    public T detectIfNone(Predicate<? super T> predicate, Function0<? extends T> function)
+    public <P> T detectWith(Predicate2<? super T, ? super P> predicate, P parameter)
     {
-        T result = this.detect(predicate);
-        return result == null ? function.value() : result;
+        return InternalArrayIterate.detectWith(this.items, this.items.length, predicate, parameter);
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayList.java
@@ -288,6 +288,17 @@ final class ImmutableArrayList<T>
     }
 
     @Override
+    public Optional<T> detectOptional(Predicate<? super T> predicate)
+    {
+        return InternalArrayIterate.detectOptional(this.items, this.items.length, predicate);
+    }
+
+    @Override
+    public <P> Optional<T> detectWithOptional(Predicate2<? super T, ? super P> predicate, P parameter)
+    {
+        return InternalArrayIterate.detectWithOptional(this.items, this.items.length, predicate, parameter);
+    }
+
     public int count(Predicate<? super T> predicate)
     {
         return InternalArrayIterate.count(this.items, this.items.length, predicate);
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
@@ -18,6 +18,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.NoSuchElementException;
+import java.util.Optional;
 import java.util.Random;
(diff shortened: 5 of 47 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/CompositeIterable.java
```diff
@@ -100,6 +101,17 @@ public E detect(Predicate<? super E> predicate)
         return null;
     }
 
+    @Override
+    public Optional<E> detectOptional(Predicate<? super E> predicate)
+    {
+        for (int i = 0; i < this.iterables.size(); i++)
+        {
+            Iterable<E> eachIterable = this.iterables.get(i);
+            return Iterate.detectOptional(eachIterable, predicate);
```
motlin: This doesn't look correct to me. The loop should proceed if an intermediate result is empty. This and detect() are probably missing some edge case tests.

itohro: Good catch! I added missing test cases in `CompositeIterableTest` (and also in `AbstractLazyIterableTestCase`) and fixed this.

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/FlatCollectIterable.java
```diff
@@ -78,6 +79,24 @@ public V detect(Predicate<? super V> predicate)
     }
 
     @Override
+    public Optional<V> detectOptional(Predicate<? super V> predicate)
+    {
+        V[] result = (V[]) new Object[1];
+        Iterate.detectOptional(this.adapted, each -> {
```
motlin: The inner call to detectOptional() here, and detect() in the previous method, is just used to short-circuit. The result is not stored in a variable. I think it would be more clear to use anySatisfy() in both methods, in a way similar to the inner-most call to anySatisfy(). The way it is written will work, but it makes it seem like the Optional returned by detectOptional() is significant.

itohro: This call is actually logically significant in a way that it should throw NullPointerException when it detects the value which is null. Throwing NullPointerException is handled within Iterate.detectOptional(), when Optional.of() is called.

Having said that, we could change the detect() side as you suggest, but maybe keeping as is makes the structure consistent. Let me know.

motlin: The call to detectOptional is on the elements of type T before they are
flattened to V. If the iterable itself is null, we'll get a NPE simply by
iterating over null. Is there something else accomplished by the call?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayList.java
```diff
@@ -282,13 +282,29 @@ public T detect(Predicate<? super T> predicate)
...
 
     @Override
+    public <P> T detectWithIfNone(
+            Predicate2<? super T, ? super P> predicate,
+            P parameter,
+            Function0<? extends T> function)
+    {
+        T result = this.detectWith(predicate, parameter);
+        return result == null ? function.value() : result;
+    }
```
motlin: The override for detectWith() looks fine because it delegates to InternalArrayIterate. I don't understand why detectWithIfNone() is improved by an override. And detectIfNone() is identical to the method it overrides, should we remove it?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/AbstractArrayAdapter.java
```diff
@@ -125,13 +126,41 @@ public T detect(Predicate<? super T> predicate)
...
     @Override
+    public <P> T detectWithIfNone(
+            Predicate2<? super T, ? super P> predicate,
+            P parameter,
+            Function0<? extends T> function)
+    {
+        T result = this.detectWith(predicate, parameter);
+        return result == null ? function.value() : result;
+    }
+
```
motlin: detectIfNone() is identical to the method it overrides. detectWithIfNone is very slightly different.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
```diff
@@ -355,13 +356,41 @@ public T detect(Predicate<? super T> predicate)
...
 
     @Override
+    public <P> T detectWithIfNone(
+            Predicate2<? super T, ? super P> predicate,
+            P parameter,
+            Function0<? extends T> function)
+    {
+        T result = ListIterate.detectWith(this, predicate, parameter);
+        return result == null ? function.value() : result;
+    }
```
motlin: I don't see how this override of detectIfWithNone is better than the implementation it overrides. It goes through a different path but still winds up calling detect(). And it's one more place with a null check. Could we just skip the override of detectIfNone and detectWithIfNone?

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/FlatCollectIterable.java
itohro: Actually you are right. Here are how I fixed this.

1. I added missing test case in FlatCollectIterableTestNoIteratorTest.
2. I applied your suggestion to use anySatisfy() for both detect() and detectOptional()
3. Fixed the logic to explicitly throw NPE when iterable is null (otherwise it'll throw IllegalArgumentException)
4. Fixed the logic to explicitly throw NPE when detecting null value

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayList.java
itohro: That's true. I removed these.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/AbstractArrayAdapter.java
itohro: These has been there before I modify (I just moved method location), but agree. I removed both as detectWithIfNone also became identical after addressing another comment in `AbstractMutableList`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
itohro: Agree and removed these overrides, too.

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/FlatCollectIterable.java
```diff
@@ -78,6 +79,32 @@ public V detect(Predicate<? super V> predicate)
...
+                throw new NullPointerException();
+            }
+            return Iterate.anySatisfy(iterable, each1 -> {
+                if (predicate.accept(each1))
+                {
+                    if (each == null)
+                    {
+                        throw new NullPointerException();
+                    }
+                    result[0] = each1;
```
motlin: I don't understand checking `each == null` all the way down here. If we made it here, then `function.valueOf(each)` already returned a non-null, non-empty list. I think we should either check earlier in the outer lambda, or not at all.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
```diff
@@ -905,13 +905,41 @@ public T detect(Predicate<? super T> predicate)
...
     @Override
+    public <P> T detectWithIfNone(
+            Predicate2<? super T, ? super P> predicate,
+            P parameter,
+            Function0<? extends T> defaultValueBlock)
+    {
+        T result = this.detectWith(predicate, parameter);
+        return result == null ? defaultValueBlock.value() : result;
+    }
+
```
motlin: This implementation of detectIfNone is identical to the superclass, and the implementation of both detechIfNone and detectWithIfNone have the null check for the result. An easy way to replace all the remaining places that check `result == null` would be something like this.

```java
return this.detectOptional(predicate).orElseGet(defaultValueBlock::value);
```

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
```diff
@@ -773,6 +774,32 @@ else if (cur != null)
...
+                if (predicate.accept(each))
+                {
+                    return Optional.of(each);
+                }
+            }
+        }
+        return Optional.empty();
+    }
+
     private Object chainedDetect(ChainedBucket bucket, Predicate<? super T> predicate)
```
motlin: Very nice, the return value of chainedDetect() is already unambiguous and only returns null to mean not found.

unit-tests-java8/src/test/java/org/eclipse/collections/test/bag/immutable/sorted/ImmutableSortedBagImplTest.java
```diff
@@ -16,10 +16,11 @@
 import org.eclipse.collections.impl.block.factory.Comparators;
 import org.eclipse.collections.impl.test.junit.Java8Runner;
 import org.eclipse.collections.test.IterableTestCase;
+import org.eclipse.collections.test.NoDetectOptionalNullTestCase;
 import org.junit.runner.RunWith;
 
 @RunWith(Java8Runner.class)
-public class ImmutableSortedBagImplTest implements ImmutableSortedBagTestCase
+public class ImmutableSortedBagImplTest implements ImmutableSortedBagTestCase, NoDetectOptionalNullTestCase
 {
```
motlin: The detectOptional null test irrelevant here because ImmutableSortedBag cannot contain nulls. ImmutableSortedBag cannot contain nulls because its elements are sorted with a Comparator, which doesn't tolerate null. Isn't this assertion irrelevant for all SortedIterables? If so, you can delete NoDetectOptionalNullTestCase and move its method into SortedIterableTestCase.

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/FlatCollectIterable.java
itohro: Good catch! 

I actually meant to check `each1` here. I reviewed my test case and find one corner case is missing there. I added missing test in `FlatCollectIterableTestNoIteratorTest`, moved `each` check out of lambda and put `each1` check here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
itohro: Good suggestion. I only made this change in `AbstractRichIterable` for now. I'll do the whole refactoring in another PR.

unit-tests-java8/src/test/java/org/eclipse/collections/test/bag/immutable/sorted/ImmutableSortedBagImplTest.java
itohro: Good suggestion. I put this to SortedIterableTestCase and removed from its subclasses.

motlin:
@itohro the build is currently failing on this PR.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
itohro: There is one interesting edge case where this doesn't work.

```java
Assert.assertEquals(Integer.valueOf(3), this.newWith(Comparators.reverseNaturalOrder(), 2, 3, 4, 5).detectIfNone(Integer.valueOf(3)::equals, null));
```

When it detects an element that meets the predicate condition, it should return it even if the function passed is null. However it throws NullPointerException with the method reference approach above (it seems method reference is "evaluated" even though it is not executed in the code path). 

We can change it as below to address this particular case still using detectOptional().

```java
this.detectOptional(predicate).orElseGet(() -> (function == null) ? null : function.value());
```

However now this code behaves differently from the original code when it detects null AND null function is passed in. Original code throws NPE, while the code above returns null.

I think this behaviour was not documented or tested anywhere, but I suppose original code behaves more naturally in this case (throws NPE). 

So I'd revert this to be original null check code. Let me know what you think.

motlin: That doesn't sound correct to me. Did you make sure to pass `defaultValueBlock::value` instead of `defaultValueBlock`?

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/FlatCollectIterable.java
```diff
@@ -78,6 +79,36 @@ public V detect(Predicate<? super V> predicate)
...
+                throw new NullPointerException();
+            }
+            return Iterate.anySatisfy(iterable, each1 -> {
+                if (predicate.accept(each1))
+                {
+                    if (each1 == null)
+                    {
+                        throw new NullPointerException();
+                    }
+                    result[0] = each1;
```
motlin: If you deleted this if-statement, wouldn't we get a NullPointerException on line 108 instead, from Optional.ofNullable(null)? Seems redundant?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
itohro: Yes, this was the code that was failing with NPE (it was in `AbstractRichIterable` not in `FastList` so using `function` instead of `defaultValueBlock`).

```
    @Override
    public T detectIfNone(Predicate<? super T> predicate, Function0<? extends T> function)
    {
        return this.detectOptional(predicate).orElseGet(function::value);
    }
```

motlin: If detectOptional returns a present Optional, then function should not be
invoked and passing a null function should be ok.

itohro: This is the simplest test case to simulate it.

```java
    @Test
    public void testNullMethodRefEval()
    {
        Function0<String> nullFunction = null;
        Verify.assertThrows(NullPointerException.class, () -> this.methodRef(nullFunction::value));
    }

    private String methodRef(Supplier<String> nullFunctionSupplier)
    {
        return "I don't execute the function but it throws NPE";
    }
```

motlin: Fascinating. The method turns out to be irrelevant. This alone throws.
```java
Function0<String> nullFunction = null;
Supplier<String> supplier = nullFunction::value;
```

motlin: And this doesn't throw.
```java
Function0<String> nullFunction = null;
Supplier<String> supplier = () -> nullFunction.value();
```

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/FlatCollectIterable.java
itohro: We have to differentiate the case we detect null (we only know it by checking `each1` == null after evaluating predicate.accept(each1)) vs the case we didn't find anything (result[0] stays null).

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
itohro: This is interesting :) 

I can do this lambda way, but again, we could do this in a separate PR as there are anyways couple of places we could apply this.

Issue #126, opened 2016-07-01
Change primitive-primitive/object-primitive functional interfaces to extend the corresponding JDK functional interfades.


itohro:
The changes would look similar to [this commit](https://github.com/eclipse/eclipse-collections/pull/112/commits/383c9896623b2a6a1b4a0d6d05317efbc5baef78).

The scope of this issue would be below:
`primitiveToPrimitiveFunction.stg` (extends JDK primitive to primitive Function such as IntToLongFunction etc.)
`objectPrimitiveProcedure.stg` (extends JDK object-primitive Consumer such as ObjIntConsumer etc.)
`primitiveObjectFunction.stg` (extends JDK primitive-object Function such as IntFunction etc.)

Issue #125, opened 2016-07-01
Change primitive functional interfaces to extend the corresponding JDK functional interfaces.


itohro:
The changes would look similar to [this commit](https://github.com/eclipse/eclipse-collections/pull/112/commits/383c9896623b2a6a1b4a0d6d05317efbc5baef78).

The scope of this issue would be below:
`primitiveFunction.stg` (extends JDK primitive ToXxxFunction)
`primitiveFunction0.stg` (extends JDK primitive XxxSupplier)
`primitiveProcedure.stg` (extends JDK primitive XxxConsumer)

PR #137, opened 2016-08-20
Change primitive functional interfaces to extend the corresponding JDK functional interfaces.

itohro:
The changes are corresponding to issue #125 and #126.

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/block/function/primitiveToPrimitiveFunction.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/block/function/primitiveToPrimitiveFunction.stg
@@ -22,10 +22,19 @@ import java.io.Serializable;
  *
  * @since 3.0.
  */
+<if(primitive1.specializedStream && primitive2.specializedStream)>@FunctionalInterface<endif>
 public interface <name1>To<name2>Function
-        extends Serializable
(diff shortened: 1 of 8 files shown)


eclipse-collections-code-generator/src/main/resources/api/block/function/primitiveToPrimitiveFunction.stg
```diff
@@ -22,10 +22,19 @@ import java.io.Serializable;
  *
  * @since 3.0.
  */
+ <if(primitive1.specializedStream && primitive2.specializedStream && !sameTwoPrimitives)>@FunctionalInterface<endif>
 public interface <name1>To<name2>Function
-        extends Serializable
+        extends <if(primitive1.specializedStream && primitive2.specializedStream && !sameTwoPrimitives)>java.util.function.<name1>To<name2>Function, <endif>Serializable
 {
```
motlin: I was expecting an else clause checking for sameToPrimitives and extending `<name1>UnaryOperator`. What do you think?

motlin:
Looks pretty good, the suggestion about UnaryOperators could potentially be in a separate commit. Let me know what you think.

itohro:
Thanks! Added a change for UnaryOperators, too.

PR #144, opened 2016-09-10
Fix addAll() method in CompositeFastList to return false on isEmpty()…

### LINKED ISSUE #141: Fix addAll() method in CompositeFastList to return false on isEmpty() case.
The method currently returns true if the incoming collection isEmpty() when it should return false, as the contract for addAll says true should only be returned if the target collection has been changed.

donraab:
… case. Fixes #141.


itohro:
The changes look good to me. Once it is rebased on top of current master we could pull it.

PR #148, opened 2016-09-20
Optimize List.clear methods to avoid clearing unused memory

chrisc36:
Fixes goldmansachs/gs-collections#33

Signed-off-by: Christopher Clark <csquared@cs.washington.edu>

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/impl/list/mutable/primitiveArrayList.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/list/mutable/primitiveArrayList.stg
@@ -109,7 +109,7 @@ public class <name>ArrayList extends Abstract<name>Iterable
 
     public void clear()
     {
-        Arrays.fill(this.items, <(zero.(type))>);
+        Arrays.fill(this.items, 0, size, <(zero.(type))>);
         this.size = 0;
     }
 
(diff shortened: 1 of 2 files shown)


chrisc36:
Note if a List is constructed wrapping a user-controlled array, then the change to the clear method might be visible to users. Although that seems like a very obscure implementation detail for users to be relying on.

eclipse-collections-code-generator/src/main/resources/impl/list/mutable/primitiveArrayList.stg
```diff
@@ -109,7 +109,6 @@ public class <name>ArrayList extends Abstract<name>Iterable
 
     public void clear()
     {
-        Arrays.fill(this.items, <(zero.(type))>);
```
motlin: Not sure how I feel about this change. The difference is probably only visible through reflection. I still think I'd prefer using the 4-arg version of `Arrays.fill` here too.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
```diff
@@ -215,7 +215,10 @@ public FastList(Collection<? extends T> source)
     @Override
     public void clear()
     {
-        Arrays.fill(this.items, null);
+        for (int i = 0; i < size; i++)
+        {
+            items[i] = null;
+        }
```
chrisc36: Sure, the reason I avoided it to start is that Arrays.fill we run the same code, but with a redundant bounds check. But the readability is preferred I can switch to that.

chrisc36:
I went ahead and switched to just using Arrays.fill in both cases

PR #166, opened 2016-11-15
Integrate JaCoCo for test coverage.

nikhilnanivadekar:


performance-tests/pom.xml
```diff
@@ -104,6 +104,12 @@
         <plugins>
 
             <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <version>${jacoco.plugin.version}</version>
+            </plugin>
```
motlin: You definitely don't want to turn code coverage on for performance tests. You should consider skipping acceptance tests, really anything other than unit-tests.

pom.xml
```diff
@@ -326,15 +328,28 @@
...
+
+                <plugin>
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>2.19</version>
                     <configuration>
                         <includes>
                             <include>**/*Test.java</include>
                         </includes>
-                        <argLine>-XX:-OmitStackTraceInFastThrow</argLine>
+                        <argLine>-XX:-OmitStackTraceInFastThrow -Xms1024m -Xmx2048m @{argLine}</argLine>
```
motlin: Why @{argline}?

unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2Test.java
```diff
@@ -1513,7 +1511,7 @@ public void partitionWithParallel()
         Assert.assertEquals(expectedBag.getSelected(), actualBag.getSelected());
         Assert.assertEquals(expectedBag.getRejected(), actualBag.getRejected());
     }
-
+*/
```
motlin: Why are these tests both edited and commented out?

unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2Test1Test.java
```diff
@@ -18,47 +18,18 @@
...
 import org.eclipse.collections.impl.partition.bag.PartitionHashBag;
-import org.eclipse.collections.impl.partition.list.PartitionFastList;
-import org.eclipse.collections.impl.partition.set.PartitionUnifiedSet;
-import org.eclipse.collections.impl.test.Verify;
-import org.eclipse.collections.impl.tuple.primitive.PrimitiveTuples;
 import org.junit.Assert;
 import org.junit.Test;
 
-public final class Collectors2Test
+public final class Collectors2Test1Test
```
motlin: The name of the test includes the word Test two times.

unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2Test2Test.java
```diff
@@ -0,0 +1,646 @@
...
+    @Test
+    public void collectDoubleParallel()
+    {
+        DoubleList expected =
+                LARGE_INTERVAL.collectDouble(each -> (double) each, DoubleLists.mutable.empty());
+        DoubleList actual =
+                this.bigData.parallelStream().collect(Collectors2.collectDouble(each -> (double) each, DoubleLists.mutable::empty));
+        Assert.assertEquals(expected, actual);
+    }
+}
```
motlin: I don't get it. The bytecode is too long for Jacoco?

unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2Test.java
nikhilnanivadekar: This is a mistake. I will fix it.

unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2Test1Test.java
nikhilnanivadekar: The file name needs to end with 'Test' because of the sure-fire plugin config:

 <include>**/*Test.java</include>

unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2Test2Test.java
nikhilnanivadekar: Seems so, but it failed with a non-Jacoco error. It might be a Travis issue as well, because the build passes locally.

pom.xml
nikhilnanivadekar: argLine is set by JaCoCo to start an agent which binds and generates an execution file. This file is used by JaCoCo to analyze the coverage. Since surefire triggers the tests, the JaCoCo agent will not bind unless the argLine is set in the command executed by surefire.
Refer to this for more info: http://stackoverflow.com/a/25774988/5182052

performance-tests/pom.xml
nikhilnanivadekar: To suppress it on performance tests and acceptance tests I will have to overload the surefire plugin config for these two modules. Since, surefire plugin requires argLine, if JaCoCo is not enabled the build fails since it does not find argLine. This does not slow down the build.
I can either overload it for these two modules or keep it as is with JaCoCo running. Please let me know what is more desirable.

motlin: Coverage *must* not be on for performance tests or jmh tests. Jacoco shouldn't even be on the classpath. As long as its off for some modules, I think it's desirable to turn it off for most modules. Or rather, only turn it on in the two modules where it's relevant, unit-tests and unit-tests-java8.

nikhilnanivadekar: Ok that is easy enough to fix then. I will have to override the surefire plugin config for these two modules to use argLine.

nikhilnanivadekar:
Fixed the poms for code coverage to run only for applicable modules. 
Fixed test names to remove redundant test.

eclipse-collections-forkjoin/pom.xml
```diff
@@ -63,7 +63,21 @@
         <plugins>
 
             <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <version>${jacoco.plugin.version}</version>
```
Godin: This declaration and similar ones seem to be redundant since plugin already declared in parent POM.

eclipse-collections-forkjoin/pom.xml
```diff
@@ -63,7 +63,21 @@
...
+
+            <plugin>
                 <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.19</version>
+                <configuration>
+                    <includes>
+                        <include>**/*Test.java</include>
+                    </includes>
+                    <argLine>-XX:-OmitStackTraceInFastThrow -Xms1024m -Xmx2048m @{argLine}</argLine>
+                    <runOrder>random</runOrder>
```
Godin: Here and in other similar places everything except `argLine` (`version`, `includes`, `runOrder`) seems to be redundant since already declared in parent POM.

Godin: Also supposed to be possible to declare `argLine` in parent. IMO common declaration is better over duplication.

eclipse-collections-forkjoin/pom.xml
nikhilnanivadekar: @Godin I tried to fix it such that redundancy is avoided, and everything from the parent pom is used. But JaCoCo execution agent was not getting triggered without adding the artifact and version. May be I was missing something basic. If you get a chance can you please help to remove the redundancy? I can close this PR if you would like to raise one with JaCoCo integrated correctly. 
It is ok if you cannot, I will try to fix it, and push a new commit. 
Appreciate your review!

eclipse-collections-forkjoin/pom.xml
nikhilnanivadekar: @Godin should the argLine just be <argLine><argLine> or the whole argLine? If I understood correctly, JaCoCo will set the correct argLine for the relevant modules?

unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2AdditionalTest.java
```diff
@@ -0,0 +1,646 @@
...
+import org.eclipse.collections.impl.list.Interval;
+import org.eclipse.collections.impl.partition.bag.PartitionHashBag;
+import org.eclipse.collections.impl.partition.list.PartitionFastList;
+import org.eclipse.collections.impl.partition.set.PartitionUnifiedSet;
+import org.eclipse.collections.impl.test.Verify;
+import org.eclipse.collections.impl.tuple.primitive.PrimitiveTuples;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class Collectors2AdditionalTest
```
Godin: Just a small note: as was said in https://github.com/jacoco/jacoco/issues/462#issuecomment-261223067 you can simply exclude test class file from instrumentation instead of splitting it on two.

But I guess that this is done to also be able to use EclEmma from Eclipse? Maybe leave a comment in test, at least to prevent accidental merge back that will lead to the same question about failure?

nikhilnanivadekar: I will exclude the test class and drop this commit on my next rebase.

eclipse-collections-forkjoin/pom.xml
Godin: @nikhilnanivadekar not sure I understand the question, probably because I don't see the difficulty that you encountered. However answering on comment above: I can try to polish this, but a little bit later today.

nikhilnanivadekar: Thank you very much!

Godin:
This is cool! and will become a nice update for page ["Projects that use JaCoCo"](https://github.com/jacoco/jacoco/wiki/Projects%20that%20use%20JaCoCo) when will be merged. We are happy to help and hear your feedback. And thank you for catching bug in ASM :thumbsup:

test-coverage/pom.xml
```diff
@@ -0,0 +1,182 @@
...
+                        <id>default-report</id>
+                        <phase>test</phase>
+                        <goals>
+                            <goal>report-aggregate</goal>
+                        </goals>
+                    </execution>
+                    <execution>
+                        <id>default-check</id>
+                        <goals>
+                            <goal>check</goal>
```
Godin: `check` unlike `report-aggregate` does not cross boundaries of modules, hence useless here.

Godin: This BTW can be seen in a build log:
```
--- jacoco-maven-plugin:0.7.7.201606060606:check (default-check) @ test-coverage ---
[INFO] Skipping JaCoCo execution due to missing classes directory:/home/travis/build/eclipse/eclipse-collections/test-coverage/target/classes
```

nikhilnanivadekar: This makes sense. Thank you for catching it!

eclipse-collections-forkjoin/pom.xml
Godin: Okay, I see difficulty that you faced: undeclared property won't be expanded into empty string by default during [late property evaluation in `maven-surefire-plugin`](http://maven.apache.org/surefire/maven-surefire-plugin/faq.html#late-property-evaluation) , so that in modules where `jacoco-maven-plugin` not executed such as `scala-unit-tests` JVM will fail with message
```
Error: Could not find or load main class @{argLine}
```

This is addressed in https://github.com/Godin/eclipse-collections/commit/e31408936224c68cc859f45c0686467e49d29519#diff-600376dffeb79835ede4a0b285078036R135 by declaring empty property.

test-coverage/src/test/java/org/eclipse/collections/impl/CoverageTest.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class CoverageTest
+{
+    @Test
+    public void smoke()
+    {
+        Assert.assertTrue("Smoke Test required by JaCoCo to aggregate reports", true);
```
Godin: Yep. Addressed in https://github.com/Godin/eclipse-collections/commit/40c9b98417e9a7ea8b0be8b9a884d6a1ebf16151

test-coverage/pom.xml
Godin: Addressed in https://github.com/Godin/eclipse-collections/commit/9f6317fae9509bc61544919cf1b043cb5dbacce2

eclipse-collections-forkjoin/pom.xml
Godin: And we'd better update documentation in JaCoCo about this.

nikhilnanivadekar:
Thank you very much @Godin for your help!

nikhilnanivadekar:
I have cherry picked @Godin 's commits.

motlin:
Thanks @Godin and @nikhilnanivadekar!

Godin:
@motlin no :stuck_out_tongue: thank you for usage of JaCoCo :wink:

PR #167, opened 2016-11-17
Remove executable bit from files that are not executable

Godin:


Godin:
@nikhilnanivadekar while working on your issue with JaCoCo I came across this. Fixed only root files, and left some Java source files - not sure if size of change will fit contribution rules for non-committers:
```
$ find . -type f -perm +111 -name "*.java" -print | wc -l
2448
```
Can fix the rest, if it will fit.

nikhilnanivadekar:
LGTM. 
I cannot merge until Travis build passes. Travis seems to be stuck in queue, I have tried to restart the build couple of times. Travis had issues earlier today, may be if you can push this again, it might trigger the build.
Seems like Travis has caught up on their backlog, see [here](https://www.traviscistatus.com/incidents/jrnxmnms13g6) for incident.

Godin:
@nikhilnanivadekar re-pushed, now both in queue 😆 

image

Godin:
And what about the rest of files?

nikhilnanivadekar:
@Godin I killed the first build. Now only one is stuck.

AFAIK we do not have any contribution limits. I am fine with fixing the other files. 
Cc @motlin for his opinion.

Godin:
@nikhilnanivadekar this is interesting, I asked because of the following conversation about my first contribution to Eclipse Orbit that was bigger than 1000 LoC - https://dev.eclipse.org/mhonarc/lists/orbit-dev/msg04559.html

motlin:
Thanks @Godin! Above The 1000 line threshold we would do an extra review to
make sure it's original work with no copyright concerns. In my opinion a
1000 line fix that only affects white space would obviously not have IP
concerns. Similarly, fixing the executable bit shouldn't worry anyone about
Copyright / IP.

PR #174, opened 2016-12-17
Implement SummaryStatistics and Collectors2.summarizing.

donraab:

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
@@ -13,6 +13,8 @@ package org.eclipse.collections.impl.collector;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.Iterator;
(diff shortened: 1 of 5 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
```diff
@@ -1668,6 +1670,22 @@ private Collectors2()
                 EMPTY_CHARACTERISTICS);
     }
 
+    /**
+     * Returns a SummaryStatistics with results for int, long and double functions calculated for
+     * each element in the Stream or Collection this Collector is applied to.
+     */
+    public static <T> Collector<T, ?, SummaryStatistics<T>> summarizing(
```
gs-rezaem: I don't see how this signature is generally applicable. What are the chances of needing exactly 1 int, 1 long and 1 double summary stats? SummaryStatistics already implements a builder pattern.

donraab:
I'm looking into the compilation issue with three of the builds.

PR #178, opened 2017-01-02
Add collect<primitive> methods to primitive array list implementations.

bhav0904:
Signed-off-by: Bhavana Hindupur <bhavana.hindupur@gmail.com>


itohro:
Looks good overall! All of my comments are minor cosmetic changes, but it would be nice to be fixed.

nikhilnanivadekar:
Can we please revisit adding all `collect<Primitive>` methods on primitive collections?
The jar is already close to 9 Mb which is huge, and causes issues while running it on small JVMs (Eg: mindstorms)
I would rather have a `collect<Primitive>` where we collect from a primitive to the same primitive, and for `int`, `float`, to widen the primitive i.e. `long` and `double`.

Basically have `collectByte()` on `byte collection`s, have `collectInt()`, `collectLong()` on `int collection`s.
If a user wants to collect to different primitive than what is available with collect<Primitive> they can always use `forEach()`

donraab:
One alternative that can reduce the total number of methods required would be to implement collect<Primitive> methods with target primitive collections.  This would reduce the need for co-variant versions of all of the methods, and would allow for the implementation to be provided as default methods on each of the primitive interfaces.  This might then not require a major release of Eclipse Collections as long as all of the implementations are provided as default methods.

nikhilnanivadekar:
@bhav0904 how my LOC are we increasing with this change? Also, how much is the jar size increasing?

nikhilnanivadekar:
This is not required as users can use the target collection variant implemented in #186

PR #183, opened 2017-01-10
Change StackIterableTestCase to extend AbstractRichIterableTestCase.

donraab:

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/AbstractRichIterableTestCase.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/AbstractRichIterableTestCase.java
@@ -108,8 +108,6 @@ import org.eclipse.collections.impl.test.Verify;
 import org.junit.Assert;
 import org.junit.Test;
 
-import static org.junit.Assert.assertFalse;
-
 public abstract class AbstractRichIterableTestCase
 {
     protected abstract <T> RichIterable<T> newWith(T... littleElements);
@@ -142,8 +140,8 @@ public abstract class AbstractRichIterableTestCase
     public void containsAllIterable()
     {
         RichIterable<Integer> collection = this.newWith(1, 2, 3, 4);
-        Assert.assertTrue(collection.containsAllIterable(FastList.newListWith(1, 2)));
-        Assert.assertFalse(collection.containsAllIterable(FastList.newListWith(1, 5)));
+        Assert.assertTrue(collection.containsAllIterable(Lists.mutable.with(1, 2)));
+        Assert.assertFalse(collection.containsAllIterable(Lists.mutable.with(1, 5)));
     }
 
     @Test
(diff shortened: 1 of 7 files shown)


unit-tests/src/test/java/org/eclipse/collections/impl/AbstractRichIterableTestCase.java
```diff
@@ -1491,7 +1491,7 @@ public void chunk_single()
     {
         RichIterable<String> collection = this.newWith("1");
         RichIterable<RichIterable<String>> groups = collection.chunk(2);
-        Assert.assertEquals(FastList.newListWith(1), groups.collect(RichIterable::size));
+        Assert.assertEquals(Lists.mutable.with(1), groups.collect(RichIterable::size, Lists.mutable.empty()));
```
motlin: This is a weaker assertion. Why is it necessary?

unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2AdditionalTest.java
```diff
@@ -68,7 +68,7 @@
 public class Collectors2AdditionalTest
 {
     public static final Interval SMALL_INTERVAL = Interval.oneTo(5);
-    public static final Interval LARGE_INTERVAL = Interval.oneTo(200000);
+    public static final Interval LARGE_INTERVAL = Interval.oneTo(30000);
```
motlin: Is this for test speed?

donraab: Yes, the Travis build was timing out.

unit-tests/src/test/java/org/eclipse/collections/impl/AbstractRichIterableTestCase.java
donraab: The Stack implementations return a different type for collect.  I understood this method to be testing that the result size should be 1 even though the chunk size requested was 2.  I didn't think that testing the return type of collect was a specifically a List was necessary for this specific test.  I can revert and override this test specifically for Stacks.

donraab:
I updated the StackIterableTestCase to override all methods from the parent and deleted a few clearly duplicate test methods.  Any other feedback?

unit-tests/src/test/java/org/eclipse/collections/impl/AbstractRichIterableTestCase.java
```diff
@@ -1504,7 +1502,7 @@ public void chunk_zero_throws()
     @Test
     public void chunk_large_size()
     {
-        RichIterable<String> collection = FastList.newListWith("1", "2", "3", "4", "5", "6", "7");
+        RichIterable<String> collection = this.newWith("1", "2", "3", "4", "5", "6", "7");
         Assert.assertEquals(collection, collection.chunk(10).getFirst());
```
itohro: Since you're changing this method, it would be nice to change this line to use getOnly() instead of deprecated getFirst() method. Same case for the overridden methods in child classes.

donraab:
I updated the the one test to use getOnly instead of getFirst in the parent class and all the subclasses.

nikhilnanivadekar:
LGTM. Will merge after Travis build passes.
Separately, you have replaced `FastList.newListWith()` with `Lists.mutable.with()`, we should do that replacement globally for Lists, Sets, Bags, Maps, Multimaps.

PR #184, opened 2017-01-17
Implement sumByBigDecimal and sumByBigInteger on Collectors2.

donraab:

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/SummaryStatistics.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/SummaryStatistics.java
@@ -21,10 +21,11 @@ import org.eclipse.collections.api.block.function.primitive.LongFunction;
 import org.eclipse.collections.api.block.procedure.Procedure;
 import org.eclipse.collections.api.map.ImmutableMap;
 import org.eclipse.collections.impl.factory.Maps;
-import org.eclipse.collections.impl.tuple.Tuples;
 
 /**
  * A Summarizer can be used to aggregate statistics for multiple primitive attributes.
+ *
+ * @since 8.1
  */
 public class SummaryStatistics<T> implements Procedure<T>
 {
--- a/RELEASE_NOTE_DRAFT.md
+++ b/RELEASE_NOTE_DRAFT.md
@@ -6,6 +6,7 @@ This is the 8.1.0 minor release.
 New Functionality
 -----------------
 
+* Implement summingBigDecimal, sumByBigDecimal, summingBigInteger, sumByBigInteger on Collectors2. 
 * Implement BigDecimalSummaryStatistics and BigIntegerSummaryStatistics.
 * Implement SummaryStatistics and Collectors2.summarizing.
 * Integrate JaCoCo for test coverage.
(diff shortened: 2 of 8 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/SummaryStatistics.java
```diff
@@ -25,6 +25,8 @@
 
 /**
  * A Summarizer can be used to aggregate statistics for multiple primitive attributes.
+ *
+ * @since 8.1
  */
 public class SummaryStatistics<T> implements Procedure<T>
```
itohro: Couple of comments for this class (and a comment for other 2 summary statistics classes, too)

- The statement `import org.eclipse.collections.impl.tuple.Tuples;` is not in use. 
- This class should have `serialVersionUID`
- Ideally we should have serialization tests for all 3 SummaryStatistics classes.

donraab: I didn't make SummaryStatistics serializable because Int, Long, DoubleSummaryStatistics classes are not serializable.  I'll optimize imports on both of the classes you indicated.  Thank you.

donraab: I added serialization tests for BigDecimalSummaryStatistics and BigIntegerSummaryStatistics.

itohro: I see. However currently `SummaryStatistics` implements `Procedure` that's marked as `Serializable` so it's against the contract.

Should `SummaryStatistics` implement `Consumer` instead? Does it cause any cascading impact (I hope not)?

donraab: I think this is something we'll have to live with for now.  There is really no way to serialize the JDK provided SummaryStatistics classes today.

donraab: I may have come up with a solution to this problem.  The SummaryStatistics classes in JDK are not final, so I can extend them and should be able to make them Serializable.  I'm working on it now.  If I get it to work I will commit an update later today.

itohro: Just to clarify, what I was asking was whether it can directly implement `Consumer` instead of `Procedure` that way it doesn't need to be serializable. Would that approach not work?

donraab: I had thought it would be less than ideal, but considering all the work (including using reflection) that it will take to make the other JDK SummaryStatistics classes Serializable, it may be an ok compromise.

donraab: I will submit a separate pull request for the serialization solution for SummaryStatistics.  I think it should continue to implement Procedure.  Separately, I found this feature request for the JDK SummaryStatistics classes which I found interesting.  (https://bugs.openjdk.java.net/browse/JDK-8043747)  I also found that Oracle Coherence had to implement their own private versions of Int/Long/DoubleSummaryStatistics, I assume because of the serialization issue.

itohro:
Merging it for now and follow-up serialization issue in another PR.

Issue #193, opened 2017-02-04
Make constructors on Unmodifiable and Synchronized collections public


jankotek:
I have my own implementations of `LongLongMap` and `LongObjectMap` collections. Constructors on `UnmodifiableLongLongMap` and `SynchronizedLongLongMap` are package protected. My implementations can not instantiate those wrappers.

Is it possible to make constructor on `Unmodifiable*` amd `Synchronized*` classes public?

donraab:
I am fine with this in principle.  I looked at what is currently done in the Unmodifiable and Synchronized Object collections, and there is a pattern of having a public static `of` method which constructs an instance.  Sometimes a constructor is inconsistently public (e.g. SynchronizedBag, SynchronizedXyzMultimap).  I'm not convinced that the benefit of having a static `of` method is worth the extra cost in this case.  There are also instances already in the synchronized wrappers, where there is no public constructor for the scenario where a lock is supposed to be explicitly provided.  I've found tests for these methods, but I believe they should be public or they useless.  If you'd like to submit a pull request, we can get some feedback from others and hopefully merge it in time for the next release.

jankotek:
ok, I will write a patch

jankotek:
pull request is out, but it fails on IP stuff. I have old mail from Craig, will look howto sign the paperwork

donraab:
Info on the Eclipse Contributor Agreement is linked in [here](https://github.com/eclipse/eclipse-collections/blob/master/CONTRIBUTING.md).

jankotek:
Old PR did not had Signed-off tag. I created new one with checks for null arguments.

PR #1, opened 2015-12-09
Initial import.

motlin:
Initial import of the code. Reviewed in [IPZilla – CQ 10432](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=10432).

PR #197, opened 2017-02-06
Make constructors public on Synchronized* and Unmodifiable* collections

jankotek:
https://github.com/eclipse/eclipse-collections/issues/193

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/impl/bag/mutable/synchronizedPrimitiveBag.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/bag/mutable/synchronizedPrimitiveBag.stg
@@ -54,12 +54,12 @@ public final class Synchronized<name>Bag
 {
     private static final long serialVersionUID = 1L;
 
-    Synchronized<name>Bag(Mutable<name>Bag bag)
+    public Synchronized<name>Bag(Mutable<name>Bag bag)
     {
         super(bag);
     }
 
-    Synchronized<name>Bag(Mutable<name>Bag bag, Object newLock)
+    public Synchronized<name>Bag(Mutable<name>Bag bag, Object newLock)
     {
         super(bag, newLock);
     }
--- a/eclipse-collections-code-generator/src/main/resources/impl/bag/mutable/unmodifiablePrimitiveBag.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/bag/mutable/unmodifiablePrimitiveBag.stg
@@ -44,7 +44,7 @@ public final class Unmodifiable<name>Bag
 {
     private static final long serialVersionUID = 1L;
 
-    Unmodifiable<name>Bag(Mutable<name>Bag bag)
+    public Unmodifiable<name>Bag(Mutable<name>Bag bag)
     {
         super(bag);
     }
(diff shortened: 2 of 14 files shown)


donraab:
The basic change looks good, and it looks like you covered all of the necessary templates.

We need to decide how we want to handle null containers and null locks.  I found the following code in `AbstractSynchronizedRichIterable`, which most of the Collection based synchronized containers extend.

```
if (delegate == null)
{
    throw new IllegalArgumentException("Cannot create a AbstractSynchronizedRichIterable on a null collection");
}

this.delegate = delegate;
this.lock = lock == null ? this : lock;
```

There is a parent class for each of the synchronized primitive collections generated from `abstractSynchronizedPrimitiveCollection.stg`.  The lock code in there is the same, but the collection is not not checked, probably because null could never be passed in previously.

```
this.collection = collection;
this.lock = newLock == null ? this : newLock;
```

I'm open to suggestions here, including accepting this pull request as is, and handling this later consistently across object and primitive collections.  Thoughts?

jankotek:
I will expand my patch to check for nulls. There should be guards if
gates are open :)
> 
>   
>   
> 
> 
> 
> 
> 
> {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c5
> 5493e4bb","name":"GitHub"},"entity":{"external_key":"github/eclipse/e
> clipse-collections","title":"eclipse/eclipse-
> collections","subtitle":"GitHub repository","main_image_url":"https:/
> /cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-
> 11e6-95fc-
> 7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent
> .com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-
> b52498112777.png","action":{"name":"Open in
> GitHub","url":"https://github.com/eclipse/eclipse-
> collections"}},"updates":{"snippets":[{"icon":"PERSON","message":"@do
> nraab in #197: The basic change looks good, and it looks like you
> covered all of the necessary templates.\r\n\r\nWe need to decide how
> we want to handle null containers and null locks.  I found the
> following code in `AbstractSynchronizedRichIterable`, which most of
> the Collection based synchronized containers extend.\r\n\r\n```\r\nif
> (delegate == null)\r\n{\r\n    throw new
> IllegalArgumentException(\"Cannot create a
> AbstractSynchronizedRichIterable on a null
> collection\");\r\n}\r\n\r\nthis.delegate = delegate;\r\nthis.lock =
> lock == null ? this : lock;\r\n```\r\n\r\nThere is a parent class for
> each of the synchronized primitive collections generated from
> `abstractSynchronizedPrimitiveCollection.stg`.  The lock code in
> there is the same, but the collection is not not checked, probably
> because null could never be passed in
> previously.\r\n\r\n```\r\nthis.collection = collection;\r\nthis.lock
> = newLock == null ? this : newLock;\r\n```\r\n\r\nI'm open to
> suggestions here, including accepting this pull request as is, and
> handling this later consistently across object and primitive
> collections.  Thoughts?  \r\n\r\n\r\n"}],"action":{"name":"View Pull
> Request","url":"https://github.com/eclipse/eclipse-collections/pull/1
> 97#issuecomment-277803328"}}}

PR #200, opened 2017-02-09
Implement primitive case procedures.

donraab:


gs-rezaem:
LGTM

Slight tangent: at some level, it's a bit odd to have serialization tests right next to tests that use non-serializable lambdas. Are we setting bad precedence?  See this [stackoverflow](http://stackoverflow.com/questions/22807912/how-to-serialize-a-lambda)  for examples of serializable lambdas.

donraab:
The serialization tests use the same package, but are contained in a separate module.

PR #203, opened 2017-02-14
Implement throwing method on functional interface factories which tak…

donraab:
…es a user specified runtime exception function.


motlin:
A realistic test could take a list of String file names, call each() on the list, and read the files one at a time. We'd only want to wrap IOException. Would it save boilerplate?

Issue #223, opened 2017-03-23
Provide p2 repository for Eclipse Collections


kthoms:
In order to be usable for Eclipse Plug-in development the Eclipse Collections framework has to provide a p2 repository. This repository should be deployed to a public consumable location.

kthoms:
The issue could be solved either with a Maven Tycho based build or using [EBR](https://github.com/eclipse/ebr). PR #225 and #219 provide a proposal for both ways.

While first one is a rather frequently used approach to build plugins, features and p2 repositories, it disallows to include the build in the overall build process. This means a second build job has to be executed to do the Tycho based build. This could be the advantage of the EBR, where the p2 repo could be build within the same build job.

Result should be in both cases a zipped p2 repository, which has to be deployed to an accessible location, from where it could be used as part of a target platform definition. This would open Eclipse Collection for Eclipse plug-in development.

nikhilnanivadekar:
@kthoms Thanks for the explanation!
Pinging @guw to check-in how far he has come with #219 .

In my opinion going with the recipe and having a single build makes sense. But will this additional build mandate using Eclipse during the build? We would like to keep the builds IDE agnostic.

guw:
No, it should not require Eclipse. It will be IDE agnostic. I still haven't had time to finish the EBR stuff. It's been a pretty busy week and doesn't look to get better before April

nikhilnanivadekar:
@guw appreciate the feedback. I think we are fine to wait, if the EBR setup will work. 
@kthoms , @guw here is a tentative road map of Eclipse Collections releases:
I am planning to call a code complete for 8.2.0 next week so that I can release it in time for The Great Indian Developer Summit.
We can pick up work for p2 repo in 8.3.0 and release it by end of May in time for the release train to pick it up.
We will start 9.0.0 where we will investigate and try and support Java 9 and release it by JavaOne.

This road map is yet to be vetted with other committers but timing-wise it makes sense.

Thanks a lot to both of you for helping us out with this!

nikhilnanivadekar:
I would like to get an opinion from @kthoms , are you fine with this approach to use EBR?

kthoms:
Yes, I'm fine with it. Thanks for asking.

PR #219, opened 2017-03-23
[#223] Create a  p2 repository (based on EBR)

guw:
This adds the initial recipe for generating an OSGi bundle for Eclipse Collections. The bundle is a combination of the main library and the api. It runs as part of the main build an produces a p2 repo with what is available in the Maven reactor.

#223 

/cc @kthoms @nikhilnanivadekar

PR #225, opened 2017-03-25
[#223] Add support for a p2 repository

kthoms:
Provide projects in a new folder 'p2-releng' that build a feature
project for Eclipse Collections and create a p2 repository with Maven
Tycho.

The build in `p2-releng` has to be executed after running a `mvn install` on the parent project. Calling `mvn package` in `p2-releng` is enough then. As a result the build will produce the p2 repository to 
`p2-releng/eclipse-collections-repository/target/org.eclipse.collections.repository-8.2.0-SNAPSHOT.zip`

The content of the repository has to be uploaded to a public available place. Each release version should have its own URL, and ideally a composite repository is build around that. Lorenzo Bettini has written on how to do this on Bintray: http://www.lorenzobettini.it/2016/02/publish-an-eclipse-p2-composite-repository-on-bintray/
This could be done in a next step. For now you should just decide whether to go with this Maven Tycho based build or with EBR.


nikhilnanivadekar:
We need to decide between #219 and #225 (this one).
I do not know how far @guw has come with #219.

For the record can you please elaborate the pros and cons of both the approaches. Apologies, I am new to the concept of p2 repositories, it is a good learning experience for me.

kthoms:
Let's move the discussion of both approaches to the issue #223

nikhilnanivadekar:
@kthoms I am going to close this PR since #223 will be fixed with #219

Issue #226, opened 2017-03-25
Remove OSGi bundles for eclipse-collections-api, eclipse-collections-bundle.


kthoms:
The bundle id's are a bit strange. This will be because of default behavior of the bundle plugin. Current identifiers (`Bundle-SymbolicName`) are:
- org.eclipse.collections.eclipse-collections
- org.eclipse.collections.eclipse-collections-api
- org.eclipse.collections.eclipse-collections-testutils
- org.eclipse.collections.eclipse-collections-forkjoin

I would expect rather:
- org.eclipse.collections.impl
- org.eclipse.collections.api
- org.eclipse.collections.testutils
- org.eclipse.collections.forkjoin

Open for discussion is `org.eclipse.collections.impl`. It might be also without the `.impl` suffix.

Changing the bundle ids is a breaking change, so this would only be valid for next major version.

nikhilnanivadekar:
@kthoms as discussed before, we will not be able to change the names since it is a compatibility breaking change. We can pick it up once 9.0.0 opens. Looking forward to your contribution. :)

guw:
FWIW, #219 introduces a new bundle `org.eclipse.collections` which combines api and impl. There is no real use case for distributing two separate bundles. The approach is backwards compatible - consumers of the old bundles continue consuming it the current way. Consumers using p2 will start using the new bundle.

kthoms:
The use case could be that the bundles are typically consumed as plain Java jars and by splitting jars you can configure API only for compilation and avoid that implementation classes are resolvable in such an environment. Of course in an OSGi environment it would be valid to put both in the same bundle, and users cannot access impl classes.

nikhilnanivadekar:
For my understanding, does this mean that eclipse-collections-api and eclipse-collections-impl modules can stop exporting OSGi packages? The users can just use the new bundle. Of course we will wait till 9.0.0 to remove the old bundles. 
Is there a way to deprecate the old bundles during minor release and point the users to the new bundle?
Having multiple bundles can confuse new users.

guw:
Yes, you can stop generating OSGi metadata for the existing projects. New metadata will be generated for the `org.eclipse.collections` bundle at build time using instructions defined in `osgi.bnd` file.

nikhilnanivadekar:
Perfect! Thank you!
Once we merge #219 I will update this issue description accordingly.

nikhilnanivadekar:
I updated this issue to remove the bundles currently created.
[...]

PR #227, opened 2017-03-25
[#226] Explicit configuration of Bundle-SymbolicName

kthoms:
The Bundle-SymbolicName property is explicitly configured to use more
concise bundle ids.

Signed-off-by: Karsten Thoms <karsten.thoms@itemis.de>


nikhilnanivadekar:
Since #226 is closed. We will need a different PR to remove bundle plugins from all the modules except the p2 repository module.

PR #234, opened 2017-03-26
Yak shaving: upgrade maven-jar-plugin and maven-source-plugin

Godin:
First step regarding Java 9 - is to execute tests with it.

In absence of usage of [Maven Toolchains](https://maven.apache.org/guides/mini/guide-using-toolchains.html), build should itself be executed with JDK 9 EA because `maven-surefire-plugin` uses JDK that was used for Maven execution. But  currently it fails - see
* [log prior to upgrade of `maven-jar-plugin`](https://github.com/eclipse/eclipse-collections/files/870527/log1.txt)
* [log after upgrade of `maven-jar-plugin`, but prior to upgrade of `maven-source-plugin`](https://github.com/eclipse/eclipse-collections/files/870529/log2.txt)

After upgrade of both there are still problems with JDK 9 EA - compilation errors:
* [in module `eclipse-collections`](https://github.com/eclipse/eclipse-collections/files/870530/log3.txt)
* and even after build with JDK 8 and usage of JDK 9 EA to build [only module `unit-tests`](https://github.com/eclipse/eclipse-collections/files/870532/log4.txt)

However this might be a bug in javac (I'm investigating), so suggest to apply this change anyway, because
* first of all it allows to investigate and report to OpenJDK any problems with compilation
* it is IMO good practice to keep Maven plugins up-to-date

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -230,7 +230,7 @@
 
                 <plugin>
                     <artifactId>maven-jar-plugin</artifactId>
-                    <version>2.6</version>
+                    <version>3.0.2</version>
                 </plugin>
 
                 <plugin>
@@ -255,7 +255,7 @@
 
                 <plugin>
                     <artifactId>maven-source-plugin</artifactId>
-                    <version>2.4</version>
+                    <version>3.0.1</version>
                     <executions>
                         <execution>
                             <phase>verify</phase>


Godin:
And for the record here is a list of classes that are the roots of many unit tests failures, because they cause `java.lang.reflect.InaccessibleObjectException` with message containing `module java.base does not "opens java.util" to unnamed module` under JDK 9 EA b162:
* `org.eclipse.collections.impl.utility.ArrayListIterate` because uses reflection to break into private fields of `java.util.ArrayList`
* `org.eclipse.collections.impl.collector.SerializableDoubleSummaryStatistics` because uses reflection to break into private fields of `java.util.DoubleSummaryStatistics`
* `org.eclipse.collections.impl.collector.SerializableLongSummaryStatistics` because uses reflection to break into private fields of `java.util.LongSummaryStatistics`
* `org.eclipse.collections.impl.collector.SerializableIntSummaryStatistics` because uses reflection to break into private fields of `java.util.IntSummaryStatistics`
* `org.eclipse.collections.impl.test.Verify` because uses reflection to break into private method `clone` of `java.lang.Object`

nikhilnanivadekar:
@Godin from my understanding, Eclipse Collections will always need the module `java.util` as it extends JCF interfaces.  
I have seen similar compilation issues related to generic types due to javac before (before 1.80_31).
I am raising an issue for Eclipse Collections Java 9 support. We can continue the discussion there.

Merging this PR.

Godin:
@nikhilnanivadekar this is not about usage of module `java.base` that contains package `java.util`, which is perfectly ok on its own, but about usage of **reflection for access to private members** of it

PR #236, opened 2017-03-27
Fix sonar issues

nikhilnanivadekar:


eclipse-collections-code-generator/src/main/resources/api/block/function/primitiveFunction.stg
```diff
@@ -22,7 +22,7 @@ import java.io.Serializable;
  *
  * @since 3.0.
  */
-<if(primitive.specializedStream)>@FunctionalInterface<endif>
+@FunctionalInterface
```
motlin: This one may be a difference between Sonar static analysis and IntelliJ static analysis. In IntelliJ, go to the folder containing generated sources. Right click the folder and select unmark as generated source root. It will temporarily show up like regular source, so inspections will run. I suspect that IntelliJ will warn you that the @FunctionalInterface annotation is redundant and ask you to remove it. It is redundant, because it's on the parent interface. In cases where static analysis conflicts, we just have to pick a rule and be consistent. I tend to prefer IntelliJ rules because they make more use of the type system to eliminate redundancy. But there's an argument to be made that the increased redundancy helps readability without jumping between files.

nikhilnanivadekar: @motlin I tried what you suggested. The only inspection warning I got was about replacing the fully qualified name with import in `DoubleFunction`. I tried `CharFunction` and no inspection warnings were thrown. Is there an inspection setting I am missing?
May be to begin with the prior `if` condition was incorrect in the template file. Looking at the code:
```
public boolean hasSpecializedStream()
{
    return this == INT || this == LONG || this == DOUBLE;
}
```
That means `DoubleFunction` will get the `@FunctionalInterface` annotation, where as `CharFunction` will not get that annotation. It should be the other way around, since `CharFunction` does not inherit the annotation from any other interface.
IMO we should keep the redundant annotation. If IntelliJ shows the warning, we can suppress it accordingly in the templates 
`<if(primitive.specializedStream)>//Suppress Redundant Annotation (or whichever suppression it is once the warning comes up to figure it out)<endif>`

nikhilnanivadekar:
On a separate note @Godin would you like to fix the remaining `@Override` annotation fixes? I have a semi-complete patch for it, and can push it in this PR itself before merging.
https://sonarqube.com/component_issues?id=org.eclipse.collections%3Aeclipse-collections-parent#types=CODE_SMELL|resolved=false|rules=squid%3AS1161

eclipse-collections-code-generator/src/main/resources/api/block/function/primitiveFunction.stg
motlin: Fair enough.

Godin:
@nikhilnanivadekar thank you for the opportunity to contribute, but doubt that I'll have time to work on remaining cases in upcoming days

PR #237, opened 2017-03-27
Fix condition that triggers SonarQube analysis

Godin:
SonarQube analysis is executed for the each of 5 jobs - see https://travis-ci.org/eclipse/eclipse-collections/jobs/215394107 and https://travis-ci.org/eclipse/eclipse-collections/jobs/215394108 for example, whereas should be executed only for job that has `DESC` `unit tests`. Because of this generation of 5 reports approximately at the same time, server side processing fails for 4 out of 5 of those reports:
sonarqube

This change fixes condition, so that only one job will be doing analysis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ matrix:
         - CMD="mvn install --projects jmh-tests,performance-tests --also-make --activate-profiles all -DskipTests=true --batch-mode --show-version --errors"
 
 script:
-- if [[ "$TRAVIS_EVENT_TYPE" == "cron" && DESC="unit tests" ]];
+  if [[ "$TRAVIS_EVENT_TYPE" == "cron" && "$DESC" == "unit tests" ]];
   then mvn install --batch-mode --show-version --errors sonar:sonar -Dsonar.host.url=https://sonarqube.com -Dsonar.login=${SONARQUBE_TOKEN} -Dsonar.organization=eclipse-collections ;
   else eval $CMD;
   fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ matrix:
       env:
         - DESC="unit tests"
         - CMD="mvn install --batch-mode --show-version --errors"
+        - SONAR_COMMAND="sonar:sonar -Dsonar.host.url=https://sonarqube.com -Dsonar.login=${SONARQUBE_TOKEN} -Dsonar.organization=eclipse-collections"
 
     - jdk: oraclejdk8
       env:
@@ -34,9 +35,10 @@ matrix:
         - CMD="mvn install --projects jmh-tests,performance-tests --also-make --activate-profiles all -DskipTests=true --batch-mode --show-version --errors"
 
 script:
-  if [[ "$TRAVIS_EVENT_TYPE" == "cron" && "$DESC" == "unit tests" ]];
-  then mvn install --batch-mode --show-version --errors sonar:sonar -Dsonar.host.url=https://sonarqube.com -Dsonar.login=${SONARQUBE_TOKEN} -Dsonar.organization=eclipse-collections ;
-  else eval $CMD;
+  if ["$TRAVIS_EVENT_TYPE" == "cron"]; then
+      eval $CMD $SONAR_COMMAND;
+  else
+      eval $CMD;
   fi
 
 cache:


.travis.yml
```diff
@@ -34,7 +34,7 @@ matrix:
         - CMD="mvn install --projects jmh-tests,performance-tests --also-make --activate-profiles all -DskipTests=true --batch-mode --show-version --errors"
 
 script:
-- if [[ "$TRAVIS_EVENT_TYPE" == "cron" && DESC="unit tests" ]];
+  if [[ "$TRAVIS_EVENT_TYPE" == "cron" && "$DESC" == "unit tests" ]];
   then mvn install --batch-mode --show-version --errors sonar:sonar -Dsonar.host.url=https://sonarqube.com -Dsonar.login=${SONARQUBE_TOKEN} -Dsonar.organization=eclipse-collections ;
```
motlin: This should really be implemented at the top of the file and should build up a "sonar command." Then just change the unit tests command in the matrix above.

```
     - jdk: oraclejdk8
       env:
         - DESC="unit tests"
         - CMD="mvn install --batch-mode --show-version --errors $SONAR_COMMAND"
```

Where $SONAR_COMMAND is allowed to be empty.

motlin: And I know I'm late to the conversation, but what does "$TRAVIS_EVENT_TYPE" == "cron" mean?

Godin: @motlin `TRAVIS_EVENT_TYPE` indicates how the build was triggered - see https://docs.travis-ci.com/user/environment-variables/ , `cron` is about https://docs.travis-ci.com/user/cron-jobs/ , so that analysis runs once a day according to your setup of Travis in order to respect "fair use" of a free SonarQube.com service - see https://about.sonarqube.com/

motlin: @Godin Interesting, looks pretty new. Thanks!

Godin: @motlin not sure I understand your proposal regarding `$SONAR_COMMAND`, because I don't see a phase that is executed early enough to set variable before it is consumed in matrix - see little experiment at https://github.com/Godin/travis-playground/blob/e4859a718dd3d5eb09a7c045fcb5fa7b8dc9b61c/.travis.yml and respective build at https://travis-ci.org/Godin/travis-playground/builds/215665470

Also IMO far more readable and flexible will be to leave only `DESC` in `matrix` and do a switch on `DESC` in script. WDYT?

nikhilnanivadekar: @motlin I have committed on top of @Godin 's commit, can you please review?

.travis.yml
```diff
@@ -12,31 +12,37 @@ matrix:
...
         - DESC="compile jmh-tests and performance-tests"
         - CMD="mvn install --projects jmh-tests,performance-tests --also-make --activate-profiles all -DskipTests=true --batch-mode --show-version --errors"
+        - SONAR_COMMAND = ""
 
 script:
-  if [[ "$TRAVIS_EVENT_TYPE" == "cron" && "$DESC" == "unit tests" ]];
-  then mvn install --batch-mode --show-version --errors sonar:sonar -Dsonar.host.url=https://sonarqube.com -Dsonar.login=${SONARQUBE_TOKEN} -Dsonar.organization=eclipse-collections ;
-  else eval $CMD;
+  if ["$TRAVIS_EVENT_TYPE" == "cron"]; then
+      eval $CMD $SONAR_COMMAND;
```
motlin: What happens if you delete all the empty declarations of SONAR_COMMAND = "" but keep the rest as it is?

nikhilnanivadekar: @motlin I tried it a couple of different ways:
1) Remove SONAR_COMMAND from all places and only keep it at unit tests, the build script does not compile.
2) Declare it at the top before the matrix and then try to override it for unit tests, the override does not go through
According to me, this is the best way we can achieve the SONAR_COMMAND to be triggered only for cron events. Happy to try other options if you have something in mind.

Godin: @motlin @nikhilnanivadekar FWIW another option something like:
```
matrix:
  fast_finish: true
  include:
    ...
    - jdk: oraclejdk8
      env:
        - DESC="unit tests"
        - CMD="mvn install --batch-mode --show-version --errors"
    ...

script:
  - SONAR_COMMAND=""
  - if ["$TRAVIS_EVENT_TYPE" == "cron"]; then SONAR_COMMAND="..."; fi
  - eval $CMD;
```

And coming back to suggestion about `switch` - something like:
```
matrix:
  fast_finish: true
  include:
    ...
    - jdk: oraclejdk8
      env:
        - DESC="unit tests"
    ...

script:
  - if ["$TRAVIS_EVENT_TYPE" == "cron"]; then SONAR_COMMAND="..."; fi
  - case "$DESC" in
  "unit tests")
    CMD="..."
    ;;
  esac
  - eval $CMD $SONAR_COMMAND
```
maybe the last one might require to move the whole script into file to avoid issues with `.travis.ym` such as one command per `-` and so on:
```
script: travis.sh
```

Also please recall that this PR is not about beautification of `.travis.yml`, but mainly about fix of problem that is stated in description.

Hope this helps.

nikhilnanivadekar: Deleted SONAR_COMMAND from other places and tested it out. You can find the test results out here:
https://travis-ci.org/nikhilnanivadekar/eclipse-collections/builds/216180148

I think now it is good to merge.

For the record I had updated the If statements to be like here: https://travis-ci.org/nikhilnanivadekar/eclipse-collections/jobs/216180152#L158
Copy pasting for ease:
```
if ["$TRAVIS_EVENT_TYPE" == "cron"]; then
      eval $CMD $SONAR_COMMAND;
  else
      eval $CMD $SONAR_COMMAND;
  fi
```

motlin:
Perfect, thank you!

PR #239, opened 2017-03-28
Implement RichIterable.getOnlyOptional() (closes #124)

### LINKED ISSUE #124: Implement RichIterable.getOnlyOptional().
Implement getOnly method that returns Optional on RichIterable. The method signature would look like below.

`Optional<T> getOnlyOptional()`

PascalSchumacher:
I tried to stick as closely as possible to existing code.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
@@ -149,6 +149,31 @@ public interface RichIterable<T>
     }
 
     /**
+     * Returns the element as an optional if the iterable has exactly one element.
+     * If the iterable is empty {@link Optional#empty} is returned.
+     * Otherwise, throw {@link IllegalStateException}.
+     *
+     * @return an element of an iterable wrapped in an optional or an empty optional.
+     * @throws IllegalStateException if iterable has more than one element.
+     * @throws NullPointerException if the element is null
+     * @since 8.2
+     */
+    default Optional<T> getOnlyOptional()
+    {
+        if (this.isEmpty())
+        {
+            return Optional.empty();
+        }
+
+        if (this.size() == 1)
+        {
+            return Optional.of(this.getFirst());
+        }
+
+        throw new IllegalStateException("Size must be 0 or 1 but was " + this.size());
+    }
+
+    /**
      * Returns true if the iterable has an element which responds true to element.equals(object).
      *
      * @since 1.0
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/SynchronizedRichIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/SynchronizedRichIterable.java
@@ -171,6 +171,15 @@ public class SynchronizedRichIterable<T>
     }
 
     @Override
+    public Optional<T> getOnlyOptional()
+    {
+        synchronized (this.lock)
+        {
+            return this.iterable.getOnlyOptional();
+        }
+    }
+
+    @Override
     public boolean contains(Object object)
     {
         synchronized (this.lock)
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/UnmodifiableRichIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/UnmodifiableRichIterable.java
@@ -278,6 +278,12 @@ public class UnmodifiableRichIterable<T>
     }
 
     @Override
+    public Optional<T> getOnlyOptional()
+    {
+        return this.iterable.getOnlyOptional();
+    }
+
+    @Override
     public RichIterable<T> select(Predicate<? super T> predicate)
     {
         return this.iterable.select(predicate);
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableArrayBag.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableArrayBag.java
@@ -263,6 +263,22 @@ public class ImmutableArrayBag<T>
     }
 
     @Override
+    public Optional<T> getOnlyOptional()
+    {
+        if (this.counts.length == 0)
+        {
+            return Optional.empty();
+        }
+
+        if (this.counts.length > 1 || this.counts[0] > 1)
+        {
+            throw new IllegalStateException("Size must be 0 or 1 but was greater than 1");
+        }
+
+        return Optional.of(this.getFirst());
+    }
+
+    @Override
     public ImmutableBag<T> select(Predicate<? super T> predicate)
     {
         return this.select(predicate, HashBag.newBag()).toImmutable();
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableEmptyBag.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableEmptyBag.java
@@ -216,6 +216,12 @@ final class ImmutableEmptyBag<T>
     }
 
     @Override
+    public Optional<T> getOnlyOptional()
+    {
+        return Optional.empty();
+    }
+
+    @Override
     public Iterator<T> iterator()
     {
         return EmptyIterator.getInstance();
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableHashBag.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableHashBag.java
@@ -191,6 +191,12 @@ public class ImmutableHashBag<T>
     }
 
     @Override
+    public Optional<T> getOnlyOptional()
+    {
+        return this.delegate.getOnlyOptional();
+    }
+
+    @Override
     public T min(Comparator<? super T> comparator)
     {
         return this.delegate.min(comparator);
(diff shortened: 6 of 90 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/QuadrupletonList.java
```diff
@@ -145,6 +146,12 @@ public T getOnly()
     }
 
     @Override
+    public Optional<T> getOnlyOptional()
```
nikhilnanivadekar: Code duplication can be reduced by moving `getOptional()` to `AbstractMemoryEfficientMutableList` and overriding it only on empty and singleton

PascalSchumacher: Of course. I just mechanically added `getOnlyOptional` to any class with `getOnly`. Let me check were in the class hierarchy `getOnlyOptional` really belongs. I will update the pull request. Thanks!

PascalSchumacher:
The *.stg files part is still missing.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -147,6 +147,30 @@ default T getOnly()
...
         throw new IllegalStateException("Size must be 1 but was " + this.size());
     }
+    
+    /**
+     * Returns the element as an optional if the iterable has exactly one element.
+     * If the iterable is empty {@link Optional#empty} is returned.
+     * Otherwise, throw {@link IllegalStateException}.
+     *
+     * @return an element of an iterable wrapped in an optional or an empty optional.
+     * @throws IllegalStateException if iterable has more than one element.
```
motlin: I wrote `Iterate.getOnly()` and I've been using it for a while now, so I feel like I can criticize it. The methods which may throw, like `getOnly()` and `groupByUniqueKey()` can be pretty difficult to use, not so much because they throw, but because the type of exception thrown and the error messages they contain are not configurable. Using `Optional` as a return type can help, but in my opinion we should only return `Optional` and never throw.

```java
default Optional<T> getOnlyOptional()
{
    if (this.size() == 1)
    {
        return Optional.ofNullable(this.getFirst());
    }
    return Optional.empty();
}
```

This isn't perfect because it's still hard to throw an error message differentiating between empty and 2-or-more, but I think it's less surprising. An `Either` type would help more. Overall though I think it's surprising for a method which returns `Optional` or `Either` to ever throw.

PascalSchumacher: Thanks for the feedback!

I can not claim the same level of experience with `getOnly`, as I was just looking for an easy to contribute issue.  

I think `getOnlyOptional` returning `Optional.empty()` for multiple elements can be confusing for these reasons:
* difference to `getOnly()`
* difference to other popular implementations? ([Iterables.getOnlyElement()](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/Iterables.html#getOnlyElement-java.lang.Iterable-) or [MoreCollectors.toOptional()](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/MoreCollectors.html#toOptional--) from guava throw an `IllegalArgumentException` for multiple elements)
* three meanings of  `Optional.empty()`: empty iterable, iterable with multiple elements or iterable containing a single `Null` element

Of course I'm fine with changing the implementation to not throw if that is decided. :)

motlin: Very interesting points here. I think we would all agree that RichIterable.getOnlyOptional() would return empty() for an empty collection and a present Optional for a single element. I had assumed empty() for two or more elements but I suppose that's not the only obvious choice. Iterables.getOnlyElement() looks exactly the same as Iterate.getOnly(), in that it throws for both empty and two or more. MoreCollectors.toOptional() has Javadoc which says `The returned collector throws an IllegalArgumentException if the stream consists of two or more elements, and a NullPointerException if the stream consists of exactly one element, which is null.` That last part means they delegate to Optional.of() instead of Optional.ofNullable() which is crucial. It's the only way to differentiate between empty and a collection containing one null.

It seems like this API could use more discussion, it's more complex than it seemed when we created the issue.

PascalSchumacher: Sure. I'm fine with shelving or scraping the pull request.

motlin:
This looks pretty good. Once it uses Optional.of() instead of Optional.ofNullable() I think it's ready to go.

PascalSchumacher:
Thanks! I will update the pull request.

PascalSchumacher:
Should I reduce the pull request to a default method on `RichIterable` plus overrides for synchronized classes plus tests?

nikhilnanivadekar:
Yes you can pull all the implementations up to RichIterable, AbstractMultiReaderMutableCollection, AbstractSynchronizedRichIterable.

motlin:
Not sure I agree. If these implementations live in all the same places as getOnly(), isn't that better / more efficient than default methods?

nikhilnanivadekar:
I agree about the implementations for Immutable* . But for mutable, would default methods be that inefficient? I don't know (seriously asking).
I am fine either ways, since the LOC are not so much that it will increase the jar size.

PascalSchumacher:
Updated to use `Optional.of()` instead of `Optional.ofNullable()`.

motlin:
@nikhilnanivadekar There's always a tradeoff between more lines of code (and bigger jar size) vs hand written optimized overrides. It's not a big deal here either way. I'm just saying all these overrides are written, no need to delete them.

@PascalSchumacher this PR looks really close but the Travis build is currently broken on two unit tests.

PascalSchumacher:
Sorry for the currently broken build. I will fix it later today.

PascalSchumacher:
Build is fixed now.

nikhilnanivadekar:
This looks good to me. @motlin are you fine merging?

motlin:
```
@throws IllegalStateException if iterable has more than one element.
```

I still don't think this makes sense. The whole idea of returning Optional was to avoid throwing exceptions. If Optional doesn't work, maybe we could write an Either type.

The implementation of getOnly throws IllegalStateException, and it's not customizable, which always causes me problems. I was picturing a new api where I could customize the exception using orElseThrow.

```java
T onlyElement = iterable.getOnlyOptional().orElseThrow(() -> new RuntimeException("Expected only one T!"))l
```

This implementation of getOnlyOptional may return empty, present, or throw an exception that's not customizable. It just doesn't seem like enough of an improvement over getOnly.

nikhilnanivadekar:
We can have a custom `Exception` on the lines of `MultipleElementsFoundException` or something like that.
If the users want to suppress it then they can wrap around a `try-catch` block and ignore it.
I am fine with `orElseThrow()` but it will throw for `Optional.empty()` so it will not work for empty cases.

Taking a step back, is the motivation of `getOnlyOptional()` to not throw at all? If it is to not throw at all would it be better to rename the method as `Either<Exception, Optional> getOnlyEither()`. This can be the contract:
1) When size == 0: left = null, right=Optional.empty()
2) When size == 1; left = null, right = Optional.empty()
3) When size > 1; left = MultipleElementsException, right = Optional.empty()

We will have to implement our own `Either` since Java does not have one.

motlin:
We could change getOnly to throw some new subclass of IllegalStateException in order to make it easier to catch and ignore, or catch and customize, the exception it throws. However the mere fact that it throws makes it annoying to use. The method groupByUniqueKey is similar that way. For both getOnly and groupByUniqueKey, I'd like to provide alternatives that never throw.

In languages and libraries that have an Either type, it is not idiomatic to combine Either and Optional. Instead, I'd imagine getOnlyOptional (getOnlyEither?) would return either the only element or the size, so Either<T, Integer>. Then you could map the Integer to an Exception and throw it. Thoughts?

nikhilnanivadekar:
I am fine with `public Either<T, Integer> getOnlyEither();` as the signature. 
I don't quite follow what you mean by 
> map the Integer to an Exception and throw it

We should return the size, as it might be useful for users.

1. When size == 0: left = null, right=0
2. When size == 1; left = onlyElement, right = 1
3. When size > 1; left = MultipleElementsException, right = size

motlin:
@nikhilnanivadekar The type `Either` is *either* a result *or* an error. You can imagine `Either<S, F>` having two implementations `Success<S>` and `Failure<F>`. So we'd be returning `Success<T>` when the size is 1 or `Failure<Integer>` holding the size, more or less.

nikhilnanivadekar:
@motlin that makes sense. Thank you for the clarification. I vote of doing `getOnlyEither()` instead of `getOnlyOptional()`. Do you concur?

PascalSchumacher:
I thought getting an Exception was the reason for the existence of `getOnly`, otherwise you can just take an element e.g. with `getFirst`?

Shouldn't `Either<L, R>` have a left and a right? Here left would probably be the only element and right the number of elements.

For success/failure I guess it would be better to have something like `Try<R>` containg the result and an Exception.

nikhilnanivadekar:
@PascalSchumacher you are correct, we should get an exception when `getOnly()` is used on a collection with size > 1 . However, the current exception which is thrown is generic, so special handling is painful.
I did not quite understand what you mean by `Try<R>`. Can you please elaborate? Thanks!

PascalSchumacher:
@nikhilnanivadekar I was suggesting `Try<R>` instead of `Either<L, R>` for the failure scenario, but on further consideration `Either<L, R>` seems the best fit for this use case at least according to http://blog.xebia.com/try-option-or-either/.

nikhilnanivadekar:
@PascalSchumacher would you be able to close out this PR before June 1st? I would like to call a code complete and do a release to align with Eclipse train release.
If we cannot close it out by then, it is ok, we can pick it up for 9.0.0.

Sorry if I missed this, do you need feedback for the approach/code review from @motlin or me?

PascalSchumacher:
@nikhilnanivadekar I believe a useful `Either` implementation would require significant effort and this is beyond my current abilities.

Let's just close this pull request.

Issue #199, opened 2017-02-06
Idea project files in git repo?


jankotek:
I noticed git repo contains `.idea` folder and it is not excluded in `.gitignore`. 

First commit included dozen of `*.iml` files from subprojects. It was a bit annoying to filter them out. 

1) is there some good purpose to include project files in git repo? In general it is considered a bad practice, Idea usually imports maven projects just fine.

2) could we update `.gitignore` to exclude subproject files?

PascalSchumacher:
The same goes for eclipse project files.

motlin:
I want to keep the .idea/ folder checked in because it includes things like .idea/codeStyleSettings.xml which allow people to `Ctrl+Alt+L` reformat their code and pass the CheckStyle build with little to no effort. On the other hand, the *.iml files and the file .idea/misc.xml seem to have a lot of churn. I know that the *.iml files will be required if we set up builds in the [public TeamCity farm](https://blog.jetbrains.com/teamcity/2016/10/hosted-teamcity-for-open-source-a-new-home/). But until we do that, I could see removing the *.iml files if they're too annoying.

There's a relevant request in JetBrain's youtrack [here](https://youtrack.jetbrains.com/issue/IDEA-91607).

jankotek:
I can live with `.idea` folder, but `*.iml` files are very annoying. Just opening EC in Idea creates 40 modifications to commit. I spend 70% of time dealing with those files:

![image](https://cloud.githubusercontent.com/assets/359021/24522772/4c688b32-1599-11e7-99dc-48ac1f2852e6.png)


Maven has task to generate Idea project files. Their creation could be automated on TeamCity.

Gitignore allows to whitelist files for inclusion with `!` prefix.

> There's a relevant request in JetBrain's youtrack here.
This issue is opened for a long time. 

Anyway, I am not sending pull request. I created my own .gitignore and it works for me.

nikhilnanivadekar:
@PascalSchumacher would you like to contribute, to ignore the eclipse project files?
Thanks!

motlin:
The same point goes for Eclipse. The files that share code style settings
and stuff should be kept. If there are files that are pure noise, we can
suppress them.

PR #240, opened 2017-03-28
Fix compilation errors

Godin:
These compilation errors were revealed during attempt to use JDK 9 EA b162 for build of project - see https://github.com/eclipse/eclipse-collections/pull/234#issue-217022672

All they originate from the code that uses same pattern:
```java
class Pattern {
  interface MutableObjectByteMap {
    <V> MutableCollection<V> collect(ByteToObjectFunction<? extends V> function);
  }

  interface MutableCollection<T> {
    ImmutableCollection<T> toImmutable();
  }

  interface ImmutableCollection<T> { }

  interface ByteToObjectFunction<T> { }

  MutableObjectByteMap delegate;

  <V> ImmutableCollection<V> example(ByteToObjectFunction<? extends V> function) {
    return delegate.collect(function).toImmutable();
  }
}
```

Compilation of this `Pattern.java` with JDK 8u121 and OpenJDK 9b52 succeeds, but fails with OpenJDK 9b53:
```
Pattern.java:23: error: incompatible types: ImmutableCollection<CAP#1> cannot be converted to ImmutableCollection<V>
    return delegate.collect(function).toImmutable();
                                                 ^
  where V is a type-variable:
    V extends Object declared in method <V>example(ByteToObjectFunction<? extends V>)
  where CAP#1 is a fresh type-variable:
    CAP#1 extends V from capture of ? extends V
1 error
```

According to log of changes in OpenJDK 9 between b52 and b53 (http://hg.openjdk.java.net/jdk9/jdk9/langtools):
```
o  changeset:   2836:01d8ed7079f1
|  user:        katleman
|  date:        Thu Mar 05 11:26:21 2015 -0800
|  summary:     Added tag jdk9-b53 for changeset 99ff00581f36
|
@    changeset:   2835:99ff00581f36
|\   tag:         jdk9-b53
| |  parent:      2829:cdef738241cd
| |  parent:      2834:d6ec687ad8fb
| |  user:        lana
| |  date:        Thu Feb 26 20:17:06 2015 -0800
| |  summary:     Merge
| |
| o  changeset:   2834:d6ec687ad8fb
| |  user:        jjg
| |  date:        Wed Feb 25 14:35:39 2015 -0800
| |  summary:     8041628: Javadoc cross-compilation problem
| |
| o  changeset:   2833:f683944ffa42
| |  user:        jlahoda
| |  date:        Tue Feb 24 16:11:59 2015 +0100
| |  summary:     8067886: Inaccessible nested classes can be incorrectly imported
| |
| o  changeset:   2832:59af0866b952
| |  user:        juh
| |  date:        Mon Feb 23 11:42:16 2015 -0800
| |  summary:     8072663: Remove the sun.security.acl package which is not used in the JDK
| |
| o  changeset:   2831:e29b25f6101f
| |  user:        mcimadamore
| |  date:        Mon Feb 23 13:02:37 2015 +0000
| |  summary:     8071291: Compiler crashes trying to cast UnionType to IntersectionClassType
| |
| o  changeset:   2830:414b82835861
| |  parent:      2828:ee20efe0255d
| |  user:        dlsmith
| |  date:        Fri Feb 20 17:05:13 2015 -0700
| |  summary:     8039214: Inference should not map capture variables to their upper bounds
| |
o |  changeset:   2829:cdef738241cd
|/   user:        katleman
|    date:        Thu Feb 26 15:58:16 2015 -0800
|    summary:     Added tag jdk9-b52 for changeset ee20efe0255d
|
o    changeset:   2828:ee20efe0255d
|\   tag:         jdk9-b52
~ ~  parent:      2822:c5561bfde449
     parent:      2827:5e500700b168
     user:        lana
     date:        Thu Feb 19 16:38:54 2015 -0800
     summary:     Merge
```
this is caused by https://bugs.openjdk.java.net/browse/JDK-8039214 ( http://hg.openjdk.java.net/jdk9/jdk9/langtools/rev/414b82835861 )
Its subticket about release notes (https://bugs.openjdk.java.net/browse/JDK-8173583) states:
> The javac compiler's behavior when handling wildcards and "capture" type variables has been improved for conformance to the language specification. This improves type checking behavior in certain unusual circumstances. It is also a source-incompatible change: certain uses of wildcards that have compiled in the past may fail to compile because of a program's reliance on the javac bug.

Should be noted that this is consistent with what IntelliJ IDEA 2017.1 already reports even without JDK 9:

intellij

as well as Eclipse Compiler for Java in Eclipse Neon.2:

eclipse

After fix for this pattern for future investigations there will be only 7 errors in module `eclipse-collections` and only 2 in module `unit-tests` out of hundreds that appeared in https://github.com/eclipse/eclipse-collections/pull/234#issue-217022672

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/impl/bag/immutable/immutablePrimitiveHashBag.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/bag/immutable/immutablePrimitiveHashBag.stg
@@ -23,6 +23,7 @@ import java.io.Serializable;
 import org.eclipse.collections.api.<name>Iterable;
 import org.eclipse.collections.api.Lazy<name>Iterable;
 import org.eclipse.collections.api.bag.ImmutableBag;
+import org.eclipse.collections.api.bag.MutableBag;
 import org.eclipse.collections.api.bag.primitive.<name>Bag;
 import org.eclipse.collections.api.bag.primitive.Immutable<name>Bag;
 import org.eclipse.collections.api.bag.primitive.Mutable<name>Bag;
@@ -40,7 +41,6 @@ import org.eclipse.collections.impl.block.procedure.checked.primitive.Checked<na
 import org.eclipse.collections.impl.iterator.Unmodifiable<name>Iterator;
 import org.eclipse.collections.api.list.ImmutableList;
 import org.eclipse.collections.api.tuple.primitive.<name>IntPair;
-
 
 /**
  * Immutable<name>HashBag is the non-modifiable equivalent of {@link <name>HashBag}.
@@ -187,7 +187,8 @@ final class Immutable<name>HashBag implements Immutable<name>Bag, Serializable
     @Override
     public \<V> ImmutableBag\<V> collect(<name>ToObjectFunction\<? extends V> function)
     {
-        return this.delegate.collect(function).toImmutable();
+        MutableBag\<V> bag = this.delegate.collect(function);
+        return bag.toImmutable();
     }
 
     @Override
--- a/eclipse-collections-code-generator/src/main/resources/impl/list/immutable/immutablePrimitiveArrayList.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/list/immutable/immutablePrimitiveArrayList.stg
@@ -265,7 +265,8 @@ final class Immutable<name>ArrayList
     @Override
     public \<V> ImmutableList\<V> collect(<name>ToObjectFunction\<? extends V> function)
     {
-        return this.collect(function, FastList.newList(this.items.length)).toImmutable();
+        FastList\<V> list = this.collect(function, FastList.newList(this.items.length));
+        return list.toImmutable();
     }
 
     /**
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutableObjectPrimitiveHashMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutableObjectPrimitiveHashMap.stg
@@ -34,6 +34,7 @@ import org.eclipse.collections.api.block.procedure.primitive.<name>Procedure;
 import org.eclipse.collections.api.block.procedure.primitive.Object<name>Procedure;
 import org.eclipse.collections.api.map.primitive.Object<name>Map;
 import org.eclipse.collections.api.collection.ImmutableCollection;
+import org.eclipse.collections.api.collection.MutableCollection;
 import org.eclipse.collections.api.collection.primitive.Immutable<name>Collection;
 import org.eclipse.collections.api.collection.primitive.Mutable<name>Collection;
 import org.eclipse.collections.api.iterator.<name>Iterator;
@@ -135,7 +136,8 @@ final class ImmutableObject<name>HashMap\<K> extends AbstractImmutableObject<nam
     @Override
     public \<V> ImmutableCollection\<V> collect(<name>ToObjectFunction\<? extends V> function)
     {
-        return this.delegate.collect(function).toImmutable();
+        MutableCollection\<V> collection = this.delegate.collect(function);
+        return collection.toImmutable();
     }
 
     <(arithmeticMethods.(type))()>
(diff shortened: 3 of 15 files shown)


Godin:
I'm wondering if it makes sense to comment all the added casts to prevent regressions in absence of automated build that reveals those errors? Depending on a result of investigations of remaining ones, build with JDK 9 EA might be added.

nikhilnanivadekar:
I don't think we can add an automated build in Travis since they do not have JDK 9-EA as yet. Here are the supported versions: https://docs.travis-ci.com/user/languages/java/#Testing-Against-Multiple-JDKs

> Travis CI provides OpenJDK 6, OpenJDK 7, Oracle JDK 7, and Oracle JDK 8. Sun JDK 6 is not provided, because it is EOL as of November 2012.

Any idea where we can have a regression build with Java 9?

motlin:
Instead of casts, we could just extract local variables.

Godin:
@nikhilnanivadekar according to https://github.com/travis-ci/travis-ci/issues/5520#issuecomment-175914595 there is a way to use Oracle JDK 9 EA in Travis , as a proof - https://travis-ci.org/junit-team/junit5/jobs/215948144 :wink: But seems that it is already quite old - b140. For JaCoCo we simply always download most recent EA build :wink:  More precisely we have two additional elements in matrix - see https://github.com/jacoco/jacoco/commit/c6f2b6b7e887eb645b8aba928ff0134cfe66ec28 :
> * `9-ea` - currently under testing and hence failures are allowed while we investigate them
> * `9-ea-stable` - tested previously and hence failures are not allowed, so that to verify that new changes do not introduce regressions

But useless to do this prior to investigation of remaining issues, because it anyway will be failing due to them and so won't allow to catch any regression.

Godin:
@motlin just to be sure - you prefer
```java
    @Override
    public <V> ImmutableCollection<V> collect(ByteToObjectFunction<? extends V> function)
    {
        MutableCollection<V> collect = this.delegate.collect(function);
        return collect.toImmutable();
    }
```
over cast ?

motlin:
Yes, I think so! My main concern is that a static analysis tool, or IntelliJ, would say that the cast is redundant and helpfully remove it! Since this seems to consistently happen for the pattern `someMethod(function).toImmutable()`, then extracting a local variable called `result` or something similar is just a style choice and wouldn't result in a static analysis error. But I'm not sure if this is still relevant. When we wrote this code, I don't think IntelliJ was highlighting it as an error.

Godin:
> My main concern is that a static analysis tool, or IntelliJ, would say that the cast is redundant and helpfully remove it!

The same way static analysis tool or IDE can say that local variable is redundant and "helpfully" inline it :wink: But hopefully they are becoming better and better from day to day and don't regress.

That actually why I was wondering ( https://github.com/eclipse/eclipse-collections/pull/240#issuecomment-289920024 ) about explicit comment in code and that's why was proposing addition of build with JDK 9 EA in mid-term, which will be helpful to catch such regressions.

Nevertheless introduction of local variable seems clean(er). PR updated.

.idea/misc.xml
```diff
@@ -49,7 +56,7 @@
   <component name="ProjectResources">
     <default-html-doctype>http://www.w3.org/1999/xhtml</default-html-doctype>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" assert-keyword="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="1.8" project-jdk-type="JavaSDK">
```
motlin: Looks good overall, and generally it's ok to sync IDE project files and commit them. I'm just wondering about this one removal of assert-keyword="true". I'm going to try to get my IDE to do that locally. I just want to make sure it doesn't mean we run tests with the assert keyword disabled (by removing the -ea flag).

Godin: @motlin ouch, sorry, in fact I didn't meant to commit any IDE files and was quite annoyed that they were showing up as modified (maybe relates to #199), so just overlooked this change - let me revert IDE files back

Issue #213, opened 2017-03-09
OrderedIterable::max/min/first to return Optional instead of throwing exception


Ziemien:
Would be possible for:
OrderedIterable::max, min, getFirst, getLast etc. to return Optional<T> instead of throwing runtime exception? That would provide more concise programming experience.

motlin:
We wouldn't break the backwards compatibility of min(), max(), getFirst(), getLast(), but we'd add alternatives that return Optional. The idea is covered (partially) by #122.

Ziemien:
perfect.

nikhilnanivadekar:
This is covered by #122 and #123 . You can follow those issues.
Closing this as redundant.

Issue #122, opened 2016-07-01
Implement minOptional, maxOptional, minByOptional and maxByOptional on RichIterable.


itohro:
Implement min/max methods that returns Optional on RichIterable. The method signatures would look like below.

`Optional<T> minOptional(Comparator<? super T> comparator)`
`Optional<T> maxOptional(Comparator<? super T> comparator)`
`Optional<T> minOptional()`
`Optional<T> maxOptional()`
`<V extends Comparable<? super V>> Optional<T> minByOptional(Function<? super T, ? extends V> function)`
`<V extends Comparable<? super V>> Optional<T> maxByOptional(Function<? super T, ? extends V> function)`

PR #244, opened 2017-03-29
Implement OrderedIterable.getFirstOptional() and OrderedIterable.getL…

### LINKED ISSUE #123: Implement OrderedIterable.getFirstOptional() and OrderedIterable.getLastOptional() .
Implement getFirst and getLast methods that returns Optional on OrderedIterable. The method signatures would look like below.

`Optional<T> getFirstOptional()`
`Optional<T> getLastOptional()`

PascalSchumacher:
…astOptional() (closes #123)

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/ordered/OrderedIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/ordered/OrderedIterable.java
@@ -13,6 +13,7 @@ package org.eclipse.collections.api.ordered;
 import java.util.Collection;
 import java.util.List;
 import java.util.NoSuchElementException;
+import java.util.Optional;
 
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.block.function.Function;
@@ -67,6 +68,23 @@ public interface OrderedIterable<T> extends RichIterable<T>
     T getFirst();
 
     /**
+     * Returns the first element of an iterable as an Optional. In the case of a List it is the element at the first index.
+     * In the case of any other Collection, it is the first element that would be returned during an iteration. If the
+     * iterable is empty, {@link Optional#empty} is returned.
+     *
+     * @throws NullPointerException if the element is null
+     * @since 9.0
+     */
+    default Optional<T> getFirstOptional()
+    {
+        if (this.isEmpty())
+        {
+            return Optional.empty();
+        }
+        return Optional.of(this.getFirst());
+    }
+
+    /**
      * Returns the last element of an iterable.  In the case of a List it is the element at the last index.  In the case
      * of any other Collection, it is the last element that would be returned during an iteration.  If the iterable is
      * empty, null is returned.  If null is a valid element of the container, then a developer would need to check to
@@ -74,6 +92,23 @@ public interface OrderedIterable<T> extends RichIterable<T>
      */
     @Override
     T getLast();
+
+    /**
+     * Returns the last element of an iterable as an Optional. In the case of a List it is the element at the last index.
+     * In the case of any other Collection, it is the last element that would be returned during an iteration. If the
+     * iterable is empty, {@link Optional#empty} is returned.
+     *
+     * @throws NullPointerException if the element is null
+     * @since 9.0
+     */
+    default Optional<T> getLastOptional()
+    {
+        if (this.isEmpty())
+        {
+            return Optional.empty();
+        }
+        return Optional.of(this.getLast());
+    }
 
     @Override
     OrderedIterable<T> tap(Procedure<? super T> procedure);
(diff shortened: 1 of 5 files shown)


nikhilnanivadekar:
In order to close #123, we should implement OrderedIterable.maxOptional() and OrderedIterable.minOptional(). It can be a separate commit, but the implementation will touch the same classes.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/ordered/OrderedIterable.java
```diff
@@ -67,6 +68,17 @@
...
 
     /**
+     * Returns the first element of an iterable as an Optional. In the case of a List it is the element at the first index. In the
+     * case of any other Collection, it is the first element that would be returned during an iteration. If the
+     * iterable is empty, {@link Optional#empty} is returned. If null is a valid element of the container, then a developer would need to
+     * check to see if the iterable is empty to validate that an {@link Optional#empty} result was not due to the container being empty.
+     */
+    default Optional<T> getFirstOptional()
+    {
+        return Optional.ofNullable(getFirst());
```
nikhilnanivadekar: Please qualify with `this`:
`this.getFirst()`

motlin:
@nikhilnanivadekar I'm not sure I understand the point about maxOptional() / minOptional(). That method doesn't exist today, does it?

nikhilnanivadekar:
@motlin you are right. I was thinking about having maxOptional() and minOptional() to be just null safe. If the user has a Comparator with nulls high or nulls low respectively. But I agree with what you are hinting at, and we do not need them.

itohro:
@nikhilnanivadekar @motlin FYI we hane another issue for maxOptional/minOptional in #122 .

motlin:
Right, thanks @itohro. I wasn't hinting at anything @nikhilnanivadekar I just thought it would be a separate piece of work in a separate pull request.

nikhilnanivadekar:
If I understand correctly, this PR will partially close #213.
May be we should clean up #213 . 
Apologies for the very first comment, I got confused between #123 and #213

motlin:
I think that #213 is redundant because the same feature requests are
covered in other issues.

motlin:
Feedback similar to the discussion on #239, and it reminds me of the work @itohro did on detectOptional(). We realized that detectOptional() is more expressive than detect(), so detect() could be rewritten to delegate to detectOptional(), but not the other way around. The problem is that detect() and getFirst() can return null for two reasons, either the found element is null or the no element was found.

If we attempt to go ahead and delegate from getFirstOptional to getFirst in a default implementation, then we at least need a size check, just like the example I wrote on #239. That way, an empty collection would return Optional.empty() and a collection with the first/last element null would throw NullPointerException by attempting Optional.of(null).

PascalSchumacher:
Thanks. Than makes sense. I will update the pull request.

I really should have checked out the implementation of `detectOptional` before submitting this pull request. Sorry for wasting your time.

motlin:
Absolutely not a waste of time. The new patch looks better. It has some CheckStyle violations which I fixed in fe4748a4e65f378c3466e3560b77c4878bab56f3. It's currently [building](https://travis-ci.org/motlin/eclipse-collections/builds/217423357). If the build passes, you could include these changes in your pull request.

PascalSchumacher:
Thanks! I have updated the pull request to include your fixes.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/ordered/OrderedIterable.java
```diff
@@ -67,6 +68,23 @@
     T getFirst();
 
     /**
+     * Returns the first element of an iterable as an Optional. In the case of a List it is the element at the first index.
+     * In the case of any other Collection, it is the first element that would be returned during an iteration. If the
+     * iterable is empty, {@link Optional#empty} is returned.
+     *
+     * @throws NullPointerException if the element is null<
```
motlin: There's an extra trailing angle bracket on this line.

motlin:
Thank you very much!

PascalSchumacher:
Thanks for merging! :)

Is the since tag I used (`9.0`) correct?

nikhilnanivadekar:
@PascalSchumacher Thanks for bringing it to attention. It should be 8.2.0 which is coming up by the end of April. 
Can you please raise a PR? I will merge it.

PascalSchumacher:
On further consideration: The default methods are broken for synchronized classes. :( There is no guarantee that the collection is still non-empty when `getFirst` is called by `getFirstOptional`.

motlin:
Great point. We have some tests in the scala module that can catch this
sort of problem if we add new tests along with the older ones for getFirst
and getLast.

PascalSchumacher:
I added tests for `getFirstOptional` and `getLastOptional` to `SynchronizedStackScalaTest`, but they are unable to demonstrate a problem. The test have two threads access `getFirstOptional`/`getLastOptional` in parallel and this case safe.

PR #249, opened 2017-04-01
Fix compilation issues.

nikhilnanivadekar:
I think these are last of the compilation issues with JDK-9-EA-163.
Few notes:
`Functions`, `Predicates`, `Procedures`, `Verify`: Diamond operator is not able to infer the type arguments, so mentioning the types explicitly fixes the compilation problem.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/parallel/set/ParallelCollectIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/parallel/set/ParallelCollectIterable.java
@@ -14,6 +14,7 @@ import java.util.concurrent.ExecutorService;
 
 import org.eclipse.collections.api.LazyIterable;
 import org.eclipse.collections.api.annotation.Beta;
+import org.eclipse.collections.api.bag.MutableBag;
 import org.eclipse.collections.api.block.function.Function;
 import org.eclipse.collections.api.block.predicate.Predicate;
 import org.eclipse.collections.api.block.procedure.Procedure;
@@ -98,20 +99,23 @@ public class ParallelCollectIterable<T, V> extends AbstractParallelIterableImpl<
     public <V1> UnsortedBagMultimap<V1, V> groupBy(Function<? super V, ? extends V1> function)
     {
         // TODO: Implement in parallel
-        return this.delegate.toBag().collect(this.function).groupBy(function);
+        MutableBag<V> mutableBag = this.delegate.toBag().collect(this.function);
+        return mutableBag.groupBy(function);
     }
 
     @Override
     public <V1> UnsortedBagMultimap<V1, V> groupByEach(Function<? super V, ? extends Iterable<V1>> function)
     {
         // TODO: Implement in parallel
-        return this.delegate.toBag().collect(this.function).groupByEach(function);
+        MutableBag<V> mutableBag = this.delegate.toBag().collect(this.function);
+        return mutableBag.groupByEach(function);
     }
 
     @Override
     public <V1> MapIterable<V1, V> groupByUniqueKey(Function<? super V, ? extends V1> function)
     {
         // TODO: Implement in parallel
-        return this.delegate.toBag().collect(this.function).groupByUniqueKey(function);
+        MutableBag<V> mutableBag = this.delegate.toBag().collect(this.function);
+        return mutableBag.groupByUniqueKey(function);
     }
 }
--- a/eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
+++ b/eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
@@ -1669,7 +1669,7 @@ public final class Verify extends Assert
     {
         try
         {
-            Assert.assertTrue(message, Predicates.anySatisfy(predicate).accept(iterable));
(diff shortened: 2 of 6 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/parallel/set/ParallelCollectIterable.java
```diff
@@ -98,20 +98,20 @@ public V detect(Predicate<? super V> predicate)
     public <V1> UnsortedBagMultimap<V1, V> groupBy(Function<? super V, ? extends V1> function)
     {
         // TODO: Implement in parallel
-        return this.delegate.toBag().collect(this.function).groupBy(function);
+        return (UnsortedBagMultimap<V1, V>) this.delegate.toBag().collect(this.function).groupBy(function);
```
motlin: You tried extracting all three intermediate results as local variables? Are you confident this isn't just a bug in javac?

nikhilnanivadekar: I was missing extracting the output of `collect()`. Should be good now.

nikhilnanivadekar:
IntelliJ was not marking the errors where I had to explicitly mention generic types. Yes IntelliJ marks these as redundant and code fix prompts to remove that.

motlin:
I'm curious if the fact that this code used to compile is a spec violation
like the previous fix. If not, this is going to be an ongoing problem where
the auto fix removes the redundant types.

nikhilnanivadekar:
I will try to compile it in Eclipse and see if it is a javac issue or an IntelliJ issue. 
Any case though we can always suppress that check on specific lines in IntelliJ before merging, so that the auto-fix will not remove the redundant types.

motlin:
This all assumes javac 9 is correct. It may just be a bug we should report
to Oracle and not suppress. It's why the spec is important in this case.
Last time it was more obvious because javac 8 was the odd one out with
javac 9, ecj, and IntelliJ all agreeing that the code was invalid.

nikhilnanivadekar:
Yes that's why I want to check with ecj if that is an issue there too. If not then I will investigate the spec and we can report the bug to either Oracle or JetBrains and Eclipse. I will post the investigations later today. 
In the mean time, I am going to separate the commit and merge `ParallelCollectIterable`. Thoughts?

nikhilnanivadekar:
Verify: No inspection error.

Functions: Inspection error false positive
image

image

Predicates: Inspection error, to investigate
image

Procedures: No inspection errors.

AnagramTest: Inspection error, to investigate
image

nikhilnanivadekar:
This latest commit has those fixes which IntelliJ shows and javac also shows.

motlin:
Interesting! An inspection error where the auto-fix results in code that's
a compiler error according to the ide. We should report this to JetBrains.

nikhilnanivadekar:
Yes I will be doing to later today. Want to verify others and do together

nikhilnanivadekar:
For the `AnagramsTest` seems that it is an issue in IntelliJ. Javac is doing the correct thing.
The compilation error is as below:
```
[ERROR] C:\EclipseCollections\eclipse-collections\unit-tests\src\test\java\org\eclipse\collections\impl\AnagramTest.java:[120,39] error: incompatible types: cannot infer type-variable(s) T
[ERROR] (argument mismatch; invalid method reference
[ERROR] no suitable method found for info(Object)
[ERROR] method Logger.info(String) is not applicable
[ERROR] (argument mismatch; Object cannot be converted to String)
[ERROR] method Logger.info(String,Object...) is not applicable
[ERROR] (argument mismatch; Object cannot be converted to String)
[ERROR] method Logger.info(Marker,String,Object...) is not applicable
[ERROR] (argument mismatch; Object cannot be converted to Marker))
[ERROR] where T is a type-variable:
[ERROR] T extends Object declared in method <T>cast(Procedure<T>)
[ERROR] C:\EclipseCollections\eclipse-collections\unit-tests\src\test\java\org\eclipse\collections\impl\AnagramTest.java:[132,38] error: incompatible types: cannot infer type-variable(s) T
```

Here is a similar bug report for reference: https://bugs.openjdk.java.net/browse/JDK-8143306
Quoting from there:
> If the method reference expression has the form ExpressionName :: 
> [TypeArguments] Identifier or Primary :: [TypeArguments] Identifier, the 
> type to search is the type of the expression preceding the :: token 

In the case of `AnagramsTest` the issue is at `Procedures.cast(Logger::info)`, so the compiler is not able to figure out the type of `Procedure<String>`. Extracting that to local variable works fine. Pushed the commit.

Interestingly Eclipse does not point this as an issue.

nikhilnanivadekar:
Also, referring to `Functions` the compile errors are correct, so from that we can infer the compile errors for `Predicates` are also correct. This seems to be another IntelliJ and Eclipse issue. I will try to reproduce these and report accordingly.

nikhilnanivadekar:
Reported 2 bugs to JetBrains
1) [IDEA-170813](https://youtrack.jetbrains.com/issue/IDEA-170813)
2) [IDEA-170814](https://youtrack.jetbrains.com/issue/IDEA-170814)

motlin:
Thanks for raising these. I'd like to wait for a response from JetBrains
before merging.

nikhilnanivadekar:
Reported bug to Eclipse: [Bug-514605](https://bugs.eclipse.org/bugs/show_bug.cgi?id=514605)

PR #253, opened 2017-04-02
Implement fixes such that library is compatible with Java 8 and Java 9-EA-163.

nikhilnanivadekar:


eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/SerializableDoubleSummaryStatistics.java
```diff
@@ -61,14 +63,33 @@
...
+                min = null;
+                max = null;
+            }
+            else
+            {
+                try
+                {
+                    throw e;
+                }
+                catch (NoSuchFieldException ignored)
```
motlin: I  don't understand this part. Didn't we just catch the same exception?

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ArrayListIterate.java
```diff
@@ -109,10 +109,22 @@
...
+            catch (Exception exception)
             {
-                data = null;
-                size = null;
+                if (exception instanceof SecurityException)
+                {
+                    data = null;
+                    size = null;
+                }
+                else if (exception.getClass().getCanonicalName().equals("java.lang.reflect.InaccessibleObjectException"))
```
motlin: Combine the if and the else if?

unit-tests/src/test/java/org/eclipse/collections/impl/collector/SerializableLongSummaryStatisticsTest.java
```diff
@@ -29,6 +30,7 @@ public void valuesEqual()
         Assert.assertTrue(with.valuesEqual(without));
     }
 
+    @Ignore
```
motlin: Instead of ignoring or deleting, we could add an Assume statement using the system property for the version of Java.

unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayListTest.java
```diff
@@ -186,11 +186,17 @@ public void get()
...
     }
 
-    @Test(expected = IllegalStateException.class)
     @Override
     public void iteratorRemove()
     {
-        this.classUnderTest().iterator().remove();
+        try
+        {
+            this.classUnderTest().iterator().remove();
```
motlin: You need to call Assert.fail() in the try block to assert that an exception in fact gets thrown.

unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayListTest.java
```diff
@@ -186,11 +186,17 @@ public void get()
         Verify.assertThrows(ArrayIndexOutOfBoundsException.class, () -> list.get(-1));
     }
 
-    @Test(expected = IllegalStateException.class)
     @Override
```
motlin: Add back the @Test annotation?

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/AbstractListTestCase.java
```diff
@@ -827,7 +841,14 @@ public void getWithArrayIndexOutOfBoundsException()
     {
         Object item = new Object();
 
-        Verify.assertThrows(ArrayIndexOutOfBoundsException.class, () -> this.newWith(item).get(-1));
+        try
+        {
+            this.newWith(item).get(-1);
```
motlin: Assert.fail()

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/SerializableDoubleSummaryStatistics.java
nikhilnanivadekar: Yes. If you see previously the same exception was caught. If I remove the `throw` then it is not necessary, but it is incorrect, we should throw.
Also, it is a compiler error.

motlin:
I'll take a closer look when I'm at my computer. Thanks!

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/SerializableDoubleSummaryStatistics.java
motlin: What about this?
```
    static
    {
        Field count;
        Field sum;
        Field sumCompensation;
        Field simpleSum;
        Field min;
        Field max;
        try
        {
            count = DoubleSummaryStatistics.class.getDeclaredField("count");
            count.setAccessible(true);
            sum = DoubleSummaryStatistics.class.getDeclaredField("sum");
            sum.setAccessible(true);
            sumCompensation = DoubleSummaryStatistics.class.getDeclaredField("sumCompensation");
            sumCompensation.setAccessible(true);
            simpleSum = DoubleSummaryStatistics.class.getDeclaredField("simpleSum");
            simpleSum.setAccessible(true);
            min = DoubleSummaryStatistics.class.getDeclaredField("min");
            min.setAccessible(true);
            max = DoubleSummaryStatistics.class.getDeclaredField("max");
            max.setAccessible(true);
        }
        catch (Exception ignored)
        {
            count = null;
            sum = null;
            sumCompensation = null;
            simpleSum = null;
            min = null;
            max = null;
        }
        COUNT = count;
        SUM = sum;
        SUM_COMPENSATION = sumCompensation;
        SIMPLE_SUM = simpleSum;
        MIN = min;
        MAX = max;
    }
```

nikhilnanivadekar: Fair enough. I will change it. Any other feedback so that I can push the changes together?

nikhilnanivadekar:
@motlin as discussed offline, I have implemented all the suggested changes. Can you please review?

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/SerializableDoubleSummaryStatistics.java
```diff
@@ -61,7 +63,12 @@
             max = DoubleSummaryStatistics.class.getDeclaredField("max");
             max.setAccessible(true);
         }
-        catch (SecurityException | NoSuchFieldException ignored)
+        catch (RuntimeException e)
```
motlin: I believe catching RuntimeException is actually a regression. We should have a test that fails due to this code, perhaps using a SecurityManager. We can delete this block entirely and just catch Exception. It's one of those rare situations where catching Exception feels right, because it executes in a static block.

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ArrayListIterate.java
```diff
@@ -109,7 +109,7 @@
                 data.setAccessible(true);
                 size.setAccessible(true);
             }
-            catch (SecurityException ignored)
+            catch (Exception exception)
```
motlin: Same here. At first glance, it looks like we're treating the two kinds of exceptions differently because one silently ignores and the other sets to null. But the fields were initialized to null.

unit-tests/src/test/java/org/eclipse/collections/impl/map/fixed/TripletonMapTest.java
```diff
@@ -360,7 +361,18 @@ public void testEqualsAndHashCode()
...
     {
         MutableMap<Integer, String> map = new TripletonMap<>(1, "One", 2, "Two", 3, "Three");
-        Verify.assertShallowClone(map);
+        try
+        {
+            Verify.assertShallowClone(map);
+        }
+        catch (Exception e)
+        {
+            // Suppress if a Java 9 specific exception related to reflection is thrown.
```
motlin: Would it make sense to use Assume or check the system property for the Java version? Just asking, maybe it's overkill.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/SerializableDoubleSummaryStatistics.java
nikhilnanivadekar: The find bugs build fails when only an Exception is caught.
http://findbugs.sourceforge.net/bugDescriptions.html#REC_CATCH_EXCEPTION

motlin:
https://github.com/eclipse/eclipse-collections/blob/master/findbugs-exclude.xml

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ArrayListIterate.java
nikhilnanivadekar: I agree, it looks awkward. I can pull the try-catch block up and just catch an Exception to ignore it in the outer block. Is this what you are suggesting?

unit-tests/src/test/java/org/eclipse/collections/impl/map/fixed/TripletonMapTest.java
nikhilnanivadekar: That would be an overkill because there is a check for Java 9 specific exception being thrown. Any regression breaks should fail the test regardless of Java version.

motlin:
yes

nikhilnanivadekar:
@motlin Fixed all the review comments.

PR #255, opened 2017-04-02
Upgrade versions of several libraries.

motlin:
* checkstyle
* guava
* hppc
* jmh
* logback
* slf4j


nikhilnanivadekar:
@motlin the previous [build](https://travis-ci.org/eclipse/eclipse-collections/builds/217819433) had compiler errors since Travis was using an older version of Java. I have committed a fix to install latest JDK before running the build.

nikhilnanivadekar:
Here is how the Java version changes:
Before (stock):

> Maven home: /usr/local/maven
> Java version: 1.8.0_31, vendor: Oracle Corporation
> Java home: /usr/lib/jvm/java-8-oracle/jre

After:

> $ java -Xmx32m -version
> java version "1.8.0_121"
> Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
> Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)

nikhilnanivadekar:
@motlin can you please review the Travis yaml, so that I can merge this.

motlin:
Interesting, this is new to me. Does it affect all builds or just the Java
8 build? I'm fine with it either way, just curious.

nikhilnanivadekar:
Sorry I didn't follow the question. All our builds use Java 8, so it will affect all builds. Travis will only install the newest version of Java 8.

motlin:
Great! And it looks like we can use more memory now.

nikhilnanivadekar:
Yes memory is increased to 2g with another set of upgrades which are needed for maven and maven plugins to support Java 9 builds.

PR #256, opened 2017-04-02
Fix sonar properties.

nikhilnanivadekar:


scala-unit-tests/pom.xml
```diff
@@ -27,6 +27,7 @@
 
     <properties>
         <maven.deploy.skip>true</maven.deploy.skip>
+        <sonar.skip>true</sonar.skip>
```
motlin: Besides the Scala tests module, why would we skip sonar on any modules?

nikhilnanivadekar:
Why do you think we should run sonar analysis on acceptance, jmh tests and
code generators?

On Apr 2, 2017 7:46 PM, "Craig P. Motlin" <notifications@github.com> wrote:

*@motlin* commented on this pull request.
------------------------------

In scala-unit-tests/pom.xml
<https://github.com/eclipse/eclipse-collections/pull/256#discussion_r109328091>
:

> @@ -27,6 +27,7 @@

     <properties>
         <maven.deploy.skip>true</maven.deploy.skip>
+        <sonar.skip>true</sonar.skip>

Besides the Scala tests module, why would we skip sonar on any modules?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/eclipse/eclipse-collections/pull/256#pullrequestreview-30422119>,
or mute the thread

motlin:
I'm not convinced we should run Sonar at all! But I'm a fan of static analysis in general and I don't make a distinction between test modules and prod modules. I try not to even make a distinction between generated code and hand-written code if possible. If we run sonar on the unit-tests module which is big, why not these other modules?

nikhilnanivadekar:
That's a fair point. In that case I'll have to figure out a way to ignore
code-generator and jmh test module from test coverage.

nikhilnanivadekar:
BTW, sonar runs on generated code as well :)

motlin:
Ok I see why JMH has to be excluded from coverage. Why the code generator?
Because it will show up as 0% covered (because it only runs as part of the
build)?

nikhilnanivadekar:
Yes exactly. To cover code generators we will have to write useless tests and basically compare java files which get generated with expected. This will lead to duplication and maintenance challenge.

nikhilnanivadekar:
You can see the coverage tree map here: https://sonarqube.com/component_measures/metric/coverage/treemap?id=org.eclipse.collections%3Aeclipse-collections-parent

junit-trait-runner/pom.xml
```diff
@@ -27,6 +27,7 @@
 
     <properties>
         <maven.deploy.skip>true</maven.deploy.skip>
+        <sonar.skip>true</sonar.skip>
```
motlin: Any idea why this modules shows as uncovered? It does run as part of the build, but maybe the fact that it's invoked by JUnit code through reflection prevents coverage information from being gathered.

Anyway it's two classes in a tangential modules. I'll go ahead and merge.

PR #257, opened 2017-04-02
Upgrade Maven, plugins.

nikhilnanivadekar:


pom.xml
```diff
@@ -284,7 +283,7 @@
                                         <version>1.8.0</version>
                                     </requireJavaVersion>
                                     <requireMavenVersion>
-                                        <version>3.1.0</version>
+                                        <version>3.3.9</version>
```
motlin: Why would we require contributors to use Maven 3.3.9?

.travis.yml
```diff
@@ -5,7 +5,13 @@ addons:
     packages:
       - oracle-java8-installer
 
-before_install: echo "MAVEN_OPTS='-Dmaven.repo.local=$HOME/.m2/repository -Xmx1g -Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss:SSS'" > ~/.mavenrc
+before_install:
+  - wget https://archive.apache.org/dist/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.zip
+  - unzip -qq apache-maven-3.3.9-bin.zip
+  - export M2_HOME=$PWD/apache-maven-3.3.9
+  - export PATH=$M2_HOME/bin:$PATH
+  - echo "MAVEN_OPTS='-Dmaven.repo.local=$HOME/.m2/repository -Xmx2g -Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss:SSS'" > ~/.mavenrc
```
motlin: Why do this? Can't we build with any recent version of maven?

nikhilnanivadekar: The stock version of Travis is 3.2. This is part of building with Java 9.

pom.xml
nikhilnanivadekar: Not necessary, but if anyone wants to build with javac of Java 9 they'll need to upgrade.
I am looking at it from adding a regression build with Java 9.

motlin: If you want to build using Java 8 and a relatively recent version of maven that's not the absolute latest, I think you should be able to.

.travis.yml
motlin: What does the Java 9 build require exactly? Some new maven plugins which depend on a newer version of maven?

nikhilnanivadekar:
Yes it'll need newer versions of plugins. Maven support is still in beta
mode, so I am fine dropping that commit. Does the plugins upgrade look ok?

nikhilnanivadekar:
Ok. How about I keep Travis build with latest maven and remove the enforcer
check?

motlin:
The commit to upgrade plugin versions looks fine. If possible it would be
great to only use a downloaded copy of maven for the Java 9 build.

nikhilnanivadekar:
Rebased and dropped the commit for Maven upgrade.

eclipse-collections-code-generator-maven-plugin/pom.xml
```diff
@@ -97,6 +97,11 @@
                 <artifactId>maven-enforcer-plugin</artifactId>
             </plugin>
 
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-plugin-plugin</artifactId>
+                <version>3.5</version>
```
motlin: Just noticed this in the plugins section. Shouldn't it go in pluginManagement for consistency? Seriously asking - not sure.

pom.xml
```diff
@@ -133,7 +133,6 @@
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
         <maven.compiler.fork>true</maven.compiler.fork>
-        <maven.compiler.verbose>true</maven.compiler.verbose>
```
motlin: Also just wondering about removing this? I think I added it when we were having out-of-memory problems with javac but I'm not sure.

nikhilnanivadekar: By default it is false. You are correct, we can remove this. The new maven compiler plugin generates too much log > 4 Mb and Travis fails the build automatically.

eclipse-collections-code-generator-maven-plugin/pom.xml
nikhilnanivadekar: Moved this plugins management section. Pushed new commit, the build is passing.

PR #258, opened 2017-04-03

motlin:


motlin:
I shouldn't have written CheckStyle violations in my commit message. I haven't figured out how to get CheckStyle to fail on these. I thought it would, using the whitespace-around rule, but it doesn't. And CheckStyle can't check unused imports.

PR #259, opened 2017-04-03
Implement OrderedIterable.getFirstOptional() and OrderedIterable.getL…

PascalSchumacher:
…astOptional() (#123, #244)

Fix concurrency for synchronized classes and MultiReaderFastList.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/mutable/SynchronizedSortedBag.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/mutable/SynchronizedSortedBag.java
@@ -14,6 +14,7 @@ import java.io.Serializable;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 
 import net.jcip.annotations.GuardedBy;
@@ -654,5 +655,23 @@ public class SynchronizedSortedBag<T>
     public ParallelBag<T> asParallel(ExecutorService executorService, int batchSize)
     {
         throw new UnsupportedOperationException("asParallel() method is not supported for " + this.getClass().getSimpleName() + '.');
+    }
+
+    @Override
+    public Optional<T> getFirstOptional()
+    {
+        synchronized (this.getLock())
+        {
+            return this.getDelegate().getFirstOptional();
+        }
+    }
+
+    @Override
+    public Optional<T> getLastOptional()
+    {
+        synchronized (this.getLock())
+        {
+            return this.getDelegate().getLastOptional();
+        }
     }
 }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
@@ -20,6 +20,7 @@ import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
+import java.util.Optional;
 import java.util.Random;
 import java.util.RandomAccess;
 import java.util.concurrent.ExecutorService;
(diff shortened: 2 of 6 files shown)


PascalSchumacher:
Two questions:

`SynchronizedStack#getLast` is not synchronized, so I did add a synchronized `getLastOptional` method. Is this correct?
`SynchronizedSortedSet` has `first`/`last` methods. I guess I should add `firstOptional`/`lastOptional`?

eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/mutable/SynchronizedStack.java
```diff
@@ -579,6 +579,15 @@ public T getFirst()
...
     @Override
+    public Optional<T> getFirstOptional()
+    {
+        synchronized (this.lock)
+        {
+            return this.delegate.getFirstOptional();
+        }
+    }
+
+    @Override
```
motlin: This one needs an implementation of getLastOptional() too, right?

PascalSchumacher:
Thanks you very much for the detailed explanation. :=)!

`SynchronizedStack#getLastOptional` was missing because `SynchronizedStack#getLast` is not synchronized (see me first question ;)). I missed that these methods just throw a `UnsupportedOperationException`, so they are not required to be synchronized.

I have updated the pull request to include `SynchronizedStack#getLastOptional`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/mutable/SynchronizedSortedBag.java
```diff
@@ -655,4 +656,22 @@ public int detectLastIndex(Predicate<? super T> predicate)
...
+    }
+
+    @Override
+    public Optional<T> getLastOptional()
+    {
+        synchronized (this.getLock())
+        {
+            return this.getDelegate().getLastOptional();
+        }
+    }
```
motlin: I think these can be pulled up to `AbstractSynchronizedRichIterable`. It's hard to see in GitHub, but in general these methods should be right next to getFirst()/getLast() if they're at the right level of abstraction.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
```diff
@@ -793,6 +794,34 @@ public T get(int index)
...
+        {
+            return this.getDelegate().getLastOptional();
+        }
+        finally
+        {
+            this.unlockReadLock();
+        }
+    }
+
+    @Override
```
motlin: Same here, these can be pulled up to `AbstractMultiReaderMutableCollection`

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/mutable/SynchronizedSortedBag.java
PascalSchumacher: `AbstractSynchronizedRichIterable` does not implement `OrderedIterable` (only `RichIterable`). I added `get(First|Last)Optional` to `OrderedIterable` (as requested), while `get(First|Last)` are part of `RichIterable`. I think we would have to add another class e.g. `AbstractSynchronizedOrderedIterable` to have a central place to add `get(First|Last)Optional` for the synchronized classes.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
PascalSchumacher: `AbstractMultiReaderMutableCollection` does not implement `OrderedIterable` (see above).

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/mutable/SynchronizedSortedBag.java
motlin: Ah yes, that is a good point. I think AbstractSynchronizedOrderedIterable is a good idea, but it can come later.

PR #261, opened 2017-04-04
Implement minOptional, maxOptional, minByOptional and maxByOptional o…

### LINKED ISSUE #122: Implement minOptional, maxOptional, minByOptional and maxByOptional on RichIterable.

PascalSchumacher:
…n RichIterable (closes #122)

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
@@ -1562,6 +1562,38 @@ public interface RichIterable<T>
     T max(Comparator<? super T> comparator);
 
     /**
+     * Returns the minimum element out of this container based on the comparator as an Optional.
+     * If the container is empty {@link Optional#empty()} is returned.
+     *
+     * @throws NullPointerException if the minimum element is null
+     * @since 8.2
+     */
+    default Optional<T> minOptional(Comparator<? super T> comparator)
+    {
+        if (this.isEmpty())
+        {
+            return Optional.empty();
+        }
+        return Optional.of(this.min(comparator));
+    }
+
+    /**
+     * Returns the maximum element out of this container based on the comparator as an Optional.
+     * If the container is empty {@link Optional#empty()} is returned.
+     *
+     * @throws NullPointerException if the maximum element is null
+     * @since 8.2
+     */
+    default Optional<T> maxOptional(Comparator<? super T> comparator)
+    {
+        if (this.isEmpty())
+        {
+            return Optional.empty();
+        }
+        return Optional.of(this.max(comparator));
+    }
+
+    /**
      * Returns the minimum element out of this container based on the natural order.
      *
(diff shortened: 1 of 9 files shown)


PascalSchumacher:
Note that #122 suggested `minByOptional` and `maxByOptional`, but I added `minOptionalBy` and `maxOptionalBy` instead.

motlin:
Interesting point about the name. We have a method named detectWithOptional. I think to be consistent, it would have to be called detectOptionalWith. I'd ask the whole group to chime in about the name. Once it's in, it's hard to change.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -1562,6 +1562,38 @@ default T getOnly()
...
+     * @throws NullPointerException if the minimum element is null
+     * @since 8.2
+     */
+    default Optional<T> minOptional(Comparator<? super T> comparator)
+    {
+        if (this.isEmpty())
+        {
+            return Optional.empty();
+        }
+        return Optional.of(min(comparator));
```
nikhilnanivadekar: Please qualify with `this`. this.min()

nikhilnanivadekar:
We should be consistent with detectWithOptional() i.e. minByOptional(). Our users know the *By() methods take a function. While looking at auto-complete option they'll see the methods in this order:
min(), minBy(), minOptional() and minByOptional(). 
The above looks symmetrical as compared to:
min(), minBy(), minOptional() and minOptionalBy(). 
So, I vote for minByOptional().

PascalSchumacher:
I missed `detectWithOptional`. I agree it's important to be consistent.

Just tell me the final decision and I will happily update the pull request.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
PascalSchumacher: Sorry, to get this wrong again. :(

motlin:
I think I also prefer minByOptional/maxByOptional. I think the point @nikhilnanivadekar made about them appearing alphabetically when you press ctrl+space is a good one. Because in English I'd probably say "optionally return the min by ..." but we don't put the word optional at the beginning because it wouldn't help much with auto-complete. Want to jump in on the naming conversation @donraab?

Once we figure out naming, the PR looks great - ready to go. Thanks again!

PascalSchumacher:
Thanks for the feedback! I have changed the method names to `minByOptional` and `maxByOptional`.

PR #262, opened 2017-04-05
Issue 196 primitive multireader

jankotek:
This pull request adds MultiReader wrappers for primitive lists. Other collections will follow in future, but this is all for now.

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/impl/collection/mutable/abstractMultiReaderPrimitiveCollection.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/collection/mutable/abstractMultiReaderPrimitiveCollection.stg
@@ -1,0 +1,554 @@
+import "copyright.stg"
+import "primitiveEquals.stg"
+import "primitiveHashCode.stg"
+import "primitiveLiteral.stg"
+
+targetPath() ::= "org/eclipse/collections/impl/collection/mutable/primitive"
+
+fileName(primitive) ::= "AbstractMultiReader<primitive.name>Collection"
+
+class(primitive) ::= <<
+<body(primitive.type, primitive.name)>
+>>
+
+readLock() ::= <<
+this.acquireReadLock();
+try
+{
+>>
+
+writeLock() ::= <<
+this.acquireWriteLock();
+try
+{
+>>
+
+readUnlock() ::= <<
+}
+finally
+{
+    this.unlockReadLock();
+}
+>>
+
+writeUnlock() ::= <<
+}
+finally
+{
+    this.unlockWriteLock();
+}
+>>
+
+body(type, name) ::= <<
+<copyright()>
+
+package org.eclipse.collections.impl.collection.mutable.primitive;
+
+import java.io.Serializable;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import org.eclipse.collections.api.<name>Iterable;
+import org.eclipse.collections.api.Lazy<name>Iterable;
+import org.eclipse.collections.api.bag.primitive.Mutable<name>Bag;
+import org.eclipse.collections.api.block.function.primitive.<name>ToObjectFunction;
+import org.eclipse.collections.api.block.function.primitive.Object<name>ToObjectFunction;
+import org.eclipse.collections.api.block.predicate.primitive.<name>Predicate;
+import org.eclipse.collections.api.block.procedure.primitive.<name>Procedure;
+import org.eclipse.collections.api.collection.MutableCollection;
+import org.eclipse.collections.api.collection.primitive.Immutable<name>Collection;
+import org.eclipse.collections.api.collection.primitive.Mutable<name>Collection;
+import org.eclipse.collections.api.iterator.Mutable<name>Iterator;
+import org.eclipse.collections.api.iterator.<name>Iterator;
+import org.eclipse.collections.api.list.primitive.Mutable<name>List;
+import org.eclipse.collections.api.set.primitive.Mutable<name>Set;
+import org.eclipse.collections.impl.lazy.primitive.Lazy<name>IterableAdapter;
+import net.jcip.annotations.ThreadSafe;
+
+/**
+ * This file was automatically generated from template file abstractMultiReaderPrimitiveCollection.stg.
+ *
+ * @since 8.1.
+ */
+@ThreadSafe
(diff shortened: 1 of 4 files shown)


eclipse-collections-code-generator/src/main/resources/impl/collection/mutable/abstractMultiReaderPrimitiveCollection.stg
```diff
@@ -0,0 +1,554 @@
...
+import org.eclipse.collections.api.iterator.<name>Iterator;
+import org.eclipse.collections.api.list.primitive.Mutable<name>List;
+import org.eclipse.collections.api.set.primitive.Mutable<name>Set;
+import org.eclipse.collections.impl.lazy.primitive.Lazy<name>IterableAdapter;
+import net.jcip.annotations.ThreadSafe;
+
+/**
+ * This file was automatically generated from template file abstractMultiReaderPrimitiveCollection.stg.
+ *
+ * @since 8.1.
```
motlin: @since 8.2 probably

eclipse-collections-code-generator/src/main/resources/impl/collection/mutable/abstractMultiReaderPrimitiveCollection.stg
```diff
@@ -0,0 +1,554 @@
...
+ */
+@ThreadSafe
+public abstract class AbstractMultiReader<name>Collection
+        implements Mutable<name>Collection, Serializable
```
motlin: Lately I've been trying to avoid making abstract classes Serializable. The serialized form of the subclass will include information about the superclass, increasing the size of the serialized form in bytes and making additional refactoring more likely to break the serialized form.

eclipse-collections-code-generator/src/main/resources/impl/collection/mutable/abstractMultiReaderPrimitiveCollection.stg
```diff
@@ -0,0 +1,554 @@
...
+
+    protected void acquireReadLock()
+    {
+        this.getLock().readLock().lock();
+    }
+
+    protected void unlockReadLock()
+    {
+        this.getLock().readLock().unlock();
+    }
```
motlin: I'm going to suggest a refactoring here, but it's not a pre-requisite. We could always come back and do this later.

The lock, and these four methods for dealing with the lock, do not use the primitive type `<name>`. They could be pulled up into an abstract superclass called AbstractMultiReaderPrimitiveCollection, or something similar.

eclipse-collections-code-generator/src/main/resources/impl/collection/mutable/abstractMultiReaderPrimitiveCollection.stg
```diff
@@ -0,0 +1,554 @@
...
+
+
+    protected Mutable<name>Collection get<name>Collection()
+    {
+        return this.collection;
+    }
```
motlin: If we follow the pattern that the object collections use, this method would be called getDelegate() and it would return PrimitiveIterable in the superclass.

eclipse-collections-code-generator/src/main/resources/impl/collection/mutable/abstractMultiReaderPrimitiveCollection.stg
```diff
@@ -0,0 +1,554 @@
...
+            return this.collection.isEmpty();
+        <readUnlock()>
+    }
+
+    public boolean notEmpty()
+    {
+        <readLock()>
+            return this.collection.notEmpty();
+        <readUnlock()>
+    }
```
motlin: Since PrimitiveIterable supports size(), isEmpty(), and notEmpty(), these methods could be pulled up as well. Depending on how many methods we find that can be pulled up, it could make decent impact on lines of generated code and jar size.

eclipse-collections-code-generator/src/main/resources/impl/collection/mutable/abstractMultiReaderPrimitiveCollection.stg
```diff
@@ -0,0 +1,554 @@
...
+            return this.collection.retainAll(source);
+        <writeUnlock()>
+    }
+
+    /**
+     * Makes copy of entire collection
+     */
+    public Mutable<name>Iterator <type>Iterator()
+    {
+        final <name>Iterator iter = toImmutable().<type>Iterator();
```
motlin: The "object" collections do this, and I think we should consider doing the same here.

```
    @Override
    public Iterator<T> iterator()
    {
        throw new UnsupportedOperationException(
                "Iterator is not supported directly on MultiReader collections.  "
                        + "If you would like to use an iterator, you must either use withReadLockAndDelegate() or withWriteLockAndDelegate().");
    }
```

eclipse-collections-code-generator/src/main/resources/impl/collection/mutable/abstractMultiReaderPrimitiveCollection.stg
```diff
@@ -0,0 +1,554 @@
...
+    public void appendString(
+            Appendable appendable,
+            String start,
+            String separator,
+            String end)
+    {
+        <readLock()>
+            this.collection.appendString(appendable, start, separator, end);
+        <readUnlock()>
+    }
```
motlin: toString, makeString, and appendString can also be pulled up to a super class.

nikhilnanivadekar:
There are build failures due to checkstyle violations: https://s3.amazonaws.com/archive.travis-ci.org/jobs/218774724/log.txt

eclipse-collections-code-generator/src/main/resources/impl/collection/mutable/abstractMultiReaderPrimitiveCollection.stg
jankotek: I will leave it, there is way too many classes already.

eclipse-collections-code-generator/src/main/resources/impl/collection/mutable/abstractMultiReaderPrimitiveCollection.stg
jankotek: It is consistent with other wrappers, there is protected method  `SynchronizedLongList.getLongCollection()` etc...

jankotek:
Ok, I will add abstract class and untouchable primitive collections

PR #271, opened 2017-04-19
Pull up implementations of some methods from AbstractRichIterable to …

### LINKED ISSUE #267: Pull up implementations of some methods from AbstractRichIterable to RichIterable.
Some methods on AbstractRichIterable don't depend on anything in the impl package. For example:

```java
    @Override
    public boolean isEmpty()
    {
        return this.size() == 0;
    }
```

Methods like these can be pulled up as default methods on RichIterable.

By contract, some methods rely on static factories that don't live in the api package, so they can't be pulled up. For example:

```java
    @Override
    public MutableList<T> toList()
    {
        MutableList<T> list = Lists.mutable.empty();
        this.forEachWith(Procedures2.addToCollection(), list);
        return list;
    }
```

PascalSchumacher:
…RichIterable. (closes #267)

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
@@ -1254,7 +1254,11 @@ public interface RichIterable<T>
      *
      * @since 1.0
      */
-    T detectIfNone(Predicate<? super T> predicate, Function0<? extends T> function);
+    default T detectIfNone(Predicate<? super T> predicate, Function0<? extends T> function)
+    {
+        T result = this.detect(predicate);
+        return result == null ? function.value() : result;
+    }
 
     /**
      * Returns the first element of the iterable that evaluates to true for the specified predicate2 and parameter, or
@@ -1422,7 +1426,10 @@ public interface RichIterable<T>
      *
      * @since 1.0
      */
-    MutableList<T> toSortedList(Comparator<? super T> comparator);
+    default MutableList<T> toSortedList(Comparator<? super T> comparator)
+    {
+        return this.toList().sortThis(comparator);
+    }
 
     /**
      * Converts the collection to a MutableList implementation and sorts it based on the natural order of the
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Most methods use classes of the implementation package, so it is not possible to pull them up.

It is possible to also pull up `isEmpty`, and `toArray`, but that would require adding a lot of overrides in sub-classes to fix error like these:

>MutableList.java:[47,7] error: interface MutableList<T#1> inherits abstract and default for toArray(T#2[]) from types RichIterable and List
>MutableSortedSet.java:[49,7] error: interface MutableSortedSet<T#1> inherits abstract and default for toArray(T#2[]) from types RichIterable and Set
>FixedSizeList.java:[21,7] error: interface FixedSizeList<T> inherits abstract and default for toArray() from types RichIterable and List
>MutableMap.java:[48,7] error: interface MutableMap<K,V> inherits abstract and default for isEmpty() from types RichIterable and Map

nikhilnanivadekar:
Agree @PascalSchumacher I went through a preliminary exercise to pull up a few more methods, but that leads to unnecessary delegation. Since @motlin had a patch long ago for a similar exercise, tagging him for his input.

motlin:
If we keep the split between api and impl, then there's only so much we can
pull up into interfaces.

I experimented with the idea of extracting "trait" interfaces above the
abstract classes. So imagine an interface called RichIterableTrait above
AbstractRichIterable with lots of default methods in the impl package. I
have mixed feelings about the idea and may return to it later.

For now, I think a good next step is to complete the tickets I created
recently to get more classes to extend AbstractRichIterable. No matter what
we pull up later, we won't be getting great code reuse unless we get all
classes into a consistent hierarchy.

nikhilnanivadekar:
In that case I'll merge this PR. We can always have a new PR with more implementations pulled up.

nikhilnanivadekar:
@PascalSchumacher would you mind rebasing on master so that the graph looks clean?
Thanks!

PascalSchumacher:
@nikhilnanivadekar Rebased on master.

nikhilnanivadekar:
Thank you!

Issue #214, opened 2017-03-12
Introduce Iterate.zipWith(Iterable<T>, Iterable<TT>, Function<T, TT>)


mokah:
I propose introducing a new zipWith (or similarly named) method in the class Iterate that would accept two iterables of type T and TT, respectively, and a function whose input is the current element of both iterables and output can be of a third type, TTT. 

This would allow you to zip two iterables with the added benefit of the option of running some computation on each iteration and outputting to a collection of any type that tickles your fancy. 

Thoughts?

motlin:
I edited the title slightly. I think this is a good idea. To be consistent with other similar iteration patterns, we'd probably also pass in a target parameter of type R extends MutableCollection<T3>.

mokah:
@motlin zip accepts type R but extends Collection, not MutableCollection. Should this pattern be followed?

Also, it seems the zip and zipWithIndex have been deprecated in RichIterable. Should zipWith be declared in OrderedIterable and omitted from RichIterable?

motlin:
Yes, and yes. Good points!

PR #276, opened 2017-05-01
[#214] Implement Iterate.zipWith(Iterable, Iterable, Function)

mokah:
Signed-off-by: Mohammed Kashim <mohammed.kashim@gs.com>


.idea/libraries/Maven__ch_qos_logback_logback_classic_1_2_3.xml
```diff
@@ -1,13 +0,0 @@
-<component name="libraryTable">
```
motlin: Checking in .idea/ files is fine, but I don't think these three files should be removed.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
```diff
@@ -38,6 +38,7 @@
 import org.eclipse.collections.api.list.primitive.MutableLongList;
 import org.eclipse.collections.api.list.primitive.MutableShortList;
 import org.eclipse.collections.api.multimap.list.MutableListMultimap;
+import org.eclipse.collections.api.ordered.OrderedIterable;
```
motlin: I don't see this new import used in the delta.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/ordered/OrderedIterable.java
```diff
@@ -11,6 +11,7 @@
 package org.eclipse.collections.api.ordered;
 
 import java.util.Collection;
+import java.util.Iterator;
```
motlin: I don't see this import used in the delta either.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/ordered/OrderedIterable.java
```diff
@@ -325,6 +326,26 @@
...
+     * remaining elements are ignored.
+     *
+     * @param that The {@code Iterable} providing the second half of each result pair
+     * @param <S>  The type of the second half of the returned pairs
+     * @param function The function to be applied to each iteration
+     * @return A new {@code OrderedIterable} containing the result of applying the function on each corresponding
+     * element of this {@code OrderedIterable} and that. The length of the returned {@code OrderedIterable} is the
+     * minimum of the lengths of this {@code OrderedIterable} and that.
+     */
+    <S, Z> OrderedIterable<Z> zipWith(Iterable<S> that, Function2<T, S, Z> function);
```
motlin: The functions should all use bounded wildcards to allow the types to vary more. Specifically, `Function2<T, S, Z>` should be replaced with `Function2<? super T, ? super S, ? extends Z>` in all locations.

Another thing to consider about the api is whether to take Iterable or OrderedIterable. The reason I deprecated zip on RichIterable is that you can write set.zip(otherSet) and you get what feels like random pairs. But if you zip two ordered iterables, you'll get back pairs in a meaningful order. Those methods are deprecated but unchanged for backwards compatibility. On new methods, we can be more strict and change the type of `that` to `OrderedIterable`. This would be consistent with the newish method corresponds(OrderedIterable, Predicate2). We would not need to vary that type on overrides. What do you think?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/ordered/OrderedIterable.java
```diff
@@ -325,6 +326,26 @@
     @Override
     <S, R extends Collection<Pair<T, S>>> R zip(Iterable<S> that, R target);
 
+    /**
+     * Returns a {@code OrderedIterable} formed from this {@code OrderedIterable} and another {@code Iterable} by
+     * applying the {@code Function2}. The second {@code Iterable} should also be ordered.
```
motlin: by applying the {@code Function2} **to each corresponding pair**

I think the description on the `return` Javadoc tag is clear about that, but the summary can be more clear too.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/set/sorted/MutableSortedSet.java
```diff
@@ -166,6 +167,9 @@
     <S> MutableList<Pair<T, S>> zip(Iterable<S> that);
 
     @Override
+    <S, Z> OrderedIterable<Z> zipWith(Iterable<S> that, Function2<T, S, Z> function);
```
motlin: Did you intend to change the return type here to MutableSortedSet? As it stands, this override is redundant.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/AbstractImmutableSortedBag.java
```diff
@@ -322,6 +323,54 @@ public ImmutableShortList collectShort(ShortFunction<? super T> shortFunction)
     }
 
     @Override
+    public <S, Z> OrderedIterable<Z> zipWith(Iterable<S> that, Function2<T, S, Z> function)
+    {
+        MutableList<Z> list = FastList.newList();
+        Iterator<S> iterator = that.iterator();
+
+        this.forEachWithOccurrences((each, parameter) -> {
```
motlin: parameter should be renamed to occurrences

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/AbstractImmutableSortedBag.java
```diff
@@ -322,6 +323,54 @@ public ImmutableShortList collectShort(ShortFunction<? super T> shortFunction)
...
+    {
+        Iterator<S> iterator = that.iterator();
+
+        if (target instanceof MutableBag)
+        {
+            MutableBag<Z> targetBag = (MutableBag<Z>) target;
+            this.forEachWithOccurrences((each, occurrences) -> {
+                if (iterator.hasNext())
+                {
+                    targetBag.addOccurrences(function.apply(each, iterator.next()), occurrences);
```
motlin: Is there a test that covers this line? If I'm reading it right, it looks like the target bag will wind up with size^2 number of elements. I realize you're replicating the pattern from zip and zipWith, so I want to make sure all three are covered by tests.

motlin: This patch is big and I don't have time to review the whole thing right now. I figured I'd share the comments I have so far. This is a really big contribution and a great first commit. Thanks!

eclipse-collections-api/src/main/java/org/eclipse/collections/api/ordered/OrderedIterable.java
mokah: That makes complete sense.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/AbstractImmutableSortedBag.java
mokah: Thanks for the detailed feedback, patience, and continued assistance. I'm new to this space. I'll make these changes soon. :)

donraab:
The last comment on this thread was almost two years ago now.  Has there been any further progress on this pull request or should we close it?

nikhilnanivadekar:
closing due to inactivity. Please reopen if desired.

PR #278, opened 2017-05-09
Implement countBy and countBy with target as default methods on RichI…

donraab:
…terable.


eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -1983,6 +1985,28 @@ default void appendString(Appendable appendable, String separator)
...
 
     /**
+     * This method will count the number of occurrences of each value calculated by applying the
+     * function to each element of the collection.
+     *
+     * @since 8.2
+     */
+    default <V> Bag<V> countBy(Function<? super T, ? extends V> function)
+    {
+        return this.asLazy().<V>collect(function).toBag();
```
motlin: Since mutable and immutable subclasses have implementations, this should probably stay abstract up on RichIterable.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/ImmutableBag.java
```diff
@@ -120,6 +120,15 @@
     @Override
     <V> ImmutableBag<V> flatCollect(Function<? super T, ? extends Iterable<V>> function);
 
+    /**
+     * @since 8.2
+     */
+    @Override
+    default <V> ImmutableBag<V> countBy(Function<? super T, ? extends V> function)
+    {
+        return this.asLazy().<V>collect(function).toBag().toImmutable();
```
motlin: On bags, I'd expect lazy collect to be slower than eager collect, since we haven't done anything to optimize CollectIterable for bag targets. The test should just be on a bag with a few keys where the counts are MAX_INT. Call this method, it should be slow. Then replace it with `return this.collect(function)` and it should be fast.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MutableBag.java
```diff
@@ -89,6 +89,15 @@
     <V> MutableBagMultimap<V, T> groupByEach(Function<? super T, ? extends Iterable<V>> function);
 
     /**
+     * @since 8.2
+     */
+    @Override
+    default <V> MutableBag<V> countBy(Function<? super T, ? extends V> function)
+    {
+        return this.asLazy().<V>collect(function).toBag();
```
motlin: Same performance problem here.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/sorted/ImmutableSortedBag.java
```diff
@@ -137,6 +137,15 @@
     @Override
     ImmutableSortedBag<T> dropWhile(Predicate<? super T> predicate);
 
+    /**
+     * @since 8.2
+     */
+    @Override
+    default <V> ImmutableSortedBag<V> countBy(Function<? super T, ? extends V> function)
+    {
+        return this.asLazy().<V>collect(function).toSortedBag().toImmutable();
```
motlin: There's no constraint on V that it's `Comparable<V>`. I think this should return an unsorted bag or an ordered (but not sorted) bag.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/sorted/MutableSortedBag.java
```diff
@@ -162,6 +162,15 @@
     <V> MutableSortedBagMultimap<V, T> groupByEach(Function<? super T, ? extends Iterable<V>> function);
 
     /**
+     * @since 8.2
+     */
+    @Override
+    default <V> MutableSortedBag<V> countBy(Function<? super T, ? extends V> function)
+    {
+        return this.asLazy().<V>collect(function).toSortedBag();
```
motlin: Same here, an ordered, but not sorted bag. And the same performance problem.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/ImmutableCollection.java
```diff
@@ -147,6 +148,15 @@
     @Override
     <V> ImmutableObjectDoubleMap<V> sumByDouble(Function<? super T, ? extends V> groupBy, DoubleFunction<? super T> function);
 
+    /**
+     * @since 8.2
+     */
+    @Override
+    default <V> ImmutableBagIterable<V> countBy(Function<? super T, ? extends V> function)
+    {
+        return this.asLazy().<V>collect(function).toBag().toImmutable();
```
motlin: Can this method be abstract?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/MutableCollection.java
```diff
@@ -468,6 +469,15 @@
     <V> MutableObjectDoubleMap<V> sumByDouble(Function<? super T, ? extends V> groupBy, DoubleFunction<? super T> function);
 
     /**
+     * @since 8.2
+     */
+    @Override
+    default <V> MutableBagIterable<V> countBy(Function<? super T, ? extends V> function)
+    {
+        return this.asLazy().<V>collect(function).toBag();
```
motlin: Can this method be abstract?

eclipse-collections/src/main/java/org/eclipse/collections/impl/AbstractRichIterable.java
```diff
@@ -680,6 +681,18 @@ public boolean containsAll(Collection<?> collection)
...
+    /**
+     * This method will count the number of occurrences of each value calculated by applying the
+     * function to each element of the collection.
+     *
+     * @since 8.2
+     */
+    @Override
+    public <V> Bag<V> countBy(Function<? super T, ? extends V> function)
+    {
+        return this.countBy(function, Bags.mutable.empty());
```
motlin: Can this be abstract, with mutable and immutable implementations on subclasses?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/AbstractImmutableBag.java
```diff
@@ -122,6 +122,17 @@
...
 
+    /**
+     * @since 8.2
+     */
+    @Override
+    public <V> ImmutableBag<V> countBy(Function<? super T, ? extends V> function)
+    {
+        MutableBag<V> result = Bags.mutable.empty();
+        this.forEachWithOccurrences((each, count) -> result.addOccurrences(function.valueOf(each), count));
+        return result.toImmutable();
```
motlin: No performance problem here. The implementation looks like it's the same as collect, though we don't implement collect on this class for performance reasons. I'd say actually delegate to collect, and consider making the default methods on the interfaces above regular abstract methods.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
donraab: If it stays abstract on RichIterable, then it can't be released in 8.2.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/ImmutableBag.java
donraab: Great point.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
FactorioBlueprints: I see what you mean. If someone has implemented RichIterable, and it doesn't extend anything like AbstractMutableCollection, or implement MutableCollection or ImmutableCollection, then it fails to compile. But if we do add a method body, then we'd want to delete the body in the next major version I'd think. And yet the method is not deprecated. Just its body is probably dead code.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MutableBag.java
donraab: Agreed.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/sorted/ImmutableSortedBag.java
donraab: I would have to go with an unsorted bag today, as we do not have an ordered bag (yet).

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/ImmutableCollection.java
donraab: In order for the code to be acceptable for the 8.2 release I believe we will need the default implementation.

eclipse-collections/src/main/java/org/eclipse/collections/impl/AbstractRichIterable.java
donraab: I can't reference Bags.mutable.empty() in the interface, but thought this is a better abstract implementation than I can provide in the interface.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/AbstractImmutableBag.java
donraab: Makes sense.  Thank you.

donraab:
I will add countByWith, and will be changing the return types for countBy and countByWith to always be an unordered Bag.  Thank you for the feedback.

donraab:
We'll wait to merge this request for the 9.0 release.  I updated all of the since tags to 9.0.

nikhilnanivadekar:
Merging based on @kristeno 's review.

Issue #279, opened 2017-05-31
size method of LazyIterable returned by LazyIterable::concatenate has O(n) complexity


cinquin:
Using `LazyIterable::concatenate` is a nice way to get a unified view of two lists that are maintained separately. Unfortunately the returned `LazyIterable` has a very inefficient `size` method, which may come as a surprise to the user (and which is not mentioned in the documentation). Specifically, that method has to be implemented as `this.count(Predicates.alwaysTrue())`, since the `Iterable` interface does not have a `size` method; i.e., the method iterates over all values to count them.

I am submitting a pull request with the following overload of the original concatenate method: `LazyIterable<T> concatenate(RichIterable<T> richIterable)`.

This solves the performance problem when `*Rich*Iterable`s are available to perform the concatenation.

PR #280, opened 2017-05-31
Improve efficiency of size method for concatenated rich iterables

cinquin:
Get LazyIterableAdapter::concatenate and LazyIterate::composite to return a
new, specialized version of CompositeIterable, which implements a size
method whose complexity is O(k), where k is the number of concatenated rich
iterables, instead of O(n), where n is the total number of elements in the
concatenation.

See issue #279

Signed-off-by: Olivier Cinquin <olivier.cinquin@uci.edu>

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/CompositeIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/CompositeIterable.java
@@ -122,6 +122,12 @@ public final class CompositeIterable<E>
     }
 
     @Override
+    public int size()
+    {
+       return (int) this.iterables.sumOfInt(Iterate::sizeOf);
+    }
+
+    @Override
     public Iterator<E> iterator()
     {
         return new CompositeIterator(this.iterables);


eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/CompositeRichIterable.java
```diff
@@ -0,0 +1,182 @@
...
+    private final MutableList<RichIterable<E>> iterables;
+
+    private CompositeRichIterable(MutableList<RichIterable<E>> newIterables)
+    {
+        this.iterables = newIterables;
+    }
+
+    public CompositeRichIterable()
+    {
+        this(FastList.newList());
```
nikhilnanivadekar: We advocate using factory methods like Lists.mutable.empty(), Lists.mutable.with(...)

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/CompositeRichIterable.java
```diff
@@ -0,0 +1,182 @@
...
+
+    public static <T> CompositeRichIterable<T> with(RichIterable<T>... iterables)
+    {
+        return new CompositeRichIterable<>(FastList.newListWith(iterables));
+    }
+
+    @Override
+    public void each(Procedure<? super E> procedure)
+    {
+        this.iterables.each(iterable -> Iterate.forEach(iterable, procedure));
```
nikhilnanivadekar: Isn't `iterable` a `RichIterable`? then you can do:

`this.iterables.each(iterable -> iterable.each(procedure));`

nikhilnanivadekar:
@cinquin Thanks for the PR. This is an API change where we cannot add default methods, so we will have to wait till 9.0.0. 
The code complete for 8.2.0 is this week, with release end of June.

cinquin:
@nikhilnanivadekar 
> @cinquin Thanks for the PR.

Thanks for the review.

> This is an API change so we will have to wait till major version 9.0.0. The code complete for 8.2.0 is this week with 8.2.0 scheduled to release end of June.

Please have a look at my updated PR. I added a default method implementation that should address API compatibility concerns.

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/CompositeRichIterable.java
```diff
@@ -0,0 +1,182 @@
...
+
+    public void add(RichIterable<E> iterable)
+    {
+        this.iterables.add(iterable);
+    }
+
+    @Override
+    public int size()
+    {
+       return (int) this.iterables.sumOfInt(RichIterable::size);
```
motlin: When you move this method to CompositeIterable, you'll have to replace RichIterable::size with iterable -> Iterate.sizeOf(iterable). The cast shows that we actually have a bug - you can concatenate two Intervals of size MAX_INTEGER and easily overflow the size without getting any exception.

cinquin:
@motlin 
> The optimization is a good idea, but it must be done without adding any new classes or methods. Simply override the size() method in CompositeIterable and you're done.

That solution has the downside that it performs multiple instance checks per method call, but it does have the upside that it avoids code duplication — so probably a better solution on balance. I've updated the PR.

nikhilnanivadekar:
@cinquin can you please rebase and squash all the changes so that we can review. Thanks!

eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/CompositeIterable.java
```diff
@@ -122,6 +122,12 @@ public void add(Iterable<E> iterable)
     }
 
     @Override
+    public int size()
+    {
+       return (int) this.iterables.sumOfInt(iterable -> Iterate.sizeOf(iterable));
```
motlin: Now that I see it written out, I realize that it can be shortened to `this.iterables.sumOfInt(Iterate::sizeOf)`. Either way (with or without this little refactoring) this looks good to go now.

nikhilnanivadekar:
@cinquin can you please update the commit message to say something like 

> fixes #279

 or have 

> #279

 in the commit message, it will auto-close that issue when I merge this PR.
This is good to go. It is up to you if you want to refactor it to a method reference (we have a method reference preference 😃 ).

cinquin:
@nikhilnanivadekar  @motlin 

> we have a method reference preference

Me too.

Please see updated PR.

Thanks for your comments.

nikhilnanivadekar:
Thanks for the contribution!

PR #281, opened 2017-06-01
More fixes for Java 9 compatibility

nikhilnanivadekar:
1) Deprecate Verify#assertShallowClone()
2) Add Assume keyword for Serialization tests where reflection is used
3) Set fields as null in the ignored exception block: This is required as the exception is thrown at setAccessible(true). However, the field is not null, due to which Java 9 regression build fails due to IllegalAccessError (reflection for Java 9)


motlin:
Why deprecate assertShallowClone?

nikhilnanivadekar:
From the implementation:
```
Method method = Object.class.getDeclaredMethod("clone", (Class<?>[]) null);
method.setAccessible(true);
Object clone = method.invoke(object);
String prefix = itemName + " and its clone";
```
It uses reflection.

PR #293, opened 2017-07-16
Remove Apache Felix plugin, Add Java 9 regression build

nikhilnanivadekar:
1) Remove Apache Felix plugin:
Removal of Apache Felix plugin will result in OSGi bundles not exported under eclipse-collections, eclipse-collections-api. However, these bundles are exported in p2 repository. Users should start using the same.
This will reflect in the release notes.

This is a compatibility breaking change.

2) Java 9 Regression builds

### DIFF
--- a/.travis-command-jdk9.sh
+++ b/.travis-command-jdk9.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# Prevent accidental execution outside of Travis:
+if [ -z "${TRAVIS+false}" ]
+then
+  echo "TRAVIS environment variable is not set"
+  exit 1
+fi
+
+# Switch to desired JDK, download if required:
+function install_jdk {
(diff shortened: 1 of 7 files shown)


nikhilnanivadekar:
#226 has discussion about removing these obsolete bundles

nikhilnanivadekar:
@guw , @kthoms to confirm from our previous discussion, it is ok to not publish the OSGi bundles since, the p2 repo will have the necessary packages? 
What about non-eclipse OSGi users? Or there is no such thing as non-eclipse OSGi users, and one needs eclipse to use OSGi? 

Asking since, I am not an OSGi user/expert.

Thanks!

guw:
Yes, there is no need to add OSGi meta-data to the regular jars. non-p2 OSGi users can consume the bundle via direct download right away. However, you could also mvn deploy the OSGi bundle and have it available from Maven Central.

nikhilnanivadekar:
@guw what are the deploy steps to maven central? I would prefer doing that as well.

nikhilnanivadekar:
Raised @294 for tracking.

Issue #147, opened 2016-09-16
java 8 convergence: provide efficient implementation of Java 8 default methods 


phraktle:
JDK8 default methods on Collections such as `forEach`, `putIfAbsent` etc. should be mapped to their equivalent but more efficient implementations.

RamVaidya:
@donraab @nikhilnanivadekar @prathasirisha 
Is this still relevant?

im-prince:
can u assign me this issue

motlin:
> Is this issue still available to work upon??

@debjitl45 this is partially duplicated by https://github.com/eclipse/eclipse-collections/issues/500. Both are open, and several default methods are not properly overridden.

shreyaskar03:
Dear Sir,
Is this issue still open to work upon?

motlin:
@shreyaskar03 yes

Rishabh-mikku:
Dear Sir,
Is this issue still open to work? if yes, then can you please assign me this issue.

motlin:
@Rishabh-mikku no need to assign the issue. Feel free to go ahead and submit a pull request.

PR #296, opened 2017-07-29
Add an efficient default implementation of forEach.

cinquin:
This implementation takes a Consumer as argument instead of a Procedure.

This avoids calls to forEach falling back on the Java 8 Iterable default
implementation, which instantiates an iterator.

Note that, because this is a default interface implementation, the API
is effectively unchanged.

This partially addresses issue #147.

Signed-off-by: Olivier Cinquin <olivier.cinquin@uci.edu>

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/InternalIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/InternalIterable.java
@@ -10,6 +10,8 @@
 
 package org.eclipse.collections.api;
 
+import java.util.function.Consumer;
+
 import org.eclipse.collections.api.block.procedure.Procedure;
 import org.eclipse.collections.api.block.procedure.Procedure2;
 import org.eclipse.collections.api.block.procedure.primitive.ObjectIntProcedure;
@@ -48,6 +50,13 @@ public interface InternalIterable<T>
      */
     @SuppressWarnings("UnnecessaryFullyQualifiedName")
     void forEach(Procedure<? super T> procedure);
+
+    @Override
+    default void forEach(Consumer<? super T> consumer)
+    {
+        Procedure<? super T> procedure = consumer::accept;
+        this.forEach(procedure);
+    }
 
     /**
      * Iterates over the iterable passing each element and the current relative int index to the specified instance of
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/AbstractRichIterableTestCase.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/AbstractRichIterableTestCase.java
@@ -11,6 +11,7 @@
 package org.eclipse.collections.impl;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.DoubleSummaryStatistics;
 import java.util.HashSet;
@@ -24,6 +25,7 @@ import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
 
 import org.eclipse.collections.api.BooleanIterable;
 import org.eclipse.collections.api.ByteIterable;
@@ -39,6 +41,7 @@ import org.eclipse.collections.api.bag.MutableBag;
 import org.eclipse.collections.api.bag.sorted.MutableSortedBag;
 import org.eclipse.collections.api.block.function.Function;
 import org.eclipse.collections.api.block.function.Function0;
+import org.eclipse.collections.api.block.procedure.Procedure;
 import org.eclipse.collections.api.collection.primitive.MutableBooleanCollection;
 import org.eclipse.collections.api.collection.primitive.MutableByteCollection;
 import org.eclipse.collections.api.collection.primitive.MutableCharCollection;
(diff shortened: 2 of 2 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/InternalIterable.java
```diff
@@ -49,6 +51,12 @@
     @SuppressWarnings("UnnecessaryFullyQualifiedName")
     void forEach(Procedure<? super T> procedure);
 
+    @Override
+    default void forEach(Consumer<? super T> action)
+    {
+        forEach((Procedure<? super T>) (x -> action.accept(x)));
```
donraab: This is a good change.  I think it would be slightly better if you break the type out onto a separate line and use a method reference as follows.  

```
Procedure<? super T> procedure = consumer::accept;
this.forEach(procedure);
```

I would also recommend duplicating the functionality in the the forEach test in `AbstractRichIterableTestCase` and call it forEachConsumer so we know this code path is explicitly covered.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/InternalIterable.java
```diff
@@ -49,6 +51,13 @@
     @SuppressWarnings("UnnecessaryFullyQualifiedName")
     void forEach(Procedure<? super T> procedure);
 
+    @Override
+    default void forEach(Consumer<? super T> consumer)
+    {
+        Procedure<? super T> procedure = consumer::accept;
+        forEach(procedure);
```
nikhilnanivadekar: To keep in line with our convention please change this to `this.forEach(procedure)`

unit-tests/src/test/java/org/eclipse/collections/impl/AbstractRichIterableTestCase.java
```diff
@@ -168,16 +171,40 @@ public void tap()
...
+        }
+        else
+        {
+            template.forEach(adder);
+        }
         Verify.assertSize(4, result);
         Verify.assertContainsAll(result, 1, 2, 3, 4);
     }
 
+    private void forEachProcedure()
```
nikhilnanivadekar: convert this to a test

```
@Test
public void forEach_procedure()
```
also missing a `this.`

unit-tests/src/test/java/org/eclipse/collections/impl/AbstractRichIterableTestCase.java
```diff
@@ -168,16 +171,40 @@ public void tap()
...
+    {
+        forEach(CollectionAddProcedure::on);
+    }
+
+    private void forEachConsumer()
+    {
+        forEach(collection -> collection::add);
```
nikhilnanivadekar: Same comment as above

unit-tests/src/test/java/org/eclipse/collections/impl/AbstractRichIterableTestCase.java
```diff
@@ -168,16 +171,40 @@ public void tap()
...
+    }
+
+    @Test
+    public void forEach()
```
nikhilnanivadekar: Can be deleted

nikhilnanivadekar:
@cinquin can you please rebase with master as well?

cinquin:
@nikhilnanivadekar The `forEach` method in AbstractRichIterableTestCase is overridden in ~9 test subclasses (not sure why things are set up that way). That's why I thought it best to keep the `forEach` method in AbstractRichIterableTestCase as such, instead of just creating two new separate tests as you suggest; otherwise the 9 subclasses need to be changed too.

nikhilnanivadekar:
@cinquin the test is overridden in 11 sub-classes by design. Here are the reasons for each override:
1) org.eclipse.collections.impl.stack.StackIterableTestCase#forEach: The iteration order changes, so that needs to be tested and asserted
2) org.eclipse.collections.impl.bag.mutable.MutableBagTestCase#forEach: Bags allow duplicates, but are hashed, the iteration over duplicates need to be tested
3) org.eclipse.collections.impl.bag.mutable.MultiReaderHashBagTest#forEach: same as 2
4) org.eclipse.collections.impl.list.mutable.MultiReaderFastListTest#forEach: need to test the order
5) org.eclipse.collections.impl.bag.immutable.ImmutableSingletonBagTest#forEach: immutable, contains only 1 element, need to test it accordingly
6) org.eclipse.collections.impl.bag.immutable.ImmutableBagTestCase#forEach: same as 2
7) org.eclipse.collections.impl.list.mutable.FastListTest#forEach: same as 4
8) org.eclipse.collections.impl.list.mutable.CompositeFastListTest#forEach: same as 4 also calls super
9) org.eclipse.collections.impl.list.fixed.ArrayAdapterTest#forEach: same as 4 also calls super
10) org.eclipse.collections.impl.set.mutable.AbstractMutableSetTestCase#forEach: asserts for collisions
11) org.eclipse.collections.impl.bag.sorted.mutable.AbstractMutableSortedBagTestCase#forEach: tests for order
There are some overrides where we should call super.forEach() in addition to special tests.

Due to the above reasons, for this change the overrides will need to be applied. Similar to what you have done in AbstractRichIterableTestCase, you can extract duplicate code and call it in test.

motlin:
One method just delegates to the other. There's basically one new statement
and the new test covers it right?

nikhilnanivadekar:
Yes

nikhilnanivadekar:
@motlin ok to merge by you?

cinquin:
Just to clarify: in its present state, my pull request adds a test for the new forEach default method. This test gets run on a number of RichIterable implementations (30 tests fail if I force the new test method to throw an exception). This test does not verify that the iteration characteristics (e.g. iteration order) match those required for the particular RichIterable implementation being tested. But the new forEach method is just a wrapper around the pre-existing forEach method, which still gets tested as it did before. One could add a bunch of specialized tests e.g. for iteration order for the new forEach method, but I think that would be redundant and would cause unnecessary code churn.

cinquin:
@nikhilnanivadekar  Sure I'll think about a followup pull request. (I do have another one ready to go to enforce the "this" convention you commented on.)
Thanks for your feedback.

PR #297, opened 2017-07-30
Implement flatCollect on lazy primitive iterables and cartesianProduc…

donraab:
…t on primitive sets factory.

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/factory/SetsTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/factory/SetsTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Goldman Sachs.
+ * Copyright (c) 2017 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -23,6 +23,7 @@ import org.eclipse.collections.api.block.procedure.Procedure2;
 import org.eclipse.collections.api.factory.set.FixedSizeSetFactory;
 import org.eclipse.collections.api.factory.set.ImmutableSetFactory;
 import org.eclipse.collections.api.factory.set.MutableSetFactory;
+import org.eclipse.collections.api.list.MutableList;
 import org.eclipse.collections.api.set.FixedSizeSet;
 import org.eclipse.collections.api.set.ImmutableSet;
 import org.eclipse.collections.api.set.MutableSet;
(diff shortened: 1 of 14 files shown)


unit-tests/src/test/java/org/eclipse/collections/impl/factory/SetsTest.java
```diff
@@ -762,6 +762,21 @@ public void cartesianProduct()
...
+        MutableSet<Integer> set1 = UnifiedSet.newSetWith(1, 2);
+        MutableSet<Integer> set2 = UnifiedSet.newSetWith(2, 3, 4);
+        MutableBag<Pair<Integer, Integer>> expectedCartesianProduct = Bags.mutable.of(
+                Tuples.pair(1, 2),
+                Tuples.pair(2, 2),
+                Tuples.pair(1, 3),
+                Tuples.pair(2, 3),
+                Tuples.pair(1, 4),
+                Tuples.pair(2, 4));
+        Assert.assertEquals(expectedCartesianProduct, Sets.cartesianProduct(set1, set2, Tuples::pair).toBag());
```
nikhilnanivadekar: Can you please add 3 more tests:
1) Same 2 elements in both tests
2) set2.size() > set1.size()
3) Something other than Tuples::pair, it can even be first + second

unit-tests/src/test/java/org/eclipse/collections/impl/factory/SetsTest.java
```diff
@@ -762,6 +762,21 @@ public void cartesianProduct()
     }
 
     @Test
+    public void cartesianProductWithFunction()
+    {
+        MutableSet<Integer> set1 = UnifiedSet.newSetWith(1, 2);
```
nikhilnanivadekar: `Sets.mutable.with()`

unit-tests/src/test/java/org/eclipse/collections/impl/factory/SetsTest.java
donraab: Yes, agreed.

unit-tests/src/test/java/org/eclipse/collections/impl/factory/SetsTest.java
donraab: Eventually, yes.  I held off doing this for now as it is a lot of work, and would make a much larger review.  I can submit one or two more pull requests implementing the new cartesianProduct on our Set interfaces.  This can include the primitive interfaces as well now that we have implementations available on IntSets.

donraab: I'll change the constructor to `Sets.mutable.with()`.

nikhilnanivadekar: Regarding adding the API on Set interfaces, makes sense to first add it as a static utility.
Thanks!

unit-tests/src/test/java/org/eclipse/collections/impl/factory/SetsTest.java
donraab: I added the tests.

nikhilnanivadekar: LGTM

PR #302, opened 2017-08-08
Implement spliterator, stream, parallelStream and castToCollection …

### LINKED ISSUE #208: Implement spliterator, stream, parallelStream and castToCollection on ImmutableCollection
These methods should be added as default methods on ImmutableCollection, and then overridden as necessary on interfaces like ImmutableList and implemented in abstract classes and subclasses as necessary.

This feature can only be done in a major version release, since adding the default methods to ImmutableCollection for spliterator, stream and parallelStream would be ambiguous for implementations since they also have to extend Collection, List or Set.

### LINKED ISSUE #274: Stream shortcut in immutable collections
Currently the only way to create a stream of ImmutableSortedSet or ImmutableList, for example, is to do something like this

```java
final ImmutableSortedSet<Object> values = // ...
StreamSupport
    .stream(values.spliterator(), false)
    .// ...
```

Could we get an api extension for the immutable collections to support `.stream` as with other collections? Would it make sense?

donraab:
…on ImmutableCollection. Fixes #208, Fixes #274

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/ImmutableCollection.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/ImmutableCollection.java
@@ -10,6 +10,12 @@
 
 package org.eclipse.collections.api.collection;
 
+import java.util.Collection;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.bag.ImmutableBag;
 import org.eclipse.collections.api.block.function.Function;
@@ -190,4 +196,36 @@ public interface ImmutableCollection<T>
             Function<? super T, ? extends K> groupBy,
             Function0<? extends V> zeroValueFactory,
             Function2<? super V, ? super T, ? extends V> nonMutatingAggregator);
+
+    /**
+     * @since 9.0
+     */
+    default Stream<T> stream()
+    {
+        return StreamSupport.stream(this.spliterator(), false);
+    }
+
+    /**
+     * @since 9.0
+     */
+    default Stream<T> parallelStream()
+    {
+        return StreamSupport.stream(this.spliterator(), true);
+    }
+
+    /**
+     * @since 9.0
+     */
+    @Override
+    default Spliterator<T> spliterator()
+    {
+        return Spliterators.spliterator(this.iterator(), (long) this.size(), 0);
+    }
+
+    /**
+     * This can be overridden in most implementations to just return this.
+     *
+     * @since 9.0
+     */
+    Collection<T> castToCollection();
 }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/immutable/AbstractImmutableCollection.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/immutable/AbstractImmutableCollection.java
@@ -15,7 +15,11 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.Spliterator;
+import java.util.Spliterators;
 import java.util.function.BinaryOperator;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.bag.ImmutableBag;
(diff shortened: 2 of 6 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/ImmutableCollection.java
```diff
@@ -190,4 +196,37 @@
...
+    {
+        return Spliterators.spliterator(this.iterator(), (long) this.size(), 0);
+    }
+
+    /**
+     * @since 9.0
+     */
+    default Collection<T> castToCollection()
+    {
+        return (Collection<T>) this;
```
nikhilnanivadekar: Should this be an interface method without implementation? 
Collection does not come in the ImmutableCollection hierarchy. However all the implementations are indeed a Collection. What do you think?

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/immutable/AbstractImmutableCollection.java
```diff
@@ -48,7 +52,8 @@
 import org.eclipse.collections.impl.set.mutable.UnifiedSet;
 import org.eclipse.collections.impl.utility.Iterate;
 
-public abstract class AbstractImmutableCollection<T> extends AbstractRichIterable<T> implements ImmutableCollection<T>, Collection<T>
+public abstract class AbstractImmutableCollection<T>
+        extends AbstractRichIterable<T> implements ImmutableCollection<T>, Collection<T>
```
nikhilnanivadekar: can you please line break at implements to make it consistent?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/ImmutableCollection.java
donraab: This can be made abstract or default and nothing will break.  Now that we can use default methods, I find myself wanting to use them where we can when we add new APIs as it will reduce the number of places client code would have to be fixed if these interfaces are implemented.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/immutable/AbstractImmutableCollection.java
donraab: Yes, will fix.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/ImmutableCollection.java
nikhilnanivadekar: If I had something like this:
```

ImmutableFooImpl implements ImmutableCollection
{
// implementation code
}


```
and I called ImmutableFooCollections.castToCollection() would it not give me a ClassCastException?

I am fine with other methods being default methods, since all of the implementations will need to have an iterator.

donraab: Yes.  It would be wrong for this to happen though.  The only other options would be for an implementor to return null here, or return some other collection.  This method is being added for symmetry with castToList, castToSet, castToMap.  It is immediately overridden in AbstractImmutableCollection to return this.

donraab: I could just remove the method altogether.

nikhilnanivadekar: I like the symmetry aspect of it. So, I would keep it, but not as a default method. :)

donraab: Ok, I will make it abstract.  We can always change it before releasing 9.0.

nikhilnanivadekar:
This closes #274 as well

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/ImmutableCollection.java
```diff
@@ -190,4 +196,36 @@
...
+     * @since 9.0
+     */
+    @Override
+    default Spliterator<T> spliterator()
+    {
+        return Spliterators.spliterator(this.iterator(), (long) this.size(), 0);
+    }
+
+    /**
+     * This can be overridden in most implementations to just return this.
```
nikhilnanivadekar: This Javadoc is correct only for Eclipse Collections implementations.

PR #315, opened 2017-08-11
Implement UnifiedMap.trimToSize(), UnifiedSet.trimToSize(). 

### LINKED ISSUE #265: Implement UnifiedMap.trimToSize or (UnifiedMap.compact).
Hi,
Is there a way of compacting a UnifiedMap? Some primitive maps and sets do have a `compact` method, but I can't find such a method (or equivalent) for UnifiedMap. Is the only way to free up memory used by the `table` field and the chained key arrays (e.g. after many elements have been inserted and then removed from the map) to create a new map from the remaining objects and discard the old one?

Would there be interest in having a `compact` method, or is it a design decision not to have one?

Incidentally, Trove also has useful autocompaction behavior, and a `retainEntries` method that helps perform automatic compaction at reasonable times.

nikhilnanivadekar:
Fixes #265

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
@@ -310,7 +310,6 @@ public class UnifiedMap<K, V> extends AbstractMutableMap<K, V>
 
     protected void computeMaxSize(int capacity)
     {
-        // need at least one free slot for open addressing
         this.maxSize = Math.min(capacity - 1, (int) (capacity * this.loadFactor));
     }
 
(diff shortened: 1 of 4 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
```diff
@@ -1330,6 +1330,115 @@ private int chainedHashCode(ChainedBucket bucket)
         while (true);
     }
 
+    public void trimToSize()
+    {
+        Object[] temp = this.table;
+        this.allocateTable(this.occupied);
+        this.occupied = 0;
```
gs-rezaem: No need to reset occupied.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
```diff
@@ -1330,6 +1330,115 @@ private int chainedHashCode(ChainedBucket bucket)
...
+    {
+        int index = this.index(key);
+        Object cur = this.table[index];
+        if (cur == null)
+        {
+            this.table[index] = UnifiedSet.toSentinelIfNull(key);
+            this.occupied++;
+            return;
+        }
+        if (cur instanceof ChainedBucket || !this.nonNullTableObjectEquals(cur, key))
```
gs-rezaem: I don't think you need this if statement. If cur != null, we will always do a chained add

nikhilnanivadekar: I think it was more to do with `this.nonNullTableObjectEquals()` but I see your point, since this is trimToSize() we will never have duplicates.

nikhilnanivadekar:
@gs-rezaem I will make these changes by EOD. Did you get a chance to look at the tests? Are they sufficient or need more work?

gs-rezaem:
I had a look; they look fine. One thing we could do is return boolean from these methods to indicate whether there was a resize or not, and make sure trimx2 doesn't do anything.

nikhilnanivadekar:
@gs-rezaem , @motlin One thing which I found during testing is that if we do operations like this:

```
map.clear();
Assert.assertTrue(map.trimToSize());
Interval.zeroTo(9).each(each -> map.put(each.toString(), each.toString()));
```
This will throw `ArrayIndexOutOfBoundsException` at `put()` since  `this.table.length == 0`
Should we introduce a size check at `put()` before finding the `index`? Or should I introduce a check for `DEFAULT_INITIAL_CAPACITY` in `trimToSize()`. I prefer adding a `DEFAULT_INITIAL_CAPACITY` check in `trimToSize()`. Otherwise this map will be useless for mutation operations. 

On the other hand, if a user is calling `clear()` and then `trimToSize()` they presumably know what is going on.

gs-rezaem:
if occupied == 0, allocate a 2 slot array.

nikhilnanivadekar:
Due to alternate key, value indexes, this.occupied < 2, then I have to allocate a 2 slot array, else put() does not work. I added a test for it.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -1430,6 +1430,94 @@ public void value(K key, V value)
         return builder.toString();
     }
 
+    public boolean trimToSize()
+    {
+        if (this.table.length <= this.occupied)
```
gs-rezaem: the size calculations and later allocation look wrong. You shouldn't be comparing table.length (because in this map there is an allocation for both keys and values, resulting in a factor of 2). You should also be taking the load factor into account. You should never call allocateTable with something that is not a power of 2.

look at the init method. It computes the correct size, given a capacity. you can refactor into a new method to return the computed size:
```
        int capacity = 1;
        while (capacity < occupied)
        {
            capacity <<= 1;
        }
```
if ((capacity << 1) < table.length) then resize (by calling allocate(capacity)).

nikhilnanivadekar: Same comment for UnifiedSet as well? Or we needn't do that for UnifiedSet since we do not need powers of 2.

nikhilnanivadekar: So, if I am not mistaken, the short-circuit should be:

`if(this.table.length/2 <= this.occupied)`

Similarly: 
`if (this.table.length/2 <= defaultSizeToAllocate)`

and then:

`this.init(Math.min(defaultSizeToAllocate, this.table.length))`

motlin:
In the current implementation, I don't see efficiency benefits over
creating a new set or map. I was expecting trimToSize to never calculate
hashCodes.

nikhilnanivadekar:
@motlin Since our index calculations are based on table.length, we need to compute hashcode. Do you have other implementation in mind which will work without hashcode computation?

motlin:
(Old table index) modulo (new table size)

nikhilnanivadekar:
This first short-circuit evaluates false for most cases:
```
if (this.fastCeil(this.table.length / this.loadFactor) <= (this.occupied << 1))
{
    return false;
}
```
For example: Consider this.occupied = 239, this.table.length = 512, this.loadFactor = 0.75
this.fastCeil = 683
this.occupied << 1 = 478 
it will not short-circuit, so even if we call trimToSize() twice back to back, it will still execute.
I think a better short-circuit is this that way we take the loadFactor in consideration and table.length is always a power of 2. Thoughts?

```
if (this.table.length <= this.getCapacity(this.fastCeil(this.occupied / this.loadFactor)))
{
    return false;
}
```

PR #317, opened 2017-08-17
Implement stream, parallelStream and spliterator on all object valued…

donraab:
… maps.


nikhilnanivadekar:
Can it be implemented without an iterator?

donraab:
Possibly, but that would be substantially more work.  I wanted to test out the API addition and it works well, and gives us good symmetry with MapIterable being Iterable on values.  We can spend time post 9.0 to implement as many optimized spliterators as we want.

nikhilnanivadekar:
Alright please rebase

Issue #319, opened 2017-08-17
Improve OSGi bundle API information


bergmanngabor:
We at the Viatra Eclipse project are now experimenting with your amazing collection library, which we consume as an OSGi bundle (via p2 update site).

**What I want to do**: 
Instantiate containers. According to your [reference guide](https://github.com/eclipse/eclipse-collections/blob/master/docs/guide.md#-creating-mutable-collections), collections shall be instantiated either by direct constructor call, or by invoking a static factory method (on _Lists_, _Maps_, etc.). 

**What I experience instead**: 
The OSGi bundle _exports_ neither the actual container implementation classes, nor the static factory classes. Therefore the recommended ways of instantiation is classified as non-API, and Eclipse PDE reports this problem ("discouraged access"). Depending on settings, this may pass as a mere warning at compile time, and be tolerated even at runtime, but not quite ideal.

**What I suggest**: 
I guess you should add _Export-Package_ directives for all these _impl_ packages into your MANIFEST.MF descriptor. Since I believe you encourage advanced users to subclass your collections anyway, this is probably the right choice here, as far as I understand your policy.

Bonus questions: will any older Java 7 compatible versions (or their current maintenance builds) be made available via p2 update sites as well? How about future nightly/milestone pre-release versions?

nikhilnanivadekar:
@bergmanngabor thank you for experimenting with Eclipse Collections!
To be honest, none of us maintainers are OSGi experts, so we have a learning curve ahead of us. If there is something which you need in terms of OSGi bundles, please feel free to contribute. If there is something which we can improve for the bundles, please guide us so we can do the right thing.

This is our p2 build: https://github.com/eclipse/eclipse-collections/blob/master/p2-repository/pom.xml
Seems like we are exporting everything. Is there a setting we are missing?
@guw is there something we might be missing?

For the milestone versions I am working on getting it pushed to maven central as well as p2. However, currently I am encountering issues with jar signing. See Issue #287, #294 . If you have any insights for fixing either of those, please feel free to give your suggestions so that we can get right for 9.0.0 release.
The 9.0.0.M2 (Milestone 2) is scheduled for 8/19, I would like to have these issues sorted out before that.

If you really need older Java 7 and below compatible versions we can discuss that. Currently, we don't have any plans to do a service release for 7.x for p2 repository. The 7.x OSGi bundles were exported via Apache Felix, so may be you can use them but you won't find them in p2 repo.

guw:
[...]

PR #326, opened 2017-08-21
[#319] Export all packages to OSGi as non-internal

bergmanngabor:
Task-Url: https://github.com/eclipse/eclipse-collections/issues/319
Signed-off-by: bergmann <bergmann@mit.bme.hu>

### DIFF
--- a/p2-repository/org.eclipse.collections/osgi.bnd
+++ b/p2-repository/org.eclipse.collections/osgi.bnd
@@ -5,7 +5,6 @@ jcip-annotations-version=${version;===;1.0}
 Export-Package: \
  !about.html,!about_files, \
  *.internal*;x-internal:=true;version="${package-version}", \
- *.impl*;x-internal:=true;version="${package-version}", \
  *;version="${package-version}"
 
 Import-Package: \


nikhilnanivadekar:
The build timeouts sometimes. I have restarted it. Let's see if it goes through.

p2-repository/org.eclipse.collections/osgi.bnd
```diff
@@ -4,8 +4,6 @@ jcip-annotations-version=${version;===;1.0}
 
 Export-Package: \
  !about.html,!about_files, \
- *.internal*;x-internal:=true;version="${package-version}", \
```
guw: Why dropping the `internal` package?

p2-repository/org.eclipse.collections/osgi.bnd
```diff
@@ -4,8 +4,6 @@ jcip-annotations-version=${version;===;1.0}
...
- *.impl*;x-internal:=true;version="${package-version}", \
```
guw: @nikhilnanivadekar dropping the `impl` rule is sufficient as the following catch-all rule `*` exports everything using the correct version. However, I would still recommend exporting the `internal` packages with the` x-internal` flag set.

nikhilnanivadekar: Makes sense thanks!

bergmanngabor:
@nikhilnanivadekar @guw 

OK, now only *internal packages are marked as internal, and everything is exported.

PR #330, opened 2017-08-24
Implement distinctBy on ListIterable.

donraab:

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableListTestCase.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableListTestCase.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Goldman Sachs.
+ * Copyright (c) 2017 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -18,6 +18,7 @@ import java.util.ListIterator;
 import java.util.NoSuchElementException;
 
 import org.eclipse.collections.api.RichIterable;
+import org.eclipse.collections.api.block.HashingStrategy;
 import org.eclipse.collections.api.block.function.Function;
 import org.eclipse.collections.api.collection.ImmutableCollection;
 import org.eclipse.collections.api.collection.MutableCollection;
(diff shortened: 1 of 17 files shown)


unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/AbstractListTestCase.java
```diff
@@ -380,6 +380,17 @@ public void distinctWithHashingStrategy()
...
 
+    /**
+     * @since 9.0.
+     */
+    @Test
+    public void distinctBy()
+    {
+        ListIterable<String> list = this.newWith("a", "A", "b", "C", "b", "D", "E", "e");
+        ListIterable<String> actual = list.distinctBy(String::toLowerCase);
+        Verify.assertListsEqual(FastList.newListWith("a", "b", "C", "D", "E"), actual.toList());
```
nikhilnanivadekar: If you use Assert.assertEquals() no need for call to toList(). It is not about efficiency, but if users refer to our tests and see something like this, it will inefficient usage.

unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableListTestCase.java
```diff
@@ -305,14 +305,28 @@ public void distinct()
...
+     * @since 9.0.
+     */
+    @Test
+    public void distinctBy()
+    {
+        MutableList<String> strings = Lists.mutable.with("A", "b", "a", "c", "B", "D", "e", "D", "e", "E");
+        ImmutableList<Integer> integers = this.classUnderTest();
+        ImmutableList<String> letters =
+                strings.subList(0, integers.size()).toImmutable().distinctBy(String::toLowerCase);
+        List<String> expectedLetters = strings.subList(0, integers.size()).distinctBy(String::toLowerCase);
```
nikhilnanivadekar: This test looks weird to me. You are testing `distinctBy()` with `distinctBy()`. 

`List<String> expected = strings.collect(String::toLowerCase).distinct();`

`List<String> actual = letters.collect(String::toLowerCase);`

It is sort of difficult to test with capital and lower case.

May be have the expected pre-populated like you do in a test below.

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/UnmodifiableMutableListTest.java
```diff
@@ -274,18 +274,29 @@ public void selectInstancesOf()
     @Test
     public void distinct()
     {
-        MutableList<Integer> list = UnmodifiableMutableList.of(FastList.newListWith(3, 1, 2, 2, 1, 3));
+        MutableList<Integer> list = UnmodifiableMutableList.of(Lists.mutable.with(3, 1, 2, 2, 1, 3));
         Verify.assertListsEqual(FastList.newListWith(3, 1, 2), list.distinct());
```
nikhilnanivadekar: Change to `Lists.mutable.with()` here as well.

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/AbstractListTestCase.java
donraab: The toList is necessary because the return type is `ListIterable`.  `ListIterable` does not work with `Verify.assertListsEqual()`.

donraab: Fixed.

unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableListTestCase.java
donraab: Fixed.

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/UnmodifiableMutableListTest.java
donraab: Fixed.

PR #343, opened 2017-09-03
Implement PrimitiveIterable.chunk().

kristeno:
Signed-off-by: Kristen O'Leary <kristen.oleary@gs.com>

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/lazyPrimitiveIterable.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/lazyPrimitiveIterable.stg
@@ -52,6 +52,12 @@ public interface Lazy<name>Iterable extends <name>Iterable
      */
     \<V> LazyIterable\<V> flatCollect(<name>ToObjectFunction\<? extends Iterable\<V>\> function);
 
+    @Override
+    default RichIterable\<PrimitiveIterable> chunk(int size)
+    {
+        throw new UnsupportedOperationException(this.getClass().getSimpleName() + ".chunk() not implemented yet");
+    }
+
     <collectPrimitive(name, type, "Boolean", "boolean")>
 
     <collectPrimitive(name, type, "Byte", "byte")>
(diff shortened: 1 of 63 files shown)


eclipse-collections-code-generator/src/main/resources/api/lazyPrimitiveIterable.stg
```diff
@@ -52,6 +52,12 @@ public interface Lazy<name>Iterable extends <name>Iterable
      */
     \<V> LazyIterable\<V> flatCollect(<name>ToObjectFunction\<? extends Iterable\<V>\> function);
 
+    @Override
+    default RichIterable\<PrimitiveIterable> chunk(int size)
+    {
+        throw new UnsupportedOperationException(this.getClass().getSimpleName() + ".chunk() not implemented yet");
```
motlin: I see a lot of UnsupportedOperationExceptions. Is this a temporary thing for the 9.0 release?

eclipse-collections-code-generator/src/main/resources/impl/list/immutable/immutablePrimitiveArrayList.stg
```diff
@@ -724,6 +726,12 @@ final class Immutable<name>ArrayList
             return next;
         }
     }
+
+    @Override
+    public RichIterable\<PrimitiveIterable> chunk(int size)
+    {
+        throw new UnsupportedOperationException(this.getClass().getSimpleName() + ".chunk() not implemented yet");
```
motlin: Why is this one unsupported? Doesn't it inherit an implementation from an abstract superclass?

eclipse-collections-code-generator/src/main/resources/api/lazyPrimitiveIterable.stg
donraab: I think the UnsupportedOperationExceptions are ok to throw in the default implementations so this can now be introduced in a minor release like 9.1.  The implementation classes should provide overrides with the appropriate implementations.

eclipse-collections-code-generator/src/main/resources/impl/list/immutable/immutablePrimitiveArrayList.stg
donraab: This should have an appropriate implementation provided.

nikhilnanivadekar:
Will be fixed via #483

PR #357, opened 2017-09-27
Implement factory class for Strings.

donraab:

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Strings.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Strings.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2017 BNY Mellon.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompany this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.impl.factory;
+
+import org.eclipse.collections.impl.string.immutable.CharAdapter;
+import org.eclipse.collections.impl.string.immutable.CodePointAdapter;
+
+/**
+ * This class can be used to create instances of CharAdapter and CodePointAdapter.
+ * <p/>
+ * <pre>
+ * CharAdapter chars1 = Strings.asChars("Hello World!");
+ * CharAdapter chars2 = Strings.toChars('a', 'b', 'c');
+ *
+ * CodePointAdapter codePoints1 = Strings.asCodePoints("Hello World!");
+ * CodePointAdapter codePoints2 = Strings.toCodePoints((int)'a', (int)'b', (int)'c');
+ * </pre>
+ */
+public final class Strings
+{
+    private Strings()
+    {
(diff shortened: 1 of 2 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Arrays.java
```diff
@@ -0,0 +1,34 @@
...
+{
+    private Arrays()
+    {
+        throw new AssertionError("Suppress default constructor for noninstantiability");
+    }
+
+    @SafeVarargs
+    public static <T> ArrayAdapter<T> asList(T... array)
+    {
+        return ArrayAdapter.adapt(array);
```
motlin: ArrayAdapter.adapt() already takes varargs.

unit-tests/src/test/java/org/eclipse/collections/impl/factory/ArraysTest.java
```diff
@@ -0,0 +1,26 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ArraysTest
+{
+    @Test
+    public void adapt()
+    {
+        ArrayAdapter<String> adapted = Arrays.asList("1", "2", "3");
+        Assert.assertEquals(java.util.Arrays.asList("1", "2", "3"), adapted);
```
motlin: Both the class and the method are named the same as the ones in java.util.Arrays.asList().

eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Arrays.java
donraab: Yes... This addition was purely to provide symmetry with the methods we have on Lists, Sets and Maps.  The method name could either be asList() or adapt().  The method adapt() would be more consistent with the other factory types.  But having a class called Arrays, I thought it might work better to have the same method signature as Arrays.asList() as this could then be used in place of Arrays.asList() which is quite commonly used.

donraab: We have symmetry in the Iterate classes (Iterate, ListIterate, MapIterate, SetIterate, ArrayIterate, StringIterate).  We did not have full symmetry in the factory classes.

motlin:
Lists.fixedSize.with(...) already returns an ArrayAdapter. I could see this
adding symmetry if it were named with() and if it returned T[].

eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Arrays.java
donraab: @motlin @nikhilnanivadekar  Any thoughts on the naming of asList vs. adapt on Arrays?  I am happy to rename it adapt if you think being more consistent with the other factories is better than being consistent with the java.util.Arrays naming.

nikhilnanivadekar: I don't really like the name Arrays. It is going to be confusing to the users, as it is the same name as Arrays in JDK. Do we really need this symmetry?

donraab:
@motlin This is a great point.  Lists.fixedSize already does already cover this.  It might still be harder to discovered than Arrays.with().  Adding these methods to the factory classes is purely to help developers discover APIs for creating types, without requiring them to have to first know the type name.  I can either remove Arrays or change the name to Arrays.with().  Thoughts?

donraab:
@motlin @nikhilnanivadekar I removed the Arrays and ArraysTest classes from the commit.  The Arrays class from JDK will suffice, along with Lists.fixedSize.with().

PR #364, opened 2017-10-03
Add withInitialCapacity to Lists.mutable.

bhav0904:
Signed-off-by: Bhavana Hindupur <bhavana.hindupur@gmail.com>

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MutableListFactoryImpl.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MutableListFactoryImpl.java
@@ -49,6 +49,23 @@ public enum MutableListFactoryImpl implements MutableListFactory
     }
 
     @Override
+    public <T> MutableList<T> ofInitialCapacity(int capacity)
+    {
+        return this.withInitialCapacity(capacity);
+    }
+
+    @Override
+    public <T> MutableList<T> withInitialCapacity(int capacity)
+    {
+        if (capacity < 0)
+        {
+            throw new IllegalArgumentException("initial capacity cannot be less than 0");
+        }
+
+        return FastList.newList(capacity);
+    }
+
+    @Override
     public <T> MutableList<T> ofAll(Iterable<? extends T> iterable)
     {
         return this.withAll(iterable);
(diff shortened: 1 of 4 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
```diff
@@ -167,6 +167,11 @@ public FastList(Collection<? extends T> source)
 
     public static <E> FastList<E> newList(int initialCapacity)
     {
+        if (initialCapacity < 0)
```
nikhilnanivadekar: This is an API breaking changing on implementation. This check needs to move up in the factory class and for next major version we should move it down to FastList.

unit-tests/src/test/java/org/eclipse/collections/impl/factory/ListsTest.java
```diff
@@ -135,6 +138,74 @@ public void emptyList()
...
+        this.assertPresizedListEquals(32, (FastList<String>) list5);
+
+        Verify.assertThrows(IllegalArgumentException.class, () -> Lists.mutable.withInitialCapacity(-6));
+    }
+
+    private void assertPresizedListEquals(int initialCapacity, FastList<String> list)
+    {
+        try
+        {
+            Field itemsField = FastList.class.getDeclaredField("items");
```
nikhilnanivadekar: These will work with Java-9 currently but will show a warning. We need to add this check:
`Assume.assumeTrue(System.getProperty("java.version").startsWith("1.8."));`

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MutableListFactoryImpl.java
```diff
@@ -13,6 +13,8 @@
 import org.eclipse.collections.api.block.function.Function0;
 import org.eclipse.collections.api.factory.list.MutableListFactory;
 import org.eclipse.collections.api.list.MutableList;
+import org.eclipse.collections.api.set.MutableSet;
```
nikhilnanivadekar: There are unused imports. Please remove them.

nikhilnanivadekar:
I have approved this. Can you please rebase?

PR #377, opened 2017-10-23
Implement removeIf for mutable primitive collections.

donraab:


nikhilnanivadekar:
I am fine with this change and the code looks good. 
I would like to have another set of test cases which will remove more than 1 element. We can have `1, 2, 3, 4` and remove even or odd elements.

donraab:
Agreed.  I will update the tests as you have suggested.

Issue #288, opened 2017-06-24
Enable release train repository aggregation for Eclipse


nikhilnanivadekar:
Cc: @guw

guw:
FWIW, I'm not sure if there is better documentation but this one is a good intro:
https://wiki.eclipse.org/Simrel/Contributing_to_Simrel_Aggregation_Build
https://wiki.eclipse.org/SimRel/Simultaneous_Release_FAQ

Basically, you need to make the p2 repo url available to the B3 aggregation process. This is the one that is aggregating from all participating project.


Certain requirements must be met:
https://wiki.eclipse.org/SimRel/Simultaneous_Release_Requirements

ujhelyiz:
Hi,

we at the [VIATRA project](https://www.eclipse.org/viatra/) are really interested in Eclipse Collections. However, given we are part of the Photon release train, for that we have to ensure that Eclipse Collections is available from the aggregator. Do you have any plans how and when would you like to join?

If you need any help, I also volunteer to help; I have followed the required process for the VIATRA project last year.

Thank you for your help!

nikhilnanivadekar:
@ujhelyiz we are definitely interested in adding Eclipse Collections to the release train. We wanted to do that for Oxygen but we were too late to get to the process. We would like to join as soon as it is possible and feasible.
Before we get on the aggregator we will need our p2 artifacts to be signed. I tried a lot to get that going as a part of 9.0.0 but I was not successful. If you could help us with that as a first step it will get one deliverable out of the way. We can do service releases for the signing purposes.
Is there something in Eclipse Collections which would need to be fixed or will need approvals from the Photon release group then we can get that going as well? 
What are the timelines we want to target the integration?

Thanks again for your interest in Eclipse Collections!

ujhelyiz:
The most important target dates are available in the following two sites:

 * https://wiki.eclipse.org/SimRel/Simultaneous_Release_Requirements
 * https://wiki.eclipse.org/Photon/Simultaneous_Release_Plan

As far as I can see, most things you are doing are fine for inclusion into the simultaneous release, no specific approval is required. There are some steps to be done (mostly related to stating your intent):
  1. Create a release record at PMI for the build that corresponds to the version that will be included in Photon. It can be reversioned later, if your plans are not finalized yet, e.g. https://projects.eclipse.org/projects/modeling.viatra/releases/1.8.0 is still a stub because we have an interim release this december.
[...]

PR #383, opened 2017-10-27
Provide a feature project for release train

ujhelyiz:
As discussed in #288 Eclipse Collections does not provide an Eclipse Feature, used to provide end-user visible display in the Simultaneous Release repository.

The EBR based repository generation does not seem to provide any way to inject a feature generation, so the only way I could see was to provide another maven configuration that relies on the previously created repository to provide a new one with a corresponding feature project.

This configuration can be started by executing `mvn clean install` from the new folder `p2-repository/feature-repository` folder, but only after the original build executed successfully. After the new tasks execute the created repository will be available from `p2-repository/feature-repository/org.eclipse.collections.repository/target/repository` folder.

### DIFF
--- a/p2-feature/.project
+++ b/p2-feature/.project
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>p2-feature</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.m2e.core.maven2Nature</nature>
+	</natures>
+</projectDescription>
--- a/p2-feature/.settings/org.eclipse.core.resources.prefs
+++ b/p2-feature/.settings/org.eclipse.core.resources.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+encoding/<project>=UTF-8
--- a/p2-feature/org.eclipse.collections.feature/.project
+++ b/p2-feature/org.eclipse.collections.feature/.project
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.collections.feature</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.pde.FeatureBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.pde.FeatureNature</nature>
+	</natures>
+</projectDescription>
--- a/p2-feature/org.eclipse.collections.feature/build.properties
+++ b/p2-feature/org.eclipse.collections.feature/build.properties
@@ -1,0 +1,3 @@
+bin.includes = feature.xml,\
+               feature.properties
+src.includes = feature.properties
--- a/p2-feature/org.eclipse.collections.feature/epl-v10.html
+++ b/p2-feature/org.eclipse.collections.feature/epl-v10.html
@@ -1,0 +1,328 @@
+<html xmlns:o="urn:schemas-microsoft-com:office:office"
+xmlns:w="urn:schemas-microsoft-com:office:word"
+xmlns="http://www.w3.org/TR/REC-html40">
+
+<head>
+<meta http-equiv=Content-Type content="text/html; charset=windows-1252">
+<meta name=ProgId content=Word.Document>
+<meta name=Generator content="Microsoft Word 9">
+<meta name=Originator content="Microsoft Word 9">
+<link rel=File-List
+href="./Eclipse%20EPL%202003_11_10%20Final_files/filelist.xml">
+<title>Eclipse Public License - Version 1.0</title>
+<!--[if gte mso 9]><xml>
+ <o:DocumentProperties>
+  <o:Revision>2</o:Revision>
+  <o:TotalTime>3</o:TotalTime>
+  <o:Created>2004-03-05T23:03:00Z</o:Created>
+  <o:LastSaved>2004-03-05T23:03:00Z</o:LastSaved>
+  <o:Pages>4</o:Pages>
+  <o:Words>1626</o:Words>
+  <o:Characters>9270</o:Characters>
+   <o:Lines>77</o:Lines>
+  <o:Paragraphs>18</o:Paragraphs>
+  <o:CharactersWithSpaces>11384</o:CharactersWithSpaces>
+  <o:Version>9.4402</o:Version>
+ </o:DocumentProperties>
+</xml><![endif]--><!--[if gte mso 9]><xml>
+ <w:WordDocument>
+  <w:TrackRevisions/>
+ </w:WordDocument>
+</xml><![endif]-->
+<style>
+<!--
+ /* Font Definitions */
+@font-face
+	{font-family:Tahoma;
+	panose-1:2 11 6 4 3 5 4 4 2 4;
+	mso-font-charset:0;
+	mso-generic-font-family:swiss;
+	mso-font-pitch:variable;
+	mso-font-signature:553679495 -2147483648 8 0 66047 0;}
+ /* Style Definitions */
+p.MsoNormal, li.MsoNormal, div.MsoNormal
+	{mso-style-parent:"";
+	margin:0in;
+	margin-bottom:.0001pt;
+	mso-pagination:widow-orphan;
+	font-size:12.0pt;
+	font-family:"Times New Roman";
+	mso-fareast-font-family:"Times New Roman";}
+p
+	{margin-right:0in;
+	mso-margin-top-alt:auto;
+	mso-margin-bottom-alt:auto;
+	margin-left:0in;
+	mso-pagination:widow-orphan;
+	font-size:12.0pt;
+	font-family:"Times New Roman";
+	mso-fareast-font-family:"Times New Roman";}
+p.BalloonText, li.BalloonText, div.BalloonText
+	{mso-style-name:"Balloon Text";
+	margin:0in;
+	margin-bottom:.0001pt;
+	mso-pagination:widow-orphan;
+	font-size:8.0pt;
+	font-family:Tahoma;
+	mso-fareast-font-family:"Times New Roman";}
+@page Section1
+	{size:8.5in 11.0in;
+	margin:1.0in 1.25in 1.0in 1.25in;
+	mso-header-margin:.5in;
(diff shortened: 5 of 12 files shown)


guw:
@ujhelyiz - one trick you could try is defining the feature dependencies as Maven dependency in the pom.xml

guw:
Ah right, forgot about that. It used to work with Tycho Bundle project types only. 

Here is another thing that could work: Instead of building the p2 repo from a feature you could build one by _mirroring_ the `org.eclipse.collections` bundle into a new (empty) repo. A the bundle could be added to a category directly. Note, the p2 repo of the bundle will be generated during the build.

ujhelyiz:
I am sorry but I don't understand how mirroring changes the setup here: the feature project requires the OSGi bundle available at the start of its reactor (given it is not a Tycho bundle) so moving the feature creation and the EBR bundle into the same reactor does not seem possible. On the other hand, the current setup reuses the repository right from the previously run maven build at least (naturally, it could reuse a previously published repository as well without issues if necessary - that is one reason why the repository url is defined as a property).

guw:
I think it would be better to take a step back and define the goals. From my understanding, the goal is to contribute the Eclipse Collections bundle to the release train. Is that correct?

FWIW, I'm challenging two things:
- a second separate rector build with a feature project
- participation in release train repository with "end user visibility" 

(1) In order to contribute Eclipse Collections to the release train repository, the following steps are necessary.
- publish the p2 repo containing the bundle to download.eclipse.org
- create an `aggrcon` file in the `org.eclipse.simrel.build` Git repo which
   - points to the p2 repo
   - lists the bundle to be included in the aggregated repository
- define a category (eg., Eclipse Collections Target Components) for Eclipse Collections in `simrel.aggr` to make the bundles visible

Looking at the aggregator documentation, it seems that only the last steps require a feature. Just contributing to the release train repository does not, i.e. one can contribute only [bundles](https://wiki.eclipse.org/CBI/aggregator/manual#Bundle).

So yes - for full visibility within a category a second build is necessary which created the feature required for the aggregation process.

(2) "End User Visibility"

Thinking about that requirement, I begin to winder about the motivation. The only thing that makes sense is making Eclipse Collections available as Target Components so that's discoverable and consumable in a target platform.

However, there are the following drawbacks that need to be considered:
- (slower) update frequency of release train repo (fewer, fixed dates)
- additional work of contributing regularly to the release train repository
- additional requirements to be allowed to participate in the release train repository

Given all the things mentioned above, it sounds way easier from a project perspective to publish a p2 repo from within the project - independent from the release train. In the beginning we also participated in the common repo but then dropped out as we released the extra work is sort of wasted because clients consumed our target components directly from the project URL rather then through the common repo.

FWIW, this is a nice URL and it works perfect with target platforms:
`http://download.eclipse.org/collections/9.0.0/repository`

Instead of contributing to the release train repo, you should think about creating a composite repository like:
`http://download.eclipse.org/collections/repository`

ujhelyiz:
Thanks for the detailed comments, I agree with your reasoning here. I started this entire discussion a few weeks ago from a selfish motivation to have Eclipse Collections available in simrel, and because I saw that there is an open issue for that, thought I can do that for the Collections project instead for ours.

----

(1) Release train participation

I have already provided a gerrit change https://git.eclipse.org/r/#/c/110114/ that already pushes the OSGi bundle to the simrel repository. If it is ok for the Eclipse Collections project that it is not visible in any category there, the gerrit change should be merged and this PR should be abandoned.

A minor remark (that does not change the original concept here): in the simrel repository it is not allowed to create a new category, you have to put everything into the predefined categories.

(2) I agree that there is _little_ (but not _no_) value in directly installing Eclipse Collections into an IDE for a project that uses explicit target platforms (as recommended). However, Eclipse by default creates a target platform from the bundles already installed; so for those using the default target platform there might be valid uses for explicitly installing from simrel. Regardless, it is quite fine not to support this use case by Eclipse Collections, especially given how it complicates the build process of the project.

I guess, everything depends on how the project wants to handle this. @nikhilnanivadekar what do you think?

nikhilnanivadekar:
Thank you @ujhelyiz and @guw for the explanations.
From what I understood, I have few questions (please correct me where I am wrong, I am still learning):

1) We can aggregate as a bundle in simrel but users will not be able to see Eclipse Collections so they cannot download it as a dependency.  
Our main motivation to contribute to simrel was to leverage the momentum which each Eclipse release has, standardize our release schedules and increase end user visibility and adoption.
So, what can we do to increase visibility?
Regarding the gerrit review, I can see it, but since I am not an expert I do not know what and how to review. Please guide me in what generally a reviewer looks at for aggregation builds.

2) Currently the way Eclipse Collections is setup, it is difficult to create a feature out of it. Can we have the feature be a part of a separate repo which has it's own separate builds so we need not complicate the existing builds? Will this help solve the problem of discoverability.

3) What is a composite repository? Should we create one irrespective of whether we contribute to release train? Is this the composite repo that we are referring to: https://help.eclipse.org/neon/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2Fp2_composite_repositories.htm

4) I did not follow this point:

> Eclipse by default creates a target platform from the bundles already installed; so for those using the default target platform there might be valid uses for explicitly installing from simrel.

ujhelyiz:
1. If your repository has a feature defined, it can be added to one of the categories in the simrel repository where it will be visible by default. Otherwise it will be there but requires non-trivial steps to make it visible. However, it is important to note that the increased visibility provided by the simrel repository is only relevant if (a) you want to provide your framework for OSGi/Eclipse plug-in development (e.g. the VIATRA project will be one of its users), or (b) you want to provide some tools inside Eclipse, e.g. a specific editor (not relevant to your project). For other users, you want to make the framework available from a maven repository or as a direct download instead (both already available). 
About the simrel, the following information has to be provided (and correct):
    * A project-specific file that contains the contribution.
    * Correct contact information (in case anything goes wrong with the build, the persons mentioned here will be notified to take steps).
    * The repository URL where the contribution has to be downloaded from.
    * The features and/or plug-ins to be downloaded, together with their specific version (the automatic builds validate that it can be downloaded successfully - but it does not check whether all version numbers are fully specified).
    * A category specification about where to put the results (currently this part is missing, as it is not really applicable for OSGi bundles, only features).
2. The change in this PR can be used to this effect. It provides a feature project and scripts that downloads the previously built repository. Basically, the `collections.repository.url` property describes the repository URL; by default it is bound to a workspace-relative path, but it can rely _any_ available URL as a source. It might be beneficial to move the new projects to another directory or rename them etc., but the functionality is already there.
3. Yes. The basic idea here is that you provide one (version-independent) repository URL that references all released versions; so users who always want the latest version look at the composite that will point them to all the specific urls. As an example, I can point you to the release repository of VIATRA: http://download.eclipse.org/viatra/updates/release/?d - here the `compositeArtifacts.xml` and `compositeContent.xml` files point to all version-specific urls. The xml structure is simple enough that it can easily be manually edited for release builds.
4. Ok, that was not entirely clear; I try to have another go. When developing Eclipse plug-ins, you have to have a target platform defined that is used for compiling and running your new plug-ins. If you don't specify any, an implicit target platform is provided that is all the plug-ins installed in your Eclipse. By installing Eclipse Collections to your Eclipse IDE, it will also be available in this implicit target platform. The way I said this use case might not be important is that it is not recommended to rely on this implicit target platform but specify one for each project - the main reason for this is reproducibility.

nikhilnanivadekar:
Thank you very much for the explanation @ujhelyiz . 
From the explanation, for visibility we will need a feature so, we will need this PR 😃 . However, to successfully integrate this PR:
- we will need a new Hudson build which will build and push this feature. 
- this build can refer to the published bundle on download.eclipse.org or it can refer to the bundle in build workspace and create a feature. 
- why is it in the same module as p2-repository? Should it be in a different module say "p2-feature"? 
- can p2-feature refer to the bundle built in p2-repository? Technically everything should be available in the build workspace. 
Or I missed something there?
Sorry I forgot to answer @guw before, yes only the bundle is copied in to the downloads area.

ujhelyiz:
The build job (or an update to an existing build job) will be necessary; by the use of `collections.repository.url` property it can refer to any published repository, e.g. one created by the current hudson job; one archived by a previous hudson job or one already published on download.eclipse.org. 

E.g. to reuse the 9.0.0 release build, the following command can be used:
`mvn clean install -Dcollections.repository.url=http://download.eclipse.org/collections/9.0.0/repository`

The reason I have put everything into the p2-repository module is that I tried to make it work in a single reactor, but that proved impossible. After that, I postponed the module reorganization until I learned how would you like to have it. I have update this PR with a new commit that moves the new stuff into a top-level `p2-feature` module.

Referring to the bundle built by p2-repository is also possible; if you look at the default value of the `collections.repository.url` property, it does that by referring to the p2 repository built in the workspace.

nikhilnanivadekar:
@ujhelyiz I see the use of p2-repository. To be honest, now I am beginning to understand the changes 😄 . 
I presume, we will have to configure a build on Hudson which might take a few iterations. 
For my understanding, where and how is the feature pushed/contributed to? I mean after the build is completed, will we get a jar or a zip for p2-feature? What do we do after that build has finished?

We will need to add a dependency between p2-repository module and p2-feature, so that p2-repository is built first.

ujhelyiz:
Sorry for not being more clear about this change. To be honest, we are using similar schemes for years, so a lot of things that are trivial to me are most likely not to you. :)

Publishing the repository happens basically the same way as you do in https://hudson.eclipse.org/collections/job/publish-p2-repo : you copy the directory from `p2-feature/org.eclipse.collections.repository/target/repository` (and `p2-feature/org.eclipse.collections.repository/target/p2-repository-«version».zip` if you also want to publish the zipped version as well) into the directory `/home/data/httpd/download.eclipse.org/collections` (in any structure you seem fit).

The easiest way I see to add this is to modify the publish-p2-repo job:

 * Keep the existing maven build step as is.
 * Add a _new_ maven build step between the existing maven build step and the shell script execution.
    * The most important thing to set up is POM file attribute (hidden behind an advanced button) - it should be set to `p2-feature/pom.xml`
    * Otherwise a default `clean install` should work fine.
 * Then update the shell scripts to use the repository created by the second maven step.



No formal dependency declaration is required between the p2-feature and p2-repository modules because they are not executed in the same maven reactor; we had a discussion before in this thread stating that would not work because the way Tycho handles dependencies. In my proposed build job this is handled by executing two maven builds one after the other.

For the copyright headers, I am sorry I have missed them. In case of the VIATRA project I am the one who usually nags others to add them, so I should have known better. :) I will add them later today.

ujhelyiz:
I have added copyright notices to the pom.xml files; I don't think there are any other files requiring these notices. If I'm wrong, please point me there.

nikhilnanivadekar:
@ujhelyiz the copyright headers look good. Can you please squash all the commits to be 1 commit?
I think it is good to merge after that.

ujhelyiz:
I have created a squashed version and rebased on current head of the master branch.

nikhilnanivadekar:
Thank you! I have merged this PR.
Let me kick off 9.1.0 release so that we can create a few milestone builds to test out.

PR #388, opened 2017-11-02
Implement collectWithIndex on ListIterate and RandomAccessListIterate.

donraab:

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ListIterate.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ListIterate.java
@@ -37,6 +37,7 @@ import org.eclipse.collections.api.block.function.primitive.IntFunction;
 import org.eclipse.collections.api.block.function.primitive.IntObjectToIntFunction;
 import org.eclipse.collections.api.block.function.primitive.LongFunction;
 import org.eclipse.collections.api.block.function.primitive.LongObjectToLongFunction;
+import org.eclipse.collections.api.block.function.primitive.ObjectIntToObjectFunction;
 import org.eclipse.collections.api.block.function.primitive.ShortFunction;
 import org.eclipse.collections.api.block.predicate.Predicate;
 import org.eclipse.collections.api.block.predicate.Predicate2;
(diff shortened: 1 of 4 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ListIterate.java
```diff
@@ -357,6 +358,36 @@ private static boolean nonRandomAccessEquals(List<?> one, List<?> two)
     }
 
     /**
+     * @since 9.1
+     */
+    public static <T, A> MutableList<A> collectWithIndex(
```
nikhilnanivadekar: In a utility like ListIterate is it better to provide only the target collection variant?

unit-tests/src/test/java/org/eclipse/collections/impl/utility/ListIterateTest.java
```diff
@@ -192,6 +196,30 @@ private void assertCollect(List<Boolean> list)
         Verify.assertListsEqual(newCollection2, FastList.newListWith("true", "false", "null"));
     }
 
+    /**
+     * @since 9.1.
+     */
+    @Test
+    public void collectWithIndex()
+    {
+        MutableList<Boolean> list = Lists.fixedSize.of(true, false, null);
```
nikhilnanivadekar: Can you please add 1 more test case which uses ArrayList<>()?

donraab:
I disagree.  I did not know that OrderedIterate existed, but looking at it, it has one method that expects OrderedIterable.  I am limiting this pull request specifically to ListIterate and RandomAccessListIterate intentionally.  Anything else is out of scope for this work.

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ListIterate.java
donraab: Good question.  The non-target collection variant is more generally useful when used as utility, which is why I included it.

unit-tests/src/test/java/org/eclipse/collections/impl/utility/ListIterateTest.java
donraab: Yes, I will add a test.

nikhilnanivadekar:
ok fair enough. But if we add `collectWithIndex()` on an interface, it will go on `OrderedIterable`

donraab:
Yes, I am planning on adding the method to the `OrderedIterable` interface.

nikhilnanivadekar:
Raised issue #395

PR #389, opened 2017-11-03
Implement groupByUniqueKey Collector on Collectors2. Resolves #363

### LINKED ISSUE #363: Implement groupByUniqueKey Collector on Collectors2.

TheLoneKing:

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
@@ -784,6 +784,40 @@ public final class Collectors2
                 EMPTY_CHARACTERISTICS);
     }
 
+    /**
+     * <p>Same as {@link #groupBy(Function, Supplier)}, except the result of evaluating groupBy function should return a
+     * unique key, or else an exception is thrown.</p>
+     *
+     * <p>Equivalent to using @{@link RichIterable#groupByUniqueKey(Function, MutableMap)}</p>
+     */
+    public static <T, K, R extends MutableMap<K, T>> Collector<T, ?, R> groupByUniqueKey(
+            Function<? super T, ? extends K> groupBy,
+            Supplier<R> supplier)
+    {
+        return Collector.of(
+                supplier,
+                (map, each) ->
+                {
+                    K key = groupBy.valueOf(each);
+                    if (map.put(key, each) != null)
+                    {
+                        throw new IllegalStateException("Key " + key + " already exists in map!");
+                    }
+                },
+                (r1, r2) ->
+                {
+                    r2.forEachKeyValue((key, value) ->
+                    {
+                        if (r1.put(key, value) != null)
+                        {
+                            throw new IllegalStateException("Key " + key + " already exists in map!");
+                        }
+                    });
+                    return r1;
+                },
+                EMPTY_CHARACTERISTICS);
+    }
+
     private static <T, K, A extends MutableMultimap<K, T>, R extends ImmutableMultimap<K, T>> Collector<T, ?, R> groupByImmutable(
             Function<? super T, ? extends K> groupBy,
             Supplier<A> supplier,
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2Test.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2Test.java
@@ -20,11 +20,13 @@ import org.eclipse.collections.api.bag.MutableBag;
 import org.eclipse.collections.api.block.function.Function;
 import org.eclipse.collections.api.collection.primitive.MutableIntCollection;
 import org.eclipse.collections.api.list.MutableList;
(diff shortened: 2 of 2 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
```diff
@@ -784,6 +784,40 @@ private Collectors2()
                 EMPTY_CHARACTERISTICS);
     }
 
+    /**
+     * <p>Same as {@link #groupBy(Function, Supplier)}, except the result of evaluating groupBy function should return a
+     * unique key, or else an exception is thrown.</p>
+     *
+     * <p>Equivalent to using @{@link RichIterable#groupByUniqueKey(Function, MutableMap)}</p>
+     */
+    public static <T, K, R extends MutableMultimap<K, T>> Collector<T, ?, R> groupByUniqueKey(
```
donraab: This is the correct return signature for groupByUniqueKey: `R extends MutableMap<K, T>`

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
```diff
@@ -784,6 +784,40 @@ private Collectors2()
...
+                    K key = groupBy.valueOf(each);
+                    if (map.containsKey(key))
+                    {
+                        throw new IllegalStateException("Key " + key + " already exists in map!");
+                    }
+                    map.put(key, each);
+                },
+                (r1, r2) ->
+                {
+                    K nonUniqueKey = r2.keysView().detect(key -> r2.get(key).size() > 1);
```
donraab: You could use `putAll` and check the size of the result and check that it has increased by the size of the Map being added.  But this will not help you figure out which key was a duplicate.  I would just use `r2.forEachKeyValue((k, v) -> code to put and check for dupes)`.

TheLoneKing:
Thanks for the review comments. I have made adjustments as suggested.

donraab:
Thank you, this looks pretty good.  My only concern is that we are probably not exercising the code path where the merge step will fail in parallel.  The only change I think is needed is to add a new test method for parallelStream() that uses LARGE_INTERVAL with maybe the id mod of 15_000 for the groupBy key so there will be collisions and to cause the collisions to happen in the merge step.

TheLoneKing:
Isn't the below test case already doing that?

`@Test(expected = IllegalStateException.class)
    public void groupByUniqueKey_parallelStream_throws_for_duplicate()
    {
        LARGE_INTERVAL.parallelStream().collect(Collectors2.groupByUniqueKey(id -> 1, Maps.mutable::empty));
    }`

donraab:
Run the test with code coverage turned on in your IDE and see if the following line is covered in the merge lambda.  Alternatively, set a breakpoint on the Exception line and run the code and verify if the debugger stops in the merge step on this line:

`throw new IllegalStateException("Key " + key + " already exists in map!");`

Without digging into the inner workings of the parallelStream code, my guess would be that the current code never makes it to the merge step, because it will fail on a duplicate in the accumulator.

I could be wrong, but I'd like to know all of the code paths are completely covered by the tests.

TheLoneKing:
You are right, the process never made it to the combiner. Exception was thrown from the accumulator. 

This makes me think if we really need the duplicate check in the combiner. Any duplicate would be caught in the accumulator itself. So, why don't we just delete the `forEachKeyValue()` logic and just use `putAll`?

Please let me know your thoughts on this.

motlin:
Duplicate checking has to happen in both places (duplicate within a thread
or across threads)

donraab:
Agreed, duplicate checking is necessary in both place.  We need tests that cover both scenarios.

TheLoneKing:
Okay, I have added a new test case to test exception from the combiner.

unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2Test.java
```diff
@@ -1140,4 +1142,38 @@ public void groupByEachParallel()
...
+    @Test(expected = IllegalStateException.class)
+    public void groupByUniqueKey_parallelStream_throws_for_duplicate()
+    {
+        LARGE_INTERVAL.parallelStream().collect(Collectors2.groupByUniqueKey(id -> 1, Maps.mutable::empty));
+    }
+
+    @Test(expected = IllegalStateException.class)
+    public void groupByUniqueKey_parallelStream_duplicate_from_combiner()
+    {
+        LARGE_INTERVAL.parallelStream().collect(Collectors2.groupByUniqueKey(id -> id == 15000 ? 1 : id, Maps.mutable::empty));
```
donraab: This should work ok, as it creates exactly one duplicate id.

nikhilnanivadekar:
Please rebase on to master.

TheLoneKing:
Rebase done.

donraab:
Great, thank you for the contribution!  I will merge as soon as the build passes.

PR #412, opened 2017-11-20
Add Javadoc to Pair, Twin, Tuples, PrimitiveTuples, ObjectPrimitivePa…

### LINKED ISSUE #390: Update Interface Javadocs to suggest possible ways of initialization.
I'm having trouble coming with a short description, but let me tell you my hardship.

I like that I can have all this new entity objects like IntObjectPair. But everything is an interface, and it's not easy to find where and what is the correct factory (method) for an instance. In this case, I had to see implementations of IntObjectPair (IntObjectPairImpl) and search for usages of IntObjectPairImpl's constructor to find out that the factory method is PrimitiveTuples.pair().

So, this information could be in the Javadoc of the interface. IntObjectPair javadoc only has "@see Pair", and Pair only has trivial information. A simple comment like "To get instances of this interface, use PrimitiveTuples.pair()" or even a "@see PrimitiveTuples.pair()" would help tremendously.

Thanks!
Leo.

donraab:
…ir, PrimitivePrimitivePair and PrimitiveObjectPair. Fixes #390.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/tuple/primitive/PrimitiveTuples.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/tuple/primitive/PrimitiveTuples.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Goldman Sachs.
+ * Copyright (c) 2017 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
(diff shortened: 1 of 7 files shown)


leomrlima:
If that's automatically created, how do you guarantee that the comments will not vanish on next change?

eclipse-collections/src/main/java/org/eclipse/collections/impl/tuple/primitive/PrimitiveTuples.java
```diff
@@ -91,6 +91,9 @@
 import org.eclipse.collections.api.tuple.primitive.ShortObjectPair;
 import org.eclipse.collections.api.tuple.primitive.ShortShortPair;
 
+/**
+ * This class should be used to create instances of object/primitive, primitive/primitive and primitive/primitive pairs.
```
leomrlima: This looks wrong: last two are the same, maybe it was supposed to be primitive/object?

donraab: good catch.  fixed.

donraab:
I changed the comments in the templates, so they get generated with the code.

leomrlima:
I see. So why doesn't it show commit changes in the generated files as well?

donraab:
The generated files are not checked in.

leomrlima:
Who/what generates them, then?

donraab:
The build process generates the primitive code using StringTemplate.

PR #414, opened 2017-11-26
Implement collectWithIndex on primitive ordered iterables.

donraab:


donraab:
I added tests for collectWithIndex collecting to a List and a Set.

PR #415, opened 2017-11-26
Implement aggregateBy Collector on Collectors2. Resolves #361

### LINKED ISSUE #361: Implement aggregateBy Collector on Collectors2.

TheLoneKing:

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
@@ -31,6 +31,7 @@ import org.eclipse.collections.api.bag.sorted.MutableSortedBag;
 import org.eclipse.collections.api.bimap.ImmutableBiMap;
 import org.eclipse.collections.api.bimap.MutableBiMap;
 import org.eclipse.collections.api.block.function.Function;
+import org.eclipse.collections.api.block.function.Function0;
 import org.eclipse.collections.api.block.function.Function2;
 import org.eclipse.collections.api.block.function.primitive.BooleanFunction;
 import org.eclipse.collections.api.block.function.primitive.ByteFunction;
@@ -55,6 +56,7 @@ import org.eclipse.collections.api.list.ImmutableList;
 import org.eclipse.collections.api.list.MutableList;
 import org.eclipse.collections.api.map.ImmutableMap;
 import org.eclipse.collections.api.map.MutableMap;
+import org.eclipse.collections.api.map.MutableMapIterable;
 import org.eclipse.collections.api.map.primitive.MutableObjectDoubleMap;
 import org.eclipse.collections.api.map.primitive.MutableObjectLongMap;
 import org.eclipse.collections.api.multimap.ImmutableMultimap;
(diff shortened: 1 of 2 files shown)


TheLoneKing:
Can someone review this please?

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
```diff
@@ -878,6 +880,29 @@ private Collectors2()
                 EMPTY_CHARACTERISTICS);
     }
 
+    /**
+     * Groups the elements using the {@code groupBy} function and all the elements that map to the same key are
+     * aggregated together using the {@code aggregator} function. The third parameter, the {@code zeroValueFactory}
```
donraab: I believe zeroValueFactory is the second parameter here not the third.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
```diff
@@ -878,6 +880,29 @@ private Collectors2()
...
+     * function, creates the initial value in each aggregation. Aggregate results are allowed to be immutable as they
+     * will be replaced in the map.
+     */
+    public static <T, K, R extends MutableMap<K, T>> Collector<T, ?, R> aggregateBy(
+            Function<? super T, ? extends K> groupBy,
+            Function0<? extends T> zeroValueFactory,
+            Function2<? super T, ? super T, ? extends T> aggregator,
+            Supplier<R> supplier)
```
donraab: Why do we need to pass in the supplier if the result is always MutableMap?  If we keep the supplier, it would be good to have a test showing how a different MutableMap is used.  It would be even better if R extends MapIterable was the return type as it is on RichIterable.  This would allow then for a Supplier that can return a MutableSortedMap, which is not in the MutableMap hierarchy.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
```diff
@@ -878,6 +880,29 @@ private Collectors2()
...
+    {
+        return Collector.of(
+                supplier,
+                (map, each) -> new NonMutatingAggregationProcedure<T, K, T>(map, groupBy, zeroValueFactory, aggregator).value(each),
```
donraab: Calling new on NonMutatingAggregationProcedure will create linear garbage.  Inline the code here from that class.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
TheLoneKing: If we use MapIterable, I am not sure how to combine the results in case of parallel stream. Because both r1 and r2 would be of type R extends MapIterable and how do we add elements from r2 to r1?

donraab: Fair point, I understand now.  Ok, the signature makes sense.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
donraab: Please see if you can inline the code here so we don't have to create a new NonMutatingAggregationProcedure on every call to the accumulator.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collector/Collectors2.java
motlin: Shouldn't it be MutableMapIterable?

TheLoneKing: MutableMapIterable would work, I believe. We can use suppliers returning different MutableMap implementations.

motlin:
Right. Otherwise we wouldn't be able to aggregate into sorted maps, for
example.

donraab:
Java 10 and 11 builds failed but all others passed.  I am going to merge this request.

donraab:
Merging with the notes below.

- While IP validation build is not running, verified author's email and sign-off message manually via ECA validation tool .
- We accept JDK 10 build failure for now.

PR #418, opened 2017-12-03
Upgrade Maven plugins, add a Java 9 build on Travis.

nikhilnanivadekar:


pom.xml
```diff
@@ -271,7 +271,7 @@
 
                 <plugin>
                     <artifactId>maven-enforcer-plugin</artifactId>
-                    <version>1.4.1</version>
+                    <version>3.0.0-M1</version>
```
donraab: Is this a temporary workaround until there is a GA release?

pom.xml
```diff
@@ -233,7 +233,7 @@
 
                 <plugin>
                     <artifactId>maven-javadoc-plugin</artifactId>
-                    <version>2.10.4</version>
+                    <version>3.0.0-M1</version>
```
donraab: Is this a temporary workaround until there is a GA release?

nikhilnanivadekar: Yes we waited for about 2 months and we don't know when the plugins will release. I think it's safe to upgrade enforcer and javadoc to milestone version

pom.xml
nikhilnanivadekar: Yes, explanation same as above

PR #419, opened 2017-12-11
Fixing feature build

ujhelyiz:
The PR #383 introduced a separate build to provide Eclipse features to the Photon Simultaneous Release repository; however, it incorrectly referenced the original parent instead of being entirely standalone.

This change fixes the issue by removing the parent declaration from the p2-feature module.

### DIFF
--- a/p2-feature/pom.xml
+++ b/p2-feature/pom.xml
@@ -11,12 +11,8 @@
   -->
     <modelVersion>4.0.0</modelVersion>
 
-    <parent>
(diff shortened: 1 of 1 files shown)


nikhilnanivadekar:
The ip validation is failing as the commit needs to be signed. Can you please sign it? I'll merge and create a milestone.

ujhelyiz:
Sorry for the trouble; I have replaced the contribution with a signed version.

nikhilnanivadekar:
No worries, thanks for fixing it.

PR #426, opened 2017-12-19
fix EntrySet.removeIf concurrency

### LINKED ISSUE #424: entrySet() and values() removeIf() methods in both CHM are not thread-safe
Basically issue is the same as https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8078645 (there are two samples there counting duplicate issue) that is still present in java8 and fixed in java9.
For eclipse-collections this issue is present in ConcurrentHashMap and ConcurrentHashMapUnsafe running on both java8 and java9.

Here is my sample revealing it - https://github.com/isopov/chm-jcstress
You can build it with
`./mvnw clean package`
And if have docker reproduce it easily on different java versions:
```docker run -v `pwd`/target:/mnt/target openjdk:8u151-jre java -jar /mnt/target/chm-jcstress.jar```
issues reproduced in both j.u.c.CHM and eclipse CHMs
``` docker run -v `pwd`/target:/mnt/target openjdk:9.0.1-11-jre java -jar /mnt/target/chm-jcstress.jar```
issues reproduced in eclipse CHMs only

gs-rezaem:
fixes #424

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
@@ -1551,8 +1551,7 @@ public final class ConcurrentHashMap<K, V>
             return e;
         }
 
-        @Override
-        public void remove()
+        protected void removeByKey()
         {
             if (this.current == null)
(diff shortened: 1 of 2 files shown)


motlin:
LGTM :+1:

PR #429, opened 2017-12-23
add maven wrapper to simplify contributions

isopov:


nikhilnanivadekar:
@isopov thank you for the PR. However, I would like to avoid having binaries in the repo. Also, installing maven is a standard procedure and most IDEs (IntelliJ, Eclipse, Netbeans) have built-in maven support.
Would you considering updating our documentation if it is missing any details regarding maven setup which will help contributors?

isopov:
@nikhilnanivadekar AFAIU there are no specific requirements regarding maven for building eclipse-collections (I built without any custom configs)

If you will need some specific version of maven (e.g. minimum required version higher than majorly used) or specific config (settings.xml can be placed in .mvn folder as can be this wrapper) - this wrapper may have more use than it has now for this project.

Personally for me using "../mvnw" for build is equally convenient as "./mvnw" so I will not pursue this.

PR #430, opened 2017-12-23
add jcstress-based tests for concurrent collections

isopov:
Tests for bug reported in #424 and partially fixed in #426 

Uses http://openjdk.java.net/projects/code-tools/jcstress/

To run them in target folder `java -jar jcstress.jar`


isopov:
I've rebased commit after #426 was merged and added descriptions for failures that are currently visible with these tests.

nikhilnanivadekar:
@isopov Thank you for the PR. I have a few questions:
1) Do you think we should have these tests run as a part of our CI builds on Travis?
2) If yes, what are the config changes in travis.yml?
3) I see a lot of .settings files, are these Eclipse IDE?
4) Are there places where we have concurrency issues per your comment?
> added descriptions for failures that are currently visible with these tests.

isopov:
@nikhilnanivadekar 

>Do you think we should have these tests run as a part of our CI builds on Travis?

I think it is a separate question - they may be too heavy for every travis build.

>If yes, what are the config changes in travis.yml?

I think it may be reasonable to add them as a separate rwo in build matrix with either packaging them as junit tests (I don\t know whether there is some API for this) or `mvn install some_options_to_build_onle_jcstress_tests_with_deps && java -jar jcstress-tests/target/jcstress.jar`

>I see a lot of .settings files, are these Eclipse IDE?

They are independent from any IDE, but I developed them with the help of Eclipse IDE. Since any project files (.idea, *.iml, .classpath, .project, .settings) are not added to .gitignore and present in all other modules, I left them in the PR.

>Are there places where we have concurrency issues per your comment?

Yes, I reported issue in #424 and it was fixed in #426 But unfortunately that fix is incomplete. I have failed to reproduce remaining problems without jcstress, so that is why I submitted this PR.

nikhilnanivadekar:
We can have a cron job which runs every night. If you look at the history of travis.yml you'll see a similar effort was done for code coverage build. Adding it as a separate row matrix is reasonable. 
Regarding the incomplete fix, would you mind pointing to what is missing? 
Cc @gs-rezaem

isopov:
@nikhilnanivadekar @gs-rezaem 

> Regarding the incomplete fix, would you mind pointing to what is missing?

Here is the test (it is actually the only test now - it is copied 4 times for 2 implementations of map and values() + entrySet() views):
```
@Actor public void setFoo(ZZZZ_Result r) { r.r1 = null == map.put(1, false); }
 
@Actor public void removeIf(ZZZZ_Result r) {
         r.r2 = null == map.put(1, true);
         r.r3 = map.entrySet().removeIf(Map.Entry::getValue);
}
 
@Arbiter public void after(ZZZZ_Result r) { r.r4 = map.containsKey(1); }
```

And here are 2 unexpected test results:
-  `false, true, true, true` means that `map.put(1, true)` got executed first. But after that both removing it with `map.entrySet().removeIf(Map.Entry::getValue)` reported success and replacing it with false via `map.put(1, false);` reported that it was there and got replaced. So basically it was successfully removed twice.

- `true, false, false, true` means that `map.put(1, false);` got executed first and `map.put(1, true)` replaced it with true after that. But then `map.entrySet().removeIf(Map.Entry::getValue)` failed to remove the value.

In both descriptions I ignored the fourth reported by the test value, since issues are clearly present without considering it.

gs-rezaem:
@isopov Sorry for not responding earlier. This got lost in the holiday shuffle ☹️ 
The real problem is in the default method in `java.util.Collection` and the only way to "fix" it is to override in our code. Specifically, the problem is:
```java
                each.remove();
                removed = true;
```
It's possible that each.remove() doesn't actually remove the entry. However, I put "fix" in quotes above, because it's not clear to me what the contract of this method's return value is for concurrent access.

The javadoc simply says "true if any elements were removed". But removed by whom? Is it not sufficient that the method observed a value that fit the filter and while the method was running, that value was removed?

Many concurrent data structures are based on cooperative work completion (e.g. the famous concurrent queue). When such a structure returns success, it's not signalling that it did the work, but merely that the work was done.

Said differently, it's not clear to me how this return value could cause a problem for client code (that's meant to work with a concurrent structure).

isopov:
@gs-rezaem What I beleave to be a bug is that it is possible to execute code
```
map.put(1, true);
map.entrySet().removeIf(Map.Entry::getValue)
```
and have the map containing pair `<1,true>` after that while no other thread is putting `true` into the map.

isopov:
@gs-rezaem See please this simplified test:
```
@JCStressTest
@Outcome(id = "false", expect = Expect.ACCEPTABLE)
@Outcome(id = "true", expect = Expect.FORBIDDEN)
@State
public class ConcurrentHashMapEntrySetRemoveIfTest
{
    private final ConcurrentHashMap<Integer, Boolean> map = new ConcurrentHashMap<>();

    @Actor
    public void setFoo(Z_Result r)
    {
        map.put(1, false);
    }

    @Actor
    public void removeIf(Z_Result r)
    {
        map.put(1, true);
        map.entrySet().removeIf(Map.Entry::getValue);
    }

    @Arbiter
    public void after(Z_Result r)
    {
        r.r1 = map.containsKey(1) && map.get(1);
    }
}
```

I think that getting resulting `true` is clear bug.

gs-rezaem:
I simplified this to the core problem:

```java
@JCStressTest
@Outcome(id = ".*, .*, true, true", expect = Expect.ACCEPTABLE, desc = "")
@Outcome(expect = Expect.FORBIDDEN, desc = "Other cases are forbidden.")
@State
public class ConcurrentHashMapSizeTest
{
    private final ConcurrentHashMap<Integer, Boolean> map = new ConcurrentHashMap<>();

    @Actor
    public void setFoo(ZZZZ_Result r)
    {
        r.r1 = null == map.put(1, false);
    }

    @Actor
    public void removeKeyValue(ZZZZ_Result r)
    {
        r.r2 = null == map.put(1, true);
        r.r3 = 1 == map.size();
    }

    @Arbiter
    public void after(ZZZZ_Result r)
    {
        r.r4 = map.containsKey(1);
    }
}
```

this fails on true, false, false, true: the map is observed with zero size.
Incidentally,JDK's ConcurrentHashMap has the same problem.

We implement `put` as `addEntry`+`incrementSize`. What's happening here is, the first thread has added an entry, but hasn't incremented the size yet. The second thread comes in, replaces the entry and doesn't touch size, observing a zero size.

The way this affects the iterator is because the iterator doesn't do anything if it thinks the size is zero. We can fix the iterator by removing that optimization, but fixing size is much harder.

gs-rezaem:
prelim fix in gs-rezaem/eclipse-collections:concitrfix

BTW, the JDK has this to say about the size method:
```
 * Bear in mind that the results of aggregate status methods including
 * {@code size}, {@code isEmpty}, and {@code containsValue} are typically
 * useful only when a map is not undergoing concurrent updates in other threads.
 * Otherwise the results of these methods reflect transient states
 * that may be adequate for monitoring or estimation purposes, but not
 * for program control.
```

isopov:
@gs-rezaem So than maybe not use size() in tests in Actors and only in Arbiter?

I don't think that you example and last my are about one and the same thing. Your example demonstrate unexpected behavior that is present in j.u.c.CHM, is part of the spec and is not a bug.

My example is not reproducible on j.u.c.CHM (even on Java8 version) but shows this strange behavior on eclipse CHM. I suggest that if eclipse CHM cannot be used as a drop-in replacement for j.u.c.CHM - incompatibilities should be clearly stated in docs.

Actually, this particular PR is not about revealing such bugs, but about new module for testing them. @nikhilnanivadekar if this module is not needed - maybe we should close this PR?

gs-rezaem:
@isopov your example and the size code get to the same root cause, given the optimization in the iterator. Removing that optimization (as in gs-rezaem/eclipse-collections:concitrfix) should fix your problem, but not the size problem.

I'm not sure if the JDK disclaimer is really good enough for this case, but I'm inclined not to "fix" the size issue.

I found the module useful for tracking this down, so I would have no objection about merging it, but it should come with at least a few passing tests :smile:

isopov:
@gs-rezaem I've updated PR with latest and simplest tests (that should pass after your fix but currently fail) and resolved minor conflict.

nikhilnanivadekar:
Sent an email to Eclipse Licensing team for guidance since jcstress is GPL licensed and EPL and GPL are not compatible. Please follow the thread here: http://dev.eclipse.org/mhonarc/lists/collections-dev/msg00033.html

nikhilnanivadekar:
We have received an OK from the Eclipse License team: http://dev.eclipse.org/mhonarc/lists/collections-dev/msg00034.html

@isopov can you please rebase and we will merge. Thanks for your patience!

isopov:
@nikhilnanivadekar rebased.

nikhilnanivadekar:
@isopov I was about to merge this, but I see that the copyrights are Goldman Sachs. The copyright needs to be updated to be either your employer or you according to the Eclipse Contributor Agreement that you signed. Please refer to the Copyright Header section in the Eclipse Project Handbook: http://www.eclipse.org/projects/handbook/#ip-copyright-headers

isopov:
@nikhilnanivadekar changed to "Ivan Sopov and others" (I hope these tests will be usable not only for me)) and will be modified and extended by others ). Ok?

jcstress-tests/pom.xml
```diff
@@ -0,0 +1,100 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright (c) 2017 Goldman Sachs.
```
nikhilnanivadekar: Copyright Goldman Sachs

motlin:
I interpreted the response to mean the code cannot be checked into an
Eclipse Foundation git repository. No?

jcstress-tests/pom.xml
isopov: Sorry, fixed

isopov:
@nikhilnanivadekar @motlin AFAIK jmh is identical openjdk-project in terms of license. So if this PR cannot be merged, modules based on jmh must be removed too.

nikhilnanivadekar:
@isopov we can add them in the repo, however, it just needs to go through the CQ. Please refer to the email thread from PMC: http://dev.eclipse.org/mhonarc/lists/collections-dev/msg00040.html

nikhilnanivadekar:
@isopov a CQ was filed: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16093

We will hold off on the merge till the CQ is approved and we get a go ahead from PMC. Thanks for understanding!

isopov:
@nikhilnanivadekar rebased. Thanks for you patience, too!

PR #439, opened 2018-01-18
Implement OrderedMapAdapter, the first implementation of MutableOrderedMap.

motlin:


unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
```diff
@@ -837,6 +852,65 @@ default void RichIterable_iterationOrder()
...
+                    sumByLongIterationOrder1.add(each);
+                    return each;
+                },
+                each -> {
+                    sumByLongIterationOrder2.add(each);
+                    return 0L;
+                });
+        assertEquals(expectedIterationOrder, sumByLongIterationOrder1);
+        assertEquals(expectedIterationOrder, sumByLongIterationOrder2);
+        */
```
motlin: While writing tests to cover OrderedMapAdapter, I found some unrelated problems. Here's the first one. The sumBy* methods were essentially uncovered by tests. Now that they are covered, the tests fail for all bag implementations, since the key function is being invoked once per occurrence rather than once per unique item.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/immutable/ImmutableUnifiedMapTest.java
```diff
@@ -40,6 +43,37 @@
...
+        MapIterable<String, Integer> map = this.newWithKeysValues("Three", 3, "Two", 2, "One", 1);
+        MapIterable<Integer, String> result = map.flipUniqueValues();
+
+        // TODO: Use IterableTestCase.assertEquals instead, after setting up methods like getExpectedTransformed, but for maps.
+        Assert.assertEquals(
+                UnifiedMap.newWithKeysValues(3, "Three", 2, "Two", 1, "One"),
+                result);
+
+        // TODO: Fix bug, flipUniqueValues is implemented incorrectly in immutable maps. Compare ImmutableDoubletonMap and DoubletonMap for example.
+        this.newWithKeysValues(1, "2", 2, "2").flipUniqueValues();
```
motlin: This is a real bug. Calling flipUniqueValues() on a map with duplicate values is supposed to throw. It fails for immutable fixed size maps. When we fix it, we also need to add coverage here. This one only catches Doubleton.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MutableMapIterableTestCase.java
```diff
@@ -0,0 +1,234 @@
...
+        assertEquals(this.newWithKeysValues("3", 3, "2", 2, "1", 1, "4", 14), map3);
+
+        // TODO: This method is strange. Should we deprecate it? I could see a hyptothetical getIfAbsentPutWithKeyWith being used, but I'm having trouble imagining a use-case for getIfAbsentPutWith.
+
+        MutableMapIterable<String, Integer> map4 = this.newWithKeysValues("3", 3, "2", 2, "1", 1);
+        assertEquals(3, map4.getIfAbsentPutWith("3", x -> x + 10, 4));
+        assertEquals(this.newWithKeysValues("3", 3, "2", 2, "1", 1), map4);
+
+        assertEquals(14, map4.getIfAbsentPutWith("4", x -> x + 10, 4));
+        assertEquals(this.newWithKeysValues("3", 3, "2", 2, "1", 1, "4", 14), map4);
```
motlin: Usually the *With methods are alternatives that take a Function2 instead of a Function. Here, the With method takes a Function instead of a Function0. I'm struggling to think of an example when it would be useful. If there are examples, that's fine, otherwise I think we should consider deprecating the method.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/UnifiedMapNoIteratorTest.java
```diff
@@ -76,6 +94,26 @@ public void RichIterable_getLast_empty_null()
         // Not applicable
     }
 
+    @Override
+    public void MapIterable_detectOptional()
+    {
+        /**
+         * TODO: {@link UnifiedMap#detectOptional(org.eclipse.collections.api.block.predicate.Predicate2)} should be optimized to not use an iterator
+         */
```
motlin: This test fails for UnifiedMap.detectOptional(Predicate2), which unnecessarily delegates to an iterator. Note that detect(Predicate), detect(Predicate2), and detectOptional(Predicate) are all already tested and optimized.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MutableMapIterableTestCase.java
gs-rezaem: There are real use cases where the key can be used to create the value. This happens when the value needs to keep a reference to its key (the code looks like ```x -> new Foo(x)```). 1 see one usage in real code.

Conceptually, there is another usage, where the key can be used to lookup an initial value for the value. If the rest of the context is static, the lambda can also remain static.

motlin: To clarify, the comment is about getIfAbsentPutWith, not getIfAbsentPutWithKey. It doesn't get access to the key in its lambda. It only gets access to the param, the second argument. A hypothetical new method getIfAbsentPutWithKeyWith (taking both the key and a param) makes more sense to me. I never really paid attention to this method until I was trying to cover it in a test.

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
nikhilnanivadekar: Sorry! I don't quite follow this. If we are doing a `sumByPrimitive()` on a List, we will evaluate the function over each element, even duplicates and do an update to the value as in org.eclipse.collections.impl.block.factory.PrimitiveFunctions#sumByDoubleFunction. For Bags also it should be the same no?
Let's consider this:
```
Bag<Integer> bag = Bags.mutable.with(1, 2, 2, 2, 3, 3, 3, 3);
MutableObjectDoubleMap result = bag.sumByDouble(each -> each, each -> 1);
MutableObjectDoubleMap expected = ObjectDoubleMaps.mutable.empty();

// Should the expected be this:
expected.put(1, 1);
expected.put(2, 3);
expected.put(3, 4);
// Or should the expected be this:
expected.put(1, 1);
expected.put(2, 1);
expected.put(3, 1);
```
I think it should be the former. But I could be wrong. What do you think @donraab , @motlin ?

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/immutable/ImmutableUnifiedMapTest.java
nikhilnanivadekar: This I presume we will have to fix as a part of major release as it might break existing users who do not have exception handling for duplicates, or can we fix it in a minor release?

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/UnifiedMapNoIteratorTest.java
nikhilnanivadekar: Issue raised: #442

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
motlin: This isn't about the value of the expect result. It's about how many times the function gets called.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/immutable/ImmutableUnifiedMapTest.java
motlin:
```java
    /**
     * Return the MapIterable that is obtained by flipping the direction of this map and making the associations from value to key.
     * @throws IllegalStateException if the MapIterable contains duplicate values.
     * @since 5.0
     */
    MapIterable<V, K> flipUniqueValues();
```

Given that the implementation violates its own contract, I think its fine to include in a bug-fix release if that's how you want to proceed.

It's an old bug that's been there since the first implementation of flipUniqueValues, so it's not urgent either.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
```diff
@@ -0,0 +1,809 @@
...
+    {
+        return this.delegate.hashCode();
+    }
+
+    @Override
+    public MutableOrderedMap<K, V> newEmpty()
+    {
+        return OrderedMapAdapter.adapt(new LinkedHashMap<>());
+    }
+
```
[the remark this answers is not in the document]
motlin: This is an adapter, so there are no public constructors. This comment is on newEmpty(), which is awkward on adapters.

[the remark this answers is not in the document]
motlin: I just looked and MapAdapter does have newEmpty(capacity), but it doesn't come from any interface. Thoughts?

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
nikhilnanivadekar: Ah! If I understand correctly, in the implementation, we would should call the `groupBy` function and `DoubleFunction` just once, but the result of addition has to be compounded over the number of occurrences for each distinct item. Or I am missing something?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
```diff
@@ -0,0 +1,809 @@
...
+    @Override
+    public MutableOrderedMap<K, V> take(int count)
+    {
+        if (count < 0)
+        {
+            throw new IllegalArgumentException("Count must be greater than zero, but was: " + count);
+        }
+
+        MutableOrderedMap<K, V> output = this.newEmpty();
```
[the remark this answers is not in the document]
motlin: This implementation is consistent with TreeSortedMap. That's a type that cannot be pre-sized though. In general the abstract maps aren't as mature as the abstract collections, and can't even be fixed until a major version. I suppose this will be consistent in all transformation methods.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/immutable/ImmutableUnifiedMapTest.java
nikhilnanivadekar: 10.0.0 will open in May/June I would rather wait on the bug fix till then.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
```diff
@@ -0,0 +1,809 @@
...
+    public MutableOrderedMap<K, V> withKeyValue(K key, V value)
+    {
+        this.put(key, value);
+        return this;
+    }
+
+    @Override
+    public MutableOrderedMap<K, V> withAllKeyValues(Iterable<? extends Pair<? extends K, ? extends V>> keyValues)
+    {
+        for (Pair<? extends K, ? extends V> keyVal : keyValues)
```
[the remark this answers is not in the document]
motlin: Could be. Again, just copied the implementation in AbstractMutableSortedMap#withAllKeyValues, but couldn't use an abstract superclass.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/immutable/ImmutableUnifiedMapTest.java
nikhilnanivadekar: Raised #443

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
```diff
@@ -0,0 +1,809 @@
...
+    public MutableOrderedMap<K, V> withoutKey(K key)
+    {
+        this.removeKey(key);
+        return this;
+    }
+
+    @Override
+    public MutableOrderedMap<K, V> withoutAllKeys(Iterable<? extends K> keys)
+    {
+        for (K key : keys)
```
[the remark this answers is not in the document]
motlin: AbstractMutableSortedMap#withoutAllKeys

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MutableMapIterableTestCase.java
gs-rezaem: ```getIfAbsentPutWith``` is quite useful on its own. The typical example is aggregation where the sum object is different from the individual value objects. For plain numbers, that would be MutableInteger vs Integer, etc. For more complex domains, the sum object may have more information, but they almost always need the current value.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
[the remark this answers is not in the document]
nikhilnanivadekar: For minor release let us add it on OrderedMapAdapter, and then eventually move it up to MutableMapIterable in 10.0.0. How does that sound?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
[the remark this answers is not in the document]
nikhilnanivadekar: May be we should then fix it for OrderedMapAdapter in 9.2.0 since it is a new implementation any case. In 10.0.0 seems like we have a ton of work to reduce the tech debt. What do you think?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
```diff
@@ -0,0 +1,809 @@
...
+    @Override
+    public <P> MutableList<V> rejectWith(Predicate2<? super V, ? super P> predicate, P parameter)
+    {
+        return this.reject(Predicates.bind(predicate, parameter));
+    }
+
+    @Override
+    public MutableList<V> select(Predicate<? super V> predicate)
+    {
+        return this.select(predicate, FastList.newList(this.size()));
```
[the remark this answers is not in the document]
nikhilnanivadekar: Let's fix it on OrderedMapAdapter. I agree AbstractMutableSortedMap is out of scope.
We should create issues to track all the tech debt which needs to be removed. If we consider OrderedMapAdapter to be the pristine implementation, then we can bring AbstractMutableSortedMap closer to it.

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
motlin: Yes. Look at AbstractBag. Almost every method delegates to forEachWithOccurrences instead of forEach. That includes sumOf*. sumBy* need similar implementations in AbstractMutableBagIterable or below.

motlin: Looks like the aggregateBy methods have the same problem.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MutableMapIterableTestCase.java
motlin: Not sure if I follow, so I'll add some code examples.

I think the form of `getIfAbsentPut` that takes a `Function` is useful for aggregations.
```java
Counter counter = this.items.getIfAbsentPut(item, Counter::new);
counter.increment();
```

I don't understand how `getIfAbsentPutWith` can be useful because its function takes *neither* the key nor the value.

```
<P> V getIfAbsentPutWith(K key, Function<? super P, ? extends V> function, P parameter);
```

All that said, I looked through every usage of getIfAbsentPutWith in the library, and I found one usage I wrote myself in AbstractMutableMultimap.

```
    private C getIfAbsentPutCollection(K key)
    {
        return this.map.getIfAbsentPutWith(key, this.createCollectionBlock(), this);
    }
```

I'll remove this comment from the code. Thanks for the feedback.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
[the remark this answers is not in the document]
motlin: Since this is an adapter, the usual use-case will probably be wrapping existing LinkedHashMaps, which already have their own capacity. Eventually we'd want to implement our own MutableOrderedMap which isn't an adapter, and then transformation methods would return this type. Just like how ListAdapter.select() return a FastList, not a ListAdapter. That type would have a constructor with a capacity.

The other odd thing about putting it on the interface is that several collections cannot support a capacity, specifically the ones based on trees. That said, it's reasonable to just ignore the capacity.

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
donraab: Sounds like a reasonable performance fix.

nikhilnanivadekar: I think we can only do the fix in major version. Seems like we have a lot of changes for 10.0.0 (October '18). To give us time we can close 9.2.0 in April and open 10.0.0 and start fixing all these issues. Thoughts?

motlin:
I think it's fine to do fixes like these immediately, or in any release.

motlin:
I've addressed all comments, except I haven't created a version of newEmpty() that takes a capacity. I did pre-size target collections wherever possible.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/AbstractMutableMapIterable.java
```diff
@@ -154,12 +154,14 @@ public V updateValue(K key, Function0<? extends V> factory, Function<? super V,
         return LazyIterate.adapt(this.entrySet()).collect(AbstractImmutableEntry.getPairFunction());
     }
 
+    // TODO: push down in next major release. Return type of MutableMap prevents this from being a superclass of MutableOrderedMaps.
```
nikhilnanivadekar: Raised #447

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
nikhilnanivadekar: Raised #448 and #449

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
```diff
@@ -0,0 +1,803 @@
...
+    @Override
+    public String toString()
+    {
+        return this.delegate.toString();
+    }
+
+    @Override
+    public MutableOrderedMap<K, V> clone()
+    {
+        throw new UnsupportedOperationException();
```
nikhilnanivadekar: Won't this be `new MutableOrderedMapAdapter.adapt(this.delegate.clone)` ?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
```diff
@@ -0,0 +1,803 @@
...
+    public MutableOrderedMap<K, V> withoutAllKeys(Iterable<? extends K> keys)
+    {
+        keys.forEach(this::removeKey);
+        return this;
+    }
+
+    @Override
+    public MutableOrderedMap<K, V> asUnmodifiable()
+    {
+        throw new UnsupportedOperationException(this.getClass().getSimpleName() + ".asUnmodifiable() not implemented yet");
```
nikhilnanivadekar: Do we plan to have `UnmodifiableMutableOrderedMap` in the future? I presume it will be after we have our own implementation of `OrderedMap`

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
```diff
@@ -0,0 +1,803 @@
...
+    @Override
+    public <VV> MutableListMultimap<VV, V> groupByEach(Function<? super V, ? extends Iterable<VV>> function)
+    {
+        return this.groupByEach(function, FastListMultimap.newMultimap());
+    }
+
+    @Override
+    public <VV> MutableOrderedMap<VV, V> groupByUniqueKey(Function<? super V, ? extends VV> function)
+    {
+        // TODO: Make this work in a major version by changing {@link org.eclipse.collections.api.RichIterable#groupByUniqueKey}'s type R from MutableMap to MutableMapIterable.
```
nikhilnanivadekar: Raised #450

eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/OrderedMaps.java
```diff
@@ -0,0 +1,34 @@
+
```
nikhilnanivadekar: Extra blank line is failing checkstyle build.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/ordered/OrderedMapAdapterTest.java
```diff
@@ -0,0 +1,64 @@
+
```
nikhilnanivadekar: Extra blank line

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/ordered/MutableOrderedMapTestCase.java
```diff
@@ -0,0 +1,26 @@
+
```
nikhilnanivadekar: Extra blank line

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MutableMapIterableTestCase.java
```diff
@@ -0,0 +1,233 @@
+
```
nikhilnanivadekar: Extra blank line

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/OrderedMapIterableTestCase.java
```diff
@@ -0,0 +1,80 @@
+
```
nikhilnanivadekar: Extra blank line

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
motlin: Sure

PR #456, opened 2018-02-02
Add UnifiedMapWithHashingStrategyNoIteratorTest.

itohiro73:
I find several methods in UnifiedMapWithHashingStrategy could be optimized, so I added UnifiedMapWithHashingStrategyNoIteratorTest in prep for that work. 

I also found a missing null check bug in `index(K key)` and fixed it, too.

### DIFF
--- a/unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/UnifiedMapWithHashingStrategyNoIteratorTest.java
+++ b/unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/UnifiedMapWithHashingStrategyNoIteratorTest.java
@@ -1,0 +1,256 @@
+/*
+ * Copyright (c) 2018 Hiroshi Ito.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompany this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.test.map.mutable;
+
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Random;
+import java.util.Set;
+
+import org.eclipse.collections.api.map.MutableMap;
+import org.eclipse.collections.impl.block.factory.HashingStrategies;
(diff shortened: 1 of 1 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
```diff
@@ -361,7 +361,7 @@ protected final int index(K key)
         // This function ensures that hashCodes that differ only by
         // constant multiples at each bit position have a bounded
         // number of collisions (approximately 8 at default load factor).
-        int h = this.hashingStrategy.computeHashCode(key);
+        int h = key == null ? 0 : this.hashingStrategy.computeHashCode(key);
```
nikhilnanivadekar: I thought the null check was purposely skipped and delegated to users to handle the hashCode for null keys. But this predates my involvement in the project, so I could be mistaken.

itohiro73: That sounds to be contradicting with UnifiedMap#index(key) implementation. Hopefully @donraab or @motlin could clarify?

```
    protected final int index(Object key)
    {
        // This function ensures that hashCodes that differ only by
        // constant multiples at each bit position have a bounded
        // number of collisions (approximately 8 at default load factor).
        int h = key == null ? 0 : key.hashCode();
        h ^= h >>> 20 ^ h >>> 12;
        h ^= h >>> 7 ^ h >>> 4;
        return (h & (this.table.length >> 1) - 1) << 1;
    }
```

nikhilnanivadekar: it will throw a NPE for the hashCode() call. So, it is necessary to handle that.

itohiro73: Ah now I understand what you mean. The decision should be made in each hashingStrategy I guess?

nikhilnanivadekar: precisely

itohiro73:
I have reverted the null check in prod code, and used a null safe hashing strategy in test code instead.

motlin:
👍 

Agree with @nikhilnanivadekar that we pass null right into the `HashingStrategy` and the collection is null-safe if the strategy is null-safe.

LGTM!

PR #460, opened 2018-02-06
Make primitive classes non-final. Closes #459.

### LINKED ISSUE #459: Make primitive classes non-final.
This is similar to #453. I find number of primitive impl stg templates are defined to generate final classes.

image

I'm not sure if there was a particular reason for that. If no valid reason, I'd suggest to remove `final` as similar to issue #453. 

Thoughts @donraab @nikhilnanivadekar @motlin @gs-rezaem ?

itohiro73:
Refer to issue #459.

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/impl/bag/mutable/primitiveHashBag.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/bag/mutable/primitiveHashBag.stg
@@ -56,7 +56,7 @@ import org.eclipse.collections.impl.tuple.primitive.PrimitiveTuples;
  *
  * @since 3.0.
  */
-public final class <name>HashBag extends Abstract<name>Iterable implements Mutable<name>Bag, Externalizable
+public class <name>HashBag extends Abstract<name>Iterable implements Mutable<name>Bag, Externalizable
 {
     private static final long serialVersionUID = 1L;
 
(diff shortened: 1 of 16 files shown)


nikhilnanivadekar:
can you please rebase? This is fine by me unless @motlin has additional comments.

itohiro73:
Rebased.

Actually you can now leverage `Rebase and merge` functionality in GitHub. 

Now that we have this functionality in GitHub, it's not very recommended to ask contributors to rebase when the commit was originally properly made on top of master and nothing else has to be changed. 

image

nikhilnanivadekar:
@itohiro73 I am aware of that option. However, if we use it, it makes the graph look weird. 
For example: this is the rebase, merge commit which was done recently. The commit is shown as if it was directly on the master branch (similar to the way Hudson commits) vs a merge commit. 
image

motlin:
Saying the same thing another way... if you use this functionality, there's
no merge commit so it's not obvious who clicked the button and did the
rebase.

itohiro73:
It actually is though. You can see who did the rebase in the commit history. 

image

I always don't like the fact we ask contributors to rebase even when they made proper commit to begin with. That makes contributors feel annoyed. I'd hope we should make the EC contribution environment more welcoming and encouraging, and this kind of rebase request **after they sent proper PR** is the one that I think we should consider improving. 

Anyways I rebased again, hope there won't be any further PRs or commits goes between if you don't like to use the functionality :P

PR #461, opened 2018-02-07

motlin:


unit-tests/src/test/java/org/eclipse/collections/impl/multimap/bag/sorted/mutable/AbstractMutableSortedBagMultimapTestCase.java
```diff
@@ -110,7 +110,7 @@ public void serialization()
     }
 
     @Override
-    public void testClear()
```
itohiro73: Should there be @Test annotation?

itohiro73:
Very nice! Thanks for taking care of these @motlin .

There seem to be one issue in dbd14a3. Please have a look.

motlin:
@nikhilnanivadekar I don't see any context with your comments so I don't know which inspections they are about.

motlin:
Yes, though the point of this PR is really to “fix the broken builds.” The
version of CheckStyle we were on must never fail and is just buggy. And
we’re not really running an inspections build anymore but if we were it
would be failing. There is no inspection for missing @Test annotations and
no great way to find them all.

nikhilnanivadekar:
Updated the context inline with the comments.

motlin:
@itohiro73 One more bit of context. Although I did add a few annotations manually in dbd14a3, I think there are hundreds or thousands more places missing the annotations and no automated way to change them.

@nikhilnanivadekar In some recent MRs, I've been asking "why didn't the CheckStyle build fail?" I suspect that the version we were on was buggy and was never failing no matter what. Merely upgrading caused a bunch of whitespace errors to get correctly flagged. I didn't attempt to turn on any newly implemented rules, just get the build correctly failing, and then passing again.

@nikhilnanivadekar The inspection "Data flow issues | Scope of variable is too broad" is not critical, but it sometimes flags code where we allocate an object and then return or throw so that it immediately gets garbage collected. I remember this one as a warning, not something that should break the build. I'm fine dropping that commit if you'd prefer.

nikhilnanivadekar:
I am fine with the commit, was wondering the reason behind it.
Glad that the upgrade caught the whitespace issues!
This is ok to merge by me, will wait on @itohiro73 if he has more comment.

itohiro73:
No further comment from me. LGTM! Let me merge.

PR #471, opened 2018-02-28
Fix toImmutable and asUnmodifiable implementations in synchronized pr…

donraab:
…imitive collections and maps.


motlin:
Are these methods covered by synchronized tests? These changes don't hurt of course, but I'm curious if they make methods safe which used to be dangerous.

I think toImmutable() is already synchronized in practice because it probably winds up delegating to forEach(). I don't think it matters if asUnmodifiable() is synchronized because it merely constructs a wrapper and doesn't really touch the underlying collection.

donraab:
I looked and I don't think we have a suite of synchronized traits for primitive collections as we do for object collections.  This is definitely something I think we should address.  Do you think we can do this in Java 8 instead of Scala this time?

I opted for the safest approach for both asUnmodifiable and toImmutable which is just to take the lock.

motlin:
I'm sure Java 8 is fine. Makes sense.

PR #479, opened 2018-03-14
Tokenize Java-10-EA, Java-11-EA URLs.

nikhilnanivadekar:


.travis-command-ea-builds.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -euo pipefail
+set +u
```
motlin: I don't know too much about bash, but would you want to keep the `set -e` part?

nikhilnanivadekar: Fixed.

PR #484, opened 2018-03-19
Fix PrimitiveHashSet#iterator()#remove() to not rehash. Fixes #481

### LINKED ISSUE #481: Primitive iterator looses sync after too many elements are removed
Consider this code:

```.java
MutableIntSet s = IntSets.mutable.of();

int max = 100000;

for(int i = 0; i < max; i++) {
    s.add(i);
}

MutableIntIterator iterator = s.intIterator();

while (iterator.hasNext()) {
    // remove 2/3 of elements
    if (iterator.next() < ( max * 2 / 3)) {
        iterator.remove();
    }
}
```
this will cause this exception:
```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 262144
	at org.eclipse.collections.impl.set.mutable.primitive.IntHashSet$InternalIntIterator.next(IntHashSet.java:1624)
```
It works for smaller numbers of elements like max = 1k or 10k but starts to fail for 100k and more.

And I found the cause: 
```InternalIntIterator.remove()``` calls the standard methods ```IntHashSet.remove();``` which does rehashing when there is too many entries occupied with sentinels.

```.java
...
if (this.occupiedWithSentinels > this.maxOccupiedWithSentinels()) {
    this.rehash();
}
...
```
and this makes iterator invalid.

nikhilnanivadekar:


nikhilnanivadekar:
I will create service releases for 7.x and 8.x. The service releases can be pushed anytime, so we can push them sooner than April 18.

eclipse-collections-code-generator/src/main/resources/impl/set/mutable/primitiveHashSet.stg
```diff
@@ -1610,7 +1615,7 @@ public class <name>HashSet extends Abstract<name>Set implements Mutable<name>Set
             {
                 removeValue = <name>HashSet.this.table[this.position - 1];
             }
-            <name>HashSet.this.remove(removeValue);
+            <name>HashSet.this.remove(removeValue, false);
```
motlin: While this is good enough to fix the bug, I think there's a better implementation that we should add to the backlog.

We should basically inline the call to remove here, removing the call to rehash, and directly mutating the backing table. When we delegate to remove(), we'll be calling probe() again. We already know the position in the table, because we're iterating through positions and have the position field.

This would also be more consistent with UnifiedSet.

This MR doesn't seem to address the same bug in primitive maps, where they delegate to removeKey and rehash.

gs-rezaem: I don't see a rehash in the maps. I checked 7.x, 8.x and 9.x

motlin:
Looking now, maps never seem to rehash during removal. Perhaps that was my
source of confusion. So no bug, which is good, though surprisingly
inconsistent with sets.

nikhilnanivadekar:
Last night I had verified all Maps, Sets and Bags. Only primitiveSets has this issue.
Looking at the history it was added at a part of GS Collections [GSCOLLECT-1086]
https://github.com/goldmansachs/gs-collections/commit/590786fbc1a0118bbad89d52de747a5b045ff438#diff-e9eedad3e71045072ffec0aac02aa241R368

PR #485, opened 2018-03-20
Optimize PrimitiveHashSet#iterator#remove().

nikhilnanivadekar:


eclipse-collections-code-generator/src/main/resources/impl/set/mutable/primitiveHashSet.stg
```diff
@@ -1615,7 +1615,21 @@ public class <name>HashSet extends Abstract<name>Set implements Mutable<name>Set
...
-            <name>HashSet.this.remove(removeValue, false);
+            if (<name>HashSet.isBetweenZeroAndThirtyOne(removeValue))
+            {
+                <name>HashSet.this.removeZeroToThirtyOne(removeValue);
+            }
+            else if (<(equals.(type))({<name>HashSet.this.table[this.position - 1]}, "removeValue")>)
+            {
+                if (<name>HashSet.this.copyOnWrite)
+                {
+                    <name>HashSet.this.copyTable();
```
nikhilnanivadekar: @motlin @gs-rezaem this was copied from the `remove()` implementation, but I do not think this if statement is necessary. `this.copyOnWrite` will always be false inside the iterator. I tried writing a test case which will get it to be true but did not find a condition. May be I am missing something. Happy to remove the if statement or add the test case which will enter the branch.

nikhilnanivadekar:
Once this is merged I will back-port it to 7.x, 8.x and [GS Collections](https://github.com/goldmansachs/gs-collections/issues/35)

nikhilnanivadekar:
Added another commit which adds test coverage. Can you please review?

Issue #491, opened 2018-03-26
Primitive Hash Maps addToValue returns wrong value after rehash and grow


bryce-pilcher:
`addToValue` returns the wrong value if the data structure is resized as a result of the add.

It appears as though the index in `addToValue` is stale after a rehash and grow operation.  And this is causing the value of the key to be misrepresented.  In my case, it is simply returning a 1 or zero, but it could return a high value for a key that shouldn't have a high value, or vice-versa.  

I have put together three junit tests:
1. `testAddToValueObj()` tests adding a list of strings to an ObjectLongHashMap and looking at the return values.
2.  `testAddToValueLong()` tests adding a list of longs (with the same value as the strings above) to a LongLongHashMap, and looks at the return values.
3. `testAddObj()`  adds a string to the map, then gets the value of that to check that it is 1.  This test passes, because it is reading the correct index of the last string `9986389012`, whereas the previous two are using the index from before the grow operation.  

[Gist](https://gist.github.com/bryce-pilcher/90ddaee905f394a3ead77b3e54658533)

PR #492, opened 2018-03-27
Fix for #491 - if map was grown as a result of addToValue, recompute …

jdimeo:
…index or else it will be stale and return value will be aribitrary

Signed-off-by: John Dimeo <dimeo@elderresearch.com>

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/objectPrimitiveHashMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/objectPrimitiveHashMap.stg
@@ -558,7 +558,7 @@ public class Object<name>HashMap\<K> implements MutableObject<name>Map\<K>, Exte
             return this.values[index];
         }
         this.addKeyValueAtIndex(key, toBeAdded, index);
-        return this.values[index];
+        return toBeAdded;
     }
 
     @Override
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/objectPrimitiveHashMapWithHashingStrategy.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/objectPrimitiveHashMapWithHashingStrategy.stg
@@ -581,7 +581,7 @@ public class Object<name>HashMapWithHashingStrategy\<K> implements MutableObject
             return this.values[index];
         }
         this.addKeyValueAtIndex(key, toBeAdded, index);
-        return this.values[index];
+        return toBeAdded;
     }
 
     @Override
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitivePrimitiveHashMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitivePrimitiveHashMap.stg
@@ -532,7 +532,7 @@ public class <name1><name2>HashMap extends AbstractMutable<name2>ValuesMap imple
             return this.<valueArray>[index<valueIndex>];
         }
         this.addKeyValueAtIndex(key, toBeAdded, index);
-        return this.<valueArray>[index<valueIndex>];
+        return toBeAdded;
     }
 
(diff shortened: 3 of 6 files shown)


gs-rezaem:
Sorry, this has a much simpler fix. Just "return value;".
:-1:

jdimeo:
OK so what's the next step? new PR? change this PR? you guys do it instead of using PR? it's a pretty high priority fix, yes?

nikhilnanivadekar:
Simply drop this commit and make the fix and push to the same branch. It'll update the PR and we will remove.

jdimeo:
Will do that hopefully tonight.

gs-rezaem:
Please add a test case as well and fix any checkstyle issues (if any).

jdimeo:
@gs-rezaem I'm having a hard time locating in which submodule and which test the right place to add the test would be...

nikhilnanivadekar:
@jdimeo please push what you have with the fix, I can cherry-pick the commit and help out from there. Similar to what I did here: https://github.com/nikhilnanivadekar/eclipse-collections/commit/6f9ea99f701d380127e87c3d3ab3f1c3b584cfd7

jdimeo:
Well, I finally found where to add the test, but when I converted the `long` test values to `int`s to conform to how the existing `addToValue` test was set up, it no longer demonstrated the bug.

I've attached a patch with how far I got adding the test (which unfortunately passes before the fix was implemented, so the correct values will have to be found).
[patch.diff.txt](https://github.com/eclipse/eclipse-collections/files/1854286/patch.diff.txt)

jdimeo:
Oh dear, now I seem to have broken the build but I'm not sure how. All unit tests passed but the exception was thrown at the end of the acceptance test suite:
`Caused by: java.lang.RuntimeException: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?`

nikhilnanivadekar:
@jdimeo that is fine. It happens for acceptance suite since it is a long running build. I have re-triggered it.
I am checking the test cases.

nikhilnanivadekar:
In the test case, you need to pre-size the map so that insert in Map will trigger a rehash.
Here are failing tests:

```
@Test
public void addToValue()
{
    ObjectIntHashMap<String> emptyMap = this.getEmptyMap();
    IntInterval interval = IntInterval.zeroTo(10);
    interval.forEachWithIndex(new IntIntProcedure() {
        @Override
        public void value(int each, int index)
        {
            Assert.assertEquals("Key:" + each, each + index, emptyMap.addToValue(String.valueOf(each), each + index));
        }
    });
}
```
```
@Test
public void addToValue()
{
    MutableLongLongMap emptyMap = this.getEmptyMap();
    MutableLongList longList = LongLists.mutable.with(936628237L, 4889384619L, 8733915902L, 2377747912L,
            277382636L, 593670575L, 296725141L, 7131901003L, 9986389012L);
    longList.forEachWithIndex((each, index) ->
            Assert.assertEquals("Key:" + each, each + index, emptyMap.addToValue(each, each + index)));
}
```

nikhilnanivadekar:
@jdimeo was this helpful?

jdimeo:
@nikhilnanivadekar Yes, I think so. Sorry, I've been really slammed this week and haven't had a chance to revisit this. Are you waiting for me to add a test to this PR? I will do that at my next opportunity.

nikhilnanivadekar:
I understand. I am will wait to merge, so that I can back port it and create service releases.
Thanks for your help!

jdimeo:
OK, I added tests to the unit test templates that I confirmed failed prior to my fix and pass after it. 
@nikhilnanivadekar on 3. above, doesn't it not apply to PrimitiveObjectHashMap since that class doesn't have `addToValue`?

jdimeo:
I think I have done that now. I saw a button to do it when I first requested the merge, but now I had to do it with git commands and then force push. Please let me know if I didn't do it correctly.

nikhilnanivadekar:
@jdimeo Thanks for your contribution!

Issue #467, opened 2018-02-20
Implement variants of groupBy that group at multiple levels.


motlin:
`Collectors.groupingBy()` has an overload which accepts another Collector, allowing multi-level groupBys.

[This guide](http://www.baeldung.com/java-groupingby-collector) includes this example:

```java
Map<String, Map<BlogPostType, List>> map = posts.stream()
  .collect(groupingBy(BlogPost::getAuthor, groupingBy(BlogPost::getType)));
```

RichIterable could have additional variants of groupBy that take 2 or 3 Functions and return a Map of Multimaps or a Map of Maps of Multimaps respectively.

Alternately, we could have a form of groupBy that takes a single key Function, and a second Function that transforms the resulting grouped RichIterables into anything, and returns a MapIterable.

PR #494, opened 2018-04-01
Implement #467 groupBy at multi levels

canthonyl:
Hi, please find implementation for #467 groupBy at multiple levels for your review, where the return types are Map(s) of Multimaps.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
@@ -13,8 +13,10 @@ package org.eclipse.collections.api;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.DoubleSummaryStatistics;
+import java.util.HashMap;
 import java.util.IntSummaryStatistics;
 import java.util.LongSummaryStatistics;
+import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.function.BiConsumer;
@@ -1997,6 +1999,69 @@ public interface RichIterable<T>
      * @since 1.0
      */
     <V> Multimap<V, T> groupBy(Function<? super T, ? extends V> function);
+
+    /**
+     * For each element of the iterable, the functions are evaluated and the results of these valuations are collected
+     * into a new map of Multimap, where the first and second functions are evaluated to produce the keys to the map and the inner
+     * Multimap respectively, and the original values are added to the same (or similar) species of collection as the
+     * source iterable.
+     * <p>
+     * Example using a Java 8 method reference:
+     * <pre>
+     * Map&lt;String, Multimap&lt;String, Person&gt;&gt; peopleByCityAndLastName =
+     *     people.groupBy(Person::getCity, Person::getLastName);
+     * </pre>
+     * <p>
+     * Example using an anonymous inner class:
+     * <pre>
+     * Map&lt;String, Multimap&lt;String, Person&gt;&gt; peopleByCityAndLastName =
+     *     people.groupBy(new Function&lt;Person, String&gt;()
+     *     {
+     *         public String valueOf(Person person)
+     *         {
+     *             return person.getCity();
+     *         }
+     *     },
+     *     new Function&lt;Person, String&gt;()
+     *     {
+     *         public String valueOf(Person person)
+     *         {
+     *             return person.getLastName();
+     *         }
+     *     }
+     *     );
+     * </pre>
+     *
+     * @since 9.x
+     */
+    default <V, W> Map<V, Multimap<W, T>> groupBy2(Function<? super T, ? extends V> function1, Function<? super T, ? extends W> function2)
+    {
+        Map<V, Multimap<W, T>> result = new HashMap<>();
+        Multimap<V, T> map1 = this.groupBy(function1);
+        map1.forEachKey(v -> {
+            Multimap<W, T> multimap2 = map1.get(v).groupBy(function2);
+            result.put(v, multimap2);
+        });
+        return result;
+    }
+
+    /**
+     * Similar to {@link #groupBy2(Function, Function)}, except the third function evaluate the key to the innermost Multimap.
+     *
+     * @since 9.x
+     */
+    default <V, W, X> Map<V, Map<W, Multimap<X, T>>> groupBy3(Function<? super T, ? extends V> function1, Function<? super T, ? extends W> function2, Function<? super T, ? extends X> function3)
+    {
+        Map<V, Map<W, Multimap<X, T>>> result = new HashMap<>();
+        Map<V, Multimap<W, T>> map1 = this.groupBy2(function1, function2);
+        map1.forEach((v, map2) -> {
+            map2.forEachKey(w -> {
+                Multimap<X, T> multimap3 = map2.get(w).groupBy(function3);
+                result.computeIfAbsent(v, key -> new HashMap<>()).put(w, multimap3);
+            });
+        });
+        return result;
+    }
 
     /**
      * This method will count the number of occurrences of each value calculated by applying the
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
@@ -13,10 +13,8 @@ package org.eclipse.collections.api;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.DoubleSummaryStatistics;
-import java.util.HashMap;
 import java.util.IntSummaryStatistics;
 import java.util.LongSummaryStatistics;
-import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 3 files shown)


nikhilnanivadekar:
A very initial feedback (I didn't get a chance to look at the entire PR). We would want to return a `MutableMap` instead of just a Map. Hence we won't use a HashMap but a UnifiedMap.
Looping in @motlin

canthonyl:
I have made some changes based on your comment (remove the hashmap and revised the implementation a bit).

canthonyl:
@nikhilnanivadekar The build is failing for Java 11 regression, shall I try close and reopen the PR?

nikhilnanivadekar:
you can ignore the Java 11 build for now. I will take a look at it later today. It would likely be updating the download url.

nikhilnanivadekar:
I am looping in others to decide on the name of the API. `groupBy2`, `groupBy3` might lead to confusion by the users. Feel free to suggest names as well.
@donraab @motlin @itohiro73 @gs-rezaem What do you think will be good names for this API.
@motlin can you please review the implementation as well?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -1998,6 +1998,55 @@ default void appendString(Appendable appendable, String separator)
...
+     * <p>
+     * Example using a Java 8 method reference:
+     * <pre>
+     * Map&lt;String, Multimap&lt;String, Person&gt;&gt; peopleByCityAndLastName =
+     *     people.groupBy(Person::getCity, Person::getLastName);
+     * </pre>
+     * <p>
+     * Example using an anonymous inner class:
+     * <pre>
+     * Map&lt;String, Multimap&lt;String, Person&gt;&gt; peopleByCityAndLastName =
```
motlin: The examples in Javadoc still refer to Map instead of MutableMap or MapIterable

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -1998,6 +1998,55 @@ default void appendString(Appendable appendable, String separator)
...
+     *         {
+     *             return person.getLastName();
+     *         }
+     *     }
+     *     );
+     * </pre>
+     *
+     * @since 9.x
+     */
+    default <V, W> MutableMap<V, Multimap<W, T>> groupBy2(Function<? super T, ? extends V> function1, Function<? super T, ? extends W> function2)
```
motlin: Since we're returning a type with nested generics, we lose the opportunity to use type variance, so no MutableMultimap on mutable types and ImmutableMultimap on immutable types. This is similar to chunk(). At runtime, they should probably really be mutable or immutable as appropriate. The outer type can still vary normally. So MapIterable here, MutableMap on MutableCollection, ImmutableMap on ImmutableCollection, etc. The overrides should go in the same places where groupBy is already overridden.

motlin: We've used the type parameters `T` and `V` since before I've been involved in the project. I think `V` must be for `Value` or something. Instead of adding `W`, I'd recommend numbered types like `V1` and `V2` or something similar.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -1998,6 +1998,55 @@ default void appendString(Appendable appendable, String separator)
...
+    {
+        return this.<V>groupBy(function1).toMap().collectValues((v, t) -> t.groupBy(function2));
```
motlin: Maybe the default implementation should just throw?

nikhilnanivadekar:
The Java 11 build is fixed

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
canthonyl: Happy to make the changes- although my original thought was the implementing classes of RichIterable can eseentially get them "for free" (implement groupBy and get the additional levels of grouping).  Would that be of a good reason to keep it?

motlin: Let me give a little more context about this comment.

I think that the covariant return types of this method should closely follow groupByUniqueKey. I previously mentioned groupBy, but groupByUniqueKey is even closer since it returns a map. With that in mind, RichIterable.groupBy2 should return MapIterable (not MutableMap), MutableCollection.groupBy2 should return MutableMap, and ImmutableCollection.groupBy2 should return ImmutableMap.

This implementation is only appropriate for some RichIterables, specifically the mutable ones, so it should move to MutableCollection. With that in mind, I think it makes sense for the method to remain abstract on RichIterable, or if we want to include this change in a minor release, then it should throw.

This implementation delegates to methods that create new collections, and thus it creates extra garbage and rehashes each T. I think we could follow an approach like groupByUniqueKey (GroupByUniqueKeyProcedure) for efficiency.

The implementation is also not overridden for Bags - we usually put in an optimization so that we call hashCode just once for each T instead of each occurrence of T, and test it in RichIterableTestCase.RichIterable_iterationOrder.

canthonyl: Thanks for the clarification and the pointers for the implementation

canthonyl: For GroupBy2Procedure, in addition to the Mutable target map, should it also take in a supplier to create suitable type of Multimap?

nikhilnanivadekar: Sorry I don't quite follow this. What will the signature be?

canthonyl: Looking back at my own comment, I just realized I should have included more details (apologies!).  I have created a GroupBy2Procedure class under package org.eclipse.collections.impl.block.procedure, for use in AbstractRichIterable (similar to how GroupByUniqueKeyProcedure is used for groupByUniqueKey).  

In GroupBy2Procedure the overridden value method, it also instantiate a new Multimap as it encounters a new first level key:

    @Override
    public void value(T object)
    {
        K1 key1 = this.keyFunction1.valueOf(object);
        K2 key2 = this.keyFunction2.valueOf(object);
        this.map.computeIfAbsent(key1, k -> /* instantiate a multimap */ ).put(key2, object);
    }

I just wonder if the procedure can use a FastListMultimap for all cases, or if it has to also vary depending on the original data structure (e.g. similar to how MutableSet's groupBy returns MutableSetMultimap , MutableSet's multi-level groupBys also return MutableSetMultimap under the hood). 

Let me find a good cut-off point (I'm working on the List-based structures) and push the code as it may be easier for code review.

nikhilnanivadekar: According to me, it should vary depending on covariant overrides. ListMultimap for List, BagMultimap for Bag, SetMultimap for Set.

motlin: Agree, the runtime types should vary based on the original collection. So set.groupBy2() should return a Map of SetMultimaps. Unfortunately, Java won't allow us to similarly vary the compile time types since generics are invariant.

donraab: Just a thought.  Could we not take the same approach as we did for partition and introduce a new type?  With a method named groupBy2 and groupBy3, you could return a custom structure which hides the underlying map details and makes the generics flat.

It's more work, and we wouldn't want to to make the structure Mutable or Immutable except for perhaps at the bottom Multimap layer.

I haven't thought much yet about the implications of this... but there might be some benefits having an api that perhaps has get(leveOneKey, levelTwoKey) instead of get(levelOneKey).get(levelTwoKey).  It will be safer and more likely to preserve the structure than exposing a Map at each level, especially if the Map is sparse and Mutable where you can just then null out values arbitrality.

One of the benefits I always teach people about Multimap is that it can deal with sparseness and doesn't let you just put null in for a value for a list.  It preserves its shape, and hides lack of a key showing up as a null.

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableUniqueTestCase.java
```diff
@@ -678,6 +679,43 @@ default void RichIterable_groupBy()
...
+                UnifiedMap.newWithKeysValues(
+                        Boolean.TRUE, UnifiedMap.newWithKeysValues(
+                            Boolean.TRUE, this.newWith(2, 1).groupBy(isOdd),
+                            Boolean.FALSE, this.newWith(4, 3).groupBy(isOdd)),
+                        Boolean.FALSE,  UnifiedMap.newWithKeysValues(
+                            Boolean.TRUE, this.newWith(-2, -1).groupBy(isOdd),
+                            Boolean.FALSE, this.newWith(-4, -3).groupBy(isOdd)));
+
+        assertEquals(expectedGroupBy, iterable.groupBy3(isPositive, inRange, isOdd));
+    }
```
donraab: This should work for a 3 level groupBy:

`iterable.reduceInPlace(Collectors.groupingBy(inRange, Collectors2.toListMultimap(isOdd)))`

The reason I point this out is that it might be useful for testing.  The underlying call in toListMultimap calls a groupBy call with a Supplier.  It is however harder to make utility calls like this covariant as you have with your new groupBy3 implementation.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
nikhilnanivadekar: @motlin I was not suggesting that in the method signature, but more so at runtime, if a user casts to a the appropriate type it should not throw a ClassCastException. It will be similar to `chunk` wherein the outer most `RichIterable` is always a `MutableList` but the inner batches are of the appropriate type and casts at runtime work fine.

I like @donraab 's idea but I am not too sure how it will work with the overload which accepts a target multimap. It makes sense for `partition()` to not have the flexibility of providing a target collection. However, for groupBy2, groupBy3, we will need a target variant to give the flexibility of controlling the value Multimap.

motlin:
I think @donraab’s idea is worth pursuing. @nikhilnanivadekar I understand
the concern about overrides that accept targets. I think it will be
difficult either way, with a special type or not.

nikhilnanivadekar:
resurrecting this thread.
@canthonyl do you need any help/direction from us? 
Thanks for your contribution!

canthonyl:
Thanks all for your comments, let me try the approache suggested above

nikhilnanivadekar:
closing due to inactivity. Please reopen if desired.

PR #497, opened 2018-04-02
Fix concurrent map iterator with size close to zero

gs-rezaem:
Make removeIf/removeAll a bit more predictable

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMapUnsafe.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMapUnsafe.java
@@ -26,11 +26,13 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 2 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMapUnsafe.java
```diff
@@ -909,11 +911,12 @@ public V put(K key, V value)
         if (o == null)
         {
             Entry<K, V> newEntry = new Entry<>(key, value, null);
+            this.addToSize(1);
             if (ConcurrentHashMapUnsafe.casArrayAt(currentArray, index, null, newEntry))
             {
-                this.addToSize(1);
                 return null;
             }
+            this.addToSize(-1);
```
motlin: This means if you insert one key, then insert a second colliding key, you could observe for a moment a size of 2?

gs-rezaem: Correct. That was preferable to seeing an empty map (size 0) in the same thread that's doing a simple put.

nikhilnanivadekar:
Merging this. Please remove concitrfix branch post the merge.

PR #498, opened 2018-04-04
Implement countByEach on RichIterable. Resolves #476

### LINKED ISSUE #476: Implement countByEach on RichIterable.
This can be implemented as a default method on RichIterable as countBy was.  It will need appropriate co-variant overrides as default methods and optimizations for Bag implementations.  

The signature of countByEach should be:

`default <V> Bag<V> countByEach(Function<? super T, ? extends Iterable<V>> function)`

The naming is based on the pattern we used for groupBy/groupByEach.

This code in the Pet Kata can then be improved.

Before:

```
MutableList<Person> people = ...;
MutableList<PetType> petTypes = people.flatCollect(Person::getPetTypes);
Bag<PetType> petTypeCounts = petTypes.toBag();
int cats = petTypeCounts.occurrencesOf(PetType.CAT);
```

After:

```
MutableList<Person> people = ...;
Bag<PetType> petTypeCounts = people.countByEach(Person::getPetTypes);
int cats = petTypeCounts.occurrencesOf(PetType.CAT);
```
http://eclipse.github.io/eclipse-collections-kata/pet-kata/#/6/2

TheLoneKing:

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
@@ -2088,6 +2088,28 @@ public interface RichIterable<T>
     }
 
     /**
+     * This method will count the number of occurrences of each value calculated by applying the
+     * function to each element of the collection.
+     *
+     * @since 10.0.0
+     */
+    default <V> Bag<V> countByEach(Function<? super T, ? extends Iterable<V>> function)
+    {
+        return this.asLazy().flatCollect(function).toBag();
+    }
+
+    /**
+     * This method will count the number of occurrences of each value calculated by applying the
+     * function to each element of the collection.
+     *
+     * @since 10.0.0
+     */
+    default <V, R extends MutableBagIterable<V>> R countByEach(Function<? super T, ? extends Iterable<V>> function, R target)
+    {
+        return this.flatCollect(function, target);
+    }
+
+    /**
      * Same as {@link #groupBy(Function)}, except that the results are gathered into the specified {@code target}
      * multimap.
      * <p>
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/stack/ImmutableStack.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/stack/ImmutableStack.java
@@ -162,6 +162,15 @@ public interface ImmutableStack<T> extends StackIterable<T>
         return this.asLazy().<P, V>collectWith(function, parameter).toBag().toImmutable();
     }
 
+    /**
+     * @since 10.0.0
+     */
+    @Override
+    default <V> ImmutableBag<V> countByEach(Function<? super T, ? extends Iterable<V>> function)
+    {
+        return this.asLazy().flatCollect(function).toBag().toImmutable();
+    }
+
     @Override
     <V> ImmutableListMultimap<V, T> groupByEach(Function<? super T, ? extends Iterable<V>> function);
 
--- a/unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
+++ b/unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
@@ -83,6 +83,7 @@ import org.eclipse.collections.impl.factory.SortedSets;
 import org.eclipse.collections.impl.factory.primitive.ObjectDoubleMaps;
 import org.eclipse.collections.impl.factory.primitive.ObjectLongMaps;
 import org.eclipse.collections.impl.list.Interval;
+import org.eclipse.collections.impl.list.primitive.IntInterval;
 import org.eclipse.collections.impl.map.mutable.UnifiedMap;
 import org.eclipse.collections.impl.map.sorted.mutable.TreeSortedMap;
 import org.eclipse.collections.impl.multimap.bag.HashBagMultimap;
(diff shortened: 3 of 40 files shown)


nikhilnanivadekar:
Since 10.0.0 is a major release, I think we can do away with all the default implementations. Having default implementations as well as concrete implementations leaves the default implementations untested, susceptible to bugs and possible inefficiencies. 
Looping in others to understand their view @donraab @motlin

motlin:
Depends what we did in the minor release. Did we avoid altering the normal
abstract classes, like AbstractMutableCollection? If so, we could move
default methods down to abstract classes.

nikhilnanivadekar:
Yup, I would like to move them down and remove the default methods.

nikhilnanivadekar:
on second thoughts, we can actually keep the default methods for now and remove them sometime later during dev work for 10.0.0

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -2058,6 +2058,26 @@ default void appendString(Appendable appendable, String separator)
         return this.collectWith(function, parameter, target);
     }
 
+    /**
+     * This method will count the number of occurrences of each value calculated by applying the
+     * function to each element of the collection.
+     *
+     */
+    default <V> Bag<V> countByEach(Function<? super T, ? extends Iterable<V>> function)
```
nikhilnanivadekar: needs an @ since 10.0.0 tag
Please add the since tag for each interface level override.

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
```diff
@@ -1633,6 +1634,18 @@ default void RichIterable_countByWith()
         Assert.assertEquals(3, evensAndOdds2.occurrencesOf(0));
     }
 
+    @Test
+    default void RichIterable_countByEach()
```
nikhilnanivadekar: How is the coverage?
Do we have all implementations covered?

nikhilnanivadekar:
For the since tags, they should be updated to be 10.0, to be consistent with rest of the Javadocs.

For coverage, you have a few options, 
1) if you use IntelliJ, you can run the tests in coverage mode: https://www.jetbrains.com/help/idea/code-coverage.html
2) if you use Eclipse, you use Eclipse, you can analyze using EclEmma: https://www.eclemma.org/
3) Lastly, our unit-test build runs JaCoCo coverage, so if you run the maven build with code coverage and the coverage analysis will be generated in the test-coverage module in the project.

I would suggest either 1 or 2, since it is easier to navigate to the implementations once the tests are run, versus going through the entire coverage report.

It might be difficult to add coverage for default methods. So, I will not stress that too much, but each implementation should have coverage.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -2059,6 +2059,28 @@ default void appendString(Appendable appendable, String separator)
...
+
+    /**
+     * This method will count the number of occurrences of each value calculated by applying the
+     * function to each element of the collection.
+     *
+     * @since 10.0.0
+     */
+    default <V, R extends MutableBagIterable<V>> R countByEach(Function<? super T, ? extends Iterable<V>> function, R target)
+    {
+        return this.<V, R>flatCollect(function, target);
```
motlin: This overload `countByEach(function, target)`, just delegates to `flatCollect(function, target)`. I don't think we should add the method because it doesn't do any additional work and it's not a more intuitive alias than flatCollect, in my opinion. Thoughts?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/stack/ImmutableStack.java
```diff
@@ -162,6 +162,15 @@
         return this.asLazy().<P, V>collectWith(function, parameter).toBag().toImmutable();
     }
 
+    /**
+     * @since 10.0.0
+     */
+    @Override
+    default <V> ImmutableBag<V> countByEach(Function<? super T, ? extends Iterable<V>> function)
+    {
+        return this.asLazy().<V>flatCollect(function).toBag().toImmutable();
```
motlin: I think all the usages of generic types specified on methods are redundant. See 8e14f7f60.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
nikhilnanivadekar: The target may add symmetry, I don't quite have a strong opinion in either direction.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/stack/ImmutableStack.java
nikhilnanivadekar: Does the static analysis actually show a warning and IntelliJ does not show a compiler error? It should be a compiler error if just <V> is getting inferred.
flatCollect() has expanded scopes, so it will return of type <? extends V> and toBag() and toImmutable() will cascade to <? extends V>, we have to force the scope to <V>. So I think the generic types are necessary.

motlin: The generic types are not necessary. I created that commit to show that Travis still passes. IntelliJ warns about the redundant types. Presumably, the TeamCity build would fail too.

TheLoneKing:
Checked the test coverage. All the implementations of the countByEach method had atleast one hit. I am attaching the jacoco.exec files if you want to double check.

[jacoc-exec.zip](https://github.com/eclipse/eclipse-collections/files/2498404/jacoc-exec.zip)

eclipse-collections-api/src/main/java/org/eclipse/collections/api/stack/ImmutableStack.java
TheLoneKing: IntelliJ does warn about redundant types. I see below that some TeamCity builds have failed too. So, should the generic types be removed?

motlin: Yes please!

nikhilnanivadekar:
@TheLoneKing this PR slipped through my to-do list. Apologies for that! 
Can you please rebase on to master? Our Team City builds now show code coverage as well, so I can just verify it there. 
Thanks for understanding!

eclipse-collections-api/src/main/java/org/eclipse/collections/api/stack/ImmutableStack.java
TheLoneKing: I have removed the generics.

TheLoneKing:
@nikhilnanivadekar I have rebased. Please verify.

PR #501, opened 2018-04-09
Resolve #475

### LINKED ISSUE #475: index method is final in both UnifiedMap and UnifiedSet
While the classes UnifiedMap and UnifiedSet aren't final, their index method is final. This issue is that sometimes index is the perfect method to Override in those classes as opposed to all the other methods that call index. Would it be possible to remove final on the index method? If not what are the design considerations for keeping it final, given that put/get can be overridden and can alter the behaviour of these classes as drastically as say index.

gjajoo:
Some unit tests depend on the ability of collections to create copies of themselves or empty instances of the same class so most of the code changes are around that.


gjajoo:
Will create another pull requests with signed-off commits

PR #503, opened 2018-04-20
 Implement Bag.selectDuplicates(). Fixes #499

### LINKED ISSUE #499: Add method Bag.selectDuplicates().
I searched my codebase for all usages of Bag.selectByOccurrences(). There are only a few usages, but all of them are of the form `bag.selectByOccurrences(occurrences -> occurrences > 1)`. I would suggest adding a convenience method `Bag.selectDuplicates()` which just delegates to `this.selectByOccurrences(occurrences -> occurrences > 1)`. Thoughts?

nikhilnanivadekar:


eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/Bag.java
```diff
@@ -116,6 +116,26 @@
...
+    /**
+     * Returns all elements of the bag that have more than one occurrence.
+     *
+     * @since 9.2.0
+     */
+    default Bag<T> selectByDuplicates()
+    {
+        return this.selectByOccurrences(occurrences -> occurrences > 1);
+    }
+
```
motlin: Let's talk about naming.

* In the commit messages and PR, selectUniques and selectDuplicates. SelectDuplicates sounds ok, but selectUniques sounds a little awkward. Maybe selectUnique?
* In the code, selectByUniques and selectByDuplicates. The plural is definitely awkward. Maybe selectByUnique and selectByDuplicate?
* Some other similar choices, selectUnique/selectDuplicate? selectDuplicates/rejectDuplicates?

nikhilnanivadekar: I'll make the commit message and method names consistent.
Having plural is awkward I agree, but saying selectByUnique might allude to someone that they are getting only 1 element back. That said I don't really have strong feelings about any of the names.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/ImmutableBag.java
```diff
@@ -60,6 +60,18 @@
     @Override
     ImmutableBag<T> selectByOccurrences(IntPredicate predicate);
 
+    @Override
+    default ImmutableBag<T> selectByUniques()
```
donraab: Add the since tag in overrides

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/Bag.java
donraab: I think selectByUnique and selectByDuplicate... it's a shorthand for selectByUniqueOccurrences and selectByDuplicateOccurrences.

motlin: I think "By" is a stretch. I think conversationally, it would be more natural to drop "By". Also, our APIs are pretty consistent today in that By means the method takes a function, which normally returns an attribute or some calculated value to go by.

My ranked preferences:
* selectDuplicates/rejectDuplicates
* selectUnique/selectDuplicate
* selectUniqueOccurrences/selectDuplicateOccurrences

nikhilnanivadekar: How about selectDuplicates (plural), selectUnique  (singular)
In the Javadoc we can clarify to say duplicates returns a bag of all repeating elements and unique returns a bag of all unique elements

motlin:
Why would we return a bag of unique elements? They all have count one.
Should be a set/sortedset/orderedset.

nikhilnanivadekar:
then it won't work efficiently with default methods. selectByOccurrences() returns a Bag, so I will have to call a toSet() on it. LazyIterable, does not have selectByOccurrences().

donraab:
I think it's more important to get the API right then worry about a default method being optimal.  LazyIterable should not have selectByOccurrences.  This is a Bag only API.

motlin:
So skip it completely in 9.2? Select duplicates is the more useful variant
in my opinion.

nikhilnanivadekar:
Yes I know it's a Bag only API and I am suggesting to add it on LazyIterable. I am fine calling a toSet(), without default methods it becomes an overkill of an implementation. 
selectUnique is quite useful but it should be fine to skip it for 9.2
If you want me to remove it let me know
Also have we decided on the name?

donraab:
How about selectUniqueValues / selectDuplicateValues... they can both return Bag as you have it anyway.  Bag and Set are kind of equivalent here anyway at least in terms of iterating.

PR #514, opened 2018-05-19
Remove GA script from site htmls in prep for GDPR.

itohiro73:


nikhilnanivadekar:
LGTM
We need to wait for 10.0.0 target to open. However, the CI instance seems to be down for me create the target. I will reply on this PR when it is good to rebase. Thanks!

itohiro73:
@nikhilnanivadekar Rebased!

PR #522, opened 2018-05-26
Fix #428. Wildcard type in MutableMapIterable#putPair & add methods.

henohenotsuyoshi:
This is for issue https://github.com/eclipse/eclipse-collections/issues/428

PR #523, opened 2018-05-27
Wildcard type in MultableMultimap#putAllPairs & add methods

henohenotsuyoshi:
This is similar to https://github.com/eclipse/eclipse-collections/pull/522

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
@@ -104,7 +104,7 @@ public abstract class AbstractSynchronizedMultimap<K, V> implements MutableMulti
     }
 
     @Override
-    public boolean add(Pair<K, V> keyValuePair)
+    public boolean add(Pair<? extends K, ? extends V> keyValuePair)
     {
         synchronized (this.lock)
         {
@@ -122,7 +122,7 @@ public abstract class AbstractSynchronizedMultimap<K, V> implements MutableMulti
     }
 
     @Override
-    public boolean putAllPairs(Pair<K, V>... pairs)
+    public boolean putAllPairs(Pair<? extends K, ? extends V>... pairs)
     {
         synchronized (this.lock)
         {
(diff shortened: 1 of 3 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
```diff
@@ -131,7 +131,7 @@ public boolean putAllPairs(Pair<K, V>... pairs)
     }
 
     @Override
-    public boolean putAllPairs(Iterable<Pair<K, V>> pairs)
+    public boolean putAllPairs(Iterable<? extends Pair<? extends K, ? extends V>> pairs)
```
motlin: I have mixed feelings about `? extends Pair`. I guess it's technically correct.

It would be good to see the unit tests change to use some supertype/subtype interaction. Maybe something like a Map with Number keys where we add a Pair<Integer, V>

nikhilnanivadekar: even though it is technically correct, do you think it is functionally correct to have `? extends Pair` ?

henohenotsuyoshi: I think it is functionally correct as Collections generally allow to add subclass objects.
Please see use case in AbstractMutableMultimapTestCase.java.

unit-tests/src/test/java/org/eclipse/collections/impl/multimap/AbstractMutableMultimapTestCase.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Goldman Sachs.
+ * Copyright (c) 2018 Goldman Sachs and others.
```
nikhilnanivadekar: Copyright should remain only GS.

nikhilnanivadekar: Also, it will fail checkstyle build due to space after parenthesis.

nikhilnanivadekar:
@motlin this one looks good to me and ready. Ok by you as well?

motlin:
Yes looks good. I’m surprised that the generics work so well.

nikhilnanivadekar:
@henohenotsuyoshi can you please rebase?

henohenotsuyoshi:
I rebased . Thanks!

nikhilnanivadekar:

PR #531, opened 2018-06-09
Fix #529. Populate CollectPrimitiveProcedure.java by stg file

### LINKED ISSUE #529: CollectPrimitiveProcedure.stg to populate CollectPrimitiveProcedure.java
Hi All, 
Currently all CollectPrimitiveProcedure has actual java files respectively in code base.
CollectBooleanProcedure
CollectByteProcedure
CollectCharProcedure
CollectDoubleProcedure
CollectFloatProcedure
CollectIntProcedure
CollectLongProcedure
CollectShortProcedure

I think better to replace those with CollectPrimitiveProcedure.stg file. What do you think ?

Thanks,
Tsuyoshi.

henohenotsuyoshi:
This is for issue #529 
Signed-off-by: sone tsuyoshi <Tsuyoshi.Sone@gs.com>


nikhilnanivadekar:
This looks good to me from a code perspective. 
However, we need to make sure that serialization is not broken. Since it is a major version, I am fine to break serialization, however, I would like to point it in the release notes.
Seems like there are no serialization tests for CollectPrimitiveProcedures. Please add serialization tests similar to org.eclipse.collections.impl.block.procedure.CollectProcedureSerializationTest#serializedForm
and let me know if serialization is broken. Also, if it is broken if I am not mistake you will have to increment the serial version UUID as well. You can also code generate the tests 😄 
I am totally fine if for transparency, you want to create another PR just for the serialization tests and get it merged before checking on this PR.

henohenotsuyoshi:
@nikhilnanivadekar I see. I will create another PR for serialization test then. Thanks!

nikhilnanivadekar:
I merged the Serialization Tests. Please rebase and push. That should tell us if serialization is broken.

nikhilnanivadekar:
Accepting the build failure as it is failed for OSGi bundle for Java 10 and Java 11-EA.
Serialization tests have passed even with the new implementation denoting that there are no serialization changes. Verified that this is not a breaking change as the package names are kept same and the functionality is kept same. 

Thanks for your contribution! 👍

PR #542, opened 2018-06-28
Remove unused ant build and upgrade maven plugins.

nikhilnanivadekar:
The Ant Code Generator is not used as the entire build is now maven based. This PR is to remove the unnecessary ant builds and the Ant Code Generator module.
Also upgrade few maven plugins.


motlin:
Our position used to be that you could use any or maven, whichever you
want. If anyone wants to add cradle support, it should be a welcome
contribution. If any doesn’t work, it’s just a bug.

On the other hand, I doubt many people care about ant. Still, it worked
well before we moved to travis.

motlin:
Deleting .idea/libraries will break the TeamCity builds when running with the IntelliJ or Inspections runners.

nikhilnanivadekar:
IMO having ant build there is 1 more thing we have to maintain. If you look at the build.xmls they are fairly dated and unmaintained. All our documentation suggests to use maven and as a project we can take the decision that we want to use maven as the default build tool. Later we might change that to be grade if there is a valid reason for it. It is fairly common for projects to set this convention. Also, all our CI, Release builds run using maven. 
I understand your point that we should be fine with a contributor using whichever build tool they desire. But I would like to reduce the number of things which need to be maintained, upgraded, tested and documented.

motlin:
@donraab @mohrezaei what do you think?

mohrezaei:
I don't mind.

donraab:
I'm not aware of anyone using the ant build.  I've never used it.  I'm ok with removing it.  If someone raises an issue, then we can always bring it back.

nikhilnanivadekar:
Thanks @mohrezaei , @donraab for the confirmation. If it is fine by the group and no one in the community is raising any objections, we should be fine to merge this PR. Thanks!

Issue #450, opened 2018-01-23
Expand scope of org.eclipse.collections.api.RichIterable#groupByUniqueKey from MutableMap to MutableMapIterable


nikhilnanivadekar:

nikhilnanivadekar:
Fixed via #549

PR #549, opened 2018-07-06
Change groupByUniqueKey() so that its target extends MutableMapIterable instead of MutableMap.

motlin:
This fixes a mistake, but it's a breaking change.


motlin:
Fixes #450

nikhilnanivadekar:
it is breaking because we will lose the API from MutableMap correct?
Or there is some other reason that I am missing?

motlin:
It's source compatible, but any change like this breaks binary compatibility.

PR #553, opened 2018-07-10
Add missing bounded wildcards on generic types.

motlin:


nikhilnanivadekar:
How did you find all the missing bounded wildcards? I am curious if there was a tool or a trick you used?

motlin:
Great question. I requested a new inspection in 2015 and it was just added!

https://youtrack.jetbrains.com/issue/IDEA-142020

PR #556, opened 2018-07-12
Change math function factories to use Math.*Exact methods.

motlin:


mohrezaei:
Why do we want exceptions on overflow? It's unclear such behavior should be default when the rest of Java doesn't handle it as such, and ignoring overflow behavior is actually desired behavior in many cases (e.g. hashing, compression, cryptographic functions, etc).

motlin:
The place where I really want this behavior is in Bags, since we use these functions to manipulate the number of occurrences. Though I suppose I could make the change just in bags to reduce impact.

mohrezaei:
Overflow checks are of course necessary in many places, but since this is a backward breaking change, I would suggest minimizing the impact.

motlin:
I'll close this and do it differently in Bags.

PR #559, opened 2018-07-13
Delete default implementations that were added for backwards compatibility.

motlin:


nikhilnanivadekar:
Looping in @donraab , @mohrezaei 
I really wanted to remove the default implementations which just throw. But seems like we have users who will start breaking. That is why I had shelved the work I did on this.

mohrezaei:
what prevents these from being properly implemented in the default implementation?

nikhilnanivadekar:
they need dependencies from eclipse-collections-impl

mohrezaei:
I assume that's just for construction? If so, can a little bit of reflection deal with that?

motlin:
I have this idea I've been meaning to try forever. If the collection factories registered themselves with Java's ServiceLocator, we could move Lists, Sets, Maps (the factories) into api. Then a whole bunch more behavior could live in the interfaces.

nikhilnanivadekar:
@motlin I think this PR can be reopened. I think it is fine to break backwards compatibility. I will mention it in the release notes with all the methods for which default implementations were removed. After all new API is being added without default implementation which throws, so backwards compatibility will already be broken.
@donraab @mohrezaei thoughts?

mohrezaei:
I like @motlin's idea of using the ServiceLocator. It allows a decent default implementation of these methods and furthermore, makes it a lot easier for EC users to use EC interfaces for their own collections.

mohrezaei:
Maybe we could open an issue for the ServiceLocator as a prerequisite of adding these implementations.

nikhilnanivadekar:
In my mind ServiceLocator can be picked up in minor release, but removing default methods is a major release change. So, I would rather remove these default implementations and then do the ServiceLocator of the same release or a future release.
The breaking change will be at compile time, with an auto-fix using the IDE, so I am less worried.

mohrezaei:
I think the point is, once we do the ServiceLocator, we'll put these methods back, so why remove them now and cause trouble for our users?
You're right that the ServiceLocator can be released in a minor version. Interestingly, so can the actual implementation of these default methods.

PR #562, opened 2018-07-14
Fix whitespace problems.

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -1515,7 +1515,8 @@ public void readExternal(ObjectInput in) throws IOException, ClassNotFoundExcept
     {
         int size = in.readInt();
         this.loadFactor = in.readFloat();
-        this.init(Math.max((int) (size / this.loadFactor) + 1,
+        this.init(Math.max(
```
nikhilnanivadekar: should be `Math.max` be in a new line?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
```diff
@@ -1559,7 +1559,8 @@ public void readExternal(ObjectInput in) throws IOException, ClassNotFoundExcept
         this.hashingStrategy = (HashingStrategy<? super K>) in.readObject();
         int size = in.readInt();
         this.loadFactor = in.readFloat();
-        this.init(Math.max((int) (size / this.loadFactor) + 1,
+        this.init(Math.max(
```
nikhilnanivadekar: should the Math.max() be in a new line?

unit-tests/src/test/java/org/eclipse/collections/impl/PersonAndPetKataTest.java
```diff
@@ -410,14 +408,12 @@ public void groupPeopleByLastName()
     @Test
     public void groupPeopleByLastNameUsingStreams()
     {
-        Map<String, List<Person>> byLastNameStream =
-                this.people.stream().collect(
-                        Collectors.groupingBy(Person::getLastName));
+        Map<String, List<Person>> byLastNameStream = this.people.stream()
```
nikhilnanivadekar: should .stream() be on a new line?

unit-tests/src/test/java/org/eclipse/collections/impl/PersonAndPetKataTest.java
```diff
@@ -410,14 +408,12 @@ public void groupPeopleByLastName()
...
+                .collect(Collectors.groupingBy(Person::getLastName));
         Verify.assertIterableSize(3, byLastNameStream.get("Smith"));
 
-        Map<String, MutableBag<Person>> byLastNameStreamTargetBag =
-                this.people.stream().collect(
-                        Collectors.groupingBy(Person::getLastName, Collectors.toCollection(Bags.mutable::empty)));
+        Map<String, MutableBag<Person>> byLastNameStreamTargetBag = this.people.stream()
```
nikhilnanivadekar: should .stream() on new line

unit-tests/src/test/java/org/eclipse/collections/impl/PersonAndPetKataTest.java
```diff
@@ -566,12 +562,10 @@ public void getCountsByPetType()
...
     public void getCountsByPetTypeUsingStreams()
     {
-        Map<PetType, Long> countsStream =
-                Collections.unmodifiableMap(
-                        this.people.stream()
-                                .flatMap(person -> person.getPets().stream())
-                                .collect(Collectors.groupingBy(Pet::getType,
-                                        Collectors.counting())));
+        Map<PetType, Long> countsStream = Collections.unmodifiableMap(
+                this.people.stream()
```
nikhilnanivadekar: .stream() on new line

unit-tests/src/test/java/org/eclipse/collections/impl/PersonAndPetKataTest.java
```diff
@@ -648,12 +642,10 @@ public void getCountsByPetAge()
...
     public void getCountsByPetAgeUsingStreams()
     {
-        Map<Integer, Long> countsStream =
-                Collections.unmodifiableMap(
-                        this.people.stream()
-                                .flatMap(person -> person.getPets().stream())
-                                .collect(Collectors.groupingBy(Pet::getAge,
-                                        Collectors.counting())));
+        Map<Integer, Long> countsStream = Collections.unmodifiableMap(
+                this.people.stream()
```
nikhilnanivadekar: .stream() on new line

unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/UnmodifiableBiMapTest.java
```diff
@@ -271,7 +271,8 @@ public void withoutAllKeys()
     @Test
     public void withAllKeyValues()
     {
-        Verify.assertThrows(UnsupportedOperationException.class,
+        Verify.assertThrows(
+                UnsupportedOperationException.class,
                 () -> this.newMapWithKeysValues("A", 1, "B", 2).withAllKeyValues(
```
nikhilnanivadekar: .withAllKeysValues() should be on new line.

unit-tests/src/test/java/org/eclipse/collections/impl/list/fixed/ArrayAdapterTest.java
```diff
@@ -404,19 +409,22 @@ public void rejectWith()
     {
         super.rejectWith();
 
-        Verify.assertContainsAll(ArrayAdapter.newArrayWith(1, 2, 3, 4).rejectWith(Predicates2.lessThan(), 3),
+        Verify.assertContainsAll(
+                ArrayAdapter.newArrayWith(1, 2, 3, 4).rejectWith(Predicates2.lessThan(), 3),
                 3,
```
nikhilnanivadekar: 3, 4 should be in one line to be in line with line 419

motlin:
Some long statements have many method calls and many parameters in a list. We don't have any rules about whether to prioritize wrapping on the method calls (each dot) or the parameters (each comma). However, one rule we generally follow is that if any items in a list are line wrapped, all items in the list should be line wrapped. This is one IntelliJ can do automatically, by turning on the setting "newline after '('". Most of these changes were done automatically by the IDE. There were a few I cleaned up by hand. I didn't try to always wrap method calls in this commit.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
motlin: I don't think so. init() only has a single parameter. Imagine extracting it as local variable.

```java
        int initialCapacity = Math.max(
                (int) (size / this.loadFactor) + 1,
                DEFAULT_INITIAL_CAPACITY);
        this.init(initialCapacity);
```

vs

```java
        this.init(
                initialCapacity);
```

unit-tests/src/test/java/org/eclipse/collections/impl/PersonAndPetKataTest.java
motlin: Doesn't matter much to me. collect() only takes a single parameter, so I thought it was weird to wrap that like it was a list. What do you think?

motlin:
I removed everything that I think is debatable. Should be ready for review again.

PR #487, opened 2018-03-25
Add test coverage, enable checkstyle for generated code and fix offenders, enable JDK 10 GA builds.

nikhilnanivadekar:

PR #564, opened 2018-07-16
Fix indentation problems.

motlin:


nikhilnanivadekar:
Shouldn't this be a part of #562 ? Just trying to avoid git churn.

motlin:
I kept them separate on purpose. The wrapping problems don't cause any
build to fail. I found them using IntelliJ. The indentation problems I
found using CheckStyle. They will cause the build to fail as soon as I
reenable CheckStyle.

nikhilnanivadekar:
Commented on #565 if there is a possibility to enable and fix offenders as a single PR but multiple commits with fixes for different break. Similar to #487

PR #565, opened 2018-07-16
Fix static analysis violations in generated code.

motlin:


nikhilnanivadekar:
Were you able to run inspections on your local IntelliJ with generated code? I started it, but it took forever and had to kill it.
We probably need a separate nightly build for inspections of all generated code. Running it as a part of PR or CI is too time consuming and will frustrate contributors.

motlin:
No. These are CheckStyle violations, not IntelliJ Inspection violations.

Our CheckStyle build already runs on all generated code. It's never failing because it's misconfigured, but the time that the build takes is realistic, because it's running mvn install checkstyle:check.

motlin:
Oh, you saw that I removed `final` and stuff in a few places. What I do is I right click the generated directory in IntelliJ and unmark it as generated. This makes it so I see warnings in a file I'm already reading. But I never run Inspections globally on Eclipse Collections either. Besides the project being too big, you'd also see each error 64x times for each primitive-primitive map.

nikhilnanivadekar:
How you mean misconfigured?
Also if that is the case, I think we should take the same approach as #487 where the checkstyle build was enabled and the violators were fixed as 1 PR but multiple commits.
It will reduce your rebasing efforts and help with 1 single monolithic review.
That said, I am fine reviewing and merging one fix at a time, but that leaves us susceptible to other changes being merged which have violations and then going and fixing them again.

motlin:
I can combine or split the PRs however you'd like, but the build won't pass either way if I turn on the Indent rule. These PRs fix many but not all of the violations.

nikhilnanivadekar:
my only concern is that once we merge these issues and if the rule is not turned on, it keeps the library open for these issues and again someone will have to go through the process of fixing it. We can pair up on fixing the violations if there are too many after these are merged.

motlin:
Yes, that is the case, and I think it's been like that for a long time, at
least a year. Unless there is a problem with this patch, I think it makes
sense to merge and continue.

PR #569, opened 2018-07-19
Build fixes and improvements.

motlin:
* Remove unused dependencies from maven build.
* Fix configuration of maven's pluginRepositories to only include repo.eclipse.org if the release-artifacts profile is enabled.
* Fix 'maven install' for newer versions of Java by using maven profiles to link modules with the version of Java that can build them.
* Move maven modules 'test-coverage' and 'p2-repository' from the default list into the 'all' profile.
* Fix Travis builds to cache the maven local cache without including Eclipse Collections jars.
* Change Travis builds to not use a custom version of maven.
* Change Travis job which only compiles to not include jmh-tests, since it gets compiled in other builds.


pom.xml
```diff
@@ -142,28 +142,6 @@
         <project.previous.version>9.2.0</project.previous.version>
     </properties>
 
-    <pluginRepositories>
```
nikhilnanivadekar: have you tested this with eclipse? That the project clean installs.
I think this was put in specifically for eclipse users.

pom.xml
```diff
@@ -567,6 +565,8 @@
                 <module>serialization-tests</module>
                 <module>acceptance-tests</module>
                 <module>performance-tests</module>
+                <module>test-coverage</module>
```
nikhilnanivadekar: should this be squashed with 1c2a0f0d7a26aca1dc957ddc6aed4914e490ccfb ?

.travis.yml
```diff
@@ -8,10 +8,6 @@ addons:
       - oracle-java9-installer
 
 before_install:
-  - wget https://archive.apache.org/dist/maven/maven-3/3.5.0/binaries/apache-maven-3.5.0-bin.zip
```
nikhilnanivadekar: I had added this for stability especially for the newer JDK versions. I would keep this because Travis upgrades the stock maven really slowly, and we will have to add similar steps in the future.
This does speed up the build and I think saves 2-3 minutes, however, longer term I think we should stick to the custom maven version.

pom.xml
motlin: I did not test in eclipse. I'd like to understand why it would affect eclipse though, since the pluginRepositories section was edited relatively recently and we've supported eclipse for years.

pom.xml
motlin: It's debatable but I think it's thematically different. The modules that rely on Scala don't compile in Java 9, 10, or 11 regardless of other profiles or configuration. Turning on Jacoco does cause Java 11 to fail, but not just on the test-coverage module. So I do think there ought to be a new Jacoco maven profile, but it wouldn't have any automatic activation criteria. And I don't think we run coverage builds in travis, so it's outside my current focus.

pom.xml
motlin: I can try this out in Eclipse shortly.

pom.xml
nikhilnanivadekar: I am fine it not being squashed. Was just asking!

motlin:
The script download and unzips maven 3.5.0. Deleting that brings us forward
to 3.5.2.

.travis.yml
motlin: The script download and unzips maven 3.5.0. Deleting that brings us forward
to 3.5.2.

motlin: Even if we keep it, it should probably have a link to https://github.com/travis-ci/enterprise-installation/issues/14 for credit

nikhilnanivadekar:
Is this PR still valid? I see others raised which I am reviewing.

PR #572, opened 2018-07-24
Update IntelliJ code style settings and add run configurations.

motlin:


motlin:
This one is good to go now. You can see that the IntelliJ builds are running successfully again.

https://teamcity.jetbrains.com/project.html?projectId=OpenSourceProjects_EclipseCollections&branch_OpenSourceProjects_EclipseCollections=572

.idea/modules.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nikhilnanivadekar: I thought this file was no longer needed. From the comment on #581 but that applies to only projects which do an auto-import. I am not quite sure if it will work on TC.

.idea/scala_compiler.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="ScalaCompilerConfiguration">
-    <profile name="Maven 1" modules="jmh-scala-tests,scala-unit-tests" />
```
nikhilnanivadekar: is this expected churn?

motlin:
I made these changes doing repeated pushes to watch TeamCity build.
Removing any of these files breaks the TeamCity build. #581 is from the
perspective of a team that uses maven and IntelliJ but not TeamCity.

PR #574, opened 2018-07-24
Change Travis builds to get the version of maven from the environment.

motlin:


.travis.yml
```diff
@@ -8,9 +8,9 @@ addons:
       - oracle-java9-installer
 
 before_install:
-  - wget https://archive.apache.org/dist/maven/maven-3/3.5.0/binaries/apache-maven-3.5.0-bin.zip
-  - unzip -qq apache-maven-3.5.0-bin.zip
-  - export M2_HOME=$PWD/apache-maven-3.5.0
+  - wget https://archive.apache.org/dist/maven/maven-3/${MAVEN_VERSION}/binaries/apache-maven-${MAVEN_VERSION}-bin.zip
```
nikhilnanivadekar: I do not think this will work for builds on contributor branches. Each contributor will have to set their build environment variable no? I do not think Travis has a way to actually add a default option.

nikhilnanivadekar:
I am fine with any version. I am not fine having it as an environment variable as it creates a maintenance nightmare. Remember the JDK-11-EA urls 😉 
The only point I will say is that we should have control of our version. We might fall behind the default, but then we can upgrade. I have seen that we do build improvements and build hygiene in pretty much every release. I think we should hard code the maven version.

PR #575, opened 2018-07-24
Remove maven option --batch-mode since it disables colorized output.

motlin:


nikhilnanivadekar:
should this be changed in the Jenkins builds on Eclipse Foundation as well?

motlin:
@nikhilnanivadekar Jenkins doesn't support ansi colors in the output by default. It has a plugin to add that functionality. Do you know if it's installed?

I saw that Travis supports colors out of the box, so this change was a freebie.

PR #578, opened 2018-07-24
Remove obsolete maven prerequisites.

motlin:


motlin:
Maven prerequisites are intended to be used by plugin modules, not plain jar modules. We've been logging errors for a warning, which suggest to only use the enforcer plugin instead.

PR #579, opened 2018-07-24
Fix 'maven install' for newer versions of Java by using maven profiles to link modules with the version of Java that can build them.

motlin:


nikhilnanivadekar:
I have approved this, I presume that this will not impact the Jenkins builds we use for releases on Eclipse Foundation.

PR #582, opened 2018-07-26
Add .mvn/maven.config with options --show-version and --errors.

motlin:
This change makes it so that the flags --show-version and --errors are always on, so you don't need to specify them on the command line.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,2 @@
+--show-version
+--errors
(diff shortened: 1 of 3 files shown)


nikhilnanivadekar:
Is this a part of build sanitization exercise?

motlin:
Yes. My goal is that unless you want to run something different (coverage, performance tests, etc) you just run `mvnw` with no args.

PR #584, opened 2018-07-30
Add maven wrapper and configure builds to use it.

motlin:


nikhilnanivadekar:
Since this is an Apache License we will have to get it through CQ. Since you are a committer you can also file a CQ with the Eclipse Foundation.

.travis.yml
```diff
@@ -8,10 +8,7 @@ addons:
       - oracle-java9-installer
 
 before_install:
-  - wget https://archive.apache.org/dist/maven/maven-3/3.5.0/binaries/apache-maven-3.5.0-bin.zip
-  - unzip -qq apache-maven-3.5.0-bin.zip
-  - export M2_HOME=$PWD/apache-maven-3.5.0
-  - export PATH=$M2_HOME/bin:$PATH
+  - mvn -N io.takari:maven:wrapper
```
motlin: I don't feel like going through a CQ is worth it, so I removed the Apache code and binaries. Instead, I'm now treating the maven-wrapper maven plugin just like an improved maven downloader, and replacing the wget stuff with it.

The advantages are:
* It automatically downloads the latest maven. We can lock it to a single version if we want, but I don't particularly want to. Our builds should be able to run on pretty much any maven 3.x.
* It downloads maven jars into .m2/ which gets cached. And the wrapper plugin will also get cached.
* This will bump us to maven 3.5.4 which has improved logs, which I do actually want in our travis builds.

nikhilnanivadekar:
Can you please squash the commits?

motlin:
I believe that the maven wrapper is interfering with ansi colors, so I may scrap this.

motlin:
I think we should merge #591 instead, for now.

nikhilnanivadekar:
I like the idea of scrapping this :)

motlin:
I think there's a bug in maven 3.5.4 related to ansi coloring. If there weren't, I think this is clearly the way to go. I will pursue the maven wrapper with a fixed version of maven.

donraab:
@motlin @nikhilnanivadekar Is this being scrapped?

motlin:
I hope not. I think we should keep progressing toward less custom bash
script and more common build tools.

motlin:
Any additional feedback? This has been pending for a long time and I believe it's good to go.

nikhilnanivadekar:
The custom bash script should also change no?

motlin:
The custom bash script did change. There are two places where I changed it to use mvnw instead of mvn. No other changes should be necessary since the before_install runs regardless of whether we're invoking an inline command or a separate script. In fact, the line echoing the maven options in the custom script is redundant, but that will be a fix for another day.

PR #585, opened 2018-07-31
Change collection factories to be services that get loaded by the ServiceLoader in the api package. Move a few methods up from abstract classes to default interface implementations.

motlin:


eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/Lists.java
```diff
@@ -0,0 +1,89 @@
...
+ * FixedSize Examples:
+ *
+ * <pre>
+ * FixedSizeList&lt;String&gt; emptyList = Lists.fixedSize.empty();
+ * FixedSizeList&lt;String&gt; listWith = Lists.fixedSize.with("a", "b", "c");
+ * FixedSizeList&lt;String&gt; listOf = Lists.fixedSize.of("a", "b", "c");
+ * </pre>
+ */
+@SuppressWarnings("ConstantNamingConvention")
+public final class Lists
```
motlin: This is a *copy* of Lists with changes. The main copy lives in impl, this one lives in api. I'll mark the main differences below.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/Lists.java
```diff
@@ -0,0 +1,89 @@
...
+{
+    public static final ImmutableListFactory immutable;
+    public static final MutableListFactory mutable;
+    public static final FixedSizeListFactory fixedSize;
```
motlin: The copy in impl has a fourth factory, for multi-reader collections. That factory also implements MutableListFactory.

When I first wrote Lists, I meant to use it this way, as a dependency breaking technique. But it existed for a long time and became a convenient one-stop-shop for List stuff. I don't think MultiReader collections belong here because they're not popular, and the "mu" prefix messes with auto-complete. But I realize that there's no good alternative place for it to go.

If we want MultiReader here, then I'd have to add another method to the MutableListFactory interface, like getKey() to return a unique identifier to discriminate between mutable and multiReader.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/Lists.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private Lists()
+    {
+        throw new AssertionError("Suppress default constructor for noninstantiability");
+    }
+
+    private static <T> T loadServiceClass(Class<T> serviceClass)
+    {
+        List<T> factories = new ArrayList<>();
+        for (T factory : ServiceLoader.load(serviceClass))
```
motlin: Here's the ServiceLoader stuff. On failure, I could probably give better error messages.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
```diff
@@ -61,24 +63,43 @@
     MutableList<T> withoutAll(Iterable<? extends T> elements);
 
     @Override
-    MutableList<T> newEmpty();
+    default MutableList<T> newEmpty()
+    {
+        return Lists.mutable.empty();
```
motlin: Here's the first example of using the Lists factory in a default method.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
```diff
@@ -61,24 +63,43 @@
...
+    }
 
     MutableList<T> clone();
 
     @Override
-    MutableList<T> tap(Procedure<? super T> procedure);
+    default MutableList<T> tap(Procedure<? super T> procedure)
+    {
+        this.forEach(procedure);
+        return this;
```
motlin: While looking for methods that could be moved up, I found plenty like this one where it could already be moved up and doesn't need the new Lists factory at all.

eclipse-collections/pom.xml
```diff
@@ -42,6 +42,13 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.google.auto.service</groupId>
+            <artifactId>auto-service</artifactId>
+            <version>1.0-rc4</version>
+            <optional>true</optional>
+        </dependency>
```
motlin: auto-service was a convenient way for me to register stuff with the ServiceLoader, but it doesn't carry its own weight. And it depends on Guava. The production implementation of this should just mess with META_INF directly, without using a framework.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/FixedSizeListFactoryImpl.java
```diff
@@ -10,10 +10,12 @@
...
 
+import com.google.auto.service.AutoService;
 import org.eclipse.collections.api.factory.list.FixedSizeListFactory;
 import org.eclipse.collections.api.list.FixedSizeList;
 import org.eclipse.collections.impl.utility.Iterate;
 
+@AutoService(FixedSizeListFactory.class)
 public enum FixedSizeListFactoryImpl implements FixedSizeListFactory
 {
     INSTANCE;
```
motlin: This actually doesn't work yet because I left these Impl classes as singleton enums. I need to convert them to plain old classes with a zero-arg constructor so that they can be instantiated through reflection. this is a fairly breaking change so I wanted to highlight it.

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/FunctionalInterfaceSetTest.java
```diff
@@ -323,9 +324,9 @@ public void test_parallel_lazy_jdk()
     @Benchmark
     public MutableSet<Integer> serial_eager_ec()
     {
-        FastList<Integer> select1 = this.integersEC.select(each -> each % 10_000 != 0);
-        FastList<String> collect1 = select1.collect(String::valueOf);
-        FastList<Integer> collect2 = collect1.collect(Integer::valueOf);
+        MutableList<Integer> select1 = this.integersEC.select(each -> each % 10_000 != 0);
+        MutableList<String> collect1 = select1.collect(String::valueOf);
+        MutableList<Integer> collect2 = collect1.collect(Integer::valueOf);
```
motlin: When I pulled some methods up to interfaces, I deleted some overrides that returned concrete classes. This is debatably the correct thing to do, but also is a not-totally-necessary source-breaking change that I could revert.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
```diff
@@ -61,24 +63,43 @@
     MutableList<T> withoutAll(Iterable<? extends T> elements);
 
     @Override
-    MutableList<T> newEmpty();
+    default MutableList<T> newEmpty()
```
mohrezaei: Not sure about this. Isn't the method's intent to return the same type as the one being called? If so, this default implementation seems to break that contract. Was this forced by the calls to this.newEmpty() in the other default methods?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
```diff
@@ -273,20 +309,35 @@
      * Returns a new MutableList in reverse order.
      */
     @Override
-    MutableList<T> toReversed();
+    default MutableList<T> toReversed()
+    {
+        return this.toList().reverseThis();
```
mohrezaei: This is also surprising, given the impl of ```toList()```. If I call toReversed() on a list, I'd want the same type back, not one from the factory.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/FixedSizeListFactoryImpl.java
mohrezaei: Why is that a breaking change?

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/FunctionalInterfaceSetTest.java
mohrezaei: Seems to be a theme here: do we want to mentally reinforce that collections return their own types or not? I think previous to this, the answer has been "yes."

mohrezaei:
My one overarching question is whether this change breaks the independence of the api jar. Specifically, can I have the api jar in my classpath, compile against it, run with it, without needing the impl jar?
Can there be a fallback in the loading where we gracefully deal with the lack of the impl jar, and maybe delegate the exception throwing to the actual method call (e.g. ```newEmpty()```) instead of class loading time?

We've faced this scenario in Reladomo, where we depend on the GSC api jar, but not the impl.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
motlin: This is optional. It's just the implementation from AbstractMutableList moved up. I could remove it.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
motlin: While newEmpty had a strange implementation, presumably toList() would be implemented to return the same type, and then this one would overall return the correct type.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/Lists.java
mohrezaei: See my comment below. Maybe on failure we can have a default factory that throws when called (and a warning message at load time)?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
mohrezaei: I vote to remove.

motlin:
@mohrezaei This change would let you compile against api, without impl on the classpath of javac. However, it doesn't remove any runtime dependency. Without the impl jar, you'd get an error like this.

```
java.lang.ExceptionInInitializerError
	at ExampleWrongDepTest.wrongDependency(ExampleWrongDepTest.java:9)
Caused by: java.lang.RuntimeException: Could not find any implementations of ImmutableListFactory. Check that eclipse-collections.jar is on the classpath and that its META-INF/services directory is intact.
	at org.eclipse.collections.api.factory.list.Lists.loadServiceClass(Lists.java:74)
	at org.eclipse.collections.api.factory.list.Lists.<clinit>(Lists.java:54)
	... 23 more
```

mohrezaei:
I don't like that error. I think that should just be a warning. The error should happen if/when the factory method is called.

mohrezaei:
One other concern: how does this affect the fat-jar people?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/Lists.java
mohrezaei: Are you proposing to have two factories or remove the impl one eventually?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/FixedSizeListFactoryImpl.java
motlin: I guess I can create a public INSTANCE constant to keep it source compatible.

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/FunctionalInterfaceSetTest.java
motlin: Implementations almost always return the same time, except some adapters. Return types have been inconsistent in interfaces. Plus only some methods return the same type. Filtering methods (select, reject, partition, selectInstancesOf) do return the same time. Transforming methods (collect, flatCollect) only sometimes return the same type. Often this is obvious from the signature. MapIterable.collect(Function) returns RichIterable, for example. But when an interface method returns that same interface, I guess the assumption is it's also going to be the same concrete type.

motlin:
We've talked about more ambitious changes where we have a module for container implementations (like abstract array-based list), another for api (RichIterable, MutableList), and another that pulls together both (FastList). That's a lot of work though. Absent that, I'm not sure how we'd turn this error into a warning? The actual lists (like FastList) and the list factory are both not on the classpath if you're seeing this error.

ServiceLoader just looks for META_INF/services on the classpath, reads it, finds class names in it, and looks for those on the classpath too. So it shouldn't affect uber jars at all. Even shading should still work. I'd have to confirm that it doesn't mess with OSGi.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/FixedSizeListFactoryImpl.java
mohrezaei: Not that it matters, but what breaks binary compatibility? Am I misunderstanding that all you're doing is removing the private constructor?

mohrezaei:
How to turn error into warning: have a default implementation of the factory that throws?

motlin:
Ah, I think I understand. The errors still trigger when you call Lists.mutable.empty(), but it would happen during the execution of empty(), not during static class loading of Lists. And that would lead to a more straight-forward exception. Is that your line of thought? I wouldn't be able to return some other list other than FastList.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/Lists.java
motlin: Good question. I could see deprecating the impl one and eventually deleting it. It's pretty disruptive to import statements though and pretty widely used. I could see keeping both, but then the new ones should probably have different names (ListFactory instead of Lists, or something like that). The downside is that the old one has the better name, and the good one is the one we'd encourage to use. What do you think?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/FixedSizeListFactoryImpl.java
motlin: I am not 100% sure that it's a binary compatibility break. But I know that using the enum keyword changes the bytecode in a few ways. Now that I'm thinking about it, I know a few more things I can do to improve compatibility.

motlin:
I will move the exception into a ThrowingMutableListFactory, or something like that. I'm not sure about logging a warning during class initialization, simply because we don't depend on any logger today. We don't even have any calls to System.out today.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/Lists.java
mohrezaei: I'm not convinced having two is a bad thing. The impl has the advantage that it never throws. Maybe if we play our cards right, we can get the api one to call the impl one, so that the api one exists for the default method implementation and the use cases where another framework doesn't want a dependency on the impl at compile time (with runtime being potentially optional).

mohrezaei:
I'm fine with no warning at load time.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/Lists.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+@SuppressWarnings("ConstantNamingConvention")
+public final class Lists
+{
+    public static final ImmutableListFactory immutable =
+            Lists.loadServiceClass(ImmutableListFactory.class, ThrowingImmutableListFactory::new);
+    public static final MutableListFactory mutable =
+            Lists.loadServiceClass(MutableListFactory.class, ThrowingMutableListFactory::new);
+    public static final FixedSizeListFactory fixedSize =
+            Lists.loadServiceClass(FixedSizeListFactory.class, ThrowingFixedSizeListFactory::new);
```
motlin: Following @mohrezaei's suggestions, I changed the implementation of loadServiceClass to use a fallback implementation. Now it will never fail at class loading time. The exception message will be the same, but the stack trace will look a little more understandable.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/ThrowingMutableListFactory.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public ThrowingMutableListFactory(String error)
+    {
+        this.error = error;
+    }
+
+    @Override
+    public <T> MutableList<T> empty()
+    {
+        throw new IllegalStateException(this.error);
```
motlin: The Throwing*Factories implement the same interfaces, like MutableListFactory in this case. I pass in the error message in the constructor, since I use a different error message for 0 implementations found, and for >2 implementations found. All methods throw IllegalStateException, but I'm not sure if that's the right choice.

mohrezaei:
Latest commit looks pretty good 😄

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/ThrowingFixedSizeListFactory.java
```diff
@@ -0,0 +1,137 @@
+/*
```
mohrezaei: I wonder if we should hide these classes. Inner classes somewhere? Maybe even a dynamic proxy with a trivial invocation handler (always throws)?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MutableListFactoryImpl.java
```diff
@@ -14,9 +14,9 @@
 import org.eclipse.collections.api.factory.list.MutableListFactory;
 import org.eclipse.collections.api.list.MutableList;
 
-public enum MutableListFactoryImpl implements MutableListFactory
+public class MutableListFactoryImpl implements MutableListFactory
 {
-    INSTANCE;
+    public static final MutableListFactory INSTANCE = new MutableListFactoryImpl();
```
motlin: I kept a constant named `INSTANCE` to improve backwards compatibility. However, this class is no longer a singleton. At runtime, there will really be two instances of MutableListFactoryImpl. The one in api.Lists will be instantiated by the ServiceLoader, which instantiates the class with the zero-arg constructor. The one in impl.Lists will be this INSTANCE. And of course, it would be possible for users to construct more.

I thought about sharing an instance like this:
```java
public final class Lists
{
    public static final ImmutableListFactory immutable = org.eclipse.collections.api.factory.list.Lists.immutable;
    public static final MutableListFactory mutable = org.eclipse.collections.api.factory.list.Lists.mutable;
    public static final FixedSizeListFactory fixedSize = org.eclipse.collections.api.factory.list.Lists.fixedSize;
    public static final MutableListFactory multiReader = MultiReaderMutableListFactory.INSTANCE;
}
```
However, two instances feels simpler, and avoids rare but strange edge cases like confusion around two copies of the same jar on the classpath.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/ThrowingFixedSizeListFactory.java
mohrezaei:
```java
public class ThrowingInvocationHandler  implements InvocationHandler{
    private String error;
    public  MyInvocationHandler(String error) {
        this.error = error;
    }
    public Object invoke(Object proxy, Method m, Object[] args) throws Throwable {
        throw new IllegalStateException(this.error);
    }
}
```
creating one looks like this:
```java
(FixedSizeListFactory) Proxy.newProxyInstance(Lists.class.getClassLoader(), 
  new Class[] { FixedSizeListFactory.class }, 
  new ThrowingInvocationHandler(error));
```

Advantages: shorter code, never have to implement any new methods and totally outside the public API.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/ServiceLoaderUtils.java
```diff
@@ -0,0 +1,52 @@
...
+    public static <T> T createProxyInstance(Class<T> serviceClass, String message)
+    {
+        InvocationHandler handler = new ThrowingInvocationHandler(message);
+        Object proxyInstance = Proxy.newProxyInstance(
+                serviceClass.getClassLoader(),
+                new Class[]{serviceClass},
+                handler);
+        return serviceClass.cast(proxyInstance);
+    }
+}
```
motlin: I followed the suggestion to create a dynamic proxy. I extracted it into ServiceLoaderUtils to share between Lists, Sets, and others. This is to be more realistic about what public methods and classes will show up in public api.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/set/Sets.java
```diff
@@ -0,0 +1,59 @@
...
+    public static final MutableSetFactory mutable =
+            ServiceLoaderUtils.loadServiceClass(MutableSetFactory.class);
+    public static final FixedSizeSetFactory fixedSize =
+            ServiceLoaderUtils.loadServiceClass(FixedSizeSetFactory.class);
+
+    private Sets()
+    {
+        throw new AssertionError("Suppress default constructor for noninstantiability");
+    }
+}
```
motlin: I'm not doing anything with the Sets factory yet. I just wanted to create a second example to force more code sharing between Lists and Sets which live in different packages.

eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Sets.java
```diff
@@ -71,9 +71,9 @@
  * FixedSize Examples:
  *
  * <pre>
- * FixedSizeList&lt;String&gt; emptySet = Sets.fixedSize.empty();
- * FixedSizeList&lt;String&gt; setWith = Sets.fixedSize.with("a", "b", "c");
- * FixedSizeList&lt;String&gt; setOf = Sets.fixedSize.of("a", "b", "c");
+ * FixedSizeSet&lt;String&gt; emptySet = Sets.fixedSize.empty();
+ * FixedSizeSet&lt;String&gt; setWith = Sets.fixedSize.with("a", "b", "c");
+ * FixedSizeSet&lt;String&gt; setOf = Sets.fixedSize.of("a", "b", "c");
```
motlin: Found this documentation mistake along the way.

eclipse-collections/src/main/resources/META-INF/services/org.eclipse.collections.api.factory.set.MutableSetFactory
```diff
@@ -0,0 +1 @@
+org.eclipse.collections.impl.set.mutable.MutableSetFactoryImpl
```
motlin: I forgot to highlight this earlier. I removed the usage of auto-service and just moved these META-INF/services files into src/main/resources. Not too difficult.

missing-dependency/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
+
+    </dependencies>
+</project>
```
motlin: I will revert this change, but I wanted to create a test that fails to find the eclipse-collections.jar so that we can see the error message again with the new dynamic proxies.

missing-dependency/src/test/java/MissingDependencyTest.java
```diff
@@ -0,0 +1,11 @@
+import org.eclipse.collections.api.factory.list.Lists;
+import org.junit.Test;
+
+public class MissingDependencyTest
+{
+    @Test
+    public void missingDependency()
+    {
+        Lists.mutable.empty();
```
motlin: With the dynamic proxy, the error message shows the line which calls empty(), but no longer includes the method empty().

IntelliJ collapses the error message to:
```
java.lang.IllegalStateException: Could not find any implementations of MutableListFactory. Check that eclipse-collections.jar is on the classpath and that its META-INF/services directory is intact.

	at org.eclipse.collections.api.factory.ThrowingInvocationHandler.invoke(ThrowingInvocationHandler.java:18)
	at MissingDependencyTest.missingDependency(MissingDependencyTest.java:9)
```

The full stack trace is:
```
java.lang.IllegalStateException: Could not find any implementations of MutableListFactory. Check that eclipse-collections.jar is on the classpath and that its META-INF/services directory is intact.

	at org.eclipse.collections.api.factory.ThrowingInvocationHandler.invoke(ThrowingInvocationHandler.java:18)
	at com.sun.proxy.$Proxy5.empty(Unknown Source)
	at MissingDependencyTest.missingDependency(MissingDependencyTest.java:9)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
```

mohrezaei: I think this is pretty nice.

eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Sets.java
mohrezaei: There are a couple of unrelated fixes in here. I don't mind, but if this doesn't get merged, they'll get lost.

mohrezaei:
I don't see why we wouldn't merge this in its current state :smile:

motlin:
Thanks @mohrezaei! I'll give @nikhilnanivadekar and @donraab a chance to take a look. Next I'd either remove the missingDependency test or refactor it into a real test. Plus I'll clean up the commit messages.

We ought to do the same thing for other factories (Bags, Maps, etc.) which ought to be quick, but I agree doesn't need to be in this first PR. We can look for methods that can be pulled up into api, which will probably take a while, since there are potentially so many.

mohrezaei:
Maybe a good place to start are the default methods we were about to remove.

motlin:
Sounds good. I'm looking at Bag.selectUnique(), and here's my feeling about
it.

* We can delete the implementation in AbstractMutableBag, and add a default
to MutableBag
* We can delete the implementation in AbstractImmutableBag, and add a
default to ImmutableBag
* I think Bag.selectUnique() should be changed to be abstract. It's unusual
for interfaces above mutable/immutable to have a specific implementation,
though there are exceptions. Usually newer methods like countBy().

mohrezaei:
Looking at the core implementation of the ```selectUnique()``` in ```AbstractBag```, it seems to me like all of the cross-converting methods (bag to set, set to list, etc), create a mutable instance, have some logic to populate the instance and then sort/immutize/synchronize the result (btw, ```SynchronizedBag``` seems to violate that).
Wouldn't it make sense to make that the default behavior (pun intended): the top level interface always cross converts into a mutable instance. If a subclass wants to decorate that, great. 
In ```Bag```:
```java
    default SetIterable<T> selectUnique()
    {
        MutableSet<T> result = Sets.mutable.empty();
        this.forEachWithOccurrences((each, occurrences) ->
        {
            if (occurrences == 1)
            {
                result.add(each);
            }
        });
        return result;
    }
```
And in ```AbstactImmutableBag```, we would have:
```java 
    @Override
    public ImmutableSetIterable<T> selectUnique()
    {
        return Bag.super.selectUnique().toImmutable();
    }
```

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
nikhilnanivadekar: If you wouldn't mind, I can selectively pick these and create a different PR. Like @mohrezaei mentioned there are few unrelated fixes and we might lose them during rebasing, editing. If you have it cleanly segregated in your working tree then let us get these unrelated one merged.

nikhilnanivadekar:
This is what I would like to propose in terms of the steps to merge this. There is another build related PR, which has been stalled to be merged before creating a milestone.
Let me merge that, and kick off a milestone. So that we know all the build sanitation exercises are not creating unwanted effect on Jenkins. After 10.0.0-M1 is pushed let us have this one merged so that we can create another milestone and have it tested via the Kata, and some other test projects. We can presumably ask some of our OSGi users to verify it in their projects.

PR #586, opened 2018-07-31
Delete default implementations that were added for backwards compatibility.

motlin:


nikhilnanivadekar:
I think there should be more than these. I will try to find the other implementations and post here.

mohrezaei:
We're already discussing a much better way to handle this over at #585. I'd suggest we try to concentrate on that instead to reduce wasted effort (whether we merge #585 or not).

PR #589, opened 2018-08-02
Change IntelliJ copyright plugin setting to format header with the message "Goldman Sachs and others" to make it more convenient for external contributors.

motlin:
The idea here is to flip the default way that the copyright plugin is configured. At this point, most contributors are outside GS, so the default copyright header probably ought to be "Goldman Sachs and others". Those contributing from within GS can still change the configuration of this plugin, but they'll have to always decide between the two strings on a file by file basis anyway.


nikhilnanivadekar:
I understand the motivation behind this PR, however, I feel that most of us use the command line and this is not really a cumbersome process to update the copyright as and when required. Often times I just update it by hand.
Do we really need this? I was thinking about going a step further and removing the copyright config altogether, so that there is no ambiguity.

motlin:
If you don't use this, you should go ahead and approve it. It's only
helpful for non-GS people like myself.

nikhilnanivadekar:
I am saying about removing the config altogether. Also I wonder if we should do similar to what projects like JaCoCo do, where in every year they update the year.

motlin:
I might be the only person who uses this, but I find it very helpful. This
change would allow more people to easily use it.

PR #590, opened 2018-08-02
Remove some IntelliJ project files from git and add them to .gitignore.

motlin:


motlin:
This change broke the TeamCity builds. I'll plan on partially reverting it.

nikhilnanivadekar:
I think we should get help for it, so that we know what is the correct and expected gitignore.

PR #594, opened 2018-08-03
Implement UnmodifiableMutableOrderedMap.

motlin:


nikhilnanivadekar:
1) Please update the copyrights.
2) Secondly, I always miss/forget this, so please remind me, don't we want our Unmodifiable data structures to return unmodifiable on operation? For Example: select() should return an UnmodifiableMap?

motlin:
Unmodifiable wrappers are interesting in that they already violate their
own interface, in that they aren't mutable. If we changed the methods like
select() to return unmodifiable, but mutable collections, that would be a
much more confusing violation of the api, to the point where you might as
well just call the thing immutable. Anyway, we're very consistent.
Unmodifiable and Synchronized wrappers have methods that never return
wrappers. Only their views (like keySet) are wrapped the same way.

nikhilnanivadekar:
ok makes sense! thank you!

nikhilnanivadekar:
@motlin may be there should be a blog about, how and why the decision was made. It is interesting problem to have and solve.

motlin:
done

nikhilnanivadekar:
Merging this even though TC builds are failing as an active branch is not created on TC since this is an old PR (I think).

PR #596, opened 2018-08-08
Change Travis job which only compiles to not include jmh-tests, since it gets compiled in other builds.

motlin:


nikhilnanivadekar:
a similar request was merged once no? Or I am missing something?

motlin:
I think I had this change in one of the bigger PRs that I closed and split into pieces.

nikhilnanivadekar:
@motlin Can you rebase this? Either of us can merge after that.

PR #597, opened 2018-08-08
 Line-wrap ternary statements.

motlin:


nikhilnanivadekar:
is there a checkstyle we can have to go with this to enforce the wrapping?

motlin:
CheckStyle will enforce that if you line-wrap, that the `?` and the `:` appear at the beginning of a line. But we're not running any rule which enforces a line length, or prioritizes what to wrap (between method chaining, parameters in a list, etc.) when wrapping is necessary.

nikhilnanivadekar:
So, will we rely on code reviews in the future to ensure consistency?

nikhilnanivadekar:
are we merging this one? Since there are no inspections or codestyle checks, I would rather keep the code as is. But if we can have automated checks to ensure consistency I am up for merging this PR.

motlin:
CheckStyle and Inspections cannot enforce this. But the style matches what
ctrl alt L would do.

nikhilnanivadekar:
ok fair enough.

PR #602, opened 2018-08-27
Override replaceAll and sort on List implementations.

donraab:

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
@@ -168,13 +168,24 @@ public interface MutableList<T>
 
     /**
      * Sorts the internal data structure of this list and returns the list itself as a convenience.
+     *
+     * @since 10.0 - Added default implementation.
      */
-    MutableList<T> sortThis(Comparator<? super T> comparator);
+    default MutableList<T> sortThis(Comparator<? super T> comparator)
+    {
+        this.sort(comparator);
+        return this;
+    }
 
     /**
      * Sorts the internal data structure of this list and returns the list itself as a convenience.
+     *
+     * @since 10.0 - Added default implementation.
      */
-    MutableList<T> sortThis();
+    default MutableList<T> sortThis()
+    {
+        return this.sortThis(null);
+    }
 
     /**
      * Sorts the internal data structure of this list based on the natural order of the attribute returned by {@code
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/block/factory/Comparators.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/block/factory/Comparators.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Goldman Sachs.
+ * Copyright (c) 2018 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -29,6 +29,7 @@ import org.eclipse.collections.impl.block.comparator.FunctionComparator;
 
 public final class Comparators
 {
+    private static final SerializableComparator<?> COMPARABLE_COMPARATOR = new ComparableComparator();
     private static final SerializableComparator<?> NATURAL_ORDER_COMPARATOR = new NaturalOrderComparator();
     private static final SerializableComparator<?> REVERSE_NATURAL_ORDER_COMPARATOR = new ReverseComparator(NATURAL_ORDER_COMPARATOR);
     private static final SerializableComparator<?> POWER_SET_COMPARATOR = new PowerSetComparator();
@@ -38,6 +39,22 @@ public final class Comparators
     private Comparators()
     {
         throw new AssertionError("Suppress default constructor for noninstantiability");
+    }
+
+    /**
+     * @since 10.0
+     */
+    public static <T> SerializableComparator<T> comparableComparator()
+    {
+        return (SerializableComparator<T>) COMPARABLE_COMPARATOR;
+    }
+
+    /**
+     * @since 10.0
+     */
+    public static <T> Comparator<? super T> comparableComparatorIfNull(Comparator<? super T> comparator)
+    {
+        return comparator == null ? Comparators.comparableComparator() : comparator;
     }
 
     /**
@@ -442,6 +459,17 @@ public final class Comparators
         public int compare(Pair<?, T> p1, Pair<?, T> p2)
         {
             return this.comparator.compare(p1.getTwo(), p2.getTwo());
+        }
+    }
+
+    private static class ComparableComparator<T> implements SerializableComparator<T>
+    {
+        private static final long serialVersionUID = 1L;
+
+        @Override
+        public int compare(T one, T two)
+        {
+            return ((Comparable<T>) one).compareTo(two);
         }
     }
 }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/AbstractArrayAdapter.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/AbstractArrayAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Goldman Sachs.
+ * Copyright (c) 2018 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -13,11 +13,13 @@ package org.eclipse.collections.impl.list.fixed;
 import java.lang.reflect.Array;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Optional;
 import java.util.RandomAccess;
+import java.util.function.UnaryOperator;
 
 import org.eclipse.collections.api.block.HashingStrategy;
 import org.eclipse.collections.api.block.function.Function;
@@ -353,6 +355,24 @@ public abstract class AbstractArrayAdapter<T>
     public boolean retainAllIterable(Iterable<?> iterable)
     {
         throw new UnsupportedOperationException("Cannot call retainAllIterable() on " + this.getClass().getSimpleName());
+    }
+
+    /**
+     * @since 10.0 - Overridden for efficiency
+     */
+    @Override
+    public void replaceAll(UnaryOperator<T> operator)
+    {
+        InternalArrayIterate.replaceAll(this.items, this.items.length, operator);
+    }
+
+    /**
+     * @since 10.0
+     */
+    @Override
+    public void sort(Comparator<? super T> comparator)
+    {
+        Arrays.sort(this.items, 0, this.size(), comparator);
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/AbstractMemoryEfficientMutableList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/AbstractMemoryEfficientMutableList.java
@@ -11,6 +11,8 @@
 package org.eclipse.collections.impl.list.fixed;
 
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.ListIterator;
 import java.util.RandomAccess;
 
@@ -19,6 +21,7 @@ import org.eclipse.collections.api.block.predicate.Predicate2;
 import org.eclipse.collections.api.block.procedure.Procedure;
 import org.eclipse.collections.api.list.FixedSizeList;
 import org.eclipse.collections.api.list.MutableList;
+import org.eclipse.collections.impl.block.factory.Comparators;
 import org.eclipse.collections.impl.factory.Lists;
 import org.eclipse.collections.impl.list.mutable.AbstractMutableList;
 import org.eclipse.collections.impl.utility.Iterate;
@@ -122,6 +125,33 @@ public abstract class AbstractMemoryEfficientMutableList<T>
     public void clear()
     {
         throw new UnsupportedOperationException("Cannot clear a fixed size list: " + this.getClass());
+    }
+
+    /**
+     * This method checks if comparator is null and use a ComparableComparator if it is.
+     *
+     * @since 10.0
+     */
+    @Override
+    public void sort(Comparator<? super T> comparator)
+    {
+        this.insertionSort(Comparators.comparableComparatorIfNull(comparator));
+    }
+
+    private void insertionSort(Comparator<? super T> comparator)
+    {
+        for (int i = 0; i < this.size(); i++)
+        {
+            for (int j = i; this.isPreviousGreaterThanCurrent(comparator, j); j--)
+            {
+                Collections.swap(this, j - 1, j);
+            }
+        }
+    }
+
+    private boolean isPreviousGreaterThanCurrent(Comparator<? super T> comparator, int index)
+    {
+        return index > 0 && comparator.compare(this.get(index - 1), this.get(index)) > 0;
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/ArrayAdapter.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/ArrayAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Goldman Sachs.
+ * Copyright (c) 2018 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -14,7 +14,6 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.Spliterator;
@@ -183,11 +182,7 @@ public final class ArrayAdapter<T>
     @Override
     public ArrayAdapter<T> sortThis(Comparator<? super T> comparator)
     {
-        if (this.size() > 1)
-        {
-            Arrays.sort(this.items, 0, this.size(), comparator);
-        }
-        return this;
+        return (ArrayAdapter<T>) super.sortThis(comparator);
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/QuintupletonList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/QuintupletonList.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Goldman Sachs.
+ * Copyright (c) 2018 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -14,6 +14,7 @@ import java.io.Externalizable;
 import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
+import java.util.function.UnaryOperator;
 
 import org.eclipse.collections.api.block.procedure.Procedure;
 import org.eclipse.collections.api.block.procedure.Procedure2;
@@ -131,6 +132,19 @@ final class QuintupletonList<T>
             default:
                 throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + this.size());
         }
+    }
+
+    /**
+     * @since 10.0 - Overridden for efficiency
+     */
+    @Override
+    public void replaceAll(UnaryOperator<T> operator)
+    {
+        this.element1 = operator.apply(this.element1);
+        this.element2 = operator.apply(this.element2);
+        this.element3 = operator.apply(this.element3);
+        this.element4 = operator.apply(this.element4);
+        this.element5 = operator.apply(this.element5);
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableList.java
@@ -19,6 +19,7 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.RandomAccess;
 import java.util.concurrent.ExecutorService;
+import java.util.function.UnaryOperator;
 
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.block.HashingStrategy;
@@ -664,6 +665,24 @@ abstract class AbstractImmutableList<T>
     public T remove(int index)
     {
         throw new UnsupportedOperationException("Cannot call remove() on " + this.getClass().getSimpleName());
+    }
+
+    /**
+     * @since 10.0 - Override for correctness
+     */
+    @Override
+    public void replaceAll(UnaryOperator<T> operator)
+    {
+        throw new UnsupportedOperationException("Cannot call replaceAll() on " + this.getClass().getSimpleName());
+    }
+
+    /**
+     * @since 10.0 - Override for correctness
+     */
+    @Override
+    public void sort(Comparator<? super T> c)
+    {
+        throw new UnsupportedOperationException("Cannot call sort() on " + this.getClass().getSimpleName());
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 Goldman Sachs and others.
+ * Copyright (c) 2018 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -606,63 +606,6 @@ public abstract class AbstractMutableList<T>
     public MutableList<T> asSynchronized()
     {
         return SynchronizedMutableList.of(this);
-    }
-
-    @Override
-    public MutableList<T> sortThis(Comparator<? super T> comparator)
-    {
-        if (this.size() < 10)
-        {
-            if (comparator == null)
-            {
-                this.insertionSort();
-            }
-            else
-            {
-                this.insertionSort(comparator);
-            }
-        }
-        else
-        {
-            this.defaultSort(comparator);
-        }
-        return this;
-    }
-
-    /**
-     * Override in subclasses where it can be optimized.
-     */
-    protected void defaultSort(Comparator<? super T> comparator)
-    {
-        Collections.sort(this, comparator);
-    }
-
-    private void insertionSort(Comparator<? super T> comparator)
-    {
-        for (int i = 0; i < this.size(); i++)
-        {
-            for (int j = i; j > 0 && comparator.compare(this.get(j - 1), this.get(j)) > 0; j--)
-            {
-                Collections.swap(this, j, j - 1);
-            }
-        }
-    }
-
-    private void insertionSort()
-    {
-        for (int i = 0; i < this.size(); i++)
-        {
-            for (int j = i; j > 0 && ((Comparable<T>) this.get(j - 1)).compareTo(this.get(j)) > 0; j--)
-            {
-                Collections.swap(this, j, j - 1);
-            }
-        }
-    }
-
-    @Override
-    public MutableList<T> sortThis()
-    {
-        return this.sortThis(Comparators.naturalOrder());
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/CompositeFastList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/CompositeFastList.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Goldman Sachs.
+ * Copyright (c) 2018 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -20,6 +20,7 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.NoSuchElementException;
 import java.util.concurrent.ExecutorService;
+import java.util.function.UnaryOperator;
 
 import org.eclipse.collections.api.block.function.Function;
 import org.eclipse.collections.api.block.function.Function2;
@@ -432,6 +433,25 @@ public final class CompositeFastList<E>
     }
 
     /**
+     * @since 10.0
+     */
+    @Override
+    public void replaceAll(UnaryOperator<E> operator)
+    {
+        this.lists.forEachWith(List::replaceAll, operator);
+    }
+
+    @Override
+    public void sort(Comparator<? super E> comparator)
+    {
+        FastList<E> list = comparator == null
+                ? (FastList<E>) this.toSortedList()
+                : (FastList<E>) this.toSortedList(comparator);
+        this.lists.clear();
+        this.lists.add(list);
+    }
+
+    /**
      * a list iterator is a problem for a composite list as going back in the order of the list is an issue,
      * as are the other methods like set() and add() (and especially, remove).
      * Convert the internal lists to one list (if not already just one list)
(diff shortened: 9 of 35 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/ArrayAdapter.java
```diff
@@ -180,6 +182,17 @@ public T set(int index, T element)
...
 
+    /**
+     * Overrides default method from List.
+     *
+     * @since 10.0
+     */
+    @Override
+    public void sort(Comparator<? super T> comparator)
+    {
+        Arrays.sort(this.items, 0, this.size(), comparator);
```
motlin: I was expecting sort() would delegate to sortThis(). I was expecting an attempt to add the implementation to MutableList. It's possible that would lead to diamond inheritance problems and need to get pushed down, but it probably still makes sense to delegate from one to the other. Thoughts? In this case, the implementation of sort() and sortThis() in ArrayAdapter are different and probably shouldn't be. Either they should both get the optimization for small sizes, or neither.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/ArrayAdapter.java
```diff
@@ -205,6 +218,15 @@ public T set(int index, T element)
         return result;
     }
 
+    /**
+     * @since 10.0
+     */
+    @Override
+    public void replaceAll(UnaryOperator<T> operator)
+    {
+        InternalArrayIterate.replaceAll(this.items, this.size(), operator);
```
motlin: It seems unlikely that this is the right place for this implementation, simply because you had to add an import for InternalArrayIterate. This should be pulled up to whichever superclass already imports InternalArrayIterate.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/ArrayAdapter.java
motlin: Alternately, sortThis could delegate to sort(). Either way, it should appear that there are no new implementations of sorting.

unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableListTestCase.java
```diff
@@ -429,28 +430,40 @@ public void toSortedListBy()
...
-    @Test
+    @Test(expected = UnsupportedOperationException.class)
     public void set()
     {
-        Verify.assertThrows(UnsupportedOperationException.class, () -> this.classUnderTest().castToList().set(0, 1));
+        this.classUnderTest().castToList().set(0, 1);
     }
 
-    @Test
+    @Test(expected = UnsupportedOperationException.class)
```
motlin: I would prefer to keep using Verify.assertThrows() in favor of @Test(expected = class). It works better with multiple assertions in a method, and it works better when delegating to super.sameTestMethod().

unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableListTestCase.java
```diff
@@ -429,28 +430,40 @@ public void toSortedListBy()
...
-        Verify.assertThrows(UnsupportedOperationException.class, () -> this.classUnderTest().castToList().addAll(0, Lists.fixedSize.of()));
+        this.classUnderTest().castToList().addAll(0, Lists.fixedSize.of());
+    }
+
+    @Test(expected = UnsupportedOperationException.class)
+    public void sort()
+    {
+        this.classUnderTest().castToList().sort(Comparator.naturalOrder());
+    }
+
```
motlin: When we have methods named sort() and sortThis(), as we probably do in this JUnit test, we should probably keep them together.

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/AbstractListTestCase.java
```diff
@@ -1132,4 +1134,20 @@ public void forEachWithIndex()
...
+        Assert.assertEquals(Lists.mutable.with(2, 4, 6, 8), integers);
+    }
+
+    @Test
+    public void sort()
+    {
+        MutableList<Integer> integers = this.newWith(1, 2, 3, 4);
+        integers.sort(Comparator.reverseOrder());
+        Assert.assertEquals(Lists.mutable.with(4, 3, 2, 1), integers);
+    }
```
motlin: I think the thread-safety tests might be missing.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/ArrayAdapter.java
donraab: I tried this initially, but ran into issues with subList().sort().  This method caused StackOverflows errors.

```
protected void defaultSort(Comparator<? super T> comparator)
{
    Collections.sort(this, comparator);
}
```
I just tried this approach again, and remembered why I opted to go with explicit implementations.  I'll see if there is something to do that will fix this.  Help appreciated.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/ArrayAdapter.java
donraab: Fair point.  I moved it up to the abstract class.

donraab: Strike the last comment.  Moving this up and removing the method causes duplication of the code in FastList as well.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/ArrayAdapter.java
donraab: I'm going to try and implement sort and delegate to sort from sortThis.  I think this conceptually makes the most sense.

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/AbstractListTestCase.java
nikhilnanivadekar: Do the thread-safety tests need to be in Scala?

motlin:
Historically they were but they don’t need to be.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
```diff
@@ -262,18 +263,33 @@ public int getBatchCount(int batchSize)
...
+
+    /**
+     * @since 10.0
+     */
     @Override
     public FastList<T> sortThis(Comparator<? super T> comparator)
     {
-        Arrays.sort(this.items, 0, this.size, comparator);
-        return this;
+        return (FastList<T>) super.sortThis(comparator);
```
nikhilnanivadekar: these overrides are because we return FastList. Should we break compatibility and replace it with a MutableList? It will be safe at runtime if users cast. We can then replace it by the default method.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/ListAdapter.java
```diff
@@ -226,17 +224,31 @@ public void forEach(int fromIndex, int toIndex, Procedure<? super T> procedure)
...
+
+    /**
+     * @since 10.0
+     */
     @Override
     public ListAdapter<T> sortThis(Comparator<? super T> comparator)
     {
-        Iterate.sortThis(this.delegate, comparator);
-        return this;
+        return (ListAdapter<T>) super.sortThis(comparator);
```
nikhilnanivadekar: Same comment as FastList.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/RandomAccessListAdapter.java
```diff
@@ -252,17 +250,22 @@ public void forEach(int fromIndex, int toIndex, Procedure<? super T> procedure)
...
-        Iterate.sortThis(this.delegate, comparator);
-        return this;
+        return (RandomAccessListAdapter<T>) super.sortThis(comparator);
     }
 
     @Override
     public RandomAccessListAdapter<T> sortThis()
     {
-        return this.sortThis(Comparators.naturalOrder());
+        return (RandomAccessListAdapter<T>) super.sortThis();
```
nikhilnanivadekar: Same comment as FastList

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ArrayListIterate.java
```diff
@@ -1658,15 +1656,7 @@ private ArrayListIterate()
      */
     public static <T> ArrayList<T> sortThis(ArrayList<T> list, Comparator<? super T> comparator)
```
nikhilnanivadekar: seems like we do not have replaceAll() on ArrayListIterate. Do we want to add it as a separate commit?

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/UnmodifiableMutableListTest.java
```diff
@@ -205,6 +206,18 @@ public void removeFromIndex()
         Verify.assertThrows(UnsupportedOperationException.class, () -> this.unmodifiableList.remove(0));
     }
 
+    @Test
+    public void replaceAll()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.unmodifiableList.replaceAll(e -> e));
```
nikhilnanivadekar: I prefer having Verify.assertThrows() but if rest of the tests are @ throws, then these ones should also be consistent.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
```diff
@@ -168,13 +168,23 @@
...
+    }
 
     /**
      * Sorts the internal data structure of this list and returns the list itself as a convenience.
+     * @since 10.0
      */
-    MutableList<T> sortThis();
+    default MutableList<T> sortThis()
+    {
+        this.sort(null);
```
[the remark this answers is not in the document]
donraab: I'm not sure it's necessary.  I believe this produces the same result.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
```diff
@@ -168,13 +168,23 @@
 
     /**
      * Sorts the internal data structure of this list and returns the list itself as a convenience.
+     * @since 10.0
```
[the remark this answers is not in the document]
donraab: I wasn't sure at first but decided to add it since I added the default implementation that wasn't there before.

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/AbstractListTestCase.java
donraab: We should eventually convert them all to Java.  I added them here to be consistent for now as we don't have an alternative yet.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
donraab: I think it's important not to break backwards compatibility here.  I was torn on whether to use the cast or reimplement the method as it is simple enough to do, but I didn't want to duplicate code, and the cast is safe here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/ListAdapter.java
donraab: See above.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/RandomAccessListAdapter.java
donraab: See above.

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ArrayListIterate.java
donraab: What would replaceAll do on ArrayListIterate?  Delegate to ArrayList?

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/UnmodifiableMutableListTest.java
donraab: Not sure what you want me to do here.  Is this inconsistent in this class or with other classes?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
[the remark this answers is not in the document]
nikhilnanivadekar: Yes, just verified

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
nikhilnanivadekar: 👍

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ArrayListIterate.java
nikhilnanivadekar: Sorry, I thought it was ArrayListAdapter. My bad.

nikhilnanivadekar:
This looks good to me. @motlin can you please review? Since this is fairly central change more eyes the better. Thanks!

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/QuintupletonList.java
```diff
@@ -133,6 +136,34 @@ public T set(int index, T element)
...
+    @Override
+    public void sort(Comparator<? super T> comparator)
+    {
+        T[] array = (T[]) new Object[]{this.element1, this.element2, this.element3, this.element4, this.element5};
+        Arrays.sort(array, comparator);
+        this.element1 = array[0];
+        this.element2 = array[1];
+        this.element3 = array[2];
+        this.element4 = array[3];
+        this.element5 = array[4];
```
motlin: Is this significantly different than the default implementation?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableList.java
```diff
@@ -666,6 +667,24 @@ public T remove(int index)
...
+
+    /**
+     * @since 10.0
+     */
+    @Override
+    public void sort(Comparator<? super T> c)
+    {
+        throw new UnsupportedOperationException("Cannot call sort() on " + this.getClass().getSimpleName());
+    }
+
```
motlin: Where is sortThis()? I was expecting it would already exist in this class. I'm sure it always threw UnsupportedOperationException, but was it throwing an unhelpful message like "Cannot call set()" instead of "Cannot call sortThis()"?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
```diff
@@ -608,63 +608,6 @@ public long injectInto(long injectedValue, LongObjectToLongFunction<? super T> f
...
-            }
-        }
-    }
-
-    @Override
-    public MutableList<T> sortThis()
-    {
-        return this.sortThis(Comparators.naturalOrder());
-    }
-
```
motlin: Did this stuff move?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/CompositeFastList.java
```diff
@@ -430,6 +431,25 @@ public int lastIndexOf(Object o)
         return -1;
     }
 
+    /**
+     * @since 10.0
+     */
+    @Override
+    public void replaceAll(UnaryOperator<E> operator)
+    {
+        this.lists.forEachWith(FastList::replaceAll, operator);
```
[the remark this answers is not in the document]
motlin: +1. I would expect `List::replaceAll`

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
```diff
@@ -262,18 +263,33 @@ public int getBatchCount(int batchSize)
...
+     * @since 10.0
+     */
+    @Override
+    public void sort(Comparator<? super T> comparator)
+    {
+        if (this.size > 1)
+        {
+            Arrays.sort(this.items, 0, this.size, comparator);
+        }
+    }
```
motlin: It's not clear to me where this was moved from. The superclass doesn't know it can delegate to the backing array, right? What was the old behavior?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
```diff
@@ -917,6 +918,40 @@ public int lastIndexOf(Object o)
...
+        this.acquireWriteLock();
+        try
+        {
+            this.delegate.sort(comparator);
+        }
+        finally
+        {
+            this.unlockWriteLock();
+        }
+    }
```
motlin: Is sortThis() overridden? I was expecting sort() to be the method before sortThis().

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/UnmodifiableMutableList.java
```diff
@@ -221,6 +222,24 @@ public void forEachWithIndex(int fromIndex, int toIndex, ObjectIntProcedure<? su
...
+
+    /**
+     * @since 10.0
+     */
+    @Override
+    public void sort(Comparator<? super T> comparator)
+    {
+        throw new UnsupportedOperationException("Cannot call sort() on " + this.getClass().getSimpleName());
+    }
+
```
motlin: Is sortThis() overridden?

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ArrayListIterate.java
```diff
@@ -1658,15 +1656,7 @@ private ArrayListIterate()
...
-        int size = list.size();
-        if (ArrayListIterate.canAccessInternalArray(list))
-        {
-            Arrays.sort(ArrayListIterate.getInternalArray(list), 0, size, comparator);
-        }
-        else
-        {
-            Collections.sort(list, comparator);
-        }
+        list.sort(comparator);
```
motlin: It's already efficient?

scala-unit-tests/src/test/scala/org/eclipse/collections/impl/list/mutable/SynchronizedMutableListScalaTest.scala
```diff
@@ -28,6 +28,15 @@ class SynchronizedMutableListScalaTest extends SynchronizedMutableCollectionTest
...
 
+    @Test
+    def sort_synchronized
+    {
+        this.assertSynchronized
+        {
+            this.classUnderTest.sort(null)
+        }
+    }
+
```
motlin: Is sortThis tested in the same class?

unit-tests/src/test/java/org/eclipse/collections/impl/list/fixed/AbstractMemoryEfficientMutableListTestCase.java
```diff
@@ -61,6 +62,31 @@ public void testGetClass()
...
+        this.list.shuffleThis();
+        this.list.sort(null);
+        Assert.assertEquals(
+                this.getNStrings(),
+                this.list);
+    }
+
     @Test
     public void sortThis()
     {
```
motlin: Is sortThis() missing coverage for the override that takes a comparator?

unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableListTestCase.java
motlin: Still relevant I think.

unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableListTestCase.java
motlin: Still relevant I think.

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/UnmodifiableMutableListTest.java
motlin: Other classes should also use assertThrows

motlin:
We've also had trouble in the past with coverage tools when one test
overrides another and one has an expected exception and the other does not.
I don't think JUnit 5 uses the expected mechanism either.

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/UnmodifiableMutableListTest.java
donraab: Please feel free to fix those other classes.

unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableListTestCase.java
donraab: Fixed.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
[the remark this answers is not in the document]
donraab: The default implementation was added in 10.0.  Please suggest a longer description you think would be acceptable and I will add it.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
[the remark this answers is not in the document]
donraab: Good point.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/AbstractArrayAdapter.java
```diff
@@ -355,6 +357,29 @@ public boolean retainAllIterable(Iterable<?> iterable)
         throw new UnsupportedOperationException("Cannot call retainAllIterable() on " + this.getClass().getSimpleName());
     }
 
+    /**
+     * @since 10.0
```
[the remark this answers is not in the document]
donraab: Fixed.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/CompositeFastList.java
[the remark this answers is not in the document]
donraab: ```private final FastList<FastList<E>> lists = FastList.newList();```

unit-tests/src/test/java/org/eclipse/collections/impl/list/fixed/AbstractMemoryEfficientMutableListTestCase.java
donraab: I'm looking at overall coverage for this entire hierarchy now.  Stay tuned.

nikhilnanivadekar:
This I think is ready to merge, unless @motlin has any concerns.

nikhilnanivadekar:
Please do not merge this till M2 is done.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableList.java
motlin: Do our tests check the message, or just the exception type?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
motlin: Why is this stuff gone?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/CompositeFastList.java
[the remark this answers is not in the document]
motlin: Do you get a compiler error when you change it to `List::replaceAll`?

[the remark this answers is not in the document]
donraab: I don't get a compiler error now.  This is not a CompositeList.  It is a CompositeFastList.  I can probably replace with a higher interface.  I can keep it with the concrete implementation as I have it.  I'm not convinced there is a benefit either way, so I'm not compelled to change it.  What do you see as the benefit of the change?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableList.java
donraab: I have to admit to not understanding this question at first.  The method `sortThis()` does not exist in `AbstractImmutableList` because it does not extend `MutableList`.  As you know, our `ImmutableList` implementations have to implement `List` and `sort` is a mutating method on `List` so it should throw `UnsupportedOperationException`.  It was previously throwing an unhelpful message as you point out which is why I provided an override.

donraab: The tests just check the exception type, not the message.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
donraab: I removed it.  There have been improvements to the `sort` implementation in the JDK with the addition of `TimSort`.  I don't think this code pulls its weight since we started compiling with Java 8.  Now that I am overriding sort in all of the `AbstractMemoryEfficientMutableList` subclasses, this code is even less beneficial.

If you think it is a mistake this remove this optimization, let's discuss how we should add it back.  It should probably move from `sortThis` to `sort`.

donraab: I'm looking to move this code down to `AbstractMemoryEfficientMutableList` removing the size check and deleting all of the current specialized implementations in the subclasses.  Make sense?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/AbstractMemoryEfficientMutableList.java
```diff
@@ -124,6 +127,33 @@ public void clear()
...
 
+    /**
+     * This method checks if comparator is null and use a ComparableComparator if it is.
+     *
+     * @since 10.0
+     */
+    @Override
+    public void sort(Comparator<? super T> comparator)
+    {
+        this.insertionSort(Comparators.comparableComparatorIfNull(comparator));
```
motlin: Do you have a JMH test for this? The extra virtual method calls from using a comparator may hurt performance.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableList.java
motlin: Right, I forgot that the interfaces are different here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/AbstractMutableList.java
motlin: Not trying to encourage a particular optimization, quite the opposite. I find this stuff hard to review. The most important issue is that without overriding sort() and replaceAll(), behavior is wrong on thread-safe collections. So I came into this expecting additional overrides.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
```diff
@@ -262,18 +263,33 @@ public int getBatchCount(int batchSize)
...
 
+    /**
+     * Overrides default method from List.
+     *
+     * @since 10.0
+     */
+    @Override
+    public void sort(Comparator<? super T> comparator)
+    {
+        if (this.size > 1)
```
motlin: ComparableTimSort already has this optimization.

```java
        int nRemaining  = hi - lo;
        if (nRemaining < 2)
            return;  // Arrays of size 0 and 1 are always sorted
```

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
```diff
@@ -1419,6 +1454,15 @@ public void forEachWithIndex(int fromIndex, int toIndex, ObjectIntProcedure<? su
...
+         * @since 10.0
+         */
+        @Override
+        public void sort(Comparator<? super T> comparator)
+        {
+            this.getDelegate().sort(comparator);
+        }
+
         @Override
         public MutableList<T> sortThis()
```
motlin: I see sort and sortThis together in the inner class. They should be together in the outer class too.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/SynchronizedMutableList.java
```diff
@@ -216,6 +217,30 @@ public T get(int index)
...
+     */
+    @Override
+    public void sort(Comparator<? super T> comparator)
+    {
+        synchronized (this.getLock())
+        {
+            this.getDelegate().sort(comparator);
+        }
+    }
+
```
motlin: Keep sort() with sortThis() on line 440?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/UnmodifiableMutableList.java
motlin: Same here line 183

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ArrayListIterate.java
motlin: Should this be marked deprecated and inlineable?

unit-tests/src/test/java/org/eclipse/collections/impl/list/fixed/AbstractMemoryEfficientMutableListTestCase.java
```diff
@@ -227,8 +266,116 @@ public void aggregateByNonMutating()
...
         Function<String, String> groupBy = Functions.getStringPassThru();
         Function2<Integer, String, Integer> sumAggregator = (aggregate, value) -> aggregate + Integer.parseInt(value);
-        MapIterable<String, Integer> actual = this.classUnderTest().aggregateBy(groupBy, () -> 0, sumAggregator);
-        MapIterable<String, Integer> expected = FastList.newList(this.classUnderTest()).aggregateBy(groupBy, () -> 0, sumAggregator);
+        MapIterable<String, Integer> actual =
+                this.classUnderTest().aggregateBy(groupBy, () -> 0, sumAggregator);
+        MapIterable<String, Integer> expected =
+                FastList.newList(this.classUnderTest()).aggregateBy(groupBy, () -> 0, sumAggregator);
```
motlin: A few unrelated changes here?

donraab: Yes.  A general rule I like to apply.  Don't make me scroll.

scala-unit-tests/src/test/scala/org/eclipse/collections/impl/list/mutable/SynchronizedMutableListScalaTest.scala
donraab: Good point.  Added tests for both forms of sortThis.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/fixed/AbstractMemoryEfficientMutableList.java
donraab: No. This code is only for FixedSizedLists of size 2 to 6.  I can try and measure it if you think it is important.

eclipse-collections/src/main/java/org/eclipse/collections/impl/block/factory/Comparators.java
```diff
@@ -444,4 +461,15 @@ public int compare(Pair<?, T> p1, Pair<?, T> p2)
...
+    private static class ComparableComparator<T> implements SerializableComparator<T>
+    {
+        private static final long serialVersionUID = 1L;
+
+        @Override
+        public int compare(T one, T two)
+        {
+            return ((Comparable<T>) one).compareTo(two);
+        }
+    }
```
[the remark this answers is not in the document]
donraab: There are actually more.  I went searching to see if one existed already.  There is a private ComparableComparator I found in `javax.swing.table.TableRowSorter`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
donraab: Nice.  I will remove the optimization then.

PR #604, opened 2018-08-30
Implement ObjectPrimitiveMap.from(Iterable, KeyFunction, ValueFunction), PrimitiveObjectMap.from(Iterable, KeyFunction, ValueFunction), PrimitivePrimitiveMap.from(Iterable, KeyFunction, ValueFunction). Fixes #547

### LINKED ISSUE #547: Implement ObjectPrimitiveMap.from(Iterable, KeyFunction, ValueFunction), PrimitiveObjectMap.from(Iterable, KeyFunction, ValueFunction), PrimitivePrimitiveMap.from(Iterable, KeyFunction, ValueFunction).
This can be a utility on Iterate to make it easy to create Primitive Maps or even on the Factory classes.
Currently, there is no easy way to create a PrimitiveMap from an Iterable/RichIterable.

nikhilnanivadekar:


nikhilnanivadekar:
Motivation for this feature is to make it easy for users to construct primitive maps from an Iterable<T>. To construct a map or get a mapping of some sort, we do not always have the key-value pairs with us. On RichIterable we have `toMap()` API which serves this exact purpose and allows us to create the required map. We need a convenience API on primitive maps to create the required map. However, due to the interlaced dependency of primitive and non-primitive side, it made sense to add it as a static factory method. This will enable the users to directly create a primitive map instead of initializing the map, writing a forEach() loop and populating the map.
In my experience I have had to populate similar maps at multiple places, also a couple of users have approached me offline for the convenience methods.

nikhilnanivadekar:
@motlin or @donraab can either of you please review. I would like this to be merged. Separately I am investigating the Java-11 build failures

eclipse-collections-code-generator/src/main/resources/test/map/immutableObjectPrimitiveMapFactoryImplTest.stg
```diff
@@ -1,52 +0,0 @@
...
-
-import org.eclipse.collections.impl.factory.primitive.Object<name>Maps;
-import org.eclipse.collections.impl.map.mutable.primitive.Object<name>HashMap;
-import org.junit.Assert;
-import org.junit.Test;
-
-/**
- * This file was automatically generated from template file immutableObjectPrimitiveMapFactoryImplTest.stg.
- */
-public class ImmutableObject<name>MapFactoryImplTest
```
motlin: Why delete these tests?

motlin: I think they are just renamed? Is there any way to get the diff to look a little simpler in the review?

p2-feature/pom.xml
```diff
@@ -48,6 +48,18 @@
...
+                        <groupId>org.eclipse.tycho</groupId>
+                        <artifactId>tycho-core</artifactId>
+                        <version>${tycho-version}</version>
+                    </dependency>
+                    <dependency>
+                        <groupId>org.eclipse.tycho</groupId>
+                        <artifactId>tycho-p2-facade</artifactId>
+                        <version>${tycho-version}</version>
+                    </dependency>
+                </dependencies>
```
motlin: The tycho stuff should probably be in its own review and be merged first, before new features?

nikhilnanivadekar: I can move it out. But they are a separate commit so thought might as well merge it with this PR. Let me know what do you prefer.

eclipse-collections-code-generator/src/main/resources/test/map/immutableObjectPrimitiveMapFactoryImplTest.stg
nikhilnanivadekar: The tests for these factories were non-non-existent and lacked coverage. If you review the whole PR instead of 1 commit at a time it'll be easier. Splitting the PR will be difficult. For transparency: I have moved all the tests from previous test classes and added more tests such that the factories are now covered.

motlin:
Sounds good, regarding the tests. I do think the build stuff ought to be in
a separate PR if you don't mind.

nikhilnanivadekar:
Fair enough. Raised #608

PR #608, opened 2018-09-01
Upgrade version of p2 feature to 10.0.0, added necessary dependencies for Tycho maven plugin.

nikhilnanivadekar:


motlin:
Thank you. Merging. I'd love to know more about the missing dependencies for the plugin, how we got so far without them, how you knew to add them, etc.

PR #609, opened 2018-09-02
Change Travis builds to use available jdks when possible.

motlin:


.travis.yml
```diff
@@ -1,11 +1,6 @@
 language: java
 dist: trusty
 sudo: false
-addons:
-  apt:
-    packages:
-      - oracle-java8-installer
-      - oracle-java9-installer
```
motlin: I believe these are not required anymore. Java 8 and 9 are just stable, included JDKs.

.travis.yml
```diff
@@ -17,85 +12,76 @@ before_install:
 install:
   -
 
-env:
-  -JDK=Java8
-  -JDK=Java9
-  -JDK=Java10
-  -JDK=Java11-EA
-  -JDK=Java12-EA
-
```
motlin: Typically, declaring an environment up front like this is used for an automatic build matrix, where builds run for all n environments for all m jdks. Anyway, the environment can still be declared lazily it seems, for the Java 11 and 12 EA builds.

.travis.yml
```diff
@@ -17,85 +12,76 @@ before_install:
...
         - DESC="compile performance-tests"
-        - JDK=Java8
         - CMD="mvn install --projects performance-tests --also-make --activate-profiles all -DskipTests=true"
 
     - jdk: oraclejdk9
       env:
-        - DESC="unit tests Java 9"
-        - JDK=Java9
+        - DESC="unit tests oraclejdk9"
         - CMD="mvn install"
```
motlin: Removing as many JDK declarations as possible, and changing the DESCriptions to use the Travis name of the jdk in as many places as possible, for consistency.

.travis.yml
```diff
@@ -17,85 +12,76 @@ before_install:
...
+    - jdk: openjdk10
+      env:
+      - DESC="unit tests openjdk10"
+      - CMD="mvn install"
+
+    - jdk: oraclejdk10
+      env:
+      - DESC="unit tests oraclejdk10"
+      - CMD="mvn install"
+
```
motlin: Here we probably don't need the openjdk10 build, but I just wanted to show how Travis has Java 8, 9, and 10 choices built in, with no addons or downloads required.

.travis.yml
```diff
@@ -17,85 +12,76 @@ before_install:
...
+    - jdk: openjdk11
       env:
-        - DESC="Java 10 regression"
-        - JDK=Java10
-        - CMD=""
+      - DESC="unit tests openjdk11"
+      - CMD="mvn install"
```
motlin: openjdk11 is also included by default, but I don't know if it's a great test. I can remove this one.

.travis.yml
```diff
@@ -17,85 +12,76 @@ before_install:
...
-        - CMD=""
+        - CMD="./.travis-command-ea-builds.sh"
 
     - jdk: oraclejdk8
       env:
-        - DESC="Java 12 regression"
+        - DESC="unit tests Java12-EA"
         - JDK=Java12-EA
-        - CMD=""
+        - CMD="./.travis-command-ea-builds.sh"
```
motlin: This is the crucial change that I really wanted to do. Launching an external script interferes with logging. The entire command is evaluated, and then only when its finished will anything appear in the logs. I want to launch an external script only when necessary. And when we launch an external script, it can just be the regular old CMD.

.travis.yml
```diff
@@ -17,85 +12,76 @@ before_install:
...
     env:
-      - DESC="Java 11 regression"
       - JDK=Java11-EA
-      - CMD=""
 
   - jdk: oraclejdk8
     env:
-      - DESC="Java 12 regression"
       - JDK=Java12-EA
-      - CMD=""
```
motlin: To allow failures, I believe only the jdk and JDK (environment variable) are required.

.travis.yml
```diff
@@ -17,85 +12,76 @@ before_install:
...
 
-script:
-  - ./.travis-command-ea-builds.sh
-  - eval travis_wait 55 $CMD
+script: eval travis_wait 30 $CMD
```
motlin: This may have been broken before. travis_wait is a command that needs to wrap another command. The way it was written, I think the first command would run for up to 10 minutes, and then we'd wait up to 55 for a noop.

.travis.yml
```diff
@@ -17,85 +12,79 @@ before_install:
...
+    - jdk: oraclejdk10
       env:
-        - DESC="Java 10 regression"
-        - JDK=Java10
-        - CMD=""
+        - DESC="unit tests oraclejdk10"
+        - CMD="mvn install"
+
+    - jdk: openjdk11
+      if: type = cron
```
motlin: Adding the condition `if: type = cron` means these builds will never run or even show up except on a schedule. Since we only run these builds for information and not to block commits, I think it makes sense to run them nightly. This should make builds faster and reduce confusion for contributors.

.travis.yml
nikhilnanivadekar: is oraclejdk9 needed in the DESC if it is mentioned in the jdk?

.travis.yml
nikhilnanivadekar: I disagree with this. I would rather have it run for every PR. It helps us isolate issues in the PRs. With a cron job we will definitely miss investigations

.travis.yml
nikhilnanivadekar: From the docs to allow failures the "allow_failures" block has to be exactly same as the matrix: https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail
It was broken before, but after #607 was merged you can see the build matrix changed on Travis to show the builds which allowed failures.
If you remove the cron condition, you should be able to see them in a different matrix labeled "Allowed Failures" that will tell us if this works or not.

.travis.yml
motlin: I guess not. It's just for clarity and appearance on this page:
https://travis-ci.org/eclipse/eclipse-collections

WDYT?

.travis.yml
motlin: Another option is `if: type = push AND fork = false`. I just had trouble getting it to work properly. Then it should happen on every new change, but contributors would never see it. I figured cron is good enough because our pace of change isn't that fast, so daily would catch almost everything. And my top priorities are not making contributors wait, or see failing builds they don't need to worry about.

WDYT?

.travis.yml
motlin: I can double check. I had tried it with an earlier commit, before adding the `if:` condition and I saw it work without DESC.

.travis.yml
nikhilnanivadekar: Makes sense. I thought it was some connection between the mentioned jdk and DESC. 👍

.travis.yml
nikhilnanivadekar: This is fair. 
I do need to check the policy of early adopters (since Eclipse Collections is a part of it) if they can be fine running it at a frequency desired by the project. It is up to us to decide then we can change it to a cron interval of once every day.

nikhilnanivadekar: Actually I am debating this. If there are build failures due to issues with URLs we can always explain it to the contributors. But in case there are actual issues with the code, they should be fixed. I think we should keep it for all PRs. Latest example of my own PR 😄 : https://travis-ci.org/eclipse/eclipse-collections/builds/427021454

motlin:
Your example broke java 9 through 12. Only 11 and 12 would have been
suppressed. I realize that it’s theoretically possible for only 11 or 12 to
break but I haven’t seen it happen, and even if it does happen it probably
shouldn’t bother contributors.

nikhilnanivadekar:
once 11 is released we would remove 9 and 10 because 11 is LTS and have 11, 12-EA, so on and so forth. At that time I would want 9 and 10 to run as cron probably. But IMO current and future JDK builds should run for every PR.

motlin:
Why would we remove 9 and 10?

motlin:
I believe this is now good to go.

.travis.yml
nikhilnanivadekar: 👍

.travis.yml
```diff
@@ -17,85 +12,83 @@ before_install:
...
 
     - jdk: oraclejdk9
       env:
-        - DESC="unit tests Java 9"
-        - JDK=Java9
+        - DESC="unit tests oraclejdk9"
         - CMD="mvn install"
 
-    - jdk: oraclejdk8
+    - jdk: openjdk10
```
nikhilnanivadekar: does declaring it as openjdk10 work? I do not know, asking for information.

motlin: The travis docs are a little scattered. This works:

https://docs.travis-ci.com/user/languages/java/#using-java-10-and-later

I interpret the docs to mean that these jdks are installed lazily on demand. Not that we have to invoke `install-jdk.sh` because we definitely don't.

PR #610, opened 2018-09-03
Fix violations found for Inspections3 configuration via TeamCity builds.

nikhilnanivadekar:


nikhilnanivadekar:
I am not able to trigger remote builds for my branch, so all the commits are labeled as WIP.

nikhilnanivadekar:
TC build for inspections3 is fixed: https://teamcity.jetbrains.com/viewType.html?buildTypeId=OpenSourceProjects_EclipseCollections_Inspections3ConsistentStyle&branch_OpenSourceProjects_EclipseCollections=610&tab=buildTypeStatusDiv

@motlin can you please review.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/Interval.java
```diff
@@ -712,6 +712,7 @@ public boolean equals(Object otherList)
         return !listIterator.hasNext();
     }
 
+    @SuppressWarnings("PointlessNullCheck")
     private boolean intObjectEqual(int i, Object object)
     {
         return object == null || !(object instanceof Integer) || ((Integer) object).intValue() != i;
```
motlin: The method is mis-named I guess. It's intObjectNotEqual really.

I believe the inspection is correct and should not be suppressed. null is not an instance of Integer, so the second clause makes the first clause redundant.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/CompositeFastList.java
```diff
@@ -178,7 +179,7 @@ public void reverseForEachWithIndex(ObjectIntProcedure<? super E> procedure)
     @Override
     public boolean isEmpty()
     {
-        return this.lists.allSatisfy(FastList<E>::isEmpty);
+        return this.lists.allSatisfy(AbstractRichIterable::isEmpty);
```
motlin: This one is something else, but I like it and would turn it on in an inspection profile. Could you identify where it comes from?

nikhilnanivadekar: This came from: "Instance field access not qualified with 'this'"
Also, this was not an auto-fix, I had to convert it to anonymous inner class and then switch it to a method reference. It is a good inspection to have, I will raise a bug to Jetbrains, may be they can add it.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/Interval.java
nikhilnanivadekar: yeah I figured that, but did not know if I should rename it to be more sensible and then fix the inspection. 
I can do that in this commit or have a separate commit, PR. Let me know. I have it fixed (without the suppression) in my local, will just have to push it.

nikhilnanivadekar:
All but Inspections 4 have passed. I think this one is ready

nikhilnanivadekar:
@motlin good to merge Inspections3?

PR #611, opened 2018-09-06
Fix violations found for Inspections4 configuration via TeamCity builds. 

nikhilnanivadekar:


nikhilnanivadekar:
The inspections build is fixed: https://teamcity.jetbrains.com/viewLog.html?buildId=1603026&tab=Inspection&buildTypeId=OpenSourceProjects_EclipseCollections_Inspections4NoAutoFix
Can you please review @motlin

motlin:
In TeamCity, if the inspection name doesn't have "(Errors)" next to it, then the violation is only a warning. For profile 4, the only errors are:

* Redundant local variable
* Raw use of parameterized class
* Declaration has Javadoc problems

I would like to treat "Explicit type can be replaced with <>" as an error (an increase its severity in the profile). Earlier versions of IntelliJ/TeamCity had false positives for this inspection, but I think they are fixed.

For the other three inspections, I'd recommend splitting the commits into a separate pull request, and avoiding suppressions. We could also turn off the inspections inside the profile that runs on TeamCity if we don't believe they are worth fixing or alerting on. As long as they show up inside the IDE, that's probably enough.

nikhilnanivadekar:
I am fine splitting in a separate PR. But I do not quite see the benefit of it. I can rename the commits to be called "Inspection Warning"

I have fixed as many as possible before deciding to suppress. I have suppressed if and only if it would create a change in logic, which was the case of "Assignment to method parameter". If there are any which you think are fixable let me know, happy to fix them.

I will increase the severity of explicit type so that it shows up as an error.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
```diff
@@ -51,7 +51,7 @@
 import org.eclipse.collections.impl.utility.MapIterate;
 import org.eclipse.collections.impl.utility.internal.IterableIterate;
 
-@SuppressWarnings({ "rawtypes", "ObjectEquality" })
+@SuppressWarnings({"rawtypes", "ObjectEquality"})
```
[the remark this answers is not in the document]
motlin: No it's great. This was a long-standing annoyance with CheckStyle. It's possible I fixed it with my recent changes without realizing.

unit-tests/src/test/java/org/eclipse/collections/impl/block/factory/ComparatorsTest.java
```diff
@@ -201,11 +201,11 @@ public void fromFunctionsWithTwoArgs()
     {
         Person raab = new Person("Don", "Raab", 0);
         Person white = new Person("Barry", "White", 0);
-        Person manilow = new Person("Barry", "Manilow", 0);
```
motlin: Most of the "narrow scope" fixes are great. But this is one where I felt it was more readable before than after, and I was reluctant to make this inspection a breaking one.

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/AbstractListTestCase.java
```diff
@@ -925,10 +925,9 @@ public void getWithIndexOutOfBoundsException()
     @Test
     public void getWithArrayIndexOutOfBoundsException()
     {
-        Object item = new Object();
-
         try
         {
+            Object item = new Object();
```
motlin: Here's another scope reduction that I don't agree with.

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/AggregateByTest.java
```diff
@@ -678,12 +678,12 @@ public boolean equals(Object o)
...
-            temp = Double.doubleToLongBits(this.getMin());
-            result = 31 * result + (int) (temp ^ (temp >>> 32));
-            temp = Double.doubleToLongBits(this.getMax());
-            result = 31 * result + (int) (temp ^ (temp >>> 32));
+            long temp1 = Double.doubleToLongBits(this.getSum());
+            result = 31 * result + (int) (temp1 ^ (temp1 >>> 32));
+            long temp2 = Double.doubleToLongBits(this.getMin());
+            result = 31 * result + (int) (temp2 ^ (temp2 >>> 32));
+            long temp3 = Double.doubleToLongBits(this.getMax());
+            result = 31 * result + (int) (temp3 ^ (temp3 >>> 32));
```
motlin: I like all the fixes, not sure about the suppressions.

acceptance-tests/src/test/java/org/eclipse/collections/impl/map/mutable/UnifiedMapAcceptanceTest.java
```diff
@@ -1580,6 +1580,7 @@ private void runGetTest(Map<CollidingInt, String> map, String mapName, int size)
             long runTimes = System.nanoTime() - startTime;
             LOGGER.info("{} get: {} ns per get on map size {}", mapName, (double) runTimes / (double) n / size, size);
         }
+        //noinspection AssignmentToMethodParameter
```
motlin: This one has so many suppressions, I'd rather treat the inspection as a warning or turn it off.

nikhilnanivadekar:
Totally agree with method parameter being reassigned should be a warning. It is a good warning to have while doing code reviews and all. But all the violations are just suppressed so it is pretty useless from a TC perspective.

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/AbstractListTestCase.java
nikhilnanivadekar: I can suppress it.

unit-tests/src/test/java/org/eclipse/collections/impl/block/factory/ComparatorsTest.java
nikhilnanivadekar: I can suppress it.

motlin:
They are warnings anyway, so suppressions aren't strictly necessary.

nikhilnanivadekar:
the close and open of PR was to ensure Travis webhook was not broken. Currently working on TC webhook

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/AggregateByTest.java
nikhilnanivadekar: Fixed all suppressions, but it does impact control flow, please review.

nikhilnanivadekar:
We have found a bug in the inspections setting: https://travis-ci.org/eclipse/eclipse-collections/jobs/427021460#L824
Raised a bug for IntelliJ: 
1) Inspection: https://youtrack.jetbrains.com/issue/IDEA-198681
2) Compiler Issue: https://youtrack.jetbrains.com/issue/IDEA-198682

I am reducing the severity in the light of this bug.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayList.java
```diff
@@ -501,10 +501,9 @@ public String toString()
         buf.append('[');
 
         int localSize = this.items.length;
-        T[] localItems = this.items;
         for (int i = 0; i < localSize; i++)
         {
-            T item = localItems[i];
+            T item = this.items[i];
```
motlin: If @donraab is ok with inlining these. Otherwise, I'd leave the inspection as a warning in the editor, remove it from profile 4 completely, and neither fix nor suppress instances like this.

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/AggregateByTest.java
motlin: Not sure what you mean by affecting control flow. The ones with reverse iteration?

unit-tests/src/test/java/org/eclipse/collections/impl/block/factory/HashingStrategiesTest.java
```diff
@@ -139,25 +139,25 @@ public void fromFunctionsTwoArgs()
...
-        Person john3 = new Person("John", "Doe");
-        Person john4 = new Person("John", "Smith", 10);
 
         HashingStrategy<Person> chainedHashingStrategy = HashingStrategies.fromFunctions(Person.TO_FIRST, Person.TO_LAST, Person.TO_AGE);
         Assert.assertEquals(john1.hashCode(), chainedHashingStrategy.computeHashCode(john1));
+        Person john2 = new Person("John", "Smith");
         Assert.assertTrue(chainedHashingStrategy.equals(john1, john2));
+        Person john3 = new Person("John", "Doe");
         Assert.assertFalse(chainedHashingStrategy.equals(john1, john3));
+        Person john4 = new Person("John", "Smith", 10);
```
motlin: I'd consider reverting this whole file.

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/AbstractListTestCase.java
```diff
@@ -926,7 +926,6 @@ public void getWithIndexOutOfBoundsException()
     public void getWithArrayIndexOutOfBoundsException()
     {
         Object item = new Object();
-
```
motlin: whitespace only left

unit-tests/src/test/java/org/eclipse/collections/impl/map/fixed/TripletonMapTest.java
```diff
@@ -360,9 +360,9 @@ public void testEqualsAndHashCode()
     @Test
     public void testClone()
     {
-        MutableMap<Integer, String> map = new TripletonMap<>(1, "One", 2, "Two", 3, "Three");
         try
         {
+            MutableMap<Integer, String> map = new TripletonMap<>(1, "One", 2, "Two", 3, "Three");
```
motlin: This is similar. I see no point reducing scope into the beginning of a try block.

motlin:
Looks pretty good. I had some minor comments but if you want to get parts of this committed, the commits which address inspection errors are ready to be merged. I only had comments about the warnings (which is expected, it's why they are only warnings).

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayList.java
nikhilnanivadekar: TBH I did not quite understand why we had a local copy created. Since it is immutable, I would expect the array not change, or may be I am missing something.
@donraab thoughts?

unit-tests/src/test/java/org/eclipse/collections/impl/block/factory/HashingStrategiesTest.java
nikhilnanivadekar: Reverted

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/AbstractListTestCase.java
nikhilnanivadekar: Reverted

unit-tests/src/test/java/org/eclipse/collections/impl/map/fixed/TripletonMapTest.java
nikhilnanivadekar: Reverted

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/AggregateByTest.java
nikhilnanivadekar: Yes the one with reverse iteration:
Basically we cap the iteration at 0.
https://github.com/eclipse/eclipse-collections/pull/611/commits/86ea7d1e0889036ce22c49297aa233bc875a7d9f#diff-1a2b9df9d2e5582ddda1ef2be9dba77bR332

https://github.com/eclipse/eclipse-collections/pull/611/commits/86ea7d1e0889036ce22c49297aa233bc875a7d9f#diff-f73899c981d66f3bf32157aca70650ecR391

Also, the compensation for Kahan summation algorithm: 
https://github.com/eclipse/eclipse-collections/pull/611/commits/86ea7d1e0889036ce22c49297aa233bc875a7d9f#diff-9d2319fbefc4b0d92c83470e981523bfR1397

motlin:
It is theoretically an optimization, though I doubt we’ve measured a
difference.

nikhilnanivadekar:
if this one is ok to be inlined by @donraab and @motlin let us merge and close this PR.

motlin:
If your priority is to fix the TeamCity build, then I'd recommend splitting this PR.

If you create a PR with these four commits:
* Fix inspection violation: Redundant local variable.
* Fix inspection violation: Raw use of parameterized class.
* Fix inspection violation: Explicit type can be replaced with <>.
* Fix inspection violation: Declaration has Javadoc problems.

Then I could merge those four without further discussion, and it would fix the build. The other two require discussion and are only warnings anyway.

nikhilnanivadekar:
Let us wait on @donraab till this evening if he is fine. I think the inlining is the only aspect we need to discuss. If he does not come back till then I will split it. For now <> is a warning due to the bug mentioned above.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayList.java
donraab: Write some JMH tests.  If there is no or very little performance difference, please go ahead with the change.

PR #614, opened 2018-09-11
Turn on additional CheckStyle rules and fix violations.

motlin:


nikhilnanivadekar:
LGTM. We should merge this after Inspections-4 (#611) is merged

PR #616, opened 2018-09-12

motlin:


nikhilnanivadekar:
Can we add a set of inspections for these? Inspections 4 or may be even 5?

motlin:
The redundant-type-cast inspection is unreliable. It flags code like this:

```java
    @Override
    public MutableMap<K, RichIterable<V>> toMap()
    {
        return (MutableMap<K, RichIterable<V>>) (MutableMap<?, ?>) this.map.toMap();
    }
```

nikhilnanivadekar:
@motlin please rebase and merge. Thanks!

PR #617, opened 2018-09-14
Implement MutableMultimap.getIfAbsentPut()

njcohen2:
Contributors: @tylermcdonald and @ethan-lord.

Signed-off-by: Nathaniel Cohen <njc1583@gmail.com>

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/multimap/MutableMultimap.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/multimap/MutableMultimap.java
@@ -69,6 +69,17 @@ public interface MutableMultimap<K, V>
         return changed;
     }
 
+    default Iterable<? extends V> getIfAbsentPut(K key, Iterable<? extends V> value)
+    {
+        boolean hasKey = this.containsKey(key);
+        if (hasKey)
+        {
+            this.putAll(key, value);
+            return value;
+        }
+        return this.get(key);
+    }
+
     boolean putAll(K key, Iterable<? extends V> values);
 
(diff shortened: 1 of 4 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/multimap/MutableMultimap.java
```diff
@@ -69,6 +69,17 @@ default boolean putAllPairs(Iterable<? extends Pair<? extends K, ? extends V>> p
...
 
+    default Iterable<? extends V> getIfAbsentPut(K key, Iterable<? extends V> value)
+    {
+        boolean hasKey = this.containsKey(key);
+        if (hasKey)
+        {
+            this.putAll(key, value);
+            return value;
+        }
+        return this.get(key);
```
motlin: This looks like it's only putting if the key is present, no?

unit-tests/src/test/java/org/eclipse/collections/impl/multimap/AbstractMutableMultimapTestCase.java
```diff
@@ -322,4 +328,22 @@ public void testToString()
...
+    {
+        ArrayList<String> tempArrayList1 = new ArrayList<>();
+        tempArrayList1.add("Uno");
+        tempArrayList1.add("Ichi");
+
+        ArrayList<String> tempArrayList2 = new ArrayList<>();
+        tempArrayList2.add("Dos");
+        tempArrayList2.add("Ni");
+
+        FastListMultimap<String, ArrayList<String>> multimap = new FastListMultimap<>();
```
motlin: It's confusing to use an iterable type as the value in a multimap. Why not just `MutableMultimap<String, String>`?

PR #621, opened 2018-09-17
Fix Javadoc lint errors and add a Javadoc build.

nikhilnanivadekar:


.travis.yml
```diff
@@ -40,6 +40,11 @@ jobs:
         - DESC="compile performance-tests"
         - CMD="mvn install --projects performance-tests --also-make --activate-profiles all -DskipTests=true"
 
+    - jdk: oraclejdk8
+      env:
+        - DESC="compile javadoc"
+        - CMD="mvn install javadoc:jar --projects '!unit-tests,!junit-trait-runner,!unit-tests-java8,!serialization-tests,!scala-unit-tests,!jmh-scala-tests,!jmh-tests,!jcstress-tests,!p2-repository,!test-coverage' --activate-profiles all -DskipTests=true"
```
motlin: Why activate the all profile, then turn off most projects? I can't keep track of which one is on. Instead of listing negations, why not list the ones you want on?

motlin:
Did you use a tool to make these fixes or did you do them by hand?

.travis.yml
nikhilnanivadekar: This is the exact build which runs during deploy. I tried doing without negation during deploy but the Jenkins build keeps on failing.

nikhilnanivadekar:
by hand, looking at build failures. That is why I added the Javadoc build so that we are not in the same situation in the future.

PR #625, opened 2018-09-25
Suppress Maven Analyzer warnings for JUnit, Eclipse Collections Test Utils.

nikhilnanivadekar:


nikhilnanivadekar:
```
[INFO] --- maven-dependency-plugin:3.1.1:analyze-only (analyze) @ eclipse-collections-forkjoin ---
[WARNING] Unused declared dependencies found:
[WARNING]    org.eclipse.collections:eclipse-collections-testutils:jar:10.0.0.M2:test
[WARNING]    junit:junit:jar:4.12:test
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.1.1:analyze-only (analyze) on project eclipse-collections-forkjoin: Dependency problems found -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.1.1:analyze-only (analyze) on project eclipse-collections-forkjoin: Dependency problems found
```
https://hudson.eclipse.org/collections/job/publish-p2-repo/19/console

motlin:
This seems like a bug in the p2 plugin or something like that. The dependency-plugin returns correct results in all other builds. I don't want global suppressions for a local problem.

I would recommend using skip in just this one Hudson build as the workaround.
https://maven.apache.org/plugins/maven-dependency-plugin/analyze-mojo.html

nikhilnanivadekar:
I think this is a bug in the dependency analyzer plugin. I just ran the build without skipping tests and it went through fine.
https://ci.eclipse.org/collections/job/publish-p2-repo/22/console

motlin:
Not skipping tests, skipping the dependency analyzer. It could be a bug
with either plugin, or some strange interaction between them. Either way,
we shouldn't suppress it globally.

nikhilnanivadekar:
I agree with you. I'll close this PR.
The weird part is that the analyzer fails before the eclipse plugin build is triggered so, I doubt it's related to eclipse plugin.

PR #628, opened 2018-10-18
Fix #627 - Consume every object created in a benchmark to avoid DCE

### LINKED ISSUE #627: JMH Tests should consume/use objects created in the benchmark to avoid DCE
We are conducting a scientific study to investigate bad practices/anti-patterns on creating micro-benchmarks using JMH, and we found an instance of harmfull non-consumed computation in three classes `CollectIfTest`, `RejectTest` and `FlatCollectTest`.

As a good practice, every computation performed in inside the benchmark should be consumed (see [JMH Samples](http://hg.openjdk.java.net/code-tools/jmh/file/66fb723292d4/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_08_DeadCode.java) )

The only aspect preventing the dead-code elimination here is that every benchmark reads from a field member, which in some cases is enough to prevent the compiler for completely removing the code. However, a more aggresive DCE algorithm might remove some instructions and impact the results substantially. 

Simple solutions: 
- Return the variable at the end of the benchmark
- Use `Blackhole.consume` method to sink the computation
- Consume the variable in an assert method call. 

Important to mention, from our static analysis tool all other (+500) benchmarks in this project consumes each object explicitly. We should probably do that on the above mentioned classes as well :)

DiegoEliasCosta:
Signed-off-by: DiegoEliasCosta <diegoelias1@gmail.com>


jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/FlatCollectTest.java
```diff
@@ -38,26 +38,30 @@
...
         MutableList<Integer> flatCollect = this.integersEC.flatCollect(e -> e);
+        return flatCollect;
     }
 
     @Benchmark
-    public void serial_lazy_ec()
+    public MutableList<Integer> serial_lazy_ec()
     {
         MutableList<Integer> flatCollect = this.integersEC.asLazy().flatCollect(e -> e).toList();
+        return flatCollect;
```
motlin: For all four of these, we can just inline the variable, like:
```suggestion
        return this.integersEC.asLazy().flatCollect(e -> e).toList();
```

DiegoEliasCosta:
Just inline the returned variable. 
We have extended Spotbugs with a plugin for bad practices related to JMH performance tests. 
We are still evaluating it, but it will be available soon :)

nikhilnanivadekar:
Please squash the commits. I will merge it after.

DiegoEliasCosta:
Commits squashed. It should be okay now.

PR #631, opened 2018-10-29
Remove default implementations which throw Unsupported Exceptions for

brenopessoa:
Hello , I hope to be helpful with 

Remove default implementations which throw Unsupported Exceptions for Backwards Compatibility #516

Signed-off-by: Breno Pessoa <brenopessoamelo@gmail.com>


nikhilnanivadekar:
@brenopessoa this is not the correct approach/change. You have deleted the implementations in the Factory methods. By deleting default implementations which throw, we mean to delete the `default` implementations. So, for example:
```
    default <T> MutableList<T> ofInitialCapacity(int capacity)
    {
        throw new UnsupportedOperationException("Adding default implementation so as to not break compatibility");
    }
```
should change to:

`<T> MutableList<T> ofInitialCapacity(int capacity)`

brenopessoa:
Ok , I had this doubt. 
So , i'm taking this approach to fix the issue. 

Thank's for the feedback.

nikhilnanivadekar:
Closing this as it has been already fixed by a different PR. @brenopessoa looking forward for a different contribution! Thanks for supporting the project!

PR #634, opened 2018-11-06
Implement UnifiedSetWithHashingStrategy#addOrReplace(Object). Fixes #633

### LINKED ISSUE #633: Implement UnifiedSetWithHashingStrategy#addOrReplace(Object)
I have encountered few usages where UnifiedSetWithHashingStrategy could have benefited to have replace(Object), this is useful when one of non-key the parameters needs to be overridden.
Thoughts?
Implementation will be an optimized version of this:
```
this.remove(Object);
this.add(Object);
```

nikhilnanivadekar:


nikhilnanivadekar:
Test Coverage evidence:
image

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/strategy/mutable/UnifiedSetWithHashingStrategy.java
```diff
@@ -2132,6 +2160,73 @@ private T chainedGet(T key, ChainedBucket bucket)
...
+                return key;
+            }
+            if (this.nonNullTableObjectEquals(bucket.three, key))
+            {
+                T previous = this.nonSentinel(bucket.three);
+                bucket.three = UnifiedSetWithHashingStrategy.toSentinelIfNull(key);
+                return previous;
+            }
+
+            this.chainedAdd(key, index);
```
mohrezaei: This is a bit curious. Can we extract a method for adding a new value into a new chain?

nikhilnanivadekar: You mean instead of calling the chainedAdd()? I can extract it out.

mohrezaei:
Good catch on the sentinel. What's with the teamcity builds? 

> Plugin org.apache.maven.plugins:maven-shade-plugin:2.3 or one of its dependencies could not be resolved:

nikhilnanivadekar:
The compile test build is failing because 1 test is ignored and seems like the baseline is not getting established. I need to investigate more or report to TC.
The checkstyle build is failing due to snapshot dependency, and I have tried all settings to have it refreshed. The same builds have gone through fine on Travis, so IMO it's ok to merge.
Regarding the sentinel catch, I want to fix that as a separate commit to point it on the release notes. I have to check for UnifiedSet too. I'll do it on this PR itself as the commit log will still show. Please hold off on the merge, till i break the commits.

nikhilnanivadekar:
@mohrezaei commits are split! Please review

mohrezaei:
Ignoring TC builds and merging...

Issue #615, opened 2018-09-12
Remove default methods in Factories which throw UnsupportedOperationException to maintain backwards compatibility.


nikhilnanivadekar:

nikhilnanivadekar:
Some methods in factories like `org.eclipse.collections.api.factory.list.MutableListFactory#ofInitialCapacity` have a default implementation which throws. This was added to maintain backwards compatibility and allow contributors to add API features in a minor release. Since we are preparing for major release, I would like to clean up the default implementations by removing these implementations. So, org.eclipse.collections.api.factory.list.MutableListFactory#ofInitialCapacity will look like this:

Current:
```
    default <T> MutableList<T> ofInitialCapacity(int capacity)
    {
        throw new UnsupportedOperationException("Adding default implementation so as to not break compatibility");
    }
```
Fix:
`    public <T> MutableList<T> ofInitialCapacity(int capacity)`

Daniel-Dos:
Hi, 

only remove in classes "factory" ?

PR #636, opened 2018-11-09
fix-#615- Remove default methods in Factories which throw Unsupported…

Daniel-Dos:
Signed-off-by: Daniel Dias <daniel.dias.analistati@gmail.com>

PR to #615 .

inside the collections-api factory classes, this class was the only one that had the default method with throw.

please review .


Daniel-Dos:
closing, since the merge has already been made here:

https://github.com/eclipse/eclipse-collections/pull/586

PR #637, opened 2018-11-10
Remove redundant parts of Travis custom script for EA builds.

motlin:


nikhilnanivadekar:
I would like to re-enable the EA builds for PRs. The push builds fail silently and it is difficult to keep track. As long as they are under allowed failures list, the entire build will not fail. Can you please make that change as well as a part of this PR?

motlin:
That feels unrelated to me. This PR just changes some dead code to echo that it ought to be unreachable.

nikhilnanivadekar:
why not just delete those lines and only keep 12-EA, fail for all except 12-EA

motlin:
I don't know. I was copying the pattern for Java 8 and Java 9. I still don't fully understand this script. I feel like we should just be setting JAVA_HOME and then returning control back to Travis.

nikhilnanivadekar:
The script downloads the jdk, untars it and runs clean install. If you have a way to download the JDK, set JAVA_HOME and then pass control back, it will be a great change to have.
The script was previously called for all JDK versions so the JAVA 8, 9 echo statements were necessary. But the builds were changed to not call the script for available JDKs.

motlin:
If you want JDK 12, you just add a section like this to .travis.yml

```yml
    - jdk: openjdk-ea
      env:
        - DESC="unit tests openjdk-ea"
        - CMD="mvn install"
```

Then run the build and you can see exactly which JDK it uses.

```
$ export JAVA_HOME=${TRAVIS_HOME}/openjdk-ea
$ bash install-jdk.sh -L GPL --target $JAVA_HOME --workspace ${TRAVIS_HOME}/.cache/install-jdk
install-jdk.sh 2018-10-17
Variables:
  feature = 12
  license = GPL
       os = linux-x64
      url = https://download.java.net/java/early_access/jdk12/19/GPL/openjdk-12-ea+19_linux-x64_bin.tar.gz
   status = 200
  archive = /home/travis/.cache/install-jdk/openjdk-12-ea+19_linux-x64_bin.tar.gz
Downloading JDK from https://download.java.net/java/early_access/jdk12/19/GPL/openjdk-12-ea+19_linux-x64_bin.tar.gz...
Picked up _JAVA_OPTIONS: -Xmx2048m -Xms512m
openjdk version "12-ea" 2019-03-19
OpenJDK Runtime Environment (build 12-ea+19)
OpenJDK 64-Bit Server VM (build 12-ea+19, mixed mode, sharing)
$ export PATH=$JAVA_HOME/bin:$PATH
```

Is the number 19 the build number? If so, it's already newer than the version in our environment variable.

```
https://download.java.net/java/early_access/jdk12/9/GPL/openjdk-12-ea+9_linux-x64_bin.tar.gz
```

I don't think we should be using these environment variables anymore because they're almost always stale. Is there another reason we have our own script that competes with [install-jdk.sh](https://github.com/sormuras/bach#install-jdksh)?

donraab:
@motlin @nikhilnanivadekar Is there anything holding this up?

nikhilnanivadekar:
I think this PR is obsolete.

motlin:
It replaces obsolete parts of
.travis-command-ea-builds.sh with echo statements. I think it’s likely that
the whole script .travis-command-ea-builds.sh is unnecessary at the moment
but I’m open to the idea that there will be windows of time around every
major release of Java again where it will be relevant. Basically the time
between when an EA url is created (and even then, only if the url doesn’t
match the same pattern) and when
https://github.com/sormuras/bach/blob/master/install-jdk.sh
is updated to handle the new url pattern.

.travis-command-ea-builds.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash
 
-set -e
-set +u
+set -euxo pipefail
```
motlin: I've been following the advice here in all my bash scripts. Let me know if there's a reason not to.

https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

.travis-command-ea-builds.sh
```diff
@@ -41,25 +40,12 @@ function install_jdk_and_run_ea_build {
   export PATH="${JAVA_HOME}/bin:${PATH}"
 
   $JAVA_HOME/bin/java -version
-  echo "export MAVEN_OPTS='-Dmaven.repo.local=$HOME/.m2/repository -Xmx2g -XX:MaxPermSize=2048m'\" > ~/.mavenrc"
```
motlin: The [before_install section of .travis.yml](https://github.com/eclipse/eclipse-collections/blob/master/.travis.yml#L5-L7) should still run regardless, so I think this was always redundant. The max memory is different, but doesn't need to be. MaxPermSize hasn't had any effect since Java 7.

.travis-command-ea-builds.sh
```diff
@@ -41,25 +40,12 @@ function install_jdk_and_run_ea_build {
...
-  ;;
-Java9)
-  echo "Java 9 already exists in Travis!"
-  ;;
-Java10)
-  install_jdk_and_run_ea_build "https://download.java.net/java/GA/jdk10/10/binaries/openjdk-10_linux-x64_bin.tar.gz"
-  ;;
-Java11-EA)
-  install_jdk_and_run_ea_build ${JDK11_EA_URL}
-  ;;
```
motlin: Instead of echoing "unreachable" I just deleted these. No builds invoke these anymore, and the EA builds are 12 and 13 now.

Issue #312, opened 2017-08-10
Implement reduce and reduceInPlace on primitive iterables.


donraab:
Signatures should be:

```
Optional<type> reduce(<type><type>To<type>Function);
R reduceInPlace(Function0<R>, Object<type>Procedure);
```
This would require supplementing the current OptionalInt/Long/Double with equivalents for all primitives.  There is also no BinaryOperator currently for primitive types other than int/long/double so I suggest using the primitive function equivalent in EC.

victornoel:
@donraab if I make a PR for this, will it go into v10?

And where should the method be declared (interface) and implemented (an abstract class? a default method?).

Thanks :)

donraab:
@victornoel Yes, this can make it into v10.  If it is possible to do as a default method, then I would go with that. Since 10.0 is a major release, if it is easier to implement in the abstract types then this would be fine as well.

victornoel:
@donraab @nikhilnanivadekar @motlin I think this can get closed (cf comment above)

Issue #639, opened 2018-11-12
injectInto with primitive result


victornoel:
Currently, the signature of injectInto, for example in DoubleIterable, is the following:
```java
<T> T injectInto(T injectedValue, ObjectDoubleToObjectFunction<? super T, ? extends T> function);
```

In the same manner that `collect` has variant for each of the primitives, it could be interesting to have the same with `injectInto`, for example:
```java
double injectIntoDouble(double injectedValue, DoubleBinaryOperator function);
```

This will highly improve performances I believe.

Thanks!

nikhilnanivadekar:
@victornoel feel free to contribute this feature. We can include this in v10.0

donraab:
Hi @victornoel, still interested in contributing this?

victornoel:
Hey hi :) I completely forgot about it ^^ Actually I think I can take a look at it tomorrow, I have a bit of free time :)

victornoel:
In the end I went with simply `injectInto` as the method name.

donraab:
> In the end I went with simply `injectInto` as the method name.

HI @victornoel, I actually prefer the specialized method name `injectIntoDouble` because it avoids potential issues with overloads. There are primitive `injectInto` methods on `RichIterable` that I would like to deprecate and rename matching this pattern because they have caused some ambiguity issues in the past, once lambda support was added to Java 8.

victornoel:
Good point, I've updated the PR :)

victornoel:
Done in #1209.

PR #641, opened 2018-11-13
reduce for primitive iterables

victornoel:
Hi,

This is a PR for #312, this is not finished, but I'm making this PR public in order to get feedbacks as soon as possible.

TODO:
- [x] add `reduce` and `reduceIfEmpty` to primitive iterables with simple implementation as a default method
- [x] check if there are other collections (than synchronized primitive iterable and synchronized ~~and unmodifiable~~ primitive collection) that needs a specialized implementation of reduce for correctness
- [x] make DoubleDoubleToDoubleFunction extend DoubleBinaryOperator (as well as with int and long) 
- [x] ~~introduce primitive optionals (for now I use `Optional` with primitive wrappers)~~
- [x] elegantly handle overflow problem (sum on int should give long…)
- [x] ~~add `reduceInPlace`~~

I think that after that, this can be merged and then we could open the following new issues:
- add optimized implementation of `reduce` and `reduceInPlace` in the right places
- other ideas?

Open questions:
- I feel like using a default method (as recommended, see https://github.com/eclipse/eclipse-collections/issues/312#issuecomment-438015186) is dangerous because you never know when it is needed to override or not the method. For example I almost missed abstract synchronized and unmodifiable primitive collections. I'm still not sure if I missed other implementations… If I had added the implementation in the abstract class, I would have a clearer view of the class hierarchies impacted by the addition of the method.

cc @donraab

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/test/collection/mutable/abstractPrimitiveIterableTestCase.stg
+++ b/eclipse-collections-code-generator/src/main/resources/test/collection/mutable/abstractPrimitiveIterableTestCase.stg
@@ -1005,6 +1005,28 @@ public void sumConsistentRounding()
         Assert.assertEquals(<wrapperName>.valueOf(<(literal.(type))("38")>), sum3);
     }
 
+    @Test(expected = NoSuchElementException.class)
+    public void reduceOnEmptyThrows()
+    {
+        this.newWith().reduce((<(wideType.(type))> result, <type> value) -> <(castWideType.(type))>result + <(castWideType.(type))>value + <(wideLiteral.(type))("1")>);
+    }
+
+    @Test
+    public void reduce()
+    {
+        <name>Iterable iterable1 = this.newWith(<["0", "2", "31"]:(literal.(type))(); separator=", ">);
+        <(wideType.(type))> sum1 = iterable1.reduce((<(wideType.(type))> result, <type> value) -> <(castWideType.(type))>result + <(castWideType.(type))>value + <(wideLiteral.(type))("1")>);
+        Assert.assertEquals(<(wideLiteral.(type))("35")>, sum1<if(primitive.floatingPoint)>, 0.001<endif>);
+
+        <name>Iterable iterable2 = this.newWith(<[ "1", "2", "31"]:(literal.(type))(); separator=", ">);
+        <(wideType.(type))> sum2 = iterable2.reduce((<(wideType.(type))> result, <type> value) -> <(castWideType.(type))>result + <(castWideType.(type))>value + <(wideLiteral.(type))("1")>);
+        Assert.assertEquals(<(wideLiteral.(type))("36")>, sum2<if(primitive.floatingPoint)>, 0.001<endif>);
+
+        <name>Iterable iterable3 = this.newWith(<["0", "1", "2", "31"]:(literal.(type))(); separator=", ">);
+        <(wideType.(type))> sum3 = iterable3.reduce((<(wideType.(type))> result, <type> value) -> <(castWideType.(type))>result + <(castWideType.(type))>value + <(wideLiteral.(type))("1")>);
+        Assert.assertEquals(<(wideLiteral.(type))("37")>, sum3<if(primitive.floatingPoint)>, 0.001<endif>);
+    }
+
     @Test
     public void chunk()
     {
--- a/eclipse-collections-code-generator/src/main/resources/api/block/function/primitivePrimitiveToPrimitiveFunction1.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/block/function/primitivePrimitiveToPrimitiveFunction1.stg
@@ -1,0 +1,40 @@
+import "copyrightAndOthers.stg"
+
+hasTwoPrimitives() ::= "true"
+
+targetPath() ::= "org/eclipse/collections/api/block/function/primitive"
+
+fileName(primitive1, primitive2, sameTwoPrimitives) ::= "<primitive1.name><primitive2.name>To<primitive1.name>Function"
+
+class(primitive1, primitive2, sameTwoPrimitives) ::= <<
+<body(primitive1.type, primitive2.type, primitive1.name, primitive2.name)>
+>>
+
+body(type1, type2, name1, name2) ::= <<
+<copyrightAndOthers()>
+
+package org.eclipse.collections.api.block.function.primitive;
+
+import java.io.Serializable;
+
+/**
+ * This file was automatically generated from template file primitivePrimitiveToPrimitiveFunction.stg.
+ *
+ * @since 10.0.
+ */
+@FunctionalInterface
+public interface <name1><name2>To<name1>Function
+        extends <if(primitive1.specializedStream && sameTwoPrimitives)>java.util.function.<name1>BinaryOperator, <endif>Serializable
+{
+    <type1> valueOf(<type1> left, <type2> right);
+<if(primitive1.specializedStream && sameTwoPrimitives)>
+
+    @Override
+    default <type1> applyAs<name1>(<type1> left, <type2> right)
+    {
+        return this.valueOf(left, right);
+    }
+<endif>
+}
+
+>>
--- a/eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
@@ -15,9 +15,11 @@ body(type, name) ::= <<
 package org.eclipse.collections.api;
 
 import java.util.Collection;
+import java.util.NoSuchElementException;
 <(wideStatisticsImport.(type))>
 
 import org.eclipse.collections.api.bag.primitive.Mutable<name>Bag;
+import org.eclipse.collections.api.block.function.primitive.<(wideName.(type))><name>To<(wideName.(type))>Function;
 import org.eclipse.collections.api.block.function.primitive.<name>ToBooleanFunction;
 import org.eclipse.collections.api.block.function.primitive.<name>ToByteFunction;
 import org.eclipse.collections.api.block.function.primitive.<name>ToShortFunction;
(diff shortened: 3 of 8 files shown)


victornoel:
@donraab (or other):
- could I get some feedbacks before I go further with `reduceInPlace`?
- the `Optional<type>` are not easy to approach:
  1. where should they reside? in `api.optional`?
  2. should their interface be the same as the JDK's `OptionalDouble` and co?
  3. should our `OptionalDouble` extend JDK's `OptionalDouble` or should we only generate `Optional<type>` for non-stream primitive (since I suppose the JDK's `OptionalDouble` are there because there are primitive stream for those)?

donraab:
@victornoel 

I have some mixed feelings, thoughts and open questions around this.  I think in the near term, if we want to support a primitive reduce, we should use the same approach we used for min, max and average.

```
int minIfEmpty(int defaultValue);
int maxIfEmpty(int defaultValue);
double averageIfEmpty(double defaultValue);
```
It would be great to have `Optional` for all primitives in the JDK, but I'd prefer not to have to create our own to augment the three existing primitive versions.  The `IfEmpty` approach removes the need to add these, at least in the near term.  

The question I have is around the potential widening of types in the primitive space for reduce.  We should be able to implement `sum` in terms of reduce, and this would mean that an `IntIterable` would need to return `long` for reduce to support this.

`long sum();
` 

Thoughts?

victornoel:
@donraab I like the `IfEmpty` solution, much simpler to implement ;)

So should reduce be called `reduceIfEmpty` instead?
And should we still keep a `reduce` for `Double`, `Int` and `Long` that returns the corresponding Optional since they already exist?

Concerning int to long and so on, I first I thought I would need to add all the combination of `reduceInt`, `reduceLong`, etc for each of the primitive iterables.
Then I concluded that for now, in the same way that the JDK's reduce on `Stream` and on `RichIterable` handle one type a time (since they are based on `BinaryOperator`), this was out of scope of this issue: I was planning to create a new issue afterwards to add the other `reduce` operations.

donraab:
Yes, `ifEmpty` works out nicely to simplify things with primitives.  You can start with implementing that.  I was not thinking we should have `reduceInt`, `reduceLong`.  I was thinking `reduceIfEmpty` for an `IntIterable` should return a `long` (and the same widening rules we use for all the other types).  This means it would need to take a `Function` that has two `ints` as parameters and `long` as a result.  But this would probably result in at least one of the parameters needing to be the wider type, or the accumulating result would get trimmed.

Does this make sense? 

If you get this working with the widening type approach, then reduce can probably be implemented using only the three primitive optional types as you will only need to return OptionalInt, OptionalLong or OptionalDouble for any of the primitive types.

victornoel:
>  This means it would need to take a Function that has two ints as parameters and long as a result.

@donraab you mean a function that takes a long and an int as parameters and long as result (since reduce takes the previous result and passes it to the next call of the function).

donraab:
@victornoel Yes, exactly.

nikhilnanivadekar:
I would add to this that `reduceIfEmpty()` is more intuitive and discoverable as an API. I would imagine that there will be more users who will want an API like `reduceIfEmpty()`. 
Also `injectInto()` is fairly esoteric and only few know about the powers of `injectInto()`

victornoel:
The only difference I see between `injectInto` and `reduce` is that with the second, there is no need for knowing the starting value (as @motlin perfectly explained). In the wikipedia page of [fold, aka reduce, accumulate, aggregate, compress, or inject](https://en.wikipedia.org/wiki/Fold_(higher-order_function)), it is defined as a fold on non-empty list.

So I don't really agree with the argument that "people don't know about `inject`" and `reduce` will be a more intuitive and discoverable. To me it is the same as having `collect` instead of `map`.
There is still need for reduce though, as a specialisation of `inject` for non-empty list.
By the way Smalltalk has both inject and reduce, with this semantic, according to the wikipedia page.

The difference between `reduce` and `reduceIfEmpty` is just technical (not to use `Optional`) and is, IMHO, not something that should drive decision making.
I feel like returning an `Optional` is closer to the spirit of `reduce`. An alternative could be throwing an `IllegalStateException`, but I'm not sure if it is in line with design of EC :)

victornoel:
And for the record, a primitive aware version of `injectInto` is being discussed in #639, I will probably make a PR for that too.

victornoel:
@motlin regarding your comment on the default value being 1 or 0 (for multiplication and sum), it is not exactly true: while with `injectInto` you are forced to use 1 instead of 0 as a starting value for multiplication (if not, you will always get a result of 0), with `reduceIfEmpty` you can decide to use 0 as a default value without breaking the behaviour of reducing by multiplication.

motlin:
min and max throw on empty primitive collections. It would be reasonable to
add reduce() with the same behavior.

victornoel:
@donraab @motlin @nikhilnanivadekar I've pushed a new version: there is `reduce` and `reduceIfEmpty`. They behave like `min` and `minIfEmpty` in terms of collection emptiness.

We are still missing something about overflow problems. Any opinion as what to do about that?

I wonder if this PR should cover `reduceInPlace` or if we should focus on reduce since it's big enough like this.

Also, should we prefer covering all optimization now (at the same level as `sum` or `min` for example) or later is ok?

motlin:
I don't think you can do anything about overflow when you don't know what
the reduction function will be. I guess there's one thing we could do which
is return the wider type for IntIterable and FloatIterable.

motlin:
Our IntIterable.sum() already returns long. No one has requested a version
that returns int. Personally I would use the existing method together with
an exact cast.
https://stackoverflow.com/a/32869210

victornoel:
@motlin sorry I'm not 100% clear: are you saying we should return `long` for `IntIterable` and use `toIntExact` when we only want an `int`? Or are you saying we should return `int` for `IntIterable` and then I don't understand how to use `toIntExact` in this context…

motlin:
I think we should consider:
1. Not adding reduce, reduceIfEmpty, or both, at least until we have some
concrete use-cases lined up
2. If we find a usecase is math, like "calculate variance", consider just
adding specific math methods first
3. If we add these new methods anyway, we should consider returning long
from IntIterable and double from FloatIterable

victornoel:
For the record, my use case is for sum and multiplications on doubles, but it has to go through a generic API.
I have objects that takes `DoubleBinaryOperator` and that uses EC to implement reduce on a collection of double, but in practice those objects are used with sum and multiplications on doubles.

So I could use `injectInto` and have users of the object pass the initial value, but it would be closer to my domain problem to have `reduce` and to fail on empty lists (since I don't have empty lists in this use case).

victornoel:
@motlin @donraab so what should be done? Keep it like that? Return the wide type? add a secondary method for the wide type?

donraab:
Return the wide type.

victornoel:
@donraab ok, I've implemented that, could you validate it's alright like this?

If yes, next step is `reduceInPlace` but maybe it could go in another PR, it seems to cover different needs and problems?

eclipse-collections-code-generator/src/main/resources/test/collection/mutable/abstractPrimitiveIterableTestCase.stg
```diff
@@ -1005,6 +1005,29 @@ public void sumConsistentRounding()
...
+    }
+
+    @Test
+    public void reduce()
+    {
+        <name>Iterable iterable1 = this.newWith(<["0", "2", "31"]:(literal.(type))(); separator=", ">);
+        <(wideType.(type))> sum1 = iterable1.reduce((<(wideType.(type))> result, <type> value) -> <(castWideType.(type))>(result + value + 1));
+        Assert.assertEquals(<(wideLiteral.(type))("35")>, sum1<if(primitive.floatingPoint)>, 0.001<endif>);
+        
+
```
donraab: Please remove the extra blank line.

victornoel:
@donraab I've pushed tests for boolean and removed the empty line.

Why would I need to override it for unmodifiable collection? This is not a mutable operation.

donraab:
@victornoel Thanks for the changes.  While you are correct, it is not a mutable operation, I'd recommend being consistent with other methods like max and sum in `abstractUnmodifiablePrimitiveCollection.stg` which delegate to the underlying collection.

```
    @Override
    public long sum()
    {
        return this.collection.sum();
    }

    @Override
    public int max()
    {
        return this.collection.max();
    }

```

victornoel:
> While you are correct, it is not a mutable operation, I'd recommend being consistent with other methods like max and sum in abstractUnmodifiablePrimitiveCollection.stg which delegate to the underlying collection.

Hmm, but isn't it the point of using default methods? Not to have to do that? I'm good with adding them, but this comforts me in the idea that default methods is a dangerous design tool :)

victornoel:
@donraab in the case I need to add them, is there any other place where this could be needed? Or only synchronized and unmodifiable are the two existing wrappers?

donraab:
> > While you are correct, it is not a mutable operation, I'd recommend being consistent with other methods like max and sum in abstractUnmodifiablePrimitiveCollection.stg which delegate to the underlying collection.
> 
> Hmm, but isn't it the point of using default methods? Not to have to do that? I'm good with adding them, but this comforts me in the idea that default methods is a dangerous design tool :)

We do have to be careful when using default methods with both synchronized and unmodifiable collections.  I think this is the only place where they are needed.

victornoel:
@donraab ah yes, I see what you mean! I've pushed changes for this.

donraab:
@victornoel The build is currently failing on the new file you added.  Fix the build failure and I will merge the change.  Thanks!

victornoel:
@donraab ah yes, sorry about that, it should be fixed now, thanks!

donraab:
@victornoel The Checkstyle build is failing due to some unnecessary parantheses.

`target/generated-test-sources/java/org/eclipse/collections/impl/collection/mutable/primitive/AbstractDoubleIterableTestCase.java:[1088,64] (coding) UnnecessaryParentheses: Unnecessary parentheses around expression.`

PR #642, opened 2018-11-16
Move maven modules 'test-coverage' and 'p2-repository' from the default list into the 'all' profile.

motlin:


nikhilnanivadekar:
How will this impact our release builds for normal jars and p2?

motlin:
Good question. If the p2 builds and coverage builds don't already use the profile all or list projects using mvn command line flags, they'd have to start doing that.

nikhilnanivadekar:
Can you update it on Jenkins and do a test milestone build?

donraab:
@motlin Did you try a test milestone build?

PR #643, opened 2018-11-16
Pull up several method bodies to default methods in api.

motlin:


nikhilnanivadekar:
There were build failures. Please hold off on the merges

donraab:
My bad.  Looking into the failure.  Looks like default keyword was missing on a method in RichIterable.

PR #644, opened 2018-11-16
Add openjdk-ea travis build.

motlin:


.travis.yml
```diff
@@ -65,6 +65,11 @@ jobs:
         - DESC="unit tests oraclejdk11"
         - CMD="mvn install"
 
+    - jdk: openjdk-ea
+      env:
+        - DESC="unit tests openjdk-ea"
+        - CMD="mvn install"
+
     - jdk: oraclejdk8
       if: type = push AND repo = eclipse/eclipse-collections
```
nikhilnanivadekar: can you please remove this `if` statement so that the 12-EA build runs for all PRs and merges? Since as it is the build file is getting edited.

motlin: this is redundant now

motlin:
Just commented on another PR, but this one really should be merged first.
* upstream/master: The JDK 11 build uses the copy provided by Travis and the JDK 12 EA build uses a copy downloaded from the environment variable by running the script ./.travis-command-ea-builds.sh
* this PR: Adds a JDK EA build using the install-jdk.sh script, which downloads a new EA build and will continue working past Java 12
* PR #637: Removes the lines that say they are running a JDK 10 or JDK 11 EA build by downloading from Oracle. We haven't done that for ages.
* Later: We should do the same thing for JDK 12 EA since we'll have two build configurations running EA 12.

motlin:
Note that our custom EA build is downloading obsolete version:
https://download.java.net/java/early_access/jdk12/9/GPL/openjdk-12-ea+9_linux-x64_bin.tar.gz

And the install-jdk.sh script downloads:

https://download.java.net/java/early_access/jdk12/26/GPL/openjdk-12-ea+26_linux-x64_bin.tar.gz
or
https://download.java.net/java/early_access/jdk13/2/GPL/openjdk-13-ea+2_linux-x64_bin.tar.gz

motlin:
Pushed a new commit which runs Java 12 and 13 builds using the two newer versions I mentioned, only relying on the built-in install-jdk.sh, no custom scripts or downloads. This makes our custom script redundant, because it runs an obsolete, older JDK 12 EA. However, I'm not removing the custom script or older Java 12 build in case that requires additional discussion.

PR #660, opened 2018-12-01
Implement MutableMultimap.getIfAbsentPutAll(). Fixes #566.

### LINKED ISSUE #566: Implement MutableMultimap.getIfAbsentPutAll().

nikhilnanivadekar:


motlin:
I believe this implementation violates the invariant that a multimap will
never contain an empty collection as a value in the backing map.
Specifically if you call the new method with an empty values collection

nikhilnanivadekar:
Fair point @motlin I have pushed another implementation which will not keep empty collections. I had an option to always throw in case the passed in values are empty, but I chose to evaluate the size only when required. Secondly, I remove the empty collection before throwing, so the map is accessed twice only when there will be an exception thrown and it is accessed once for non-failure scenarios. Corresponding test cases are also added.

motlin:
I wasn't suggesting that the method should throw. I think it shouldn't, for an empty set of values. Take a look at the implementations of AbstractMutableMultimap#remove and AbstractMutableMultimap#putAll(K, java.lang.Iterable<? extends V>) to see how we avoid calling getIfAbsentPutCollection unless we know the collection ought to exist in the backing map.

Also consider adding a package private method to allow asserting that a multimap contains no empty collections for values.

motlin:
```java
@Override
public RichIterable<V> getIfAbsentPut(K key, Iterable<? extends V> values)
{
    if (Iterate.isEmpty(values))
    {
        return this.get(key);
    }

    C existingValue = this.getIfAbsentPutCollection(key);
    if (existingValue.isEmpty())
    {
        int newSize = Iterate.addAllTo(values, existingValue).size();
        this.addToTotalSize(newSize);
    }
    
    return existingValue.asUnmodifiable();
}
```

motlin:
There are a bunch of missing implementations. Everywhere there's an override of MutableMultimap.get(), there should be an override of getIfAbsentPut(). For example, MutableListMultimap.getIfAbsentPut() should return MutableList<V>, not RichIterable<V>.

nikhilnanivadekar:
@motlin this seems to be good to go.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/multimap/MutableMultimap.java
```diff
@@ -77,6 +77,16 @@ default boolean putAllPairs(Iterable<? extends Pair<? extends K, ? extends V>> p
...
     RichIterable<V> removeAll(Object key);
 
+    /**
+     * Get and return a view onto the values associated with the specified key.
+     * Alternatively, if there are no values in the multimap at the key, return an unmodifiable view of specified values,
+     * and put those values in the multimap at the specified key.
+     *
+     * @throws IllegalArgumentException when there are no values for the key and the passed values are empty
+     * @since 10.0
+     */
```
motlin: I believe the Javadoc needs to be updated to not have a throws clause now. I'd suggest language like:

Puts `values` into multimap if there are no values already associated with `key`. Then returns a view of the values associated with `key`, like the result of get(K).

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Shotaro Sano.
+ * Copyright (c) 2018 Shotaro Sano and others.
```
motlin: Copyright-only change to this file. Is it intentional?

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedPutMultimap.java
```diff
@@ -96,4 +97,22 @@ private C getIfAbsentPutCollection(K key)
...
+
+        C existingValue = this.getIfAbsentPutCollection(key);
+        if (existingValue.isEmpty())
+        {
+            int newSize = Iterate.addAllTo(values, existingValue).size();
+            this.addToTotalSize(newSize);
+        }
+
+        return (C) existingValue.asUnmodifiable();
+    }
```
motlin: We don't have to grab a lock or anything? I'm concerned about the fact that we first put an empty collection into the backing map and then populate it without explicitly taking a lock.

motlin: It would be good if @donraab could look at this one file, but I suspect you just need to synchronize on existingValue. Also, should it be called `existingValues`?

unit-tests/src/test/java/org/eclipse/collections/impl/multimap/AbstractMutableMultimapTestCase.java
```diff
@@ -197,6 +197,28 @@ public void putAllFromIterable()
...
+        multimap.putAll(3, Lists.mutable.with(3, 3, 3));
+        Assert.assertEquals(this.createCollection(1), multimap.getIfAbsentPut(1, Lists.mutable.empty()));
+        Assert.assertEquals(this.createCollection(2, 2), multimap.getIfAbsentPut(2, Lists.mutable.empty()));
+        Assert.assertEquals(this.createCollection(3, 3, 3), multimap.getIfAbsentPut(3, Lists.mutable.empty()));
+        Assert.assertEquals(this.createCollection(4, 4, 4, 4), multimap.getIfAbsentPut(4, Lists.mutable.with(4, 4, 4, 4)));
+        Assert.assertEquals(4, multimap.sizeDistinct());
+        int multimapSize = this.createCollection(1).size() + this.createCollection(2, 2).size() + this.createCollection(3, 3, 3).size() + this.createCollection(4, 4, 4, 4).size();
+        Assert.assertEquals(multimapSize, multimap.size());
+
+        Verify.assertThrows(UnsupportedOperationException.class, () -> multimap.getIfAbsentPut(5, Lists.mutable.with(5)).add(5));
```
motlin: Add some assertions about adding empty values? Preferably early, so if anything is messed up the size checks have a chance of catching the problem.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedPutMultimap.java
nikhilnanivadekar: Just checked put(), we synchronize on the new empty collection.

nikhilnanivadekar: refactored and renamed. Also synchronized on existingValues.

eclipse-collections/src/main/java/org/eclipse/collections/impl/multimap/AbstractSynchronizedMultimap.java
nikhilnanivadekar: I pushed it to implementations due to covariant overrides, forgot to undo the copyright update. Fixed.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/multimap/MutableMultimap.java
nikhilnanivadekar: Fixed.

nikhilnanivadekar:
Fixed all review comments. Build is running.

nikhilnanivadekar:
@motlin all builds have passed. Please take a look

nikhilnanivadekar:
I have rebased. Is this ready to go?

nikhilnanivadekar:
Merging based on approval from @donraab

PR #664, opened 2018-12-03
Implement MutableMapIterable#newEmpty(initialCapacity). Fixes #444

### LINKED ISSUE #444: Implement MutableMapIterable#newEmpty(initialCapacity)

nikhilnanivadekar:


donraab:
Why are we adding this on `MutableMapIterable` and not `MutableList` for instance?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -194,6 +194,11 @@ default V add(Pair<? extends K, ? extends V> keyValuePair)
      */
     MutableMapIterable<K, V> newEmpty();
 
+    /**
+     * Creates a new instance of the same type, using the initial capacity.
+     */
```
motlin: `@since 10.0`

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -194,6 +194,11 @@ default V add(Pair<? extends K, ? extends V> keyValuePair)
...
+    <K, V> MutableMapIterable<K, V> newEmpty(int initialCapacity);
```
motlin: Why does this take K and V? I see two problems.

* These are the same names as the outer generic types, which shadows them and is super confusing.
* This cannot work for subclasses that take `Comparator<K>` or `HashingStrategy<K>` in their constructors.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
```diff
@@ -51,7 +51,7 @@
 import org.eclipse.collections.impl.utility.MapIterate;
 import org.eclipse.collections.impl.utility.internal.IterableIterate;
 
-@SuppressWarnings({ "rawtypes", "ObjectEquality" })
+@SuppressWarnings({"rawtypes", "ObjectEquality"})
```
motlin: whitespace only change

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
```diff
@@ -138,6 +138,12 @@ public int hashCode()
         return OrderedMapAdapter.adapt(new LinkedHashMap<>());
     }
 
+    @Override
+    public MutableOrderedMap<K, V> newEmpty(int initialCapacity)
+    {
+        return OrderedMapAdapter.adapt(new LinkedHashMap<>(initialCapacity));
```
motlin: I think the implementation for newEmpty() was already wrong, since it ignores the potential comparator.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMap.java
```diff
@@ -1394,6 +1394,12 @@ public void appendString(Appendable appendable, String start, String separator,
         return this.delegate.newEmpty();
     }
 
+    @Override
+    public MutableOrderedMap<K, V> newEmpty(int initialCapacity)
+    {
+        return this.delegate.newEmpty(initialCapacity);
```
motlin: I am stumped. How does this compile? The generic arguments are missing.

motlin: I checked and there's a warning on this line about unchecked types, as well as a warning on the parent about the shadowing.

motlin: My biggest concern is that the build passed. This should have failed TeamCity for sure, and probably a regular javac build with warnings treated as errors.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/SortedMapAdapter.java
```diff
@@ -75,6 +75,12 @@ public void forEachKeyValue(Procedure2<? super K, ? super V> procedure)
         return TreeSortedMap.newMap(this.delegate.comparator());
     }
 
+    @Override
+    public MutableSortedMap<K, V> newEmpty(int initialCapacity)
+    {
+        return TreeSortedMap.newMap(this.delegate.comparator());
```
motlin: This is unsound.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/TreeSortedMap.java
```diff
@@ -210,6 +210,12 @@ public int size()
         return new TreeSortedMap<>(this.comparator());
     }
 
+    @Override
+    public MutableSortedMap<K, V> newEmpty(int initialCapacity)
+    {
+        return this.newEmpty();
```
motlin: Also unsound.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
```diff
@@ -1436,7 +1438,9 @@ private int chainedHashCode(Object[] chain)
                 return hashCode;
             }
             Object value = chain[i + 1];
-            hashCode += this.hashingStrategy.computeHashCode(this.nonSentinel(cur)) ^ (value == null ? 0 : value.hashCode());
+            hashCode += this.hashingStrategy.computeHashCode(this.nonSentinel(cur)) ^ (value == null
+                    ? 0
+                    : value.hashCode());
```
motlin: whitespace only changes

motlin:
I'm wondering this is the right interface for the method. The sorted subclasses ignore the capacity.

nikhilnanivadekar:
what do you mean it is unsound? I am fine just throwing at those implementations. I am fine closing the PR and the issue as won't fix if it is not going to add any values.

motlin:
The generics are just wrong. You don't have to close the PR, we can just fix the generics. The IDE should be showing you, no?

nikhilnanivadekar:
I tried at first with the fixed generics  it clashes with the existing newEmpty()

motlin:
Just to make sure the problem is clear:

```java
    @Test
    @Override
    public void newEmptyWithInitialCapacity()
    {
        super.newEmptyWithInitialCapacity();

        MutableMapIterable<Integer, Integer> map = this.newMapWithKeysValues(Integer::compare, 1, 1, 2, 2);
        MutableMapIterable<Double, Double> newMap3 = map.newEmpty(7);
        Verify.assertEmpty(newMap3);

        newMap3.put(1.0, 1.0);
        newMap3.put(2.0, 2.0);
        newMap3.put(3.0, 3.0);
        Verify.assertSize(3, newMap3);
    }
```

nikhilnanivadekar:
It’s a major release we can break backwards compatibility and mention it in release notes

motlin:
Sounds good. Let me know if you need anything.

donraab:
@nikhilnanivadekar Is this ready to rebase and merge?

nikhilnanivadekar:
Nope. It needs more work which I will get to after Jan 4th

nikhilnanivadekar:
While going through the implementations, and review comments, on second thoughts, this change does not make sense and feels to be wasteful. I am going to close the PR and also the issue. Please reopen if I am mistaken.

PR #665, opened 2018-12-05
Implement MutableMapIterable.removeIf(Predicate). Fixes #662

### LINKED ISSUE #662: Implement MutableMapIterable#removeIf(Predicate<K, V>).

nikhilnanivadekar:


eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -1035,6 +1035,22 @@ public V removeKey(K key)
...
+        {
+            Entry<K, V> entry = iterator.next();
+            if (predicate.accept(entry.getKey(), entry.getValue()))
+            {
+                iterator.remove();
+            }
+        }
+        return previousSize > this.size();
+    }
+
```
donraab: Is there a more efficient way to write this for UnifiedMap?

nikhilnanivadekar: not that I know of or think of. The main problem is around, when do we actually rehash.

motlin:
The faster implementation would be to iterate through the backing data
structure and perform removes in one chain at a time without rehashing.
Once at the very end, do the normal rehash check that removeKey() does.

donraab:
@nikhilnanivadekar Please rebase and I'll merge.  The optimization for UnifiedMap can be a separate PR.

nikhilnanivadekar:
@donraab Done!

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/ordered/MutableOrderedMapTestCase.java
```diff
@@ -22,4 +28,33 @@
...
+    {
+        MutableOrderedMap<Integer, String> map = this.newWithKeysValues(1, "1", 2, "Two", 3, "Three");
+
+        Assert.assertFalse(map.removeIf(Predicates2.alwaysFalse()));
+        Assert.assertEquals(this.newWithKeysValues(1, "1", 2, "Two", 3, "Three"), map);
+        Assert.assertTrue(map.removeIf(Predicates2.alwaysTrue()));
+        Verify.assertEmpty(map);
+
+        map.putAll(Maps.mutable.with(1, "One", 2, "TWO", 3, "THREE", 4, "four"));
+        map.putAll(Maps.mutable.with(5, "Five", 6, "Six", 7, "Seven", 8, "Eight"));
```
motlin: Here you're calling putAll using a hashtable, so the keys are in random order. This means that you're not asserting anything about ordering.

PR #666, opened 2018-12-05
Implement MutableMapIterable.removeAllKeys(). Fixes #663

### LINKED ISSUE #663: Implement MutableMapIterable#removeAllKeys(Iterable<K> keys).

nikhilnanivadekar:


eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -66,6 +67,14 @@ default V add(Pair<? extends K, ? extends V> keyValuePair)
...
     V removeKey(K key);
 
+    /**
+     * Remove entries from the map at the specified {@code keys}.
+     *
+     * @return true if any entry is removed
+     * @since 10.0
+     */
+    boolean removeAllKeys(Set<? extends K> keys);
+
```
motlin: More details below, but there should be a default implementation and it should look like this:

```java
    /**
     * Remove entries from the map at the specified {@code keys}.
     *
     * @return @return <tt>true</tt> if this map changed as a result of the call
     * @since 10.0
     */
    default boolean removeAllKeys(Set<? extends K> keys)
    {
        int previousSize = this.size();
        keys.forEach(this::removeKey);
        return previousSize != this.size();
    }
```

I remain skeptical that this method is necessary. Personally, I'd just use keys.forEach(map::removeKey) instead.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/MapAdapter.java
```diff
@@ -163,6 +164,31 @@ public V removeKey(K key)
...
+            while (iterator.hasNext())
+            {
+                K key = iterator.next();
+                if (keys.contains(key))
+                {
+                    iterator.remove();
+                }
+            }
+        }
+        return previousSize > this.size();
```
motlin: This one can be removed

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -1035,6 +1036,31 @@ public V removeKey(K key)
...
+            {
+                K key = iterator.next();
+                if (keys.contains(key))
+                {
+                    iterator.remove();
+                }
+            }
+        }
+        return previousSize > this.size();
+    }
```
motlin: This one can have the optimized implementation.

```java
    @Override
    public boolean removeAllKeys(Set<? extends K> keys)
    {
        int previousSize = this.size();
        if (this.size() > keys.size())
        {
            keys.forEach(this::removeKey);
        }
        else
        {
            this.keySet().removeIf(keys::contains);
        }
        return previousSize != this.size();
    }
```

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
```diff
@@ -230,6 +232,31 @@ public V removeKey(K key)
...
+            while (iterator.hasNext())
+            {
+                K key = iterator.next();
+                if (keys.contains(key))
+                {
+                    iterator.remove();
+                }
+            }
+        }
+        return previousSize > this.size();
```
motlin: This one can be removed

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/SortedMapAdapter.java
```diff
@@ -195,6 +197,31 @@ public V removeKey(K key)
...
+            while (iterator.hasNext())
+            {
+                K key = iterator.next();
+                if (keys.contains(key))
+                {
+                    iterator.remove();
+                }
+            }
+        }
+        return previousSize > this.size();
```
motlin: This one can be removed

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/TreeSortedMap.java
```diff
@@ -216,6 +218,31 @@ public V removeKey(K key)
...
+            while (iterator.hasNext())
+            {
+                K key = iterator.next();
+                if (keys.contains(key))
+                {
+                    iterator.remove();
+                }
+            }
+        }
+        return previousSize > this.size();
```
motlin: This one can have the optimized implementation.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
```diff
@@ -1078,6 +1079,31 @@ public V removeKey(K key)
...
+            while (iterator.hasNext())
+            {
+                K key = iterator.next();
+                if (keys.contains(key))
+                {
+                    iterator.remove();
+                }
+            }
+        }
+        return previousSize > this.size();
```
motlin: this one can have the optimized implementation

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
```diff
@@ -1436,7 +1464,9 @@ private int chainedHashCode(Object[] chain)
                 return hashCode;
             }
             Object value = chain[i + 1];
-            hashCode += this.hashingStrategy.computeHashCode(this.nonSentinel(cur)) ^ (value == null ? 0 : value.hashCode());
+            hashCode += this.hashingStrategy.computeHashCode(this.nonSentinel(cur)) ^ (value == null
+                    ? 0
+                    : value.hashCode());
```
motlin: whitespace only changes

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/ordered/MutableOrderedMapTestCase.java
```diff
@@ -22,4 +27,26 @@
...
+
+    @Test
+    default void MutableOrderedMap_removeAllKeys()
+    {
+        MutableOrderedMap<Integer, String> map = this.newWithKeysValues(1, "1", 2, "Two", 3, "Three");
+
+        Verify.assertThrows(NullPointerException.class, () -> map.removeAllKeys(null));
+        Assert.assertFalse(map.removeAllKeys(Sets.mutable.with(4)));
+        Assert.assertFalse(map.removeAllKeys(Sets.mutable.with(4, 5, 6)));
+        Assert.assertFalse(map.removeAllKeys(Sets.mutable.with(4, 5, 6, 7, 8, 9)));
```
motlin: Should have an assertion for an empty set of keys.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/ordered/UnmodifiableMutableOrderedMapTest.java
```diff
@@ -69,6 +69,14 @@ public void MutableMapIterable_removeKey()
         assertThrows(UnsupportedOperationException.class, () -> map.removeKey(2));
     }
 
+    @Override
+    @Test
+    public void MutableOrderedMap_removeAllKeys()
+    {
+        MutableMapIterable<Object, Object> map = this.newWith();
+        assertThrows(UnsupportedOperationException.class, () -> map.removeAllKeys(null));
```
motlin: Why null here? It's inconsistent with UnmodifiableBiMapTest.

unit-tests/src/test/java/org/eclipse/collections/impl/map/fixed/AbstractMemoryEfficientMutableMapTest.java
```diff
@@ -99,6 +99,12 @@ public void removeKey_throws()
         this.classUnderTest().removeKey(null);
     }
 
+    @Test(expected = UnsupportedOperationException.class)
```
motlin: Avoid @Test(expected)

nikhilnanivadekar: It is consistent with other tests in this class. I am fine using Verify.assertThrows() when there are multiple test cases in the same test method. However, here it does not make sense.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentMutableHashMap.java
```diff
@@ -252,6 +253,31 @@ public V removeKey(K key)
...
+            while (iterator.hasNext())
+            {
+                K key = iterator.next();
+                if (keys.contains(key))
+                {
+                    iterator.remove();
+                }
+            }
+        }
+        return previousSize > this.size();
```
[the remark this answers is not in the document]
nikhilnanivadekar: I will throw for this class then.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
nikhilnanivadekar: I want to provide the method on the Map itself. I am fine having the default implementation.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/MapAdapter.java
nikhilnanivadekar: why can it be or rather should it be removed? Because of the default method?

motlin: Yes because of the default method. Also an adapter could be adapting a ConcurrentHashMap. In an adapter, we can't assume the optimization is safe to do since we can't assume anything about the nature of the adapted map.

nikhilnanivadekar: Ok makes sense

nikhilnanivadekar: Ok makes sense

nikhilnanivadekar:
@motlin I think I have fixed all the review comments. Please take a look. I think SortedTreeMap should just use the default implementation as well. But I have kept the optimized implementation as you mentioned.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -66,6 +67,19 @@ default V add(Pair<? extends K, ? extends V> keyValuePair)
      */
     V removeKey(K key);
 
+    /**
+     * Remove entries from the map at the specified {@code keys}.
+     *
+     * @return @return <tt>true</tt> if this map changed as a result of the call
```
motlin: Double javadoc `@return @return`

motlin: I see I had that mistake in my example, sorry

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
```diff
@@ -301,6 +302,23 @@ public V removeKey(K key)
...
+        Objects.requireNonNull(keys);
+        int previousSize = this.size();
+
+        if (this.size() > keys.size())
+        {
+            keys.forEach(this::removeKey);
+        }
+        else
+        {
+            this.keySet().removeIf(keys::contains);
```
motlin: It just occurred to me that this is only an optimization if the keys parameter is a hashtable. If it's a tree, it could be slower. It's only guaranteed to be faster if this instance is also a tree. What do you think? The whole thing is such a hypothetical that will never happen in our code.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
```diff
@@ -1079,6 +1080,22 @@ public V removeKey(K key)
...
+        if (this.size() > keys.size())
+        {
+            keys.forEach(this::removeKey);
+        }
+        else
+        {
+            this.keySet().removeIf(keys::contains);
+        }
+        return previousSize != this.size();
+    }
```
motlin: Unrelated to the current change, but we're probably getting to the point where there is enough commonality between UnifiedMap and UnifiedMapWithHashingStrategy that we could extract an AbstractUnifiedMap. I think we talked about doing it in the past but we were worried about performance. If no state were pulled up and it were just used for "default methods" like these, I think it would be fine.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
nikhilnanivadekar: If it’s a Tree we should just iterate over the keys, but what should I do an instance check with? I can check for TreeSet so, the If will get an ‘or’ as well. Any other comments? I would like to get this merged and prepare for 10.0

motlin: Well I'm really asking for your thoughts here, not suggesting you add an instanceof. If you do want to add an instanceof, maybe NavigableSet makes sense. But maybe it's ok as is since this situation seems so rare. Or maybe we should get rid of the optimization entirely because it's not "worth it". What do you think?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
nikhilnanivadekar: Raised #698

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
nikhilnanivadekar: fixed

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
nikhilnanivadekar: I think it is very rare for the keys to be Ordered or Sorted. I am going to add that in the javadocs that it might be slower. I will add the workaround as well in the Javadoc.

nikhilnanivadekar:
@motlin any other feedback or it’s good to merge?

nikhilnanivadekar:
Merging based on approval

PR #669, opened 2018-12-07
made it possible to get value from map using different from original key types but with the same equals and hashcode output

kolybelkin:
The idea is to use custom implementation of CharSequence as key (when it returns the same hashcode and it equals) for getting the value associated with String key. Currently it works for general case in get method but does not work if slowGet method is involved.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
@@ -151,7 +151,7 @@ public final class ConcurrentHashMap<K, V>
                 while (e != null)
                 {
                     K candidate = e.getKey();
-                    if (candidate.equals(key))
+                    if (key.equals(candidate))
                     {
                         return e.getValue();
                     }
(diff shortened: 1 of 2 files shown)


mohrezaei:
Why doesn't it work with `slowGet`?

kolybelkin:
If the original key is String object, then it is possible to use only String type as a key because equals method of String has only one condition for comparing - obj instanceof String

mohrezaei:
I understand the reason you're reversing the call order. What I don't understand is, you've reversed the call order in `slowGet`, and yet you said:

> Currently it works for general case in get method but does not work if slowGet method is involved.

kolybelkin:
Take a look at get method, the call order is already reversed there for the case with no resizing (no slowGet call)

mohrezaei:
Sorry, I understand now. I misread your statement.

The changes are fine. Please add test cases for all the changed methods (using CharSeq/String).

nikhilnanivadekar:
@kolybelkin any updates on this? I see that @mohrezaei had asked for tests before merging this PR. Can you please add tests? Please let us know if you need help.

kolybelkin:
Hi, I am working on it currently and do not have time for unit tests, sorry
11 февр. 2020 г., 19:08 +0400, Nikhil Nanivadekar <notifications@github.com>, писал:

kolybelkin:
Not working
11 февр. 2020 г., 19:08 +0400, Nikhil Nanivadekar <notifications@github.com>, писал:

motlin:
Should we add a ConcurrentHashingStrategyMap?

nikhilnanivadekar:
Closing due to inactivity

PR #670, opened 2018-12-15
Implement LazyIterate.cartesianProduct

victornoel:
This generalize the already available `Sets.cartesianProduct` and `SetIterables.cartesianProduct` to any `Iterable`.

This is something I did need many times and I was bothered to have to convert my two `LazyIterable`s into concrete `Set`s just to get the product of their elements.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/LazyIterate.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/LazyIterate.java
@@ -13,6 +13,7 @@ package org.eclipse.collections.impl.utility;
 import org.eclipse.collections.api.LazyIterable;
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.block.function.Function;
+import org.eclipse.collections.api.block.function.Function2;
 import org.eclipse.collections.api.block.predicate.Predicate;
 import org.eclipse.collections.api.block.procedure.Procedure;
 import org.eclipse.collections.api.tuple.Pair;
@@ -33,6 +34,7 @@ import org.eclipse.collections.impl.lazy.TakeWhileIterable;
 import org.eclipse.collections.impl.lazy.TapIterable;
 import org.eclipse.collections.impl.lazy.ZipIterable;
 import org.eclipse.collections.impl.lazy.ZipWithIndexIterable;
+import org.eclipse.collections.impl.tuple.Tuples;
 
 /**
  * LazyIterate is a factory class which creates "deferred" iterables around the specified iterables. A "deferred"
@@ -195,5 +197,33 @@ public final class LazyIterate
     public static <T> LazyIterable<T> tap(Iterable<T> iterable, Procedure<? super T> procedure)
     {
         return new TapIterable<>(iterable, procedure);
+    }
+
+    /**
+     * Create a deferred cartesian product of the two specified iterables.
+     *
+     * See {@link LazyIterate#cartesianProduct(Iterable, Iterable, Function2)} about performance and presence of duplicates.
+     *
+     * @since 10.0
+     */
+    public static <A, B> LazyIterable<Pair<A, B>> cartesianProduct(Iterable<A> iterable1, Iterable<B> iterable2)
+    {
+        return LazyIterate.cartesianProduct(iterable1, iterable2, Tuples::pair);
+    }
+
+    /**
+     * Create a deferred cartesian product of the two specified iterables.
+     *
+     * This operation has O(n^2) performance.
+     *
+     * The presence of duplicates in the resulting iterable is both dependent on the
+     * presence of duplicates in the two specified iterables, and on the behaviour
+     * of the terminating operation that is applied to the resulting lazy iterable.
+     *
+     * @since 10.0
+     */
+    public static <A, B, C> LazyIterable<C> cartesianProduct(Iterable<A> iterable1, Iterable<B> iterable2, Function2<? super A, ? super B, ? extends C> function)
+    {
+        return LazyIterate.flatCollect(iterable1, first -> LazyIterate.collect(iterable2, second -> function.value(first, second)));
     }
 }
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/utility/LazyIterateTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/utility/LazyIterateTest.java
@@ -11,17 +11,24 @@
 package org.eclipse.collections.impl.utility;
 
 import org.eclipse.collections.api.LazyIterable;
+import org.eclipse.collections.api.bag.MutableBag;
 import org.eclipse.collections.api.block.procedure.Procedure;
 import org.eclipse.collections.api.list.ImmutableList;
 import org.eclipse.collections.api.list.MutableList;
+import org.eclipse.collections.api.set.MutableSet;
+import org.eclipse.collections.api.tuple.Pair;
 import org.eclipse.collections.impl.block.factory.Predicates;
 import org.eclipse.collections.impl.block.factory.Procedures;
 import org.eclipse.collections.impl.block.function.AddFunction;
+import org.eclipse.collections.impl.factory.Bags;
+import org.eclipse.collections.impl.factory.Lists;
+import org.eclipse.collections.impl.factory.Sets;
 import org.eclipse.collections.impl.list.Interval;
 import org.eclipse.collections.impl.list.mutable.FastList;
 import org.eclipse.collections.impl.math.IntegerSum;
 import org.eclipse.collections.impl.math.Sum;
 import org.eclipse.collections.impl.test.Verify;
+import org.eclipse.collections.impl.tuple.Tuples;
 import org.junit.Assert;
 import org.junit.Test;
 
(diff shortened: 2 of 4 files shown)


victornoel:
@donraab I've updated the PR per your comments, thanks :)

unit-tests/src/test/java/org/eclipse/collections/impl/utility/LazyIterateTest.java
```diff
@@ -179,4 +184,77 @@ public void classIsNonInstantiable()
...
+
+    @Test
+    public void cartesianProduct_empty()
+    {
+        Assert.assertEquals(
+                Bags.mutable.empty(),
+                LazyIterate.cartesianProduct(
+                        Lists.mutable.with(1, 2),
+                        Lists.mutable.empty()).toBag());
+    }
```
motlin: Please test with duplicates. The reason this functionality only ever existed for sets is because it's easier to answer how it ought to behave in the case of duplicates (there aren't any). What should the cartesian product be between (1, 1) and ("same", "same")?

victornoel: hmm, very good question :) 

I would say that the result should be `((1, "same"), (1, "same"), (1, "same"), (1, "same"))` because I'm trusting the two collections to handle duplicates or not.

So the behaviour should be different depending on the collections given to the method and if you need no duplicates, then you should use sets... but maybe that's not compatible with the philosophy of EC? WDYT? @donraab @motlin

nikhilnanivadekar: As long as the behavior is true to the resulting collection in my opinion it’s fine. For lazy, we can make this to be non terminating and the resultant collection will take care of collection semantics.

nikhilnanivadekar: As long as the behavior is true to the resulting collection in my opinion it’s fine. For lazy, we can make this to be non terminating and the resultant collection will take care of collection semantics.

donraab: @victornoel Please write some tests with duplicates and outputs of Bag, Set, List.  Then it will be easier for us to see what the resulting behavior looks like.  It would be good to document our resulting agreement in the JavaDoc for the method.

victornoel: @donraab yep, good idea, I will do this!

victornoel: @donraab done

motlin:
I'm not sure if I want a lazy cartesian product to allow duplicates or
always unique. However if we do want uniquing behavior, take a look at
DistinctIterable. It is stateful and dedupes on the fly, while maintaining
laziness. In fact, we could delegate to DistinctIterable or call distinct()
before cartesianProduct.

Since this is a lazy implementation that can explode to O(n^2) results, I
think it would be good to create a test that would take ages if the
implementation weren't lazy. Something like this pseduocode:

Interval(1, 100000).asLazy().cartesianProduct(Interval(1,
100000)).contains(Pair(2, 2))

victornoel:
@motlin I see your point but maybe this means that this method is either poorly named or there are at least two distincts needs:
- in set theory, a cartesian product does not contain duplicates, since sets does not contain duplicates
- in general, a cartesian product is sometimes called a cross join (cf SQL) and in that case nothing is really said about duplicates in the result, everything depends on the two input collections (if they have duplicates, then there can be duplicate in the cross join).

As @nikhilnanivadekar said, I think it can be acceptable to consider that the behaviour is determined both by the input collections and the concretization of the lazy iterable (via `toSet`, `toList`, `toBag`, etc).

So I wouldn't want to impose on users of the method `LazyIterable.cartesianProduct` that no duplicate should exists.
Maybe we should thus change the name of this method in `LazyIterable` and call it… `crossJoin`?

nikhilnanivadekar:
Sorry to bring physics here but isn’t Cartesian product same as cross join same as cross product in this case? May be I am missing something basic here, but the availability of dups or only unique should be delegated to the final terminating collection. We should be explicit about this contract in javadoc. This is especially important when we have terminating operations like detect or count, when we have terminating operations like detect or count we should go back to the behavior of input collections. 
@victornoel I understand this is a bit more work, what do you think of this approach?
Cc @donraab , @motlin

nikhilnanivadekar:
Yup that’s it. But, please add examples in javadoc to show how it’ll behave. Documentation is extra work 😊

victornoel:
@nikhilnanivadekar ok, in that case I agree with you yes :) I will wait for feedbacks from @motlin and @donraab then!

motlin:
I thought of another valuable test. Create a bag with like a million copies
of “left” and another with a million copies of “right”, take their lazy
Cartesian product, and call toBag(). It should complete in constant time
and space.

victornoel:
@motlin so you want me to add those two tests you were talking about? Those are not typical unit tests, where should they go? I see there is a performance tests module, but I havn't look into it yet.

@motlin @nikhilnanivadekar @donraab are we good with current implementation and naming of the method? I will just add some javadoc to it to explain what @nikhilnanivadekar said about non-termination (even though it is kind of obvious since the result is a `LazyIterable` :) and that the content (in terms of duplicates) will depend on the input and resulting operations (this one is not obvious).

nikhilnanivadekar:
@victornoel I was asking only about the not so obvious part (resulting and input collections). Since the behavior differs depending on which terminating operation is called.

Regarding adding the test, I do not know where is a good place to add either. Performance tests are different and I do not think we can add this test there, but I might be wrong. You can just add it in `LazyIterateTest` and see how long it takes for the test to complete. If it takes up to a minute I would be fine with it, add a comment on why the test exists. If it takes more than a minute, we can look at reducing the size from a million to say 10,000 to run. [Fail the assertion if it takes more than a specific time](https://github.com/junit-team/junit4/wiki/timeout-for-tests).
This is an interesting API to have, and I am looking forward for it to be merged 👍

motlin:
It's a unit test, not a performance test. I know @Nikhil said 1 minute, but
I'd be fine with these taking up to 10 milliseconds :)

The one with two Bags in particular should be instantaneous.

victornoel:
@motlin are you sure?

I wrote this:
```
    @Test(timeout=10000)
    public void cartesianProductIsNotSlow()
    {
        MutableBag<Integer> iterable1 = Bags.mutable.with();
        iterable1.addOccurrences(1, 100000000);
        MutableBag<Integer> iterable2 = Bags.mutable.with();
        iterable1.addOccurrences(2, 100000000);
        MutableBag<Pair<Integer, Integer>> expectedCartesianProduct = Bags.mutable.with();
        expectedCartesianProduct.addOccurrences(Tuples.pair(1, 2), 100000000*100000000);
        Assert.assertEquals(expectedCartesianProduct, LazyIterate.cartesianProduct(iterable1, iterable2).toBag());
    }
```

I don't think it can be instantaneous (or it means we should implement `cartesianProduct` differently then…).

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/LazyIterate.java
```diff
@@ -196,4 +198,14 @@ private LazyIterate()
...
     }
+
+    public static <A, B> LazyIterable<Pair<A, B>> cartesianProduct(Iterable<A> iterable1, Iterable<B> iterable2)
+    {
+        return LazyIterate.cartesianProduct(iterable1, iterable2, Tuples::pair);
+    }
+
+    public static <A, B, C> LazyIterable<C> cartesianProduct(Iterable<A> iterable1, Iterable<B> iterable2, Function2<A, B, C> function)
+    {
+        return LazyIterate.flatCollect(iterable1, first -> LazyIterate.collect(iterable2, second -> function.value(first, second)));
```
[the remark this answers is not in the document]
victornoel: @motlin so what is the conclusion on this? Is it ok like this or should we stop writing procedural code and introduce a `cartesianProduct` (or equivalent name) method on collections interfaces in an object-oriented way to handle each implementation with respect to the concerned type?

I believe this is out of scope of this PR in the latter case... `LazyIterate.cartesianProduct` is an helper method just to simplify life of people because writing this flatCollect/collect/apply is common and boring to write.

Honestly having a `cartesianProduct` for `Bag`s or `Set`s does not interest me at all, I just want to ease my life because I often have to cross join multiple collections.

So should we continue with this? If yes I will add something about the complexity in the javadoc at least. I can't make this decision, so I'm referring to you all :)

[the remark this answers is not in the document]
motlin: I don't think those are the only two choices. The implementation for Sets returns a `FlatCollectIterable`. That's fine for sets, because they're already unique, and no additional work is required than the work done inside the lambda.

I think you should considering implementing LazyIterate.cartesianProduct() by "simply" returning a new type `CartesianProductIterable`. That would give you a new type with methods like toSet() and toBag() that you could optimize. For sets, the optimization could be to delegate to distinct() before proceeding. For bags, you could delegate to forEachWithOccurrences if available, or toBag followed by forEachWithOccurrences if not. It's a little tricky, and a lot of work for an obscure method, so I'd be open to a simpler implementation plus filing a follow-up issue for the optimization work. What do you think?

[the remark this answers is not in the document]
donraab: The `CartesianProductIterable` could also be add as an optimization later, as long as the method returns `LazyIterable`, which it does currently.

unit-tests/src/test/java/org/eclipse/collections/impl/utility/LazyIterateTest.java
donraab: @victornoel Thanks!

victornoel:
@nikhilnanivadekar @donraab sorry I thought I answered you the other day to tell you it was already rebased. Since then master changed, so I just did it :)

nikhilnanivadekar:
Awesome, thanks for your contribution!

PR #673, opened 2018-12-18
Change boxed result array to primitive array to avoid unnecessary box…

donraab:
…ing in primitive reduce.


donraab:
@victornoel I noticed that the result array in reduce was a boxed array.  I replaced it with a primitive array.  Let me know if there is any issue with this.

nikhilnanivadekar:
LGTM

victornoel:
@donraab arg, sorry about that :)

PR #674, opened 2018-12-18
add SafeVarargs annotation to of and with methods. Closes #284

### LINKED ISSUE #284: Add @SafeVarargs to Factory methods "of" and "with"
If you call `Lists.mutable.of(Lists.mutable.empty(), Lists.mutable.empty());`, java will give an error about  unchecked generic array creation for varargs parameter.

As far as I can tell, you aren't doing anything potentially unsafe (like what is shown [here](https://docs.oracle.com/javase/7/docs/api/java/lang/SafeVarargs.html)), so I think you can add the `@SafeVarargs` to all of the methods that take a generic vararg parameter.

sendilkumarn:
Fixes #284

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MutableListFactoryImpl.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MutableListFactoryImpl.java
@@ -27,6 +27,7 @@ public class MutableListFactoryImpl implements MutableListFactory
         return FastList.newList();
     }
 
+    @SafeVarargs
     @Override
     public <T> MutableList<T> with(T... items)
     {
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableBagFactoryImpl.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableBagFactoryImpl.java
@@ -49,14 +49,16 @@ public class ImmutableBagFactoryImpl implements ImmutableBagFactory
         return new ImmutableSingletonBag<>(element);
     }
 
+    @SafeVarargs
     @Override
-    public <T> ImmutableBag<T> of(T... elements)
+    public final <T> ImmutableBag<T> of(T... elements)
     {
         return this.with(elements);
     }
 
+    @SafeVarargs
     @Override
-    public <T> ImmutableBag<T> with(T... elements)
+    public final <T> ImmutableBag<T> with(T... elements)
     {
         if (elements == null || elements.length == 0)
         {
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/MultiReaderMutableBagFactory.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/MultiReaderMutableBagFactory.java
@@ -25,7 +25,8 @@ public enum MultiReaderMutableBagFactory implements MultiReaderBagFactory
     {
         return MultiReaderHashBag.newBag();
     }
-
+    
+    @SafeVarargs
     @Override
     public <T> MultiReaderBag<T> with(T... items)
     {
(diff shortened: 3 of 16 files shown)


sendilkumarn:
@nikhilnanivadekar Done 👍 

>I do not think we should add final keyword on the methods. 

The methods should either be final or static

nikhilnanivadekar:
Won’t this break the existing users who have extended the Impl classes? Also how does it impact the service loader work which @motlin did?

sendilkumarn:
> Won’t this break the existing users who have extended the Impl classes?  

The other option is considering it as `static` and rewriting the entire classes/ interfaces based on that.

> Also how does it impact the service loader work which @motlin did?

I am not sure about this one.

motlin:
I think it’s fine to add final. I’d prefer making the entire classes final,
not individual methods. The classes are not designed for extension.

I’m not sure why we think it’s ok to add SafeVarArgs. I admit I don’t
understand it well.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MutableListFactoryImpl.java
```diff
@@ -36,14 +36,16 @@
...
     {
         return this.with(items);
     }
 
+    @SafeVarargs
     @Override
-    public <T> MutableList<T> with(T... items)
+    public final <T> MutableList<T> with(T... items)
     {
         return FastList.newListWith(items);
```
motlin: My hypothesis is that the immutable factories are in fact safe, and that the mutable factories are not safe. I am most confident that this is a good example of unsafe varargs. I would ask that we reason through what we think is safe or not, rather than go with my guess that the immutable factories are ok.

sendilkumarn: @motlin I had the same thought when implementing that. I agree that we should reason through  👍

sendilkumarn: And I do agree that the whole class should be final because It makes sense and I don't expect anybody to override them. ( While I debated a lot with myself about this, I am not finding a proper use case.)  The one obvious thing is that users may need to modify some aspects of the underlying data structure, they can extend.

nikhilnanivadekar:
Let's make these classes final and not have the safevarargs annotation. I think that is my recommendation. The warnings have to be suppressed by each user in their code, since only they can know if it is safe or not.
@motlin thoughts?

motlin:
I don't think it's up to users. The @SafeVarArgs annotation says something
specific to me, which is that you won't get class-cast or array-store
exceptions, not even if you use raw types. I think it's ok to make these
classes final (in a major version of Eclipse Collections). I think it's
probably correct to add @SafeVarArgs to our immutable factories. I think
it's probably incorrect to add @SafeVarArgs to our mutable factories. I'd
only want to add the annotations if we can build confidence that it's truly
safe.

sendilkumarn:
@motlin I am more interested to learn why you (think) say it will cause problems in mutable factories. 

But AFAICT, quoting from `Effective Java` ;) 

```if the varargs parameter array is used only to transmit a variable number of arguments from the caller to the method—which is, after all, the purpose of varargs—then the method is safe.```

We are just transmitting and I am also pretty sure that we are not trying to typecast anywhere. So at this point, it is generally safe.

Making the class final, I do agree it should be done in a `major` version. At this moment, making this final will not affect most of our users.

motlin:
Even Arrays.asList(...) looks suspicious to me because you can pass in an array of a subtype, call set() on the returned list using a supertype, and get ArrayStoreException. But I guess you can't get ClassCastException which would be far worse.

Since Effective Java mentions Arrays.asList(), and the method is annotated @SafeVarargs, maybe I'm overthinking it. It seems like there's already a culture of suppressing warnings on similar methods.

donraab:
@motlin @nikhilnanivadekar Any further comments or concerns here?

sendilkumarn:
@motlin & @nikhilnanivadekar can we pursue this? Is there something  pending here?

sendilkumarn:
Closing this, since it is super inactive

PR #682, opened 2019-01-08
Use Java 8 Base64 instead of apache

mohrezaei:


motlin:
This is great, thank you.

mohrezaei:
ummm, that wasn't ready :confused: . I was waiting for the checks to pass so I could fix them (they break with high probability), which is why I didn't sign the commit message.

nikhilnanivadekar:
This failed ip validation as well. I am going to revert this commit

PR #684, opened 2019-01-08
Revert "Use Java 8 Base64 instead of apache"

nikhilnanivadekar:
Reverts eclipse/eclipse-collections#682


nikhilnanivadekar:
@mohrezaei can you merge this revert. Or approve and I’ll merge it. The PR needs to be reverted as ip validation did not pass. So for tracking we need to revert and then create a commit which is signed off. Thanks!

PR #685, opened 2019-01-08
Use Java 8 Base64 instead of apache

mohrezaei:


mohrezaei:
Everything passed, so this should be good to merge.

PR #699, opened 2019-02-16
Adding builder style construction to MapFactories

Chris-Almond:
This small convenient new feature empowers EC Map users to construct/instantiate maps with as many entries as they need, surpassing the current limit of 4. This is achieved by leveraging the builder pattern. The common use case for this feature is when users want/need to create static maps that hold configuration mappings within their applications. This feature eliminates the need to hydrate such maps inside static blocks which commonly induces code smell.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/map/ImmutableMapFactory.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/map/ImmutableMapFactory.java
@@ -13,6 +13,7 @@ package org.eclipse.collections.api.factory.map;
 import java.util.Map;
 
 import org.eclipse.collections.api.map.ImmutableMap;
+import org.eclipse.collections.api.map.builder.MapBuilder;
 
 public interface ImmutableMapFactory
 {
@@ -67,4 +68,6 @@ public interface ImmutableMapFactory
     <K, V> ImmutableMap<K, V> ofAll(Map<K, V> map);
 
     <K, V> ImmutableMap<K, V> withAll(Map<K, V> map);
+
+    <K, V> MapBuilder<K, V> newBuilderWith(K key, V value);
 }
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/map/MutableMapFactory.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/map/MutableMapFactory.java
@@ -14,6 +14,7 @@ import java.util.Map;
 
 import org.eclipse.collections.api.map.MapIterable;
 import org.eclipse.collections.api.map.MutableMap;
+import org.eclipse.collections.api.map.builder.MapBuilder;
 
 public interface MutableMapFactory
 {
@@ -77,4 +78,6 @@ public interface MutableMapFactory
     <K, V> MutableMap<K, V> ofMapIterable(MapIterable<? extends K, ? extends V> mapIterable);
 
     <K, V> MutableMap<K, V> withMapIterable(MapIterable<? extends K, ? extends V> mapIterable);
+
+    <K, V> MapBuilder<K, V> newBuilderWith(K key, V value);
 }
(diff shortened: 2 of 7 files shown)


donraab:
@Chris-Almond Thank you for the submission. There is another existing approach to constructing maps that is similar to a builder. You can use a `MutableMap` as a builder for itself or for an `ImmutableMap`. It is not as well known as using the static factory methods. 

```
@Test
public void mutableMapAsbuilder()
{
    MutableMap<Integer, String> map =
            Maps.mutable.<Integer, String>empty()
                    .withKeyValue(1, "One")
                    .withKeyValue(2, "Two")
                    .withKeyValue(3, "Three")
                    .withKeyValue(4, "Four")
                    .withKeyValue(5, "Five");
    Verify.assertSize(5, map);
    Verify.assertContainsAllKeyValues(map, 1, "One", 2, "Two", 3, "Three", 4, "Four", 5, "Five");
}
```

Chris-Almond:
> @Chris-Almond Thank you for the submission. There is another existing approach to constructing maps that is similar to a builder. You can use a `MutableMap` as a builder for itself or for an `ImmutableMap`. It is not as well known as using the static factory methods.
> 
> ```
> @Test
> public void mutableMapAsbuilder()
> {
>     MutableMap<Integer, String> map =
>             Maps.mutable.<Integer, String>empty()
>                     .withKeyValue(1, "One")
>                     .withKeyValue(2, "Two")
>                     .withKeyValue(3, "Three")
>                     .withKeyValue(4, "Four")
>                     .withKeyValue(5, "Five");
>     Verify.assertSize(5, map);
>     Verify.assertContainsAllKeyValues(map, 1, "One", 2, "Two", 3, "Three", 4, "Four", 5, "Five");
> }
> ```

Not going to lie... my initial reaction: 

_Can't_ believe I missed that 😄 . 

In fact, while just trying it out now, I see that it can be achieved without the explicit typing if we start with the `with()` function.
>```java
>    ImmutableMap<Integer, String> stepsToSuccess = Maps.mutable.with(1, "Read The Docs")
>                .withKeyValue( 2, "Double check the API")
>                .withKeyValue( 3, "???")
>                .withKeyValue( 4, "Profit")
>                .withKeyValue( 5, "Thank the author(s)").toImmutable();
>```

Thanks for letting me know.

_Hope things are well!_
-Chris

donraab:
@Chris-Almond To be fair, it is not obvious that this functionality exists, and probably could use a blog or two and maybe specific coverage in our reference guide to help folks discover it. 

I covered some of them in this [blog](https://medium.com/@donraab/preposition-preference-1f1c709b098b), but didn't provide any examples with Maps.

Your submission motivated me to tweet an example with primitive immutable Maps yesterday. 

```
@Test
public void numbers()
{
    ObjectIntMap<String> numbers =
            ObjectIntMaps.mutable.<String>empty()
                    .withKeyValue("Zero", 0)
                    .withKeyValue("One", 1)
                    .withKeyValue("Two", 2)
                    .withKeyValue("Three", 3)
                    .withKeyValue("Four", 4)
                    .withKeyValue("Five", 5)
                    .withKeyValue("Six", 6)
                    .withKeyValue("Seven", 7)
                    .withKeyValue("Eight", 8)
                    .withKeyValue("Nine", 9)
                    .withKeyValue("Ten", 10)
                    .toImmutable();

    Assert.assertEquals(0, numbers.get("Zero"));
    Assert.assertEquals(1, numbers.get("One"));
    Assert.assertEquals(10, numbers.get("Ten"));
}
 
```
Thank you for uncovering this lesser known API!

donraab:
@Chris-Almond I'm going to add another in the series of [As a matter of Factory](https://medium.com/@donraab/as-a-matter-of-factory-part-1-mutable-75cc2c5d72d9) blog series to cover this lesser known API. These were the first two blogs in the series.

nikhilnanivadekar:
I think we should have the builder pattern which Chris mentioned. Two reasons for it:
1) withKeyValue() ends up creating intermediate maps 
2) discoverability, this will make that EC is similar to Guava which follows the builder pattern
Thoughts?

donraab:
1. With MutableMap, withKeyValue mutates the existing map.  newWithKeyValue on an ImmutableMap has to copy (thus the addition of new)
2. I don't think it is worth the extra code to add.

motlin:
I use empty().withKeyValue() myself. I think the name makes it a little
hard for newcomers to find. One idea I have for an alias is putThis(key,
value). It’s not really accurate but it’s a little like sortThis,
shuffleThis, reverseThis.

PR #708, opened 2019-04-14
Add test coverage for zip with an iterable.

donraab:


nikhilnanivadekar:
@donraab For the test coverage PRs as long as the build passes, there is increase in test coverage and it’s rebased on to master feel free to merge. I don’t think we need to wait on a code review for that. Thanks for reducing the tech debt!

PR #717, opened 2019-04-24
Reduce duplicate readlock and writelock code in MultiReaderFastList, MultiReaderHashBag and MultiReaderUnifiedSet.

donraab:

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Goldman Sachs and others.
+ * Copyright (c) 2019 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -16,6 +16,7 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Spliterator;
+import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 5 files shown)


donraab:
This was a thought experiment I wanted to run by @motlin and @nikhilnanivadekar. Feedback welcome. If this works here, then we can apply the same approach to the other two MultiReader collections to further reduce duplicate code.

donraab:
Nice! The Scala tests failed for MultiReaderFastList testing that the appropriate readlocks and writelocks were taken for each method. I had changed some code that should have used a writelock and instead used a readlock. I added a few missing sortByPrimitive methods to the Scala test.

donraab:
Ok build is passing now. Checkstyle and FindBugs had a couple complaints which I addressed. @motlin  and @nikhilnanivadekar let me know what you think.

mohrezaei:
Have you looked at performance and garbage creation with these changes?

donraab:
No, but I could do that.  I think I might be able to make some of the calls not generate instances with the lambda by using the same approach we use with `with` methods.  Just requires me writing some more code. I can do that after I return from my vacation.

mohrezaei:
wouldn't this https://stackoverflow.com/questions/6965731/are-locks-autocloseable be a more straight forward solution? I think the wrapper could be created just once, but we'd have to change the signature of the getLock() method to return the wrapper instead.

donraab:
How would the LockWrapper close method be implemented?  It is a ReentrantReadWriteLock.

mohrezaei:
Two wrappers, one for read, one for write.

donraab:
Right, figured it out after I commented.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
```diff
@@ -102,6 +103,101 @@ protected void unlockReadLock()
         this.getLock().readLock().unlock();
     }
 
+    protected <R> R applyReadLock(Function0<R> function)
```
sendilkumarn: Can we try using `execute-around` pattern for these methods too?

donraab:
I'm updating all of the methods to use the new LockWrapper approach. I am changing MultiReaderFastList, MultiReaderHashBag and MultiReaderUnifiedSet to use this new pattern.

donraab:
I refactored MultiReaderFastList, MultiReaderHashBag and MultiReaderUnifiedSet to use the try with resources approach Moh suggested.  Can you have a look @mohrezaei @motlin @nikhilnanivadekar @sendilkumarn?  I think it is working out pretty well.  I got rid of the lambda approach altogether, because it's easier to read the consistent surround approach than having multiple lambda with parameter methods in order to leverage method references.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
```diff
@@ -2576,4 +1995,52 @@ public void appendString(Appendable appendable, String start, String separator,
             return map;
         }
     }
+
+    public class ReadWriteLockWrapper
```
mohrezaei: add static

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
```diff
@@ -2576,4 +1995,52 @@ public void appendString(Appendable appendable, String start, String separator,
...
+         * This method must be wrapped in a try block.
+         */
+        public LockWrapper acquireWriteLock()
+        {
+            this.writeLock.lock.lock();
+            return this.writeLock;
+        }
+    }
+
+    public class LockWrapper implements AutoCloseable
```
mohrezaei: add static

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
```diff
@@ -78,231 +79,139 @@
...
 {
-    protected abstract MutableCollection<T> getDelegate();
-
-    protected abstract ReadWriteLock getLock();
-
-    protected void acquireWriteLock()
-    {
-        this.getLock().writeLock().lock();
-    }
+    protected transient ReadWriteLock lock;
```
mohrezaei: do we still need this field?

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
```diff
@@ -78,231 +79,139 @@
...
-    }
-
-    protected void acquireReadLock()
-    {
-        this.getLock().readLock().lock();
-    }
+    protected abstract MutableCollection<T> getDelegate();
 
-    protected void unlockReadLock()
+    protected ReadWriteLock getLock()
```
mohrezaei: do we need this method?

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
donraab: There are other parallel classes which still depend on the field.  I will get rid of it once I get a chance to refactor those classes as well.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
donraab: Good catch, will do.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
donraab: I'll check for usages and remove if not used.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
donraab: done

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
donraab: removed

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
donraab: done

nikhilnanivadekar:
There are `finally` blocks removed from the code. Will that be ok? If it will be ok, how and when is the lock released? Sorry, if I am missing something obvious.

The build failures seem to be due to an issue with JDK download and is not related to the PR. Since the Oracle JDK builds for the failed Open JDK builds have gone through, I am fine from a build perspective.

mohrezaei:
@nikhilnanivadekar it's [try with resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)

nikhilnanivadekar:
@mohrezaei yup... it was so obvious in hindsight! :)

PR #734, opened 2019-07-18
Use maven's --no-transfer-progress option to reduce noise in the build logs.

motlin:


nikhilnanivadekar:
@motlin what is pending for PR?

motlin:
I think it's ready to go if you approve. The `--no-transfer-progress` was added in maven 3.6.1 and dramatically shrinks the build logs. We're already using 3.6.1 in travis.

PR #748, opened 2019-09-01
Implement #744 : Add range method to Interval

### LINKED ISSUE #744: Implement Interval.range() with similar behavior as IntStream.range()
This issue can be merged with #743 .

govi20:
Signed-off-by: govi20 <govindasakhare20@gmail.com>

Add range() method to Interval which has a behavior similar to the IntStream.range

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/Interval.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/Interval.java
@@ -85,6 +85,27 @@ public final class Interval
     }
 
     /**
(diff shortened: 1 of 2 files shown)


nikhilnanivadekar:
@govi20 sorry, I missed replying to your question. Interval is an interesting class (data structure). It behaves lazy as well as eager in some patterns. In order to create an empty Interval you might want to create a private sub-class called `EmptyInterval` and override the iteration patterns.

donraab:
@isank I think you just need to add the test @nikhilnanivadekar requested and he'll probably approve and merge.

> Need a test to compare with the behavior of `IntStream.range()`

PR #754, opened 2019-09-16
Add new APIs, `ofOccurrences` and `withOccurrences`, to bag mutable and immutable factories 

### LINKED ISSUE #739: Bag factories with occurrences
Something like the `Maps.immutable.of` factory, but for bags:

```java
Bags.immutable.ofWithOccurrences(first, 3,
                                 second, 5);
```

jiminhsieh:
Fix #739

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/bag/ImmutableBagFactory.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/bag/ImmutableBagFactory.java
@@ -13,7 +13,9 @@ package org.eclipse.collections.api.factory.bag;
 import java.util.stream.Stream;
 
 import org.eclipse.collections.api.bag.ImmutableBag;
+import org.eclipse.collections.api.bag.MutableBag;
 import org.eclipse.collections.api.factory.Bags;
+import org.eclipse.collections.api.tuple.primitive.ObjectIntPair;
 
 /**
  * A factory which creates instances of type {@link ImmutableBag}.
@@ -47,7 +49,110 @@ public interface ImmutableBagFactory
      */
     <T> ImmutableBag<T> of(T... elements);
 
+    /**
+     * Same as {@link #withOccurrences(Object, int)}.
+     *
+     * @since 10.3
+     */
+    default <T> ImmutableBag<T> ofOccurrences(T element, int occurrence)
+    {
+        return this.withOccurrences(element, occurrence);
+    }
+
+    /**
+     * Same as {@link #withOccurrences(Object, int, Object, int)}.
+     *
+     * @since 10.3
+     */
+    default <T> ImmutableBag<T> ofOccurrences(T element1, int occurrence1, T element2, int occurrence2)
+    {
+        return this.withOccurrences(element1, occurrence1, element2, occurrence2);
+    }
+
+    /**
+     * Same as {@link #withOccurrences(Object, int, Object, int, Object, int)}.
+     *
+     * @since 10.3
+     */
+    default <T> ImmutableBag<T> ofOccurrences(T element1, int occurrence1, T element2, int occurrence2, T element3, int occurrence3)
+    {
+        return this.withOccurrences(element1, occurrence1, element2, occurrence2, element3, occurrence3);
+    }
+
+    /**
+     * Same as {@link #withOccurrences(Object, int, Object, int, Object, int, Object, int)}.
+     *
+     * @since 10.3
+     */
+    default <T> ImmutableBag<T> ofOccurrences(T element1, int occurrence1, T element2, int occurrence2, T element3, int occurrence3, T element4, int occurrence4)
+    {
+        return this.withOccurrences(element1, occurrence1, element2, occurrence2, element3, occurrence3, element4, occurrence4);
+    }
+
+    /**
+     * Same as {@link #withOccurrences(ObjectIntPair[])}
+     *
+     * @since 10.3
+     */
+    default <T> ImmutableBag<T> ofOccurrences(ObjectIntPair<T>... elementsWithOccurrences)
+    {
+        return this.withOccurrences(elementsWithOccurrences);
+    }
+
     <T> ImmutableBag<T> with(T... elements);
+
+    /**
+     * @since 10.3
+     */
+    default <T> ImmutableBag<T> withOccurrences(T element, int occurrence)
+    {
+        MutableBag<T> mutableBag = Bags.mutable.withOccurrences(element, occurrence);
+        return mutableBag.toImmutable();
+    }
+
+    /**
+     * @since 10.3
+     */
+    default <T> ImmutableBag<T> withOccurrences(T element1, int occurrence1, T element2, int occurrence2)
+    {
+        MutableBag<T> mutableBag = Bags.mutable.withOccurrences(
+                element1, occurrence1,
+                element2, occurrence2);
+        return mutableBag.toImmutable();
+    }
+
+    /**
+     * @since 10.3
+     */
+    default <T> ImmutableBag<T> withOccurrences(T element1, int occurrence1, T element2, int occurrence2, T element3, int occurrence3)
+    {
+        MutableBag<T> mutableBag = Bags.mutable.withOccurrences(
+                element1, occurrence1,
+                element2, occurrence2,
+                element3, occurrence3);
+        return mutableBag.toImmutable();
+    }
+
+    /**
+     * @since 10.3
+     */
+    default <T> ImmutableBag<T> withOccurrences(T element1, int occurrence1, T element2, int occurrence2, T element3, int occurrence3, T element4, int occurrence4)
+    {
+        MutableBag<T> mutableBag = Bags.mutable.withOccurrences(
+                element1, occurrence1,
+                element2, occurrence2,
+                element3, occurrence3,
+                element4, occurrence4);
+        return mutableBag.toImmutable();
+    }
+
+    /**
+     * @since 10.3
+     */
+    default <T> ImmutableBag<T> withOccurrences(ObjectIntPair<T>... elementsWithOccurrences)
+    {
+        return Bags.mutable.withOccurrences(elementsWithOccurrences).toImmutable();
+    }
 
     /**
      * Same as {@link #withAll(Iterable)}.
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/bag/MutableBagFactory.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/bag/MutableBagFactory.java
@@ -13,6 +13,7 @@ package org.eclipse.collections.api.factory.bag;
 import java.util.stream.Stream;
 
 import org.eclipse.collections.api.bag.MutableBag;
+import org.eclipse.collections.api.tuple.primitive.ObjectIntPair;
 
 /**
  * A factory which creates instances of type {@link MutableBag}.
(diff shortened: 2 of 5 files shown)


jiminhsieh:
Signed-off-by: Jimin Hsieh jimin.hsieh.engineer@gmail.com

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/bag/ImmutableBagFactory.java
```diff
@@ -49,6 +51,18 @@
 
     <T> ImmutableBag<T> with(T... elements);
 
+    <T> ImmutableBag<T> ofWithOccurrences(T element, int occurrence);
+
+    <T> ImmutableBag<T> ofWithOccurrences(T element1, int occurrence1, T element2, int occurrence2);
```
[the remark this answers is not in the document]
jiminhsieh: I think we should have `withOccurrences` and `ofOccurrences` since we have `of` and `with` those 2 APIs. In this situation, the users will be much easier to find the new APIs.

jiminhsieh:
> it would be nice to add withOccurrences/withoutOccurrences to MutableBag and newWithOccurrences/newWithoutOccurrences to ImmutableBag.

I can understand the behavior of `withOccurrences` and `newWithOccurrences`, but I don't know the expected behavior or APIs for `withoutOccurrences` and `newWithoutOccurrences`. @donraab Could you explain more about those APIs with `withoutOccurrences`? Thanks.

I am thinking should I implement the `withOccurrences` for mutable bag factories? I already implement those APIs for the immutable bag factories, it's probably better to implement for mutable too.

donraab:
@jiminhsieh This is what I am thinking for the implementation on `MutableBag`.

```
default MutableBag<T> withOccurrences(T element, int occurrences)
{
    this.addOccurrences(element, occurrences);
    return this;
}

default MutableBag<T> withoutOccurrences(T element, int occurrences)
{
    this.removeOccurrences(element, occurrences);
    return this;
}
```

Does this make sense?

donraab:
@jiminhsieh the `withOccurrences` and `withoutOccurrences` could either added in a separate PR, or could be used to implement `withOccurrences` and `ofOccurences`.  I'll leave it up to you.  I agree with you that to be consistent we should have both `withOccurrences` and `ofOccurrences` on the factory classes.

jiminhsieh:
I will implement this PR with `withOccurrences` and `ofOccurrences` to mutable and immutable factories.

>  I was also thinking it would be nice to add withOccurrences/withoutOccurrences to MutableBag and newWithOccurrences/newWithoutOccurrences to ImmutableBag.

I think this should be another PR since this includes a new API, `withoutOccurrences`, and I need to implement for MutableBag and ImmutableBag instead of ImmutableBagFactory and MutableBagFactory which I will change this time.

jiminhsieh:
@donraab I'm thinking about rewriting history to fewer commits since I will be much easier for you to review. Is that ok with you?

nikhilnanivadekar:
I would prefer if you could squash all the commits to a single commit per feature. Thanks again for your contribution!

donraab:
> @donraab I'm thinking about rewriting history to fewer commits since I will be much easier for you to review. Is that ok with you?

Yes, please squash all the commits.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/bag/ImmutableBagFactory.java
```diff
@@ -47,8 +49,50 @@
...
+
+    <T> ImmutableBag<T> withOccurrences(T element1, int occurrence1, T element2, int occurrence2);
+
+    <T> ImmutableBag<T> withOccurrences(T element1, int occurrence1, T element2, int occurrence2, T element3, int occurrence3);
+
+    <T> ImmutableBag<T> withOccurrences(T element1, int occurrence1, T element2, int occurrence2, T element3, int occurrence3, T element4, int occurrence4);
+
+    <T> ImmutableBag<T> withOccurrences(ObjectIntPair<T>... elementsWithOccurrences);
+
+    ImmutableBag<Integer> withOccurrences(IntIntPair... elementsWithOccurrences);
```
donraab: I do not understand why `withOccurrences(IntIntPair... elementsWithOccurrences) `is necessary.  If there is a use case for this, I would rather add the factory methods to the primitive factories, so this returns an `ImmutableIntBag`.

I understand why you might want to replace this with `ObjectIntPair` here to return an `ImmutableBag<T>`, but `IntIntPair` returning a boxed `ImmutableBag<Integer>` does not make sense.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/bag/MutableBagFactory.java
```diff
@@ -48,8 +50,50 @@
...
+
+    <T> MutableBag<T> withOccurrences(T element1, int occurrence1, T element2, int occurrence2);
+
+    <T> MutableBag<T> withOccurrences(T element1, int occurrence1, T element2, int occurrence2, T element3, int occurrence3);
+
+    <T> MutableBag<T> withOccurrences(T element1, int occurrence1, T element2, int occurrence2, T element3, int occurrence3, T element4, int occurrence4);
+
+    <T> MutableBag<T> withOccurrences(ObjectIntPair<T>... elementsWithOccurrences);
+
+    MutableBag<Integer>  withOccurrences(IntIntPair... elementsWithOccurrences);
```
donraab: Same as above.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/bag/ImmutableBagFactory.java
jiminhsieh: > `withOccurrences(IntIntPair... elementsWithOccurrences)`

The reason I provided this API is I already provided the following API:
```java
withOccurrences(ObjectIntPair<T>... elementsWithOccurrences);
```

Above API will need to use:
```java
PrimitiveTuples.pair("2", 2)
```

Which will return `ObjectIntPair`.
If I use this:
```java
PrimitiveTuples.pair(2, 2)
```

Which will return `IntIntPair`.
That's why I have `withOccurrences(IntIntPair... elementsWithOccurrences);`


> IntIntPair returning a boxed ImmutableBag<Integer> doesn't make sense to me.


If the APIs with varargs make sense to you, I could see the possibility of returning `ImmutableIntBag` instead of the current one. If there is a better collection than `ImmutableIntBag`, please let me know too. Thanks.

donraab: This method would make more sense on `IntBags` in the `MutableIntBagFactory` and `ImmutableIntBagFactory`.

For Example:
```
IntBag bag = IntBags.mutable.withOccurrences(
    PrimitiveTuples.pair(1, 10), 
    PrimitiveTuples.pair(2, 20));
```
There is also `ShortBag`, `ByteBag`, `FloatBag`, `LongBag`, `DoubleBag`, `BooleanBag`.  They should have the equivalent methods.  But I would recommend doing this in a separate PR.

donraab: Any update on this @jiminhsieh?

jiminhsieh: @donraab I don't have any update yet, I was busy at the end of last year. However, this PR will be the first finished PR for 2020. :)

donraab:
Hi @jiminhsieh please squash the commits again.  Thanks!

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/bag/MutableBagFactory.java
```diff
@@ -48,8 +49,43 @@
...
+    }
+
+    default <T> MutableBag<T> ofOccurrences(ObjectIntPair<T>... elementsWithOccurrences)
+    {
+        return this.withOccurrences(elementsWithOccurrences);
+    }
+
     <T> MutableBag<T> with(T... elements);
 
+    <T> MutableBag<T> withOccurrences(T element, int occurrence);
```
donraab: We have a choice to make here.  The specification and implementation of this method looks fine, but will require the next version of Eclipse Collections to be 11.0 because the `withOccurrences` methods do not have default implementations.  As long as @nikhilnanivadekar is ok with this, then I am fine.  I would suggest adding `@since 11.0` tags to the interfaces for `MutableBagFactory` and `ImmutableBagFactory` then as this will require a major release.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableBagFactoryImpl.java
```diff
@@ -105,4 +108,55 @@
...
+        bag.addOccurrences(element1, occurrence1);
+        bag.addOccurrences(element2, occurrence2);
+        bag.addOccurrences(element3, occurrence3);
+        return bag.toImmutable();
+    }
+
+    @Override
+    public <T> ImmutableBag<T> withOccurrences(T element1, int occurrence1, T element2, int occurrence2, T element3, int occurrence3, T element4, int occurrence4)
+    {
+        MutableBag<T> bag = new HashBag<>(occurrence1 + occurrence2 + occurrence3 + occurrence4);
```
motlin: The pre-sizings are wrong. This map will have four elements, and could be pre-sized to 4, or not pre-sized. This implementation could easily lead to OutOfMemoryError and must be fixed.

My recommendation would be not to pre-size at all, nor call a constructor. You could replace this with a call to empty(), and then the implementation could be pulled up as a default method.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableBagFactoryImpl.java
```diff
@@ -105,4 +108,55 @@
...
+        bag.addOccurrences(element2, occurrence2);
+        bag.addOccurrences(element3, occurrence3);
+        bag.addOccurrences(element4, occurrence4);
+        return bag.toImmutable();
+    }
+
+    @Override
+    public <T> ImmutableBag<T> withOccurrences(ObjectIntPair<T>... elementsWithOccurrences)
+    {
+        MutableBag<T> bag = new HashBag<>();
```
motlin: This is the only location where pre-sizing may have a benefit. The size would be the array length.

unit-tests/src/test/java/org/eclipse/collections/impl/bag/mutable/MutableBagTestCase.java
```diff
@@ -450,7 +450,7 @@ public void selectDuplicates()
     @Test
     public void topOccurrences()
     {
-        MutableBagIterable<String> strings = this.newWithOccurrences(
+        MutableBagIterable<String> strings = Bags.mutable.withOccurrences(
```
motlin: Please add at least one test where the number of occurrences is Integer.MAX_VALUE. This should fail with OutOfMemoryError until the pre-sizing fix is in place.

donraab:
> Even if our next version is a major release (11.0) I still think this change should be made using default methods in the interfaces. I think it's easier than not doing it.

@jiminhsieh I agree with Craig's idea here.  This change can be made via default methods. @nikhilnanivadekar Any thoughts?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableBagFactoryImpl.java
jiminhsieh: Thanks for catching that. I will fix this soon.

unit-tests/src/test/java/org/eclipse/collections/impl/bag/mutable/MutableBagTestCase.java
jiminhsieh: I tried to change the last element of occurrences to `Integer.MAX_VALUE`. I didn't meet OutOfMemoryError even I didn't fix pre-sizing. 
Here is the change: 
```java
MutableBagIterable<String> strings = Bags.mutable.withOccurrences(
                PrimitiveTuples.pair("one", 1),
                PrimitiveTuples.pair("two", 2),
                PrimitiveTuples.pair("three", 3),
                PrimitiveTuples.pair("four", 4),
                PrimitiveTuples.pair("five", 5),
                PrimitiveTuples.pair("six", 6),
                PrimitiveTuples.pair("seven", 7),
                PrimitiveTuples.pair("eight", 8),
                PrimitiveTuples.pair("nine", 9),
                PrimitiveTuples.pair("ten", Integer.MAX_VALUE));
```

Should I still add one test for this? 

But I do understand that the reason of pre-sizing for the variable arguments. The underlying storage could grow dynamically and exceed than we need. So it will be better to pre-sizing in the initialization.

jiminhsieh:
> this change should be made using default methods in the interfaces

Sorry, I can't catch up on this one. Could someone elaborate this more to me? Thanks. :)

nikhilnanivadekar:
@motlin can you please help @jiminhsieh with your review comments, please?

motlin:
I made most of the changes I wanted to see in these commits:

* 2e5d3ceb Fix pre-sizing in MutableBagFactoryImpl.
* a6923131 Fix pre-sizing and simplify ImmutableBagFactoryImpl.
* 2f0649ee Pull methods up from MutableBagFactoryImpl to MutableBagFactory as default methods.
* 4db0ef87 Pull methods up from ImmutableBagFactoryImpl to ImmutableBagFactory as default methods.

Feel free to cherry-pick these or squash them in. I also think this could use a few additional tests to get full coverage.

motlin:
Also, I completely removed pre-sizing. If we add a method like empty(size) or something like that, we can add pre-sizing back in.

jiminhsieh:
Current code coverage of those 2 classes: 

Screen Shot 2020-02-22 at 18 28 25

nikhilnanivadekar:
@motlin can you please review this PR since you have been working on it. Also, I see multiple commits on this one, is that the plan or we want these commits to be squashed?

jiminhsieh:
@nikhilnanivadekar I can squash into one commit. But I have a question for @motlin. Is it fine for you that I use [**co-author**](https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) since you do help me with those commits. Thanks.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/bag/ImmutableBagFactory.java
```diff
@@ -47,8 +49,111 @@
...
+        return mutableBag.toImmutable();
+    }
+
+    /**
+     * @since 11.0
+     */
+    default <T> ImmutableBag<T> withOccurrences(ObjectIntPair<T>... elementsWithOccurrences)
+    {
+        return Bags.mutable.withOccurrences(elementsWithOccurrences).toImmutable();
+    }
```
motlin: The "since 11.0" is a guess. We could potentially release this in 10.3.0. Just an observation, this is fine.

nikhilnanivadekar: It should be 10.3.0. I am planning to have a minor release before a major release

jiminhsieh: Changed to `@since 10.3`.

motlin:
LGTM. Would you also rebase it on top of upstream/master?

jiminhsieh:
Rebased on the master without any conflict.

motlin:

Isammoc:
Thank you!

PR #764, opened 2019-10-24
Fixes #762. Fix fromToBy & injectInto issues with same start and end with negative step. Add unit tests to Interval as well, to cover the scenario.

### LINKED ISSUE #762: IntInterval.fromToBy(n, n, -x) doesn't work as expected
Using eclipse collections 10.0.0

`IntInterval.fromToBy(2, 2, -1).toList()` raises an exception whereas we expect `[2]`.

 Here is the exception raised:
```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at org.eclipse.collections.impl.list.primitive.IntInterval.lambda$toArray$7243f8d6$1(IntInterval.java:610)
	at org.eclipse.collections.impl.list.primitive.IntInterval.forEachWithIndex(IntInterval.java:324)
	at org.eclipse.collections.impl.list.primitive.IntInterval.toArray(IntInterval.java:610)
	at org.eclipse.collections.impl.list.mutable.primitive.IntArrayList.newList(IntArrayList.java:98)
	at org.eclipse.collections.impl.list.primitive.IntInterval.toList(IntInterval.java:900)
       [...]
```

`IntInterval.fromToBy(2,2,-2).forEach(i -> { System.out.println(i);}); ` prints an infinite sequence of decreasing integers whereas we expect just `2`.

This behaviour is not the same as `Interval.fromToBy(2,2,-2)` wich seems to be ok.

cguntur-bnym:
Signed-off-by: Chandra Guntur <chandra.guntur@bnymellon.com>

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/IntInterval.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/IntInterval.java
@@ -317,7 +317,7 @@ public final class IntInterval
     public void forEachWithIndex(IntIntProcedure procedure)
     {
         int index = 0;
-        if (this.from <= this.to)
+        if (this.goForward())
         {
             for (int i = this.from; i <= this.to; i += this.step)
             {
@@ -339,13 +339,18 @@ public final class IntInterval
         this.each(procedure);
     }
 
+    private boolean goForward()
+    {
+        return this.from <= this.to && this.step > 0;
+    }
+
     /**
      * @since 7.0.
      */
     @Override
     public void each(IntProcedure procedure)
     {
-        if (this.from <= this.to)
+        if (this.goForward())
         {
             for (int i = this.from; i <= this.to; i += this.step)
             {
@@ -615,7 +620,7 @@ public final class IntInterval
     public <T> T injectInto(T injectedValue, ObjectIntToObjectFunction<? super T, ? extends T> function)
     {
         T result = injectedValue;
-        if (this.from <= this.to)
+        if (this.goForward())
         {
             for (int i = this.from; i <= this.to; i += this.step)
             {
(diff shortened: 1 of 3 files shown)


motlin:
I never realized that this works:
`Interval.fromToBy(2, 2, 2)`

And this ought to work:
`Interval.fromToBy(2, 2, -2)`

But this is deliberately banned
`Interval.fromToBy(2, 2, 0)`

cguntur-bnym:
This PR includes a change to injectInto() method as well, where I found the same issue. Added tests for that as well.

nikhilnanivadekar:
Please break it up as separate commits or change the comment of the commit to mention that. It helps while preparing release notes to look at the commits and their comments. Thank you!

cguntur-bnym:
Will run a commit amend and re-push over the weekend.

nikhilnanivadekar:
@cguntur-bnym any updates on this? Also the commit message needs to state that it fixes which issue. Fixes #762 so that it will auto-close.

cguntur-bnym:
@nikhilnanivadekar - please let me know if you need me to do anything else?

nikhilnanivadekar:
@cguntur-bnym can you please rebase on to upstream/master . From the network graph below, your branch is a few commits behind the upstream
image

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/IntInterval.java
```diff
@@ -339,13 +339,18 @@ public void forEach(IntProcedure procedure)
         this.each(procedure);
     }
 
+    public boolean goForward()
```
nikhilnanivadekar: should be private

PR #766, opened 2019-10-27
Override Predicate.negate().

motlin:


motlin:
Without this change Predicate.negate() returns java.util.function.Predicate.

With this change, it returns org.eclipse.collections.api.block.predicate.Predicate.

The possibly awkward thing is that it's a change in api, so it doesn't return org.eclipse.collections.impl.block.factory.Predicates.NotPredicate#NotPredicate

PR #773, opened 2019-11-09
Replace usages of Comparators.nullSafeEquals() with Objects.equals().

motlin:

### DIFF
--- a/acceptance-tests/src/test/java/org/eclipse/collections/impl/map/mutable/UnifiedMapAcceptanceTest.java
+++ b/acceptance-tests/src/test/java/org/eclipse/collections/impl/map/mutable/UnifiedMapAcceptanceTest.java
@@ -19,11 +19,11 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import org.eclipse.collections.api.map.MutableMap;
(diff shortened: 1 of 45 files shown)


nikhilnanivadekar:
@motlin The code in Comparators.nullSafeEquals() should also be changed to delegate to Objects.equals().

motlin:
I didn't change `Comparators.nullSafeEquals()` out of an abundance of caution. The behavior of `Comparators.nullSafeEquals()` and `Objects.equals()` is not identical for a single case, where equals() is implemented to return false no matter what. Changing it breaks this test, which I wrote a long time ago.

```
    @Test
    public void assertNotEquals()
    {
        Object object = new Object()
        {
            @Override
            public boolean equals(Object obj)
            {
                return false;
            }
        };

        Verify.assertNotEquals(object, object);
    }
```

I may be the only one who has ever cared about this distinction. This implementation doesn't technically violate the contract of equals(), but I think that's only because it's so obvious that x.equals(x) should return true that they didn't bother to write it in the javadoc of equals().

This is a long way of saying, I thought about it and chose not to. It would be reasonable to go ahead and do your suggestion, and then I just have to delete or change the test.

nikhilnanivadekar:
Even though making the change will be sensible but it also alters the behavior and can be considered breaking change. Hence, please don’t change it. Once you have rebased on to master feel free to merge with the current changes.

PR #774, opened 2019-11-18
Add static factory methods to interfaces.

motlin:
Since we use ServiceLoaders to have factories in the api module, and since Java 8 supports static methods on interfaces, we can support factory methods on interfaces.

This PR just adds the factory methods, but you can start to see what it will look like when we use these methods in my work-in-progress commit: e7476bee1dc2f1113194c0a93d04c96108fd168b

Before:
```java
Lists.immutable.with(1, 2, 3)
```

After:
```java
ImmutableList.of(1, 2, 3)
```

Where we support both of() and with(), I only added the *of* methods. This is because the static method withAll() clashes with the non-static method with the same name. I think this is ok though since JDK interfaces consistently use "of"

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/ImmutableBag.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/ImmutableBag.java
@@ -10,6 +10,8 @@
 
 package org.eclipse.collections.api.bag;
 
+import java.util.stream.Stream;
+
 import org.eclipse.collections.api.bag.primitive.ImmutableBooleanBag;
 import org.eclipse.collections.api.bag.primitive.ImmutableByteBag;
 import org.eclipse.collections.api.bag.primitive.ImmutableCharBag;
@@ -33,6 +35,7 @@ import org.eclipse.collections.api.block.predicate.Predicate;
 import org.eclipse.collections.api.block.predicate.Predicate2;
 import org.eclipse.collections.api.block.predicate.primitive.IntPredicate;
 import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.factory.Bags;
 import org.eclipse.collections.api.factory.Sets;
 import org.eclipse.collections.api.list.ImmutableList;
 import org.eclipse.collections.api.multimap.bag.ImmutableBagMultimap;
@@ -48,6 +51,36 @@ import org.eclipse.collections.api.tuple.primitive.ObjectIntPair;
  */
 public interface ImmutableBag<T> extends UnsortedBag<T>, ImmutableBagIterable<T>
 {
+    static <T> ImmutableBag<T> empty()
+    {
+        return Bags.immutable.empty();
+    }
+
+    static <T> ImmutableBag<T> of()
+    {
+        return Bags.immutable.of();
+    }
+
+    static <T> ImmutableBag<T> of(T element)
+    {
+        return Bags.immutable.of(element);
+    }
+
+    static <T> ImmutableBag<T> of(T... elements)
+    {
+        return Bags.immutable.of(elements);
+    }
+
+    static <T> ImmutableBag<T> ofAll(Iterable<? extends T> items)
+    {
+        return Bags.immutable.ofAll(items);
+    }
+
+    static <T> ImmutableBag<T> fromStream(Stream<? extends T> stream)
+    {
+        return Bags.immutable.fromStream(stream);
+    }
+
     @Override
     ImmutableBag<T> newWith(T element);
 
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MutableBag.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MutableBag.java
@@ -10,6 +10,8 @@
 
 package org.eclipse.collections.api.bag;
 
+import java.util.stream.Stream;
+
 import org.eclipse.collections.api.bag.primitive.MutableBooleanBag;
 import org.eclipse.collections.api.bag.primitive.MutableByteBag;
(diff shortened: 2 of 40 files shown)


donraab:
Thinking through what this will look like with primitive collections:

```
MutableIntList a1 = MutableIntList.of(1).with(2).with(3);
MutableIntList a2 = IntLists.mutable.with(1).with(2).with(3);
MutableLongList b1 = MutableLongList.of(1L).with(2L).with(3L);
MutableLongList b2 = LongLists.mutable.with(1L).with(2L).with(3L);
```

motlin:
That's right. I'm working on replacing usages, which covers the new methods.

motlin:
@donraab one problem is that we never fleshed out all the potential factories in api, to match impl.

For example, there's no Multimaps factory in org.eclipse.collections.api.factory, and there are no primitive factories in org.eclipse.collections.api.factory at all.

We can do it of course, there's just so many of them.

motlin:
@donraab I changed usages of factories to go through the new factory methods, which increases coverage. There are something like 5000 usages replaced, so it's kind of hard to review. The first commit is the same as before, the second commit replaces usages.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/Bags.java
```diff
@@ -19,17 +19,17 @@
  * Mutable Examples:
  *
  * <pre>
- * MutableBag&lt;String&gt; emptyBag = Bags.mutable.empty();
- * MutableBag&lt;String&gt; bagWith = Bags.mutable.with("a", "b", "c");
```
donraab: You shouldn't be replacing in any of the factory classes.

motlin:
@donraab I'm working on the changes you suggested. I've added tests that are heavily based on test classes like ListsTest, SetsTest, etc. It's taking me some time so I wanted to show you before I'm done to make sure I'm on the right track. I still have to test Maps and SortedMaps at least, but I'll check that I got everything.

motlin:
@donraab I believe this addresses your feedback. I dropped the second commit and replaced it with a commit that adds tests in eclipse-collections/unit-tests/src/test/java/org/eclipse/collections/impl/factory

donraab:
@motlin @nikhilnanivadekar I'm still not sure if these new features will pull their weight.  There is something that fundamentally bothers me about having static methods on interfaces, as it feels like we are tying interfaces to implementations.  I know that this is done in Java itself now for List, Set, Map, so there is a precedent that exists for this change.  I also understand and appreciate that the static methods on interfaces like `MutableList` and `ImmutableList` will be more clear in terms of the runtime type actually being mutable or immutable, unlike List, Set or Map in the JDK.

Does this change provide more benefit than saving a few characters?

motlin:
I understand the concerns, especially about interfaces being tied to implementations.

I feel that interfaces are already somewhat tied to implementations through the "plural" factories. For example, `Lists.mutable.empty()` is declared to return `MutableList` but always returns `FastList`.

I implemented the new methods in terms of the old ones, so I also think the binding is no worse. For example, on MutableList:
```java
    static <T> MutableList<T> empty()
    {
        return Lists.mutable.empty();
    }
```

The factories, like `Lists.mutable` are loaded using `ServiceLoader`. So it's theoretically possible to swap in a different implementation. That's probably prohibitively hard for users, but maybe we could create an example soon. There's no problem with `FastList`, but I figured it would be useful to have a null-intolerant alternative. So Lists.mutable.empty() would return a FastList wrapped with a wrapper that throws on add(null), as one example.

It's hard to put myself in the shoes of a newcomer, but I feel like the static factory methods could help with learning. You could basically start by learning interfaces first.

I think the static factory methods help slightly with typing. I type `Lists.m` a lot, and my IDE suggests `Lists.multireader`. The auto-complete for `ML.e` is unambiguous, at least for now.

The regular factories get a little harder to use for complex types like multimaps. For example `Multimaps.mutable.list.of(1, 1, 2, 2)`. I'm not sure if it's better, but it would be "flatter" to write MutableListMultimap.of(1, 1, 2, 2). It's also not clear to me how we'll add other types of multimaps to the Multimaps factory, specifically if we add sorted key versions. `Multimaps.sortedKey.mutable.sortedSet.empty()`?

motlin:
Discussed offline with Don. He suggested an alternative, a method with a name like factory() or companion() on the interfaces. For example, on `MutableList`:

```java
    static MutableListFactory factory()
    {
        return Lists.mutable;
    }
```

I think this is a good idea. It's worth discussing because it doesn't make sense to do both. Some pros/cons that occur at first glance.
* A much smaller change, both in terms of lines of code and new api.
* More consistent api. All new methods would have the same name (factory or companion) and their return types would be interfaces we already have.
* More typing for users. `MutableList.factory().empty()` vs `MutableList.empty()`

It's also worth considering doing nothing. We already have lots of ways of creating collections.

nikhilnanivadekar:
I am a big fan of “if not broken don’t fix it” and also of “do nothing if not necessary”

donraab:
This is the obvious and most accurate option:

```
MutableList.factory().empty(); 
```

There are other shorter options:

```
MutableList.$().empty();
MutableList.f$().empty();
MutableList.make().empty();
MutableList.build().empty();
MutableList.forge().empty();
```

The upside of the dollar methods is that they will most likely never clash with any real methods.

donraab:
@motlin I think we should go with the companion object approach to start off.  We can always introduce additional static methods later, and they can call the factory method directly if we deem them valuable.  What do you think?

motlin:
```java
MutableList.empty()
Lists.mutable.empty()
MutableList.factory().empty()
```

I still prefer the 1st because it's both shorter and refers to the interface. But we don't need to change anything now. I'll close the pull request and we can continue to think about it. I'll keep the branch around to look at.

PR #780, opened 2019-12-08
use implementation instead of deprecated compile when declaring depen…

LarsEckart:
…dcy with gradle.
and bump to 10.1.0 in readme for both maven and gradle.

Signed-off-by: Lars Eckart <lars.eckart@gmail.com>

I hope i completed the registration and contributor agreement steps correctly.


LarsEckart:
will update my fork first

Issue #677, opened 2018-12-28
Optimize removeIf() for UnifiedMap.


nikhilnanivadekar:
Refer to https://github.com/eclipse/eclipse-collections/pull/665#issuecomment-449521787

sendilkumarn:
This looks like the issue is closed by 7ce7a19

Can we close the issue?

nikhilnanivadekar:
Fixed with #786

PR #786, opened 2020-01-12
Fix Bug #677 - Optimize removeIf() for UnifiedMap.

digsb:
Signed-off-by: Digant Bhavsar <digantbhavsar@gmail.com>

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
@@ -1038,17 +1038,48 @@ public class UnifiedMap<K, V> extends AbstractMutableMap<K, V>
     @Override
     public boolean removeIf(Predicate2<? super K, ? super V> predicate)
     {
-        int previousSize = this.size();
-        Iterator<Entry<K, V>> iterator = this.entrySet().iterator();
-        while (iterator.hasNext())
+        int previousOccupied = this.occupied;
+        for (int index = 0; index < this.table.length; index += 2)
         {
-            Entry<K, V> entry = iterator.next();
-            if (predicate.accept(entry.getKey(), entry.getValue()))
+            Object cur = this.table[index];
+            if (cur == null)
             {
-                iterator.remove();
+                continue;
+            }
+            if (cur == CHAINED_KEY)
+            {
+                Object[] chain = (Object[]) this.table[index + 1];
+                for (int chIndex = 0; chIndex < chain.length; )
+                {
+                    if (chain[chIndex] == null)
+                    {
+                        break;
+                    }
+                    K key = this.nonSentinel(chain[chIndex]);
+                    V value = (V) chain[chIndex + 1];
+                    if (predicate.accept(key, value))
+                    {
+                        this.overwriteWithLastElementFromChain(chain, index, chIndex);
+                    }
+                    else
+                    {
+                        chIndex += 2;
+                    }
+                }
+            }
+            else
+            {
+                K key = this.nonSentinel(cur);
+                V value = (V) this.table[index + 1];
+                if (predicate.accept(key, value))
+                {
+                    this.table[index] = null;
+                    this.table[index + 1] = null;
+                    this.occupied--;
+                }
             }
         }
-        return previousSize > this.size();
+        return previousOccupied > this.occupied;
     }
 
     private void chainedForEachEntry(Object[] chain, Procedure2<? super K, ? super V> procedure)
--- a/RELEASE_NOTE_DRAFT.md
+++ b/RELEASE_NOTE_DRAFT.md
@@ -21,6 +21,7 @@ Central theme of this release is to fix issues identified for SimRel repository.
 * Fixed generated Eclipse features for p2 repository to ensure correct EPLv1 license is downloaded.
 * Fixed generated Eclipse features for p2 repository to ensure correct signatures on artifacts.
 * Fixed repository path for Eclipse features.
+* Optimize removeIf() for UnifiedMap.
 
 # Build Changes
 -----------------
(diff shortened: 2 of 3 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -1038,17 +1038,40 @@ public V removeKey(K key)
...
+            {
+                continue;
+            }
+            if (cur == CHAINED_KEY)
             {
-                iterator.remove();
+                Object[] chain = (Object[]) this.table[index + 1];
+                for (int chIndex = 0; chIndex < chain.length; chIndex += 2)
+                {
+                    K key = this.nonSentinel(chain[chIndex]);
```
mohrezaei: there is a subtle bug here. It would be good to have a test for it. The bug is this: `chIndex` is incremented unconditionally, even if the predicate matches. So if we have 3 elements, `a`, `b`, and `c`, if `a` matches, the chain look likes `c`, `b`, but `c` is no longer put through the predicate because the index will move on.

digsb: Ah, got it. Let me add a test case for it and fix it.

mohrezaei: also, if `chain[chIndex] == null`, break the loop immediately.

digsb: Thanks for Pointing this out. Fixed it.

digsb: Thanks for Pointing this out. Fixed it.

motlin: Style nitpick... now with the refactoring of the for-loop, I think you could move the condition `chain[chIndex] != null` into the loop condition.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -1038,17 +1038,48 @@ public V removeKey(K key)
     @Override
     public boolean removeIf(Predicate2<? super K, ? super V> predicate)
     {
-        int previousSize = this.size();
-        Iterator<Entry<K, V>> iterator = this.entrySet().iterator();
-        while (iterator.hasNext())
+        int previousOccupied = this.occupied;
```
motlin: I'm curious about `occupied` vs `size()`. Is there an advantage to one over the other?

motlin:
This looks good overall. I think you could squash the three commits and we could merge it. I also think we have some next steps we should track if we're not going to do them now, and I wanted to start a conversation about that.

`UnifiedMapWithHashingStrategy` has the same implementation and same structure as `UnifiedMap`. We should repeat essentially the identical change in that class.

`ConcurrentMutableHashMap` has the same implementation but wraps a `ConcurrentMap`. I'm struggling to come up with an optimization. It may be impossible.

`ConcurrentHashMapUnsafe` has the same implementation but a radically different structure. One possible implementation would be an atomic removeIf once per partition. But that has degenerate worst-case performance and isn't obviously an optimization. I'll leave it to @mohrezaei whether that's a good idea to ever do.

`AbstractMutableBiMap` has the same implementation but has two delegate `UnifiedMaps`. Even AbstractMutableBiMap#remove has inefficiencies where it calculates the same hashcode twice so I'm guessing this one is too hard to optimize.

`TreeSortedMap` has the same implementation but wraps a `TreeMap`. We wouldn't be able to optimize until we have an implementation where we have control over the tree nodes. And the current implementation may already be efficient.

The tree adapters `MapAdapter`, `OrderedMapAdapter`, and `SortedMapAdapter` are impossible to optimize and all have the same implementation.

Since all subclasses have the same implementation, I'm wondering why we never used default methods. It seems like we could pull up the old implementation as a default method on `MutableMapIterable`. I'm wondering if there's a reason we didn't do it that way before @nikhilnanivadekar.

mohrezaei:
@motlin Both `ConcurrentHashMap` and `ConcurrentHashMapUnsafe` would be difficult to optimize. Because of the possible concurrent modifications, the code would have to repeat large portions of `HashIterator`.

mohrezaei:
@motlin I thought about this a bit more and figured I'd clarify one thing: for`FastList.removeIf`, the optimization is very important. It makes the algorithm O(n) instead of O(n^2) (worst case).  

For hashed structures, using an iterator to remove is already O(n), so the optimization is minor. I wonder if the change in this PR is even measurable. It does create a little garbage, so it's worth merging at least for that.

For tree structures, it depends on how hard it is to remove individual nodes (rebalancing, etc), so it can be potentially O(n^2) (or is it O(n^2 * log n)?). A depth first iteration would reduce the cost a bit. For predicates that remove lots of nodes, it's probably best to reconstruct the tree just once.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
digsb: size() is just a wrapper around occupied, so though of using the field directly.

motlin:
@mohrezaei I agree, the algorithm is O(n) for hash tables either way. I didn't measure but I think the constant factor could be significant, because it could mean the difference between calling or not calling hashCode() for many elements.

Considering two cases:
* Remove 25% of the elements without triggering a rehash
* Remove 90% of the elements

In the first case, the unoptimized algorithm calls hashCode() on the removed elements. The optimized algorithm doesn't call hashCode() at all.

In the second case, the unoptimized algorithm calls hashCode() on every element (90% during removal and 10% during rehashing). The optimized algorithm only calls hashCode() on the 10% that get rehashed.

mohrezaei:
@motlin Sorry, where does the original (`Iterator` based algorithm) call `hashcode()`?

motlin:
Ah, I see that it doesn't. We've been assuming it did ever since:
https://github.com/eclipse/eclipse-collections/pull/665#discussion_r243697682
and:
https://github.com/eclipse/eclipse-collections/pull/665#issuecomment-449521787
It looks like iterator never triggers rehashing either.

PR #789, opened 2020-01-15
Turn on CheckStyle check TrailingComment and fix violations.

motlin:
We were about 10 versions behind on CheckStyle. I upgraded locally to the latest and I've been experimenting with getting our configuration to be closer to CheckStyle's own configuration for its own GitHub project [here](https://raw.githubusercontent.com/checkstyle/checkstyle/checkstyle-8.28/config/checkstyle_checks.xml). This check was already available in the older version and only had a few violations. If this rule is controversial we can skip it. There are other new rules I also plan on turning on as I have time.

### DIFF
--- a/acceptance-tests/src/test/java/org/eclipse/collections/impl/concurrent/MultiReaderUnifiedSetAcceptanceTest.java
+++ b/acceptance-tests/src/test/java/org/eclipse/collections/impl/concurrent/MultiReaderUnifiedSetAcceptanceTest.java
@@ -303,7 +303,8 @@ public class MultiReaderUnifiedSetAcceptanceTest
         Assert.assertTrue(set.retainAll(toRetain));
         Assert.assertTrue(set.containsAll(toRetain));
 
-        Assert.assertFalse(set.retainAll(toRetain)); // a second call should not modify the set
+        // a second call should not modify the set
+        Assert.assertFalse(set.retainAll(toRetain));
 
         Verify.assertSize(size / 2, set);
 
@@ -343,7 +344,8 @@ public class MultiReaderUnifiedSetAcceptanceTest
         Assert.assertTrue(set.retainAll(toRetain));
         Assert.assertTrue(set.containsAll(toRetain));
 
-        Assert.assertFalse(set.retainAll(toRetain)); // a second call should not modify the set
+        // a second call should not modify the set
+        Assert.assertFalse(set.retainAll(toRetain));
 
         Verify.assertSize(size / 2, set);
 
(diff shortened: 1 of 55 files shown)


unit-tests/src/test/java/org/eclipse/collections/impl/list/fixed/FixedSizeListFactoryTest.java
```diff
@@ -126,9 +126,12 @@ public void varArgsListCreation()
     {
         String[] content = {"one", "two"};
 
-        //List<Object>   list1 = Lists.fixedSize.of(content);  // incompatible types: List<Object> vs List<String>
-        //List<String[]> list2 = Lists.fixedSize.of(content);  // incompatible types: List<String[]> vs List<String>
-        List<String[]> list3 = Lists.fixedSize.<String[]>of(content);  // correct!
+        // incompatible types: List<Object> vs List<String>
```
mohrezaei: this now is impossible to read. Use a block comment?

acceptance-tests/src/test/java/org/eclipse/collections/impl/concurrent/MultiReaderUnifiedSetAcceptanceTest.java
```diff
@@ -303,7 +303,8 @@ private static void runUnifiedSetRetainAllFromList(int shift)
         Assert.assertTrue(set.retainAll(toRetain));
         Assert.assertTrue(set.containsAll(toRetain));
 
-        Assert.assertFalse(set.retainAll(toRetain)); // a second call should not modify the set
+        // a second call should not modify the set
```
mohrezaei: It feels like all these comments should end with `:`, or they don't read quite the same.

motlin: I tried to address both readability concerns by adding newlines where they were missing so it's even more obvious that the comment applies to the next statement or expression. I'm open to adding `:` too. And I'm still open to the idea that this is not a style choice that we should enforce.

nikhilnanivadekar:
What is the latest on this?

motlin:
I tried to address readability suggestions. I have some additional CheckStyle pull requests to make. But I don't feel particularly strongly about any one rule.

nikhilnanivadekar:
should we close this PR or there is work pending on this or it needs to be merged?

motlin:
I thought it was a good change, but there doesn't seem to be any enthusiasm for it.

Issue #761, opened 2019-10-14
Add withMap() to MutableMap and MutableMapFactory


mkrivoshein:
Would be nice to have utility methods that can add contents of an another map while allowing use of the API in a builder pattern style.

It will especially be useful if you already have multiple maps and you want to create an union of these maps.

```
ImmutableMap<String, String> map = Maps.mutable
           .withMap(firstMap)
           .withMap(secondMap)
           .withMap(thirdMap)
           .toImmutable();
```

Just as a reference, similar functionality is available with Guava ImmutableMap builder.

nikhilnanivadekar:
this makes sense. We have withKeyValue() but with do not have withMap().
Would you like to contribute this enhancement? This should be relatively straightforward and quick to implement. I will be happy to guide you through.

mkrivoshein:
I'd be interesting experience. Will look into what needs to be done to become a contributor first.

nikhilnanivadekar:
Please read our contribution guide. And feel free to ask questions on this
issue. Please tag me so that they show up in my notifications. Once you
raise a PR we can discuss on the PR itself. Thanks!

victornoel:
it would make sense to have both `withMap(MapIterable)` and `withMap(Map)` methods I think, no?

PR #801, opened 2020-02-21
Add withMap to MutableMap

aboullaite:
Signed-off-by: Mohammed Aboullaite <aboullaite.mohammed@gmail.com>

This PR adds `withMap` a requested in #761

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
@@ -153,6 +153,19 @@ public interface MutableMapIterable<K, V> extends MapIterable<K, V>, Map<K, V>
     MutableMapIterable<K, V> withKeyValue(K key, V value);
 
     /**
+     * Similar to {@link #putAll(Map)}, but returns this instead of void
+     *
+     * @see #putAll(Map)
+     *
+     * @since 10.3.0
+     */
+    default MutableMapIterable<K, V> withMap(Map<? extends K, ? extends V> map)
+    {
+        this.putAll(map);
+        return this;
+    }
+
+    /**
      * This method allows mutable, fixed size, and immutable maps the ability to add elements to their existing
      * elements. In order to support fixed size maps, a new instance of a map would have to be returned including the
      * keys and values of the original plus all of the additional keys and values. In the case of mutable maps, the
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
@@ -12,6 +12,7 @@ package org.eclipse.collections.impl.bimap.mutable;
 
 import java.io.Serializable;
 import java.util.Collection;
+import java.util.Map;
 import java.util.Set;
 
 import org.eclipse.collections.api.RichIterable;
@@ -287,6 +288,16 @@ public class SynchronizedBiMap<K, V> extends AbstractSynchronizedMapIterable<K, 
         synchronized (this.lock)
         {
             this.getDelegate().put(key, value);
+            return this;
+        }
+    }
+
+    @Override
+    public MutableBiMap<K, V> withMap(Map<? extends K, ? extends V> map)
+    {
+        synchronized (this.lock)
+        {
+            this.putAll(map);
             return this;
         }
     }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
@@ -1980,6 +1980,12 @@ public final class ConcurrentHashMap<K, V>
     }
 
     @Override
+    public ConcurrentHashMap<K, V> withMap(Map<? extends  K, ? extends V> map)
+    {
+        return (ConcurrentHashMap<K, V>) super.withMap(map);
+    }
+
+    @Override
     public ConcurrentHashMap<K, V> withAllKeyValues(Iterable<? extends Pair<? extends K, ? extends V>> keyValues)
     {
         return (ConcurrentHashMap<K, V>) super.withAllKeyValues(keyValues);
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/SynchronizedSortedMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/SynchronizedSortedMap.java
@@ -13,6 +13,7 @@ package org.eclipse.collections.impl.map.sorted.mutable;
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.Map;
 import java.util.Optional;
 import java.util.SortedMap;
 
@@ -118,6 +119,16 @@ public class SynchronizedSortedMap<K, V>
         synchronized (this.lock)
         {
             this.put(key, value);
+            return this;
+        }
+    }
+
+    @Override
+    public MutableSortedMap<K, V> withMap(Map<? extends K, ? extends V> map)
+    {
+        synchronized (this.lock)
+        {
+            this.putAll(map);
             return this;
         }
     }
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/UnmodifiableBiMapTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/UnmodifiableBiMapTest.java
@@ -163,6 +163,41 @@ public class UnmodifiableBiMapTest extends AbstractMutableBiMapTestCase
 
     @Override
     @Test
+    public void withMap()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(Maps.mutable.with(1, Character.valueOf('a'))));
+    }
+
+    @Override
+    @Test
+    public void withMapEmpty()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(Maps.mutable.empty()));
+    }
+
+    @Override
+    @Test
+    public void withMapTargetEmpty()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(Maps.mutable.with(1, Character.valueOf('a'))));
+    }
+
+    @Override
+    @Test
+    public void withMapEmptyAndTargetEmpty()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.getEmptyMap().withMap(Maps.mutable.empty()));
+    }
+
+    @Override
+    @Test
+    public void withMapNull()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.getEmptyMap().withMap(null));
+    }
+
+    @Override
+    @Test
     public void withAllKeyValueArguments()
     {
         Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeysValues("A", 1, "B", 2).withAllKeyValueArguments(Tuples.pair("B", 22), Tuples.pair("C", 3)));
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
@@ -12,6 +12,7 @@ package org.eclipse.collections.impl.map.mutable;
 
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.concurrent.ConcurrentMap;
@@ -558,6 +559,51 @@ public abstract class MutableMapIterableTestCase extends MapIterableTestCase
 
         MutableMapIterable<String, Integer> mapWith2 = mapWith.withKeyValue("A", 11);
         Verify.assertMapsEqual(UnifiedMap.newWithKeysValues("A", 11, "B", 2), mapWith);
+    }
+
+    @Test
+    public void withMap()
+    {
+        MutableMapIterable<String, Integer> map = this.newMapWithKeysValues("A", 1, "B", 2);
+        Map<String, Integer> simpleMap = Maps.mutable.with("B", 22, "C", 3);
+        map.putAll(simpleMap);
+        MutableMapIterable<String, Integer> mapWith = map.withMap(simpleMap);
+        Assert.assertSame(map, mapWith);
+        Verify.assertMapsEqual(UnifiedMap.newWithKeysValues("A", 1, "B", 22, "C", 3), mapWith);
+    }
+
+    @Test
+    public void withMapEmpty()
+    {
+        MutableMapIterable<String, Integer> map = this.newMapWithKeysValues("A", 1, "B", 2);
+        MutableMapIterable<String, Integer> mapWith = map.withMap(Maps.mutable.empty());
+        Assert.assertSame(map, mapWith);
+        Verify.assertMapsEqual(UnifiedMap.newWithKeysValues("A", 1, "B", 2), mapWith);
+    }
+
+    @Test
+    public void withMapTargetEmpty()
+    {
+        MutableMapIterable<String, Integer> map = this.newMap();
+        Map<String, Integer> simpleMap = Maps.mutable.with("A", 1, "B", 2);
+        MutableMapIterable<String, Integer> mapWith = map.withMap(simpleMap);
+        Assert.assertSame(map, mapWith);
+        Verify.assertMapsEqual(UnifiedMap.newWithKeysValues("A", 1, "B", 2), mapWith);
+    }
+
+    @Test
+    public void withMapEmptyAndTargetEmpty()
+    {
+        MutableMapIterable<String, Integer> map = this.newMap();
+        MutableMapIterable<String, Integer> mapWith = map.withMap(Maps.mutable.empty());
+        Assert.assertSame(map, mapWith);
+        Verify.assertMapsEqual(UnifiedMap.newMap(), mapWith);
+    }
+
+    @Test
+    public void withMapNull()
+    {
+        Verify.assertThrows(NullPointerException.class, () -> this.newMap().withMap(null));
     }
 
     @Test
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/UnmodifiableMutableMapTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/UnmodifiableMutableMapTest.java
@@ -14,6 +14,7 @@ import org.eclipse.collections.api.block.function.Function0;
 import org.eclipse.collections.api.map.MutableMap;
 import org.eclipse.collections.impl.block.factory.Functions;
 import org.eclipse.collections.impl.factory.Lists;
+import org.eclipse.collections.impl.factory.Maps;
 import org.eclipse.collections.impl.list.mutable.FastList;
 import org.eclipse.collections.impl.test.Verify;
 import org.eclipse.collections.impl.tuple.ImmutableEntry;
@@ -282,6 +283,41 @@ public class UnmodifiableMutableMapTest extends MutableMapTestCase
     public void withKeyValue()
     {
         this.newMapWithKeysValues(1, "One", 2, "2").withKeyValue(null, null);
+    }
+
+    @Override
+    @Test
+    public void withMap()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(Maps.mutable.with(1, Character.valueOf('a'))));
+    }
+
+    @Override
+    @Test
+    public void withMapEmpty()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(Maps.mutable.empty()));
+    }
+
+    @Override
+    @Test
+    public void withMapTargetEmpty()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMap().withMap(Maps.mutable.with(1, Character.valueOf('a'))));
+    }
+
+    @Override
+    @Test
+    public void withMapEmptyAndTargetEmpty()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMap().withMap(Maps.mutable.empty()));
+    }
+
+    @Override
+    @Test
+    public void withMapNull()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMap().withMap(null));
     }
 
     @Override
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/map/sorted/mutable/UnmodifiableTreeMapTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/map/sorted/mutable/UnmodifiableTreeMapTest.java
@@ -16,6 +16,7 @@ import org.eclipse.collections.api.map.sorted.MutableSortedMap;
 import org.eclipse.collections.impl.block.factory.Functions;
 import org.eclipse.collections.impl.block.function.PassThruFunction0;
 import org.eclipse.collections.impl.factory.Lists;
+import org.eclipse.collections.impl.factory.Maps;
 import org.eclipse.collections.impl.list.mutable.FastList;
 import org.eclipse.collections.impl.test.Verify;
 import org.eclipse.collections.impl.tuple.ImmutableEntry;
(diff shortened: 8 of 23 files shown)


.idea/checkstyle-idea.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nikhilnanivadekar: Please revert this file

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
```diff
@@ -560,6 +563,18 @@ public void withKeyValue()
         Verify.assertMapsEqual(UnifiedMap.newWithKeysValues("A", 11, "B", 2), mapWith);
     }
 
+    @Test
+    public void withMap()
+    {
+        MutableMapIterable<String, Integer> map = this.newMapWithKeysValues("A", 1, "B", 2);
```
nikhilnanivadekar: Can you add a test case for empty maps?
1) Input map is empty, target map is non-empty
2) Input map is empty, target map is empty
3) Input map is non-empty, target map is empty

nikhilnanivadekar:
@aboullaite thanks for your contribution.

1. Your Eclipse Foundation ECA validation is not passing. Please refer to this to follow the instructions for getting it to pass:
2. I asked for a few more tests in the review comments. These are for regression.
3. The builds are failing due to test failures.

aboullaite:
Thanks @nikhilnanivadekar for your feedback! 1 and 2 are fixed! I am trying to get the build pass right now.
I have a question, now with `withMap` if the argument is `null`, this will trigger a `NullPointerException`. Is it the intended behavior ? should I check for nullability and trigger `IllegalArgumentException` instead.  also should I add test about this specific case ?

nikhilnanivadekar:
Not supporting nulls is an intended behavior but it’ll be great if you could throw IllegalArgumentException instead of an NPE

motlin:
NPE is also fine for null. It's nice to use Objects.requireNonNull with a
helpful error message.

nikhilnanivadekar:
But isn’t IAE better than NPE for tracing and maintenance? Asking honestly, not a rhetorical question

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
aboullaite: I added them!

aboullaite:
I've opted for IAE for the moment! If we decide to switch to a NPE with a `requireNonNull` I can do that.

WDYT?

nikhilnanivadekar:
sounds good!

motlin:
It's not a big deal either way, but Effective Java Item 72: _Favor the use of standard exceptions_ says "If a caller passes `null` in some parameter for which null values are prohibited, convention dictates that `NullPointerException` be thrown rather than `IllegalArgumentException`."

nikhilnanivadekar:
@aboullaite we are close to completing it. The Javadoc needs a little bit of editing. The implementation is correct, in the tests, I have suggested simpler and Eclipse Collections ways. Please reach out if you have any questions.

aboullaite:
@nikhilnanivadekar I'll give the javadoc a second read! However, AFAIK I've implemented the tests you've suggested! Am I missing something ?

nikhilnanivadekar:
Tests are good and correct. Just a few stylistic comments to make the code condensed and use Eclipse Collections :)

aboullaite:
I modified the javadoc for `MutableMapIterable.java`. However could you elaborate more on:
> Tests are good and correct. Just a few stylistic comments to make the code condensed and use Eclipse Collections :)

Thanks

nikhilnanivadekar:
I have already added line comments. I am away from my machine right now, but I can push the changes to test to my branch for reference. Please let me know.

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
```diff
@@ -560,6 +563,50 @@ public void withKeyValue()
...
+    }
+
+    @Test
+    public void withMapTargetEmpty()
+    {
+        MutableMapIterable<String, Integer> map = this.newMap();
+        Map<String, Integer> simpleMap = Stream.of(new Object[][] {
+                { "A", 1 },
+                { "B", 2 },
+        }).collect(Collectors.toMap(data -> (String) data[0], data -> (Integer) data[1]));
```
nikhilnanivadekar:
```suggestion
        Map<String, Integer> simpleMap = Maps.mutable.with("A", 1, "B", 2);
```

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
```diff
@@ -560,6 +563,50 @@ public void withKeyValue()
...
+        MutableMapIterable<String, Integer> mapWith = map.withMap(simpleMap);
+        Assert.assertSame(map, mapWith);
+        Verify.assertMapsEqual(UnifiedMap.newWithKeysValues("A", 1, "B", 2), mapWith);
+    }
+
+    @Test
+    public void withMapEmptyAndTargetEmpty()
+    {
+        MutableMapIterable<String, Integer> map = this.newMap();
+        MutableMapIterable<String, Integer> mapWith = map.withMap(Collections.emptyMap());
```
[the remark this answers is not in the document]
nikhilnanivadekar:
```suggestion
        MutableMapIterable<String, Integer> mapWith = map.withMap(Maps.mutable.empty());
```

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
```diff
@@ -638,6 +685,7 @@ public void rehash_null_collision()
     public void updateValue()
     {
         MutableMapIterable<Integer, Integer> map = this.newMap();
+
```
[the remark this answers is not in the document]
nikhilnanivadekar:
```suggestion
```

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/UnmodifiableMutableMapTest.java
```diff
@@ -284,6 +288,36 @@ public void withKeyValue()
         this.newMapWithKeysValues(1, "One", 2, "2").withKeyValue(null, null);
     }
 
+    @Override
+    @Test
+    public void withMap()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(
+                Stream.of(new Object[][] {{ 1, 'a' }}).collect(Collectors.toMap(data -> (Integer) data[0], data -> (Character) data[1]))));
```
nikhilnanivadekar:
```suggestion
        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(
                Maps.mutable.with(1, Character.valueOf("a"))));
```

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/UnmodifiableMutableMapTest.java
```diff
@@ -284,6 +288,36 @@ public void withKeyValue()
...
+    }
+
+    @Override
+    @Test
+    public void withMapEmpty()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(Collections.emptyMap()));
```
nikhilnanivadekar:
```suggestion
        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(Maps.mutable.empty()));
```

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/UnmodifiableMutableMapTest.java
```diff
@@ -284,6 +288,36 @@ public void withKeyValue()
...
+    }
+
+    @Override
+    @Test
+    public void withMapTargetEmpty()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMap().withMap(
+                Stream.of(new Object[][] {{ 1, 'a' }}).collect(Collectors.toMap(data -> (Integer) data[0], data -> (Character) data[1]))));
```
nikhilnanivadekar:
```suggestion
        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMap().withMap(Maps.mutable.with(1, Character.valueOf("a")));
```

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/UnmodifiableMutableMapTest.java
```diff
@@ -284,6 +288,36 @@ public void withKeyValue()
...
+    }
+
+    @Override
+    @Test
+    public void withMapEmptyAndTargetEmpty()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMap().withMap(Collections.emptyMap()));
```
nikhilnanivadekar:
```suggestion
        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMap().withMap(Maps.mutable.empty()));
```

unit-tests/src/test/java/org/eclipse/collections/impl/map/sorted/mutable/UnmodifiableTreeMapTest.java
```diff
@@ -348,6 +351,36 @@ public void withKeyValue()
         this.newMapWithKeysValues(1, "One", 2, "2").withKeyValue(null, null);
     }
 
+    @Override
+    @Test
+    public void withMap()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(
+                Stream.of(new Object[][] {{ 1, 'a' }}).collect(Collectors.toMap(data -> (Integer) data[0], data -> (Character) data[1]))));
```
nikhilnanivadekar:
```suggestion
        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(Maps.mutable.with(1, Character.valueOf("a"))));
```

unit-tests/src/test/java/org/eclipse/collections/impl/map/sorted/mutable/UnmodifiableTreeMapTest.java
```diff
@@ -348,6 +351,36 @@ public void withKeyValue()
...
+    }
+
+    @Override
+    @Test
+    public void withMapEmpty()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(Collections.emptyMap()));
```
nikhilnanivadekar:
```suggestion
        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(Maps.mutable.empty()));
```

unit-tests/src/test/java/org/eclipse/collections/impl/map/sorted/mutable/UnmodifiableTreeMapTest.java
```diff
@@ -348,6 +351,36 @@ public void withKeyValue()
...
+    }
+
+    @Override
+    @Test
+    public void withMapTargetEmpty()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMap().withMap(
+                Stream.of(new Object[][] {{ 1, 'a' }}).collect(Collectors.toMap(data -> (Integer) data[0], data -> (Character) data[1]))));
```
nikhilnanivadekar:
```suggestion
        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMap().withMap(Maps.mutable.with(1, Character.valueOf("a")));
```

unit-tests/src/test/java/org/eclipse/collections/impl/map/sorted/mutable/UnmodifiableTreeMapTest.java
```diff
@@ -348,6 +351,36 @@ public void withKeyValue()
...
+    }
+
+    @Override
+    @Test
+    public void withMapEmptyAndTargetEmpty()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMap().withMap(Collections.emptyMap()));
```
nikhilnanivadekar:
```suggestion
        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMap().withMap(Maps.mutable.empty()));
```

nikhilnanivadekar:
I forgot that I should give code suggestions on github itself. I have done that in the tests. Also, I apologize, I had forgotten to submit my previous review so it was sitting in draft forever. I have submitted it now. Hopefully the review makes it clear. 
Also, this is a great way for you to see the difference between a concise way of writing code using Eclipse Collections compared to the verbose Java Collections way.
Lastly, for the Javadoc, I dont think we are supported fixedsize and immutable maps. So, you can remove those terms.
This is almost close to a merge! Thanks for your help and contribution! 👍

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -152,6 +152,25 @@ default boolean removeIf(Predicate2<? super K, ? super V> predicate)
      */
     MutableMapIterable<K, V> withKeyValue(K key, V value);
 
+    /**
+     * This method allows mutable, fixed size, and immutable maps the ability to add elements of a Map to their existing
```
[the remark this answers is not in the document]
aboullaite: I am planning to! but I'll fix this one first, merge, and work on fixed size and immutable maps support as well.

aboullaite:
Thanks a lot @nikhilnanivadekar for your patience and support! I've made a new commit solving all the improvements you've mentioned :)

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -152,6 +152,25 @@ default boolean removeIf(Predicate2<? super K, ? super V> predicate)
...
+     * map = map.withKeyValue("new key", "new value");
+     * </pre>
+     * In the case of FixedSizeMap, a new instance will be returned by withKeyValue, and any variables that
+     * previously referenced the original map will need to be redirected to reference the new instance. In the case
+     * of a FastMap or UnifiedMap, you will be replacing the reference to map with map, since FastMap and UnifiedMap
+     * will both return "this" after calling put on themselves.
+     *
+     * @see #put(Object, Object)
+     */
+    MutableMapIterable<K, V> withMap(Map<K, V> map);
```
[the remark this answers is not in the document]
motlin: "with mutable maps"

I feel like this must mean "immutable" no?

Does the method return `this`? It should say so.

This behaves kind of like putAll() right? The link in @see should be to putAll if so.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
```diff
@@ -153,6 +153,17 @@ else if (other == value || value.equals(other))
...
+    @Override
+    public MutableBiMap<K, V> withMap(Map<K, V> map)
+    {
+        if (map == null)
+        {
+            throw new IllegalArgumentException("Cannot Add a null Map!");
+        }
+        map.forEach(this::put);
+        return this;
+    }
```
motlin: I don't think we should have a null check.

Instead of delegating to forEach, I think we should be delegating to putAll() which sometimes is better optimized (See UnifiedMap for an example).

Taken together, the implementation for all mutable maps will be:
```java
this.putAll(map);
return this;
```
Which means the implementations can be pulled up to interfaces as default methods. AbstractMutable*Maps will be unchanged.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
```diff
@@ -291,6 +292,20 @@ public V forcePut(K key, V value)
...
+    {
+        if (map == null)
+        {
+            throw new IllegalArgumentException("Cannot Add a null Map!");
+        }
+        synchronized (this.lock)
+        {
+            map.forEach(this::put);
+            return this;
+        }
```
motlin: Synchronized wrappers should delegate to methods of the same name. But then return `this`, the wrapper.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
aboullaite: The null check is following the discussion we had previously in this PR! is just to trigger IAE instead of NPE! If the default behaviour should be to trigger NPEs instead, then I remove it and update the tests! WDYT ?

motlin: Yes, I replied. Repeating here:

> It's not a big deal either way, but Effective Java Item 72: Favor the use of standard exceptions says "If a caller passes null in some parameter for which null values are prohibited, convention dictates that NullPointerException be thrown rather than IllegalArgumentException."

In addition, this method feels very similar to putAll(). If you call Map.putAll(null) on any map, including our own implementations, you'll get a NullPointerException. None of them have a null check either.

aboullaite: Got it! Thanks

aboullaite: @motlin I'm facing some test failures after removing the null check from the method! some classes (`org.eclipse.collections.impl.map.mutable.ConcurrentHashMap` for example) are triggering IAE since they use `java.util.Map#putAll` method, however others (`org.eclipse.collections.impl.bimap.mutable.HashBiMap`) trigger NPE, since `putAll` method is overridden in `org.eclipse.collections.impl.bimap.mutable.AbstractMutableBiMap`

nikhilnanivadekar:
While being pedantic and following Effective Java is great! But as a user, I feel that NPEs are super unintuitive and hard to narrow down to find the issue. I would rather have an IAE where I get to know what exactly is incorrect.

motlin:
I somewhat agree. This is a method with a single argument. The only way it
can throw NPE is if it's one argument is null. In this case, it's also just
like putAll(), should be implemented in terms of putAll(), and the
exception should even include putAll() in the stack trace.

nikhilnanivadekar:
I think we need to provide direction to @aboullaite about this. I don’t quite understand your line of thought of having putAll() in the stacktrace. 
Let’s get this PR merged by throwing an IAE and delegating to putAll() in the implementation. If there are additional improvements we can do it as a part of another PR.

motlin:
The implementation should be exactly:
```
this.putAll(arg);
return this;
```
and should be a default method on the interface. I would like to be
convinced why it should be something more complex than that.

nikhilnanivadekar:
Now I understand. @aboullaite can you please change the implementation to be only what Craig pointed out? If there are inconsistencies in the exceptions thrown then you’ll need to change the tests for this method accordingly. Thanks for your contribution @aboullaite and thanks @motlin for the clarification!

aboullaite:
Thank you both for your help! On it :)

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -161,11 +161,15 @@ default boolean removeIf(Predicate2<? super K, ? super V> predicate)
      * map = map.withMap(someMap);
      * </pre>
      *
-     * @see #put(Object, Object)
+     * @see #putAll(Map)
      *
      * @since 10.3.0
      */
-    MutableMapIterable<K, V> withMap(Map<K, V> map);
+    default MutableMapIterable<K, V> withMap(Map<K, V> map)
```
motlin: While I do think we should pull the method up, we need it to have the correct generics, which is a breaking change. It needs the bounded wildcards ? extends K, ? extends V

aboullaite: Oh yes! That is true. Fixing it right away

aboullaite:
@nikhilnaniv @motlin any required action from my side?! Just double checking 😊

nikhilnanivadekar:
@motlin can you check and approve. this looks good from my perspective.

nikhilnanivadekar:
@aboullaite can you squash all commits and rebase. It will be easier to review.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/map/MutableMapFactory.java
```diff
@@ -72,6 +72,9 @@
 
     <K, V> MutableMap<K, V> ofMap(Map<? extends K, ? extends V> map);
 
+    /**
+     * @since 10.3.0
+     */
```
motlin: Not accurate. @nikhilnanivadekar added this in 2018.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
```diff
@@ -291,6 +292,16 @@ public V forcePut(K key, V value)
         }
     }
 
+    @Override
+    public MutableBiMap<K, V> withMap(Map<? extends K, ? extends V> map)
+    {
+        synchronized (this.lock)
+        {
+            this.getDelegate().putAll(map);
```
motlin: this could be this.putAll instead of this.getDelegate().putAll but I'm not sure if that matters

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
```diff
@@ -1979,6 +1979,12 @@ public void zeroOutQueuePosition()
         return (ConcurrentHashMap<K, V>) super.withKeyValue(key, value);
     }
 
+    @Override
+    public ConcurrentHashMap<K, V> withMap(Map<? extends  K, ? extends V> map)
+    {
+        return (ConcurrentHashMap<K, V>) super.withMap(map);
```
motlin: Any reason not to do this.putAll(); return this; ?

aboullaite: I basically followed the same coding style here for all `with*` methods! only reason is consistency

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/map/MutableMapFactory.java
aboullaite: My bad! shouldn't be here! fixing it ...

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
donraab: @molin Can you take a look at this and review/approve? I took a quick look and didn't see any casts.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -152,6 +152,25 @@ default boolean removeIf(Predicate2<? super K, ? super V> predicate)
...
     MutableMapIterable<K, V> withKeyValue(K key, V value);
 
+    /**
+     * This method allows mutable maps the ability to add elements of a Map to their existing
+     * elements. The original map is modified and then returned. In order to use this method properly with mutable maps
+     * the following approach must be taken:
+     * <p>
+     * <pre>
+     * map = map.withMap(someMap);
+     * </pre>
```
motlin, line 162: The javadoc seems wrong. At first I thought this was going to be about a method on mutable/immutable maps. If you're not sure if the method is mutating or not, then the safe thing to do is:
```
newReference = oldReference.possiblyMutatingMethod(args)
```
This one is specifically on mutable maps so the guidance doesn't need to be, and should not be followed. If you just delete "In order to use this method properly" until the end, I think it's good.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/SynchronizedSortedMap.java
```diff
@@ -122,6 +123,16 @@ public SynchronizedSortedMap(MutableSortedMap<K, V> newMap, Object lock)
         }
     }
 
+    @Override
+    public MutableSortedMap<K, V> withMap(Map<? extends K, ? extends V> map)
+    {
+        synchronized (this.lock)
+        {
+            this.putAll(map);
+            return this;
```
motlin, line 132: Some of the synchronized wrappers call this.putAll and some call this.getDelegate.putAll. I prefer this.putAll, like we see here.

unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/UnmodifiableBiMapTest.java
```diff
@@ -161,6 +164,36 @@ public void withKeyValue()
         Verify.assertThrows(UnsupportedOperationException.class, () -> this.getEmptyMap().withKeyValue(1, 'a'));
     }
 
+    @Override
+    @Test
+    public void withMap()
+    {
+        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(
```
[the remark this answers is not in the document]
motlin: Instead of "mutation" I think you mean "mutable"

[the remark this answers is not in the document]
motlin: edited inline

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
aboullaite: Roger that!

aboullaite:
Thanks @motlin, I made a new commit with the requested changes (I didn't notice that @nikhilnanivadekar requested changes as well until you mentioned it :/)

nikhilnanivadekar:
@aboullaite this looks good to me. @motlin you still have any changes required or can you approve the Review as well.
Lastly @aboullaite , a little bit of housekeeping. Can you please change the commit message to have only "Add withMap() to MutableMap. Closes #761 " Please? It helps during preparing release notes. Thanks!

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -152,6 +152,20 @@ default boolean removeIf(Predicate2<? super K, ? super V> predicate)
      */
     MutableMapIterable<K, V> withKeyValue(K key, V value);
 
+    /**
+     * This method allows mutable maps the ability to add elements of a Map to their existing
+     * elements. The original map is modified and then returned.
```
motlin, line 157: The wording is confusing to me. I'd suggest something like "Copies all of the mappings from the specified map to this map and returns this map" or "Similar to putAll() but returns `this` instead of `void`"

aboullaite:
@motlin I pushed new changes based on your feedback. Can you check ?

motlin:
Looks good now. Just needs to be rebased on top of upstream/master.

motlin:
This was a tricky one. Thanks for sticking with me and getting it landed!

PR #804, opened 2020-02-24
use `mvn verify` instead of `mvninstall` in the contributing guide

aboullaite:
This is mainly a recommendation inspired from the blog series of Andres Almiray (https://blogs.oracle.com/author/e475065c-7c0f-4efc-98f4-7d6d0212138d): 
> Friends don't let friends use `mvn clean install`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,7 +38,7 @@ The Eclipse Collections build requires below as dependencies.
 The Eclipse Collections build performs code generation to create primitive collections. Run the full build once before opening your IDE.
 
 ```bash
-$ mvn install -DskipTests=true
+$ ./mvnw -DskipTests=true
 ```
 
 Now you can open the project in your IDE and it won't complain about missing files. You'll be able to use the IDE to perform incremental builds and run tests. You should rarely need to run the maven build, except when:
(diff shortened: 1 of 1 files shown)


motlin:
`verify` instead of `install` may be ok, but neither command passes. I think the command we should encourage people to run is `./mvwn` and then we can do the rest in `pom.xml`.

nikhilnanivadekar:
@motlin is that the command you are suggesting to change in the contributing.md?

motlin:
Yes. I gave the new command a try: `mvn clean verify checkstyle:check findbugs:check --projects '!scala-unit-tests,!jmh-scala-tests,!jmh-tests' -DskipTests=true`

It fails, so I tried with `mvn install` again. It also fails. So that's bad and we should fix it. In the meantime, we've also added the maven wrapper and we can configure default build targets (verify, install, or something else) inside the root pom.xml. So I think we should recommend running `./mvnw`

nikhilnanivadekar:
So only change is to run ./mvnw or there is some other change? So that it can be updated in this PR and we can close it

motlin:
Updated in-place.

donraab:
@motlin is this ok to approve and merge now?

motlin:
yes!

nikhilnanivadekar:
@aboullaite something has gone wrong in the rebase. Seems like all the merge commits are pushed. Can you please fix it and then we can merge?

aboullaite:
@nikhilnanivadekar Ughh! Sorry for this, I wanted to sync my repo and this happened! Should be fixed now

motlin:
It's back to being based on 10.2.0 instead of the latest on master. Try these commands:
```
git checkout mvn-verify
git pull https://github.com/eclipse/eclipse-collections master --rebase
git push origin mvn-verify -f
```

aboullaite:
I don't think the job failure is related to my commit!

Issue #647, opened 2018-11-16
Delegate abstract class for each of the EC iterables


victornoel:
This is a bit similar to #273 in spirit but it seems different.

My need would be to have wrapper/adapter abstract classes that I could use to have classes that implements one of the EC interface but delegating to an existing implementation.

For example:
```java
public final class MyClassWhichIsAlsoACollection extends AbstractListIterable<Stuff> {
    public MyClassWhichIsAlsoACollection() {
        this.xxxx = // create a collection or whatever;
    }

    protected ListIterable<Stuff> delegate() {
        this.xxxx;
    }
}
```

In a sense, this exactly looks like what `ListAdapter` is but it's not public and it's final!

victornoel:
For the record, the closest I found to my need is `LazyIterableAdapter`.

victornoel:
After some exploration, I think it would be great if all the adapter were non-final so they can be used as wrapper classes.

victornoel:
@motlin @nikhilnanivadekar @donraab would you be open to this? I can make a PR.

PR #806, opened 2020-03-01
Ensure all adapters and decorators can be extended

victornoel:
I got bitten by #647 again so decided to make a PR directly :)

The constructors could have been made public but the EC way seems to be using the `of()` static method for this usage, so I left it at that.


motlin:
I believe this is technically a security vulnerability, but I'm also kind of open to it. A malicious user could deserialize some custom bytes into a subclass of an unmodifiable wrapper that's not actually unmodifiable. On the other hand, I haven't used serialization in years and if you're security conscious, you probably forbid deserialization. It's something to think about though. We prevented subclassing on purpose at the time. The JDK wrappers do the same thing.

motlin:
I'm looking around in Guava, and they don't even have concrete classes. Unmodifiable wrappers are all abstract.

I don't fully understand why the concrete classes must have hidden visibility, but I think it's important.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/CollectionAdapter.java
```diff
@@ -44,7 +44,7 @@
  * {@link #wrapList(Iterable)} factory method. To wrap a collection with the MutableCollection interface alone, use
  * the {@link #adapt(Collection)} factory method.
  */
-public final class CollectionAdapter<T>
+public class CollectionAdapter<T>
         extends AbstractCollectionAdapter<T>
```
motlin: Note how CollectionAdapter extends AbstractCollectionAdapter. This is the pattern I think we should copy. The superclass is public api.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/UnmodifiableBag.java
```diff
@@ -64,7 +64,7 @@
         extends AbstractUnmodifiableMutableCollection<T>
         implements MutableBag<T>, Serializable
 {
-    UnmodifiableBag(MutableBag<? extends T> mutableBag)
+    protected UnmodifiableBag(MutableBag<? extends T> mutableBag)
```
motlin: Note how UnmodifiableBag extends AbstractUnmodifiableMutableCollection. I think the right thing to do here is extract a new superclass AbstractUnmodifiableBag, which becomes part of our public api. I think we should leave the sublcass alone.

victornoel:
@motlin in my own projects, I usually use the following pattern to solve this kind of situation:
- I have only abstract or final classes, and abstract classes have either final or abstract methods
- I have abstract "envelopes" which are classes whose sole purpose is to implement an interface, wrap an object of that interface and delegate all methods (defined as final) to the object.

With this pattern, I can define any implementation of the interfaces in terms of existing implementations. See https://github.com/victornoel/eo-envelopes for a tooling made and use to simplify my life to apply this pattern.

Writing envelopes for all the interface of EC could be a good alternative to this PR :)

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/UnmodifiableBag.java
victornoel: what would be left in the `UnmodifiableBag` class at then end? nothing except being final for security purpose?

motlin: That's right. I'm not even sure I know why, but I see that the JDK does it and Guava does it, and I think we must have done it for a similar reason.

PR #813, opened 2020-03-06
Add GitHub Actions Workflow

brunoborges:
Builds on Ubuntu and for the following versions of Java:

- 8
- 11
- 14-EA
- 15-EA

Since it uses the Setup Java action from GitHub, it leverages Azul Zulu builds by default. If needed, it is possible to download (wget) and then use the parameter `jdkFile` to tell Setup Java where to get the JDK.


brunoborges:
@nikhilnanivadekar done

nikhilnanivadekar:
LGTM

PR #819, opened 2020-03-08
Add missing methods to ImmutableMapFactory

### LINKED ISSUE #738: ImmutableMapFactory missing withMap, ofMapIterable and withMapIterable
There are three methods missing on `ImmutableMapFactory` interface.  These are the methods on `MutableMapFactory`.

```
<K, V> MutableMap<K, V> ofMap(Map<? extends K, ? extends V> map);
<K, V> MutableMap<K, V> withMap(Map<? extends K, ? extends V> map);
<K, V> MutableMap<K, V> ofMapIterable(MapIterable<? extends K, ? extends V> mapIterable);
<K, V> MutableMap<K, V> withMapIterable(MapIterable<? extends K, ? extends V> mapIterable);
```
These are currently the methods available on ImmutableMapFactory.
```
<K, V> ImmutableMap<K, V> ofMap(Map<K, V> map);
<K, V> ImmutableMap<K, V> ofAll(Map<K, V> map);
<K, V> ImmutableMap<K, V> withAll(Map<K, V> map);
```
The `ofAll` and `withAll` methods should be deprecated on `ImmutableMapFactory` and `withMap`, `ofMapIterable` and `withMapIterable` should be added.

kedar-joshi:
Additional changes include -
1. Replace usage of newly deprecated methods.
2. Update interface Javadoc to align with deprecated
implementation in (ImmutableMapFactoryImpl#ofMap).

Closes gh-738

Signed-off-by: Kedar Joshi <kdar_joshi@yahoo.com>

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/map/ImmutableMapFactory.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/map/ImmutableMapFactory.java
@@ -13,6 +13,7 @@ package org.eclipse.collections.api.factory.map;
 import java.util.Map;
 
 import org.eclipse.collections.api.map.ImmutableMap;
+import org.eclipse.collections.api.map.MapIterable;
 
 public interface ImmutableMapFactory
 {
@@ -59,12 +60,37 @@ public interface ImmutableMapFactory
 
     <K, V> ImmutableMap<K, V> with(K key1, V value1, K key2, V value2, K key3, V value3, K key4, V value4);
 
+    /**
+     * @deprecated use {@link #withMap(Map)} instead (inlineable).
+     */
+    @Deprecated
     <K, V> ImmutableMap<K, V> ofMap(Map<K, V> map);
 
     /**
      * Same as {@link #withAll(Map)}.
+     * @deprecated since 10.3.0, use {@link #withMap(Map)} instead.
      */
+    @Deprecated
     <K, V> ImmutableMap<K, V> ofAll(Map<K, V> map);
 
+    /**
+     * @deprecated since 10.3.0, use {@link #withMap(Map)} instead.
+     */
+    @Deprecated
     <K, V> ImmutableMap<K, V> withAll(Map<K, V> map);
+
+    /**
+     * @since 10.3.0
+     */
+    <K, V> ImmutableMap<K, V> withMap(Map<K, V> map);
+
+    /**
+     * @since 10.3.0
+     */
+    <K, V> ImmutableMap<K, V> ofMapIterable(MapIterable<K, V> mapIterable);
+
+    /**
+     * @since 10.3.0
+     */
+    <K, V> ImmutableMap<K, V> withMapIterable(MapIterable<K, V> mapIterable);
 }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableMapFactoryImpl.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableMapFactoryImpl.java
@@ -15,6 +15,7 @@ import java.util.Objects;
 
 import org.eclipse.collections.api.factory.map.ImmutableMapFactory;
 import org.eclipse.collections.api.map.ImmutableMap;
+import org.eclipse.collections.api.map.MapIterable;
 
 public class ImmutableMapFactoryImpl implements ImmutableMapFactory
 {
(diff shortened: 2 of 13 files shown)


kedar-joshi:
I think I should also add `@Since` on new methods. Right ?

nikhilnanivadekar:
@kedar-joshi thanks for your contribution! Before I review the PR, a little bit of admin stuff so that it is easier for me when I write the release notes. Can you break the additional changes which you have done in separate commits, please? It is fine to have it in this PR, just separate commits, so that I can look at the history easily and compile the release notes. Also, the history remains clean.
Yes, please add `since` tag as well.
Please ping me when it is done, I will review it. Cheers!
Thanks for helping us out!

kedar-joshi:
> Can you break the additional changes which you have ..

Sure.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/map/ImmutableMapFactory.java
```diff
@@ -59,12 +60,37 @@
 
     <K, V> ImmutableMap<K, V> with(K key1, V value1, K key2, V value2, K key3, V value3, K key4, V value4);
 
+    /**
+     * @deprecated use {@link #withMap(Map)} instead (inlineable).
+     */
```
nikhilnanivadekar: Mention deprecated since 10.3.0

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableMapFactoryImpl.java
```diff
@@ -184,4 +206,58 @@
...
+        return this.withMapIterable(mapIterable);
+    }
+
+    /**
+     * @since 10.3.0
+     */
+    @Override
+    public <K, V> ImmutableMap<K, V> withMapIterable(MapIterable<K, V> mapIterable)
+    {
+        if (mapIterable.isEmpty())
```
nikhilnanivadekar: The entire implementation is just:
```suggestion
      return (ImmutableMap) this.mapIterable.toImmutable();
```

nikhilnanivadekar (changes requested):

nikhilnanivadekar:
There are build failures. Also the implementation is far easier than what you have right now. Please take a look in the review. Thanks for your contribution!

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/map/ImmutableMapFactory.java
```diff
@@ -59,12 +60,37 @@
...
+    @Deprecated
```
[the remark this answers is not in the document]
kedar-joshi: It's implementation has been deprecated ever since it was committed. I have just aligned the interface with its implementation.

https://github.com/eclipse/eclipse-collections/blob/417aa68093a26897ac46fa6ec7023b8ec0ea6ca3/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableMapFactoryImpl.java#L139-L146

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/map/ImmutableMapFactory.java
```diff
@@ -59,12 +60,37 @@
...
+
+    /**
+     * @since 10.3.0
+     */
+    <K, V> ImmutableMap<K, V> ofMapIterable(MapIterable<K, V> mapIterable);
+
+    /**
+     * @since 10.3.0
+     */
+    <K, V> ImmutableMap<K, V> withMapIterable(MapIterable<K, V> mapIterable);
```
motlin: ImmutableMapFactoryImpl has a single deprecated method.

```java
    /**
     * @deprecated use {@link #ofAll(Map)} instead (inlineable)
     */
    @Override
    @Deprecated
    public <K, V> ImmutableMap<K, V> ofMap(Map<K, V> map)
    {
        return this.ofAll(map);
    }
```

This does not bring the interface in line with the implementation. The one method that ought to be deprecated doesn't match the correct suggestion.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableMapFactoryImpl.java
```diff
@@ -132,23 +133,42 @@
...
     public <K, V> ImmutableMap<K, V> withAll(Map<K, V> map)
+    {
+        return this.withMap(map);
+    }
+
+    /**
+     * @since 10.3.0
+     */
+    @Override
+    public <K, V> ImmutableMap<K, V> withMap(Map<K, V> map)
```
motlin: I feel like the change you're making is different than what the title says. You're really adding a new method called withMap and deprecating ofAll and withAll at the same time in favor of it.

motlin: In addition, this only changes ImmutableMapFactory. It misses other object maps (BiMap, hashing strategy) and primitive maps.

motlin (changes requested):
I'm marking this as "request changes" so we don't merge it yet, but let's chat about it. I'm not sure we should go this route.

nikhilnanivadekar:
Looping in @donraab as he initially opened the issue.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableMapFactoryImpl.java
kedar-joshi: > I feel like the change you're making is different than what the title says. You're really adding a new method called withMap and deprecating ofAll and withAll at the same time in favor of it.

@motlin I kept phrasing of the title similar to the original issue gh-738.

motlin:
I see the context of gh-738 now, but the deprecation still doesn't make sense as written. The methods with bad generics are deprecated, but the new methods also have bad generics. There's no version of ofMap that's not deprecated.

I think we should experiment with pull methods up to a super-interface to drive consistency (in a separate commit) to make this sort of thing enforced by the compiler.

motlin:
Here's an example, to start the conversation: https://github.com/motlin/eclipse-collections/commit/f9397b23d0f2876dce1d184540a5bc1e3a378333

motlin:
And here's a similar example for Maps. This one doesn't compile because the interfaces are truly not compatible.
https://github.com/motlin/eclipse-collections/commit/07c81aad52e4f9fb797f5e9a804fac0773447314

anishlukk123:
is this pull request abadoned?

donraab:
Closing this for now as it's been 6 years.

PR #831, opened 2020-04-02
Add method wrapCopy to primitive lists to mirror functionality in Fas…

### LINKED ISSUE #742: Add method wrapCopy to Primitive Lists to mirror the functionality in FastList
Let me take `MutableIntList` as an example.

`int[] array = { 1, 2, 3, 4 };`
`MutableIntList list = IntLists.mutable.of(array);`

In the source code of `IntArrayList`, the constructor takes an initial array as parameter and holds the reference to it in `this.items`:

`public IntArrayList(int... array)`
`{`
`    this.size = array.length;`
`    this.items = array;`
`}`

When invoke `removeAtIndex()`, the item will be removed from both the list and the initial array (this behavior is a disaster and should be documented clearly!):

`list.removeAtIndex(0);`
`System.out.println(Arrays.toString(array)); // [2,3,4,0]`

On the other hand, if invoke `add()` before `removeAtIndex()`, due to `ensureCapacityForAdd()`, `this.items` will reference to a new larger array, and `removeAtIndex()` will no longer affect the initial array:

`list.add(5);`
`list.removeAtIndex(0);`
`System.out.println(Arrays.toString(array)); // [1,2,3,4]`

So for Mutable Primitive List, pls consider
* make a array copy in the constructor instead of holding the reference
* if removing items from the initial array is by design, pls make the behavior clear in the document, and, add another version of `removeAtIndex()` which affects on the list only

donraab:
…tList. Fixes #742.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/MutableListFactory.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/MutableListFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 Goldman Sachs and others.
+ * Copyright (c) 2020 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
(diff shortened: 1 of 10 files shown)


nikhilnanivadekar:
LGTM
Feel free to merge

donraab:
Thanks for the review @nikhilnanivadekar.

PR #834, opened 2020-04-05
Remove duplicate forEach overrides.

donraab:

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
@@ -578,12 +578,6 @@ public class <name>ObjectHashMap\<V> implements Mutable<name>ObjectMap\<V>, Exte
     }
 
     @Override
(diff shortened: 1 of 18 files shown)


donraab:
This is the definition of `forEach` in `RichIterable` which all these classes extend. 

```
    @Override
    default void forEach(Procedure<? super T> procedure)
    {
        this.each(procedure);
    }
```
I was looking at code coverage, and saw that the default method in RichIterable was not covered. Now it is.

motlin:
I'm a fan of changes like this. I believe they are source compatible and not binary compatible. Should we wait until we know we're doing a major version?

donraab:
Are you sure this isn't binary compatible?

https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html

"Moving a method upward in the class hierarchy."

I didn't even move the method.  It was already defined in RichIterable as a default method.

nikhilnanivadekar:
It isn’t binary compatible because you’ll be removing implementations

donraab:
Moving a method upward in a class hierarchy is removing a method in a subclass, no?

motlin:
Interesting, I didn't know that's a binary compatible change.

PR #835, opened 2020-04-05
Use org.eclipse.collections.api.factory instead of org.eclipse.collections.impl.factory wherever possible.

motlin:


motlin:
I tried to land the same change once before but the scope got too large and now there are too many merge conflicts to deal with. This doesn't include all possible locations. I skipped performance tests and I skipped primitive collections.

PR #838, opened 2020-04-06
Use interfaces instead of implementations wherever possible without breaking backwards compatibility.

motlin:


motlin:
@donraab I am open to any advice on getting this stuff landed. I've been trying for ages now and keep closing my PRs after there are too many merge conflicts. The idea is if we only use interfaces, then we can use factories. If we can use the api.factory classes like Lists.immutable.with(...) then we can pull the bodies up as default methods and move the implementations from the impl jar to the api jar.

The next thing I'll try is deleting the parts that use interfaces like Function and Procedure and just sticking to collections types.

BTW I'm not making this changes by bulk applying an inspection. I'm doing it manually, because of performance concerns. There are comments like `Implemented to avoid megamorphic call on castProcedure` where it's important not to use the weaker type.

donraab:
If the intent is so you can lift up more behavior, then the PR should include no private methods or constructors, and only public API. Just submit smaller PRs that are focused and it will get merged gradually. That's my advice.

PR #843, opened 2020-04-10
Fix broken link in Javadoc.

motlin:


nikhilnanivadekar:
@motlin I think there was a change in the way to denote the deprecation and alternative classes in JDK11. That is why the Javadoc build is failing.

donraab:
I encountered the same problem when I went to fix the JavaDoc.  IntelliJ said the hashtag method approach was good and complained about the dot separated.  In the end I had to revert my change.

motlin:
I believe the Javadoc build is broken on master, and I'm not trying to fix it. I'm trying to fix the TeamCity build which is also failing due to the unnecessary import.

nikhilnanivadekar:
Which TeamCity build? Also, the Javadoc master build is what runs to create Javadocs for releases, so it has to pass for every PR.

motlin:
Reverted the hashes and kept the fix for the unnecessary fully-qualified name.

motlin:
@nikhilnanivadekar "Inspections - 4 No Auto-fix" sometimes fails on this deprecation line (while I'm working on other things)

nikhilnanivadekar:
ah I see. I was adding a Javadoc build to TeamCity since TC builds are much more faster. I thought for some reason that build got triggered prematurely 😄

PR #845, opened 2020-04-11
Upgrade dependencies used by tests and the build.

motlin:


nikhilnanivadekar:
@motlin the Scala plugin causes issues in Jenkins builds. So, can you please revert it back to what it was.

motlin:
Done. Do you know why it causes a problem?

nikhilnanivadekar:
It is due to incompatibility of Scala versions with JDK. It can be reproduced in our local as well. Try running the clean build/verify using JDK8 and also using JDK11.

PR #847, opened 2020-04-11
Add iml files to gitignore

nikhilnanivadekar:


nikhilnanivadekar:
The gitignore is mainly from feedback by contributors. Every IntelliJ release leads to the iml files being churned. For the project and to have it work correctly, I do not think we need to have it in there. This makes rebasing, merging, etc. painful as we have to always stash these files first.
@donraab @mohrezaei @motlin @cguntur-bnym

.gitignore
```diff
@@ -75,3 +76,18 @@ fabric.properties
 .project
 .settings/
 .classpath
+
+#IntelliJ Project files
```
prathasirisha, line 80: Missing this module's iml file. 
acceptance-tests

.gitignore
```diff
@@ -21,6 +21,7 @@ test-coverage/**
 .idea/**/tasks.xml
 .idea/**/usage.statistics.xml
 .idea/**/shelf
+.idea/*.xml
```
motlin, line 24: This is xml rather than iml and includes many files which are checked in. This line needs to be removed.

.gitignore
```diff
@@ -75,3 +76,18 @@ fabric.properties
...
+eclipse-collections/*.iml
+eclipse-collections-testutils/*.iml
+eclipse-collections-forkjoin/*.iml
+unit-tests/*.iml
+serialization-tests/*.iml
+jcstress-tests/*.iml
+junit-trait-runner/*.iml
+unit-tests-java8/*.iml
+test-coverage/*.iml
+p2-repository/*.iml
```
motlin, line 93: Replace this whole section with `*.iml`. In other words, get rid of the directory prefixes. No need for globs that only match one file.

motlin:
`git rm -f '*.iml'`

motlin:
Once we remove the iml files from version control, it will break all the TeamCity builds. But leaving them in version control and also ignoring them - I don't think that will have the desired effect but you can develop that way and let us know if it works.

motlin:
My version of the same idea: #851

.gitignore
nikhilnanivadekar: The global *.iml doesn’t seem to work. I have tried it before.

nikhilnanivadekar:
No need to remove the iml files, we just don’t need to track them

.gitignore
nikhilnanivadekar: This is not a mistake it’s on purpose. There are xml files which get added due to plugins or new versions for example encodings.xml, hydra.xml. These files should never be committed but cause issues during rebasing. So, having this line is correct

nikhilnanivadekar: This is not a mistake it’s on purpose. There are xml files which get added due to plugins or new versions for example encodings.xml, hydra.xml. These files should never be committed but cause issues during rebasing. So, having this line is correct

motlin:
> No need to remove the iml files, we just don’t need to track them

Why do you think that works? I'll show you why I saw this doesn't work.

```
$ git checkout nikhil/gitIgnore -f
HEAD is now at 8e05c4ed Add iml files to gitignore as they are a major source of annoyance and inconvenience to contributors.
$ grep iml .gitignore 
eclipse-collections-code-generator/*.iml
eclipse-collections-code-generator-maven-plugin/*.iml
eclipse-collections-api/*.iml
eclipse-collections/*.iml
eclipse-collections-testutils/*.iml
eclipse-collections-forkjoin/*.iml
unit-tests/*.iml
serialization-tests/*.iml
jcstress-tests/*.iml
junit-trait-runner/*.iml
unit-tests-java8/*.iml
test-coverage/*.iml
p2-repository/*.iml
$ echo "some deliberate change" >> eclipse-collections/eclipse-collections.iml 
$ git status
HEAD detached at nikhil/gitIgnore
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   eclipse-collections/eclipse-collections.iml

no changes added to commit (use "git add" and/or "git commit -a")
```

.gitignore
motlin: It's too broad. There are 2 or 3 xml files which can be ignored and a bunch which must be committed. Unfortunately, encodings.xml is one that is probably required by TeamCity. I'll test that after I test removing the imls.

nikhilnanivadekar: Sorry for being dense here. In none of the other projects I have worked with to get things to work correctly, was there a necessity for the xml files. If we want an off cycle change we can always work around it. I would rather be too broad than too restrictive

nikhilnanivadekar: Sorry for being dense here. In none of the other projects I have worked with to get things to work correctly, was there a necessity for the xml files. If we want an off cycle change we can always work around it. I would rather be too broad than too restrictive

nikhilnanivadekar:
I am confused, doesn’t it say no changes staged for commit? I tried it in my local to keep the iml files and use different versions of intellij. The iml files change but are never shown in git status.

donraab:
> Changes not staged for commit:

This is the desired behavior I thought?

donraab:
I think Craig is correct... we should see the files in this section right?

> Untracked files:

nikhilnanivadekar:
No they shouldn’t otherwise all files under target should also show up each time we build. Or it’s possible I am missing something

nikhilnanivadekar:
handled in #851

PR #848, opened 2020-04-11
Add covariant overrides for sortThis().

motlin:
While reviewing #801, I wanted to point out that methods that return `this` need covariant overrides on all sub-interfaces. I wanted to point to sortThis() as an example and was surprised to find that it's missing overrides.

While unlikely to cause problems for users, I believe this is a breaking change and can wait for the next major version.


nikhilnanivadekar:
We have 2 options,
1) if we can close up the feature set that is currently in rest of the PRs
I can cut a 10.3.0 release. With the new Eclipse Foundation process, we can
release whenever we are ready without the wait period.
2) The other option is to just open 11.0 right away.
I would prefer option 1.

motlin:
That sounds good to me

PR #851, opened 2020-04-11
Remove *.iml files from version control.

motlin:


nikhilnanivadekar:
Why are we removing the iml files? I mean they are fine the way they are... once ignored they won’t be causing a problem. I feel like by removing them, we will create a non-existent problem and solution

motlin:
@nikhilnanivadekar I answered on the other thread, but from what I already understood about .gitignore and from the example I posted, ignore globs have no effect on tracked files. It's in or out. No such thing as ignoring only new changes. If you figure out a way to get git to ignore only new changes, please me know / show me how.

motlin:
There's `--assume-unchanged` but it's so advanced and it's local, it's probably not a good option.

motlin:
For some reason, only Inspections - 3 failed, the others passed.

```
  Initializing project...Loaded shared project profile '3 Consistent Style'
21:50:24
  done.
21:50:24
  Inspecting with profile '3 Consistent Style'
21:50:24
  This project contains no modules. Inspection cannot proceed.
21:50:25
  Process exited with code 1
21:50:25
  Inspection output
21:50:25
  Process exited with code 1 (Step: Inspections (IntelliJ IDEA))
```

The `no modules` error is what I was concerned could happen if we deleted all the iml files. Is this what JetBrains worked on fixing?

nikhilnanivadekar:
Seems like both Inspections-1 and Inspections-4 failed due to the same error.
Let us go ahead with this approach. It at least reduces the pain for contributors. I will raise an issue with TC to see what they can do to help. Till they fix the issue or we find a workaround, we can ignore the Inspections build failures.
Thoughts?

.gitignore
```diff
@@ -52,6 +52,7 @@ cmake-build-*/
 
 # IntelliJ
 out/
+*.iml
```
nikhilnanivadekar, line 55: can you add `idea/encodings.xml` as well?

motlin: So now we know that this breaks TeamCity builds. We have enough info to report to JetBrains or update our existing report. Do we have one in progress?

motlin:
@nikhilnanivadekar I keep pushing updates to watch the TeamCity builds. They're pretty consistently failing. I'm happy to help report this to JetBrains. I don't plan on landing this in a form that breaks the TeamCity builds.

motlin:
@nikhilnanivadekar I've also been thinking about the experiment with .gitignore changes for tracked files. I'm pretty confident it has no effect on git, but maybe it has an effect on IntelliJ? I'm curious how it helps.

nikhilnanivadekar:
It is only the 2 inspections builds right? If you can do the same steps for encodings.xml, I’ll report both the issues to JetBrains. At this moment the benefit of ignoring these files is far higher than the pain of 2 broken builds.

nikhilnanivadekar:
Before I pull I report to JetBrains can you remove encodings.xml as well and see the effect. It’ll help to show them the behavior in case we have any unexpected effect. Thanks! I’ll report to JetBrains tomorrow morning.

donraab:
I would recommend we hold off until JetBrains fixes the failures. While I appreciate the pain this causes developers having to stash files (myself included), allowing the inspections builds to fail for any period of time would suggest to me that they are not useful for catching real errors and thus should be disabled full stop.

motlin:
I created a second draft PR with encodings.xml also removed so you can see
both builds outputs.

https://github.com/eclipse/eclipse-collections/pull/859

There are other files that we ought to be able to remove like compiler.xml
and modules.xml.

nikhilnanivadekar:
Let's consider only #859 for these changes.

PR #857, opened 2020-04-13
Fix inspections violations.

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/immutable/ImmutableEmptyStack.java
```diff
@@ -103,7 +103,7 @@ private Object readResolve()
     @Override
     public ImmutableStack<T> push(T item)
     {
-        return new ImmutableNotEmptyStack<T>(item, this);
+        return new ImmutableNotEmptyStack<>(item, this);
```
nikhilnanivadekar, line 106: This was a bug in previous versions which I had reported. Glad they fixed it 😄

nikhilnanivadekar:
Retriggered failing GitHub actions build. Please feel free to merge once the build passes.

eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/immutable/ImmutableEmptyStack.java
motlin: Unfortunately, this inspection is still buggy in some edge cases.

PR #859, opened 2020-04-13
Remove .idea/encodings.xml from version control and update .gitignore.

motlin:


nikhilnanivadekar:
I have updated the configuration for Inspections-4. Post that, the build seems to have succeeded. If this works, I will fix the configuration for all the other inspection builds. 
The real test will be, if you can do a commit with inspection violations for each of the profiles, let us see if they get detected?

nikhilnanivadekar:
Additionally, if you could delete and ignore all the xml files which were added only for TeamCity builds, I will check if the configurations work.

nikhilnanivadekar:
FWIW now the inspections builds are failing due to inspections violations. So, I think we are good 😄 
Let me know once you have removed all the xml files, I will check the builds and tweak the configurations accordingly.

motlin:
Once we delete the iml and xml files, we'll lose the ability to reconfigure TeamCity, no?

nikhilnanivadekar:
Didn’t follow your thought process about reconfiguring TeamCity builds. We *shouldnt* remove the *Inspections*.xml files. But all other Xmls which were historically added only for TeamCity, we should try to remove

motlin:
@nikhilnanivadekar you said you fixed the inspections builds by reimporting them. That imports not just the inspection profiles, but the iml files. Once we delete them, clicking around in TeamCity and pressing reimport project will fail.

nikhilnanivadekar:
That’s not what I did. I just removed the dependency of the imls in Inspection-4. You can check the logs for Inspection-1 where I have not removed the dependency those builds work as well. If you look at the build configs then it might make it clear.
Can you remove the unnecessary xml files and we can check if it works?
I think we are close to getting this working :)

nikhilnanivadekar:
BTW seems like the build failures are seemingly legitimate from the log but I might be mistaken as I looked at it on the phone and might have missed something

motlin:
I don't follow what you fixed? But poking around in the configs, it looks like maybe you changed the project type from IntelliJ to Maven, and only on profile 4. Is that right?

nikhilnanivadekar:
Only on profile 4, yes. There is an option under IntelliJ build to not require iml files. I just disabled that and increased the memory. I tried with Maven build and that works as well

motlin:
@nikhilnanivadekar that's excellent, that's the new information for me. Do you want me to update the other 3 profiles?

nikhilnanivadekar:
Yes please go ahead. You might need to put the compiler.xml back in. I think that’s why the jmh test module is failing for compilation

nikhilnanivadekar:
Heard back from JetBrains. Can you remove modules.xml? And convert to Maven project.

https://teamcity-support.jetbrains.com/hc/requests/2599837

motlin:
I can't see that support link. Do you want to take over the branch again
from here?

nikhilnanivadekar:
Sure will do

nikhilnanivadekar:
Moved to #866

Issue #822, opened 2020-03-14
Pull up the implementations of with(), without(), withAll(), withoutAll() as default methods.


motlin:
These four methods on MutableCollection could be pulled up as default methods on it and its sub-interfaces.

```
    MutableCollection<T> with(T element);
    MutableCollection<T> without(T element);
    MutableCollection<T> withAll(Iterable<? extends T> elements);
    MutableCollection<T> withoutAll(Iterable<? extends T> elements);
```

donraab:
The thing that always perplexed me about doing this is that you have to choose then how to co-variantly override. Choice one says call super and cast the return type (pro - no duplicate code). Choice two says reimplement in the sub-interface with the more specific return (pro - no casting). I wish there was a choice three. Let's at least agree before someone implements what the best default implementation option is in the sub-interfaces

motlin:
With these you can call super and return `this`

donraab:
Ok, perfect.  Option three: Implicit casting.

PR #862, opened 2020-04-13
Implement with(), without(), withAll(), withoutAll() as default methods

aboullaite:
This PR implements the changes discussed in #822

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MutableBag.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MutableBag.java
@@ -84,16 +84,32 @@ public interface MutableBag<T>
     }
 
     @Override
-    MutableBag<T> with(T element);
+    default MutableBag<T> with(T element)
+    {
+        this.add(element);
+        return this;
+    }
 
     @Override
-    MutableBag<T> without(T element);
+    default MutableBag<T> without(T element)
+    {
+        this.remove(element);
+        return this;
+    }
 
     @Override
-    MutableBag<T> withAll(Iterable<? extends T> elements);
+    default MutableBag<T> withAll(Iterable<? extends T> elements)
+    {
+        this.addAllIterable(elements);
+        return this;
+    }
 
     @Override
-    MutableBag<T> withoutAll(Iterable<? extends T> elements);
+    default MutableBag<T> withoutAll(Iterable<? extends T> elements)
+    {
+        this.removeAllIterable(elements);
+        return this;
+    }
 
     @Override
     MutableBag<T> newEmpty();
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/set/FixedSizeSet.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/set/FixedSizeSet.java
@@ -20,5 +20,17 @@ public interface FixedSizeSet<T>
         extends MutableSet<T>, FixedSizeCollection<T>
 {
     @Override
+    MutableSet<T> with(T element);
+
+    @Override
+    MutableSet<T> without(T element);
+
+    @Override
+    MutableSet<T> withAll(Iterable<? extends T> elements);
+
+    @Override
+    MutableSet<T> withoutAll(Iterable<? extends T> elements);
+
+    @Override
     FixedSizeSet<T> tap(Procedure<? super T> procedure);
 }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/MultiReaderHashBag.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/MultiReaderHashBag.java
@@ -482,34 +482,6 @@ public final class MultiReaderHashBag<T>
     }
 
     @Override
-    public MutableBag<T> with(T element)
-    {
-        this.add(element);
-        return this;
-    }
-
-    @Override
-    public MutableBag<T> without(T element)
-    {
-        this.remove(element);
-        return this;
-    }
-
-    @Override
-    public MutableBag<T> withAll(Iterable<? extends T> elements)
-    {
-        this.addAllIterable(elements);
-        return this;
-    }
-
-    @Override
-    public MutableBag<T> withoutAll(Iterable<? extends T> elements)
-    {
-        this.removeAllIterable(elements);
-        return this;
-    }
-
-    @Override
     public MutableMap<T, Integer> toMapOfItemToCount()
     {
         try (LockWrapper wrapper = this.lockWrapper.acquireReadLock())
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
@@ -624,34 +624,6 @@ public final class MultiReaderFastList<T>
         }
     }
 
-    @Override
-    public MutableList<T> with(T element)
-    {
-        this.add(element);
-        return this;
-    }
-
-    @Override
-    public MutableList<T> without(T element)
-    {
-        this.remove(element);
-        return this;
-    }
-
-    @Override
-    public MutableList<T> withAll(Iterable<? extends T> elements)
-    {
-        this.addAllIterable(elements);
-        return this;
-    }
-
-    @Override
-    public MutableList<T> withoutAll(Iterable<? extends T> elements)
-    {
-        this.removeAllIterable(elements);
-        return this;
-    }
-
     /**
      * This method is not supported directly on a MultiReaderFastList. If you would like to use a ListIterator with
      * MultiReaderFastList, then you must do the following:
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MutableBagIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MutableBagIterable.java
@@ -114,16 +114,32 @@ public interface MutableBagIterable<T> extends Bag<T>, MutableCollection<T>
     MutableList<ObjectIntPair<T>> bottomOccurrences(int count);
 
     @Override
-    MutableBagIterable<T> with(T element);
+    default MutableBagIterable<T> with(T element)
+    {
+        this.add(element);
+        return this;
+    }
 
     @Override
-    MutableBagIterable<T> without(T element);
+    default MutableBagIterable<T> without(T element)
+    {
+        this.remove(element);
+        return this;
+    }
 
     @Override
-    MutableBagIterable<T> withAll(Iterable<? extends T> elements);
+    default MutableBagIterable<T> withAll(Iterable<? extends T> elements)
+    {
+        this.addAllIterable(elements);
+        return this;
+    }
 
     @Override
-    MutableBagIterable<T> withoutAll(Iterable<? extends T> elements);
+    default MutableBagIterable<T> withoutAll(Iterable<? extends T> elements)
+    {
+        this.removeAllIterable(elements);
+        return this;
+    }
 
     @Override
     <V> RichIterable<V> collectWithOccurrences(ObjectIntToObjectFunction<? super T, ? extends V> function);
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/sorted/MutableSortedBag.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/sorted/MutableSortedBag.java
@@ -85,16 +85,32 @@ public interface MutableSortedBag<T>
     MutableSortedMap<T, Integer> toMapOfItemToCount();
 
     @Override
-    MutableSortedBag<T> with(T element);
+    default MutableSortedBag<T> with(T element)
+    {
+        this.add(element);
+        return this;
+    }
 
     @Override
-    MutableSortedBag<T> without(T element);
+    default MutableSortedBag<T> without(T element)
+    {
+        this.remove(element);
+        return this;
+    }
 
     @Override
-    MutableSortedBag<T> withAll(Iterable<? extends T> elements);
+    default MutableSortedBag<T> withAll(Iterable<? extends T> elements)
+    {
+        this.addAllIterable(elements);
+        return this;
+    }
 
     @Override
-    MutableSortedBag<T> withoutAll(Iterable<? extends T> elements);
+    default MutableSortedBag<T> withoutAll(Iterable<? extends T> elements)
+    {
+        this.removeAllIterable(elements);
+        return this;
+    }
 
     @Override
     MutableSortedBag<T> newEmpty();
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/FixedSizeList.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/FixedSizeList.java
@@ -22,6 +22,18 @@ public interface FixedSizeList<T>
         extends MutableList<T>, FixedSizeCollection<T>
 {
     @Override
+    MutableList<T> with(T element);
+
+    @Override
+    MutableList<T> without(T element);
+
+    @Override
+    MutableList<T> withAll(Iterable<? extends T> elements);
+
+    @Override
+    MutableList<T> withoutAll(Iterable<? extends T> elements);
+
+    @Override
     FixedSizeList<T> toReversed();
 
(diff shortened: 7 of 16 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MutableBag.java
```diff
@@ -84,16 +84,32 @@
...
+        return this;
+    }
 
     @Override
-    MutableBag<T> withoutAll(Iterable<? extends T> elements);
+    default MutableBag<T> withoutAll(Iterable<? extends T> elements)
+    {
+        MutableBagIterable.super.withoutAll(elements);
+        return this;
+    }
```
motlin, line 112: Why do these call `super`?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/MutableCollection.java
```diff
@@ -138,7 +150,11 @@
      *
      * @see #removeAll(Collection)
      */
-    MutableCollection<T> withoutAll(Iterable<? extends T> elements);
+    default MutableCollection<T> withoutAll(Iterable<? extends T> elements)
+    {
+        this.removeAllIterable(elements);
+        return this;
+    }
```
motlin, line 157: And these don't call `super`? I prefer these if I had to pick one, for consistency.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/ArrayListAdapter.java
```diff
@@ -238,13 +238,6 @@ public void sort(Comparator<? super T> comparator)
         return (ArrayListAdapter<T>) super.sortThis();
     }
 
-    @Override
-    public ArrayListAdapter<T> with(T element)
```
motlin, line 242: Unfortunately, this is a breaking change. It's a good idea, it just has to wait for the next major version.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MutableBag.java
aboullaite: I tried to follow the same patter suggested by Donald here 
https://github.com/eclipse/eclipse-collections/issues/822#issuecomment-599230918

motlin:
Will you also do MutableSortedSet and MutableSortedBag?

nikhilnanivadekar:
Please do them as separate commits, same PR is fine. It helps with release notes

aboullaite:
@motlin I can add them, yes 👍

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/ArrayListAdapter.java
nikhilnanivadekar: This needs to be removed for minor version

nikhilnanivadekar:
We should do this in 2 steps. Add default methods in minor version and then remove the implementation in major version. You can raise the other PR to remove implementation as well, we will label it as major version and merge it when I open 11.0.0 target in June

aboullaite:
@nikhilnanivadekar roger that 🙏

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MutableBag.java
donraab: > With these you can call super and return `this`

This was your suggestion in the linked comment @aboullaite pointed to above @motlin

motlin:
I think @donraab proved that this is ok in a minor release. No need to split it up.

motlin:
It also doesn't hurt to split it up, but I think only the removal of the covariant concrete types is breaking.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MutableBag.java
donraab: @motlin @nikhilnanivadekar please share your opinions on this question.  I presented two options, and a third was proposed which I agreed with and posted what the resulting solution would look like here:

https://github.com/eclipse/eclipse-collections/issues/822#issuecomment-599230918

motlin: I find that every method which returns `this` delegates to another method and then returns `this`. They're pretty much always two lines. We can basically specify the implementation in Javadoc and have a consistent implementation everywhere. That's why I slightly prefer the implementation without `super`.

I realize this doesn't matter much and both implementations are fine. It's a slight preference. I'll go with whatever you'd like here.

donraab: This is what the implementation will look like in MutableCollection:
```
default MutableCollection<T> with(T element)
{
    this.add(element);
    return this;
}
```
Duplicating the implementation looks like this:
```
@Override
default MutableList<T> with(T element)
{
    this.add(element);
    return this;
}
```
Or calling the super default implementation:
```
@Override
default MutableList<T> with(T element)
{
    MutableCollection.super.with(element);
    return this;
}
```
The question I have is what is the benefit of pulling up the implementation to `MutableCollection` if we are not going to use the pulled up implementation? This could create an uncovered or difficult to cover code path. We could leave it abstract in `MutableCollection` and enforce it being overridden at leaf interfaces. Thoughts?

motlin: Good point. When I do work like this, I really do pull the methods up, using the IntelliJ refactoring. These methods aren't implemented in AbstractMutableCollection. If they were, I'd move them with the confidence that they are already covered. +1 these should be abstract on MutableCollection.

aboullaite: So If I'm not mistaken, this would look something like this in MutableCollection
```
    abstract MutableCollection<T> withAll(Iterable<? extends T> elements);
```
and in other classes 
```
    @Override
    default MutableBagIterable<T> with(T element)
    {
        this.with(element);
        return this;
    }
```

donraab: Correct, minus the abstract as `MutableCollection` is an interface.

aboullaite:
@nikhilnanivadekar Can you advise on build failures ? I tried to understand what's happening and I think it's not related to changes I made, instead `specify sdk '1.8' for module 'acceptance-tests'`?

nikhilnanivadekar:
I’ll need to fix the config of those builds. I’ll do that later today and kick these builds off again.

motlin:
LGTM

donraab:
Hi @motlin are you approving this to merge and will it still require a major version release?

motlin:
I think it's good to go. It just needs to be rebased on top of master. I saw the "update branch" option and tried it, but it created a merge instead of a rebase. This doesn't make it harder to rebase, but it doesn't help either.

donraab:
Please rebase @aboullaite and we can merge.  Thanks!

aboullaite:
@donraab done :)

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/FixedSizeCollection.java
```diff
@@ -23,79 +23,6 @@
 public interface FixedSizeCollection<T>
         extends MutableCollection<T>
 {
-    /**
```
donraab, line 26: @aboullaite I'd like to understand why was this javadoc removed. This indicates the behavior for a `FixedSizedCollection` is different than other `MutableCollection` instances. This is an awkward behavior here to be sure, but this is the unfortunate nature of something that is fixed in size.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/MultiReaderHashBag.java
```diff
@@ -481,34 +481,6 @@ public MutableShortBag collectShort(ShortFunction<? super T> shortFunction)
         }
     }
 
-    @Override
```
donraab, line 484: I think these should be default methods in `MultiReaderBag`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
```diff
@@ -624,34 +624,6 @@ public int lastIndexOf(Object o)
         }
     }
 
-    @Override
```
donraab, line 627: I think these should be default methods in `MultiReaderList`.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/FixedSizeCollection.java
donraab: @aboullaite Now that I have taken a look, please push this JavaDoc down to the implementation classes so it is not lost. The behavior is definitely different out of necessity. I will try and clean this up a bit after we merge your PR.

motlin: I disagree that the Javadoc should be pushed down. It's on an interface where it probably belongs, and the fact that it's abstract is good. Subclasses must have an interesting implementation. I'd suggest just revert the change to FixedSized* interfaces.

Good catch @donraab

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/MultiReaderHashBag.java
motlin: @donraab why? To change the return type to MultiReaderBag? If so, that would be a good change, but one that has to happen in a major version. I think the minor version change looks good as-is, pulling these up to the MutableBag interface.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
motlin: Same comment. Presumably MultiReaderSet too?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/FixedSizeCollection.java
donraab: The only issue wit this is that the implementations can't be left abstract because they are already defined as default methods in a parent interface.

motlin: Are we talking about in the interface (FixedSizeCollection) or in the concrete classes that implement it? It's ok to override a default interface method with an abstract interface method. And before we started FixedSizeCollection's methods were abstract, so I know the concrete classes implement these methods. I'm not understanding the problem here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
donraab: Yes

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/MultiReaderHashBag.java
donraab: Fair enough. I like the idea of returning MultiReaderBag though so access to the multi-reader specific API is retained.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/FixedSizeCollection.java
donraab: In the interface.

> It's ok to override a default interface method with an abstract interface method.

My bad. I thought I remembered having an issue here overriding a default method with an abstract method in an interface. Just did some googling and this seems to be ok. If this isn't required then the interface can just be reverted.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/MultiReaderHashBag.java
motlin: I started this change just now in https://github.com/eclipse/eclipse-collections/pull/879

Once this PR lands, I can add the new methods too.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/FixedSizeCollection.java
aboullaite: @donraab @motlin The compiler complains about having an abstract and a default method with matching signatures. The problem happens in `FixedSizeSet` and `FixedSizeList` since both `MutableSet` and `FixedSizeCollection` define  methods `with()`, `without()` ... and one is default and the other is abstract, the compiler requires the subinterface to explicitly declare which one it should inherit by overriding the methods again.

Should I override the methods in sub-interfaces (FixedSizeSet, FixedSizeList) , or remove the the abstract methods from `FixedSizeCollection` and push down javadoc documentation to implementation classes

motlin:
You must override these methods anyway or they won’t have the right return
types. For example, FixedSizeSet.with() should be declared to return
FixedSizeSet, and it should be left abstract in the interface.

aboullaite:
@motlin I believe the signature should return `MutableSet`
```@Override
    MutableSet<T> without(T element);
```
Since many implementation classes override it expecting `MutableSet` (`DoubletonSet` for example)

motlin:
My mistake, you're right. A regular abstract declaration of those methods should work though, with the corrected (otherwise redundant) return type.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/set/FixedSizeSet.java
```diff
@@ -19,6 +19,18 @@
...
+    MutableSet<T> with(T element);
+
+    @Override
+    MutableSet<T> without(T element);
+
+    @Override
+    MutableSet<T> withAll(Iterable<? extends T> elements);
+
+    @Override
+    MutableSet<T> withoutAll(Iterable<? extends T> elements);
```
motlin, line 32: I'm wondering, can without and withoutAll return FixedSize*?

aboullaite: Nope, same as the others. for example `AbstractMemoryEfficientMutableList` that implements `FixedSizeList` return MutableList
```
    @Override
    public MutableList<T> withoutAll(Iterable<? extends T> elements)
    {
        if (Iterate.isEmpty(elements))
        {
            return this;
        }
        return Lists.fixedSize.ofAll(this.toList().withoutAll(elements));
    }
```

motlin: ???

This compiles fine for me...

```java
@Override
public FixedSizeList<T> withoutAll(Iterable<? extends T> elements)
{
    if (Iterate.isEmpty(elements))
    {
        return this;
    }
    return Lists.fixedSize.ofAll(this.toList().withoutAll(elements));
}
```

motlin: Anyway, this change can't be made in a minor version, so I think we're ready to land this change as-is.

aboullaite: Yes, in that cyaee, the same changes needs to be made in every class that implements `FixedSize*`! Is this something of interest? I can introduce all the discussions captured here in another PR

donraab: @aboullaite Don't make those changes yet.  Leave them as is for now.  We can fix this during a major release.

donraab: @aboullaite Yes you can do this in a separate PR, but it will have to wait for a major release.

donraab:
Hey @aboullaite @motlin, is this ready for a final review and merge?

aboullaite:
AFAIK no required changes needed.

donraab:
Thanks @aboullaite, please rebase on master and I will give a final review and merge the PR if there are no changes necessary to be included in a minor release.

aboullaite:
@donraab pushed 🙏

PR #863, opened 2020-04-14
Pull up ListIterable.binarySearch() as default methods.

motlin:


eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/ListIterable.java
```diff
@@ -279,7 +279,10 @@
...
      * @see Collections#binarySearch(List, Object, Comparator)
      */
-    int binarySearch(T key, Comparator<? super T> comparator);
+    default int binarySearch(T key, Comparator<? super T> comparator)
+    {
+        return Collections.binarySearch((List<? extends T>) this, key, comparator);
+    }
 
     /**
      * Searches for the specified object using the binary search algorithm. The elements in this list must implement
```
motlin, line 288: Note that this overload `binarySearch(T key)` already casts `this` and is already default. Not sure why we only did one of the two methods, but I'm guessing it had to do with complex generics.

donraab: Interesting.  Good catch.

motlin:
@nikhilnanivadekar the TeamCity builds are now consistently failing with: `File /mnt/agent/work/1a6b949da01cf326/test-coverage/test-coverage.iml does not exist`

This is failing on master, unrelated to this PR. What are our next steps here?

nikhilnanivadekar:
Can you try reverting the configuration to what it was before? IntellIj Project style instead of Pom. I think that’s the only thing we changed.

PR #864, opened 2020-04-14
Pull up OrderedIterable.toStack() as a default method.

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/UnmodifiableMutableList.java
```diff
@@ -210,12 +208,6 @@ public void forEachWithIndex(int fromIndex, int toIndex, ObjectIntProcedure<? su
         throw new UnsupportedOperationException("Cannot call shuffleThis() on " + this.getClass().getSimpleName());
     }
 
-    @Override
```
donraab, line 213: This implementation is subtly different than what you have pulled up.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMap.java
```diff
@@ -1492,12 +1491,6 @@ public void forEachWithIndex(
         this.delegate.forEachWithIndex(fromIndex, toIndex, objectIntProcedure);
     }
 
-    @Override
```
donraab, line 1495: This is different than what you have pulled up into the default method.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/mutable/UnmodifiableSortedSet.java
```diff
@@ -452,12 +451,6 @@ public int compareTo(SortedSetIterable<T> o)
         return this.getSortedSet().compareTo(o);
     }
 
-    @Override
```
donraab, line 455: This is different than what you have pulled up to the default method.

eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/immutable/ImmutableArrayStack.java
```diff
@@ -292,12 +290,6 @@ public boolean containsAllArguments(Object... elements)
         return this.delegate.asReversed().containsAllArguments(elements);
     }
```
donraab, line 291: This is different than what you have pulled up to the default method.

eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/mutable/UnmodifiableStack.java
```diff
@@ -586,12 +586,6 @@ public double injectInto(double injectedValue, DoubleObjectToDoubleFunction<? su
         return this.mutableStack.toSortedSet(comparator);
     }
 
-    @Override
```
donraab, line 589: This is different than what you have pulled up to the default method.

eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/mutable/ArrayStack.java
```diff
@@ -935,12 +935,6 @@ public void forEachWithIndex(ObjectIntProcedure<? super T> objectIntProcedure)
         return this.delegate.asReversed().toSortedSet(comparator);
     }
 
-    @Override
```
donraab, line 938: This is different than what you have pulled up to the default method.

eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/immutable/ImmutableNotEmptyStack.java
```diff
@@ -190,12 +188,6 @@ public T getLast()
         throw new UnsupportedOperationException("Cannot call getLast() on " + this.getClass().getSimpleName());
     }
 
-    @Override
```
donraab, line 193: This is different than what you have pulled up to the default method.

motlin:
@donraab you correctly pointed out that I didn't replace all method bodies with identical implementations. Is that a problem though? Just an observation?

I don't think there are any observable behavior changes. Please correct me if I'm wrong. For example, I was careful not to delete the overrides in synchronized wrappers, so that the lock is grabbed once.

donraab:
@motlin In bulk refactorings such as this there should be no observable behavior changes. A separate refactoring can be done to review and possibly clean up individual cases where it requires a behavior change to be reasoned about.

motlin:
@donraab I agree, but I'm also claiming that there are no observable behavior differences despite the change in text. I also don't mind splitting up major/minor changes, I'm just not sure why this would require a major version.

donraab:
Yes, I understand. I'm not saying this requires a major version. I'm just saying it's easier to review and approve if the refactoring is purely structural. If you change the logic in a few methods at the same time as pulling up and removing duplication, then we have to reason about each and every method which can be hard. There could be performance differences or hard to spot bugs.

motlin:
I narrowed this PR. Now, the PR only contains changes in the api/ module. I kept the other changes in a separate commit so I can further split it and land more of it.

PR #865, opened 2020-04-14
Pull up implementations of aggregateBy() as default methods.

motlin:


eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutablePrimitiveObjectEmptyMap.stg
```diff
@@ -854,12 +854,6 @@ final class Immutable<name>ObjectEmptyMap\<V> implements Immutable<name>ObjectMa
         return Lists.immutable.of();
     }
 
-    @Override
```
donraab, line 857: Why remove this?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
```diff
@@ -657,18 +655,6 @@ public void forEachValue(Procedure<? super V> procedure)
         return new HashBiMap<>(this.delegate.groupByUniqueKey(function));
     }
 
-    @Override
```
donraab, line 660: Similar to the other pull request that was pulling up default methods, I would prefer to remove only purely structural code.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnmodifiableMutableMap.java
```diff
@@ -1148,22 +1148,4 @@ public MutableShortBag collectShort(ShortFunction<? super V> shortFunction)
     {
         return this.getMutableMap().zipWithIndex();
     }
-
```
donraab, line 1151: Similar to the comment above. This is the more standard pattern for implementing unmodifiable behavior.

eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/mutable/UnmodifiableStack.java
```diff
@@ -951,24 +951,6 @@ public int hashCode()
         return this.mutableStack.hashCode();
     }
 
-    @Override
```
donraab, line 954: Same as the other unmodifiable comment above.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnmodifiableMutableMap.java
motlin: Catching up on your other comments, I take the point about it being more difficult to perform the code review when I remove methods, and it requires reasoning about whether that's ok. I have to point out that this is not that much of a standard pattern for implementing unmodifiable behavior. I just tried going into UnmodifiableMutableMap and pressing Cmd+I (to implement, not override) and I see many methods implemented that are otherwise not implemented using this pattern.

* flatCollectWith
* flatCollectWith (target)
* countBy
* countBy(target)
* countByWith
* countByWith(target)
* countByEach
* countByEach(target)
* forEach(Procedure)
* forEach(Consumer)
* getAny
* minOptional()
* maxOptional()
* minOptional(Comparator)
* maxOptional(Comparator)
* minByOptional
* maxByOptional
* summarizeInt
* summarizeFloat
* summarizeLong
* summarizeDouble
* reduceInPlace(Collector)
* reduceInPlace(Supplier, BiConsumer)
* reduce
* groupByAndCollect

donraab: Looking at this list of methods, most of them were probably added as default methods, which we started doing after the 8.0 release. I saw a few since 9.0 and 10.0 marks in the JavaDoc. It's an interesting question whether their behavior should be overridden in `UnmodifiableMutableMap`. If you pull this method up, then you should probably pull all of the API's up. If we agree this makes sense, I would rather this be done as a separate PR for `UnmodifiableMutableMap`.

All of these older methods are implemented directly in `UnmodifiableMutableMap`, just as `aggregateBy` is today.

```
    public <R> MutableBagMultimap<R, V> groupBy(Function<? super V, ? extends R> function)
    public <R, C extends MutableMultimap<R, V>> C groupBy(Function<? super V, ? extends R> function, C target)
    public <R> MutableBagMultimap<R, V> groupByEach(Function<? super V, ? extends Iterable<R>> function)
    public <R, C extends MutableMultimap<R, V>> C groupByEach(Function<? super V, ? extends Iterable<R>> function, C target)
    public <VV> MutableMap<VV, V> groupByUniqueKey(Function<? super V, ? extends VV> function)
    public <VV, R extends MutableMapIterable<VV, V>> R groupByUniqueKey(Function<? super V, ? extends VV> function, R target)
```

motlin: Yes, I plan on pulling all those up and more. I started with a few "easy" methods to see how it goes.

donraab: Ok, I think I can get comfortable with the unmodifiable moves here.

motlin:
I narrow this. Now, the PR only contains changes in the api/ module. I kept the other changes in a separate commit so I can further split it and land more of it.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/collection/ImmutableCollection.java
```diff
@@ -210,16 +210,35 @@
...
             Function<? super T, ? extends K> groupBy,
             Function0<? extends V> zeroValueFactory,
-            Function2<? super V, ? super T, ? extends V> nonMutatingAggregator);
+            Function2<? super V, ? super T, ? extends V> nonMutatingAggregator)
+    {
+        MutableMap<K, V> map = Maps.mutable.empty();
+        this.forEach(each ->
+        {
+            K key = groupBy.valueOf(each);
+            map.updateValueWith(key, zeroValueFactory, nonMutatingAggregator, each);
```
donraab, line 238: Why is this implementation different than the one in MutableCollection right below?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMap.java
```diff
@@ -184,16 +184,35 @@
     }
 
     @Override
-    <K2, V2> MutableMap<K2, V2> aggregateInPlaceBy(
-            Function<? super V, ? extends K2> groupBy,
```
donraab, line 188: I think this might be able to be deleted as it inherits from `MutableMapIterable` which has the same implementation.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMap.java
```diff
@@ -184,16 +184,35 @@
...
+        });
+        return map;
+    }
 
     @Override
-    <K2, V2> MutableMap<K2, V2> aggregateBy(
-            Function<? super V, ? extends K2> groupBy,
-            Function0<? extends V2> zeroValueFactory,
-            Function2<? super V2, ? super V, ? extends V2> nonMutatingAggregator);
+    default <KK, VV> MutableMap<KK, VV> aggregateBy(
```
donraab, line 203: Same as above.

Issue #726, opened 2019-05-29
Distinct element view of Bag 


victornoel:
It would be great to have a lazy view on Bags that contains the distinct elements of the Bag. It could be called `distinct` or `distinctView`.

Currently, the only way to work on the distinct elements of a Bag is to use `forEachWithOccurrences`.

An alternative could be a view which would contain `ObjectIntPair` with each distinct element and their number of occurrence.

nikhilnanivadekar:
@victornoel this is a good idea. I think `distinctView()` is a better name. 
Curious about your use case: are you trying to manipulate the elements? If you just need a view, we might want to add asSet() and just return the keySet() of the backing map. thoughts?

victornoel:
@nikhilnanivadekar I just want to iterate over the elements, or maybe collect something else based on them, so maybe I would do: `bag.distinctView().collect(e -> ...).toList()`.

So yes, it is basically a view that I am interested in: `asSet()` would totally correspond to my need.

nikhilnanivadekar:
would you like to contribute? On the interface it can be default implementation which throws but implementation will have the actual `asSet()`

victornoel:
@nikhilnanivadekar yes I can, not sure I will be able to before 2 weeks though, but I guess now that v10 is released, we have some time :)

Questions:
- the default implementation on the interface is needed for backward compat, right?
- I wonder if we should return a `MutableSet`, a `SetIterable` or a `RichIterable`?
  - If it's a `Set`, should it be an `UnmodifiableSet` like what is returned by primitive maps `keySet` method?
  - some of the implementation are not using a set-based collection inside so it's not going to be views in that case… I was thus thinking more of using `RichIterable` or a `LazyIterable` even for this, in the same way that the maps have a `keyView`.
- The actual implementations of `MutableBag` have internally a method `getKeysView` but it's not a `Set`, would it make sense to make this method be the same as the one I would be adding here?

motlin:
SetIterable is ideal. LazyIterable is easier to implement but limiting.

victornoel:
Hi, I started working on this, and I have some questions:

- To implement `asSet` for `ImmutableHashBag`, the best would be to reuse `ImmutableUnifiedSet` to wrap the `delegate.asSet()` of `ImmutableHashBag`. Of course this is not possible because the constructor of `ImmutableUnifiedSet` is private. What's the best course of action in that case?
[...]

PR #869, opened 2020-04-18
Adds asSet() and distinctView() to Bag for #726

victornoel:
This is a second tentative PR for #726.

It combines things from #797 and #861, in particular introducing `distinctView()` to avoid accessing internal arrays of some of the bags the `Set` adapters.

I still had to:
- introduce two adapters: `ImmutableSetFromBagAdapter` and `ImmutableSortedSetFromBag`
  - those are the main difference with #797 via the use of `distinctView()`
  - I couldn't implement `asParallel` for those so it will use the O(n) implementation for the superclass...
- most of the other `asSet()` implementations are directly implemented  as in #797 and `distinctView()` simply relies on `asSet()`.
- I also had to introduce `indexOf()` in the various `Iterate` classes.


victornoel:
@motlin FYI :)

victornoel:
@motlin have you had the time to take a look at this PR?

victornoel:
@motlin hi, sorry to insist, could you take a look?

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/AbstractMultiReaderSet.java
```diff
@@ -0,0 +1,1004 @@
...
+import org.eclipse.collections.api.tuple.Pair;
+import org.eclipse.collections.impl.collection.mutable.AbstractMultiReaderMutableCollection;
+import org.eclipse.collections.impl.lazy.parallel.set.MultiReaderParallelUnsortedSetIterable;
+import org.eclipse.collections.impl.utility.LazyIterate;
+
+import static org.eclipse.collections.impl.factory.Iterables.mList;
+
+abstract class AbstractMultiReaderSet<T>
+        extends AbstractMultiReaderMutableCollection<T>
+        implements RandomAccess, MultiReaderSet<T>
```
motlin, line 70: RandomAccess? That's a marker interface for a List.

Is this class extracted from the subclass automatically by the IDE?

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/AbstractMultiReaderSet.java
```diff
@@ -0,0 +1,1004 @@
...
+{
+    protected MutableSet<T> delegate;
+
+    protected abstract <K> AbstractMultiReaderSet<K> newSetFrom(MutableSet<K> newDelegate);
+
+    @Override
+    protected MutableSet<T> getDelegate()
+    {
+        return this.delegate;
+    }
```
motlin, line 80: Is it possible to implement without a getDelegate method?

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/AbstractMultiReaderSet.java
```diff
@@ -0,0 +1,1004 @@
...
+        @Override
+        public MutableSet<T> asUnmodifiable()
+        {
+            throw new UnsupportedOperationException("Cannot call asUnmodifiable() on " + this.getClass().getSimpleName());
+        }
+
+        @Override
+        public ImmutableSet<T> toImmutable()
+        {
+            return Sets.immutable.withAll(this.getDelegate());
```
motlin, line 462: These can all just be `delegate` instead of `this.getDelegate()` if we need, right?

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/MultiReaderSetAdapter.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Set;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import org.eclipse.collections.api.set.MultiReaderSet;
+import org.eclipse.collections.api.set.MutableSet;
+
+public final class MultiReaderSetAdapter<T> extends AbstractMultiReaderSet<T>
+{
+    MultiReaderSetAdapter(MutableSet<T> newDelegate, ReadWriteLock newLock)
```
motlin, line 22: package private constructor in public class. Not sure if you want this to be public or package private but the two visibilities should agree

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/SetAdapter.java
```diff
@@ -152,6 +152,16 @@ public boolean containsAll(Collection<?> collection)
...
+        if (obj == this)
+        {
+            return true;
+        }
+
+        if (!(obj instanceof Set))
+        {
+            return false;
+        }
+
```
motlin, line 164: Same here, how did it come up?

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
```diff
@@ -1282,8 +1282,24 @@ public boolean equals(Object object)
...
+            return this.containsAll(s);
+        }
+        catch (ClassCastException ignored)
+        {
+            return false;
+        }
+        catch (NullPointerException ignored)
+        {
+            return false;
+        }
```
motlin, line 1302: Why change this implementation? How did it come up?

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnmodifiableMutableSet.java
```diff
@@ -104,6 +104,16 @@
...
+        {
+            return true;
+        }
+
+        if (!(obj instanceof Set))
+        {
+            return false;
+        }
+
         return this.getMutableSet().equals(obj);
```
motlin, line 117: In wrappers, I see how checking `== this` is a nice optimization. I don't see why we'd check `instanceof Set` though. That's the second thing the delegate will do, after its own `== this` check.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/immutable/ImmutableTreeSet.java
```diff
@@ -167,6 +167,10 @@ public boolean equals(Object obj)
         {
             return false;
         }
+        catch (NullPointerException ignored)
+        {
+            return false;
+        }
```
motlin, line 173: Did this come up in a test? Is it covered?

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ArrayListIterate.java
```diff
@@ -1338,6 +1339,27 @@ private ArrayListIterate()
...
+            {
+                if (Objects.equals(elements[i], object))
+                {
+                    return i;
+                }
+            }
+            return -1;
+        }
+        return RandomAccessListIterate.indexOf(list, object);
+    }
```
motlin, line 1361: Is this an unrelated change?

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/Iterate.java
```diff
@@ -2423,6 +2423,26 @@ public static boolean notEmpty(Iterable<?> iterable)
...
+        if (iterable instanceof List<?>)
+        {
+            return ListIterate.indexOf((List<T>) iterable, object);
+        }
+        if (iterable != null)
+        {
+            return IterableIterate.indexOf(iterable, object);
+        }
+        throw new IllegalArgumentException("Cannot perform indexOf on null");
+    }
```
motlin, line 2444: I think this is to implement the sorted bag adapter. This must mean that the indexOf method is unsupported everywhere, and you could leave your new implementation unsupported as well if you'd like.

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/IterableIterate.java
```diff
@@ -1100,7 +1108,7 @@ public static boolean notEmpty(Iterable<?> iterable)
         }
     }
 
-    public static <T> void forEachWithIndex(List<T> iterable, int from, int to, ObjectIntProcedure<? super T> objectIntProcedure)
+    public static <T> void forEachWithIndex(Iterable<T> iterable, int from, int to, ObjectIntProcedure<? super T> objectIntProcedure)
```
motlin, line 1111: Oof. This is a little thing to push us to a major-version, but it's technically a breaking change. I just want @donraab and @nikhilnanivadekar to see this one.

donraab: Can't we just add a new static method which takes `Iterable`?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableEmptyBag.java
```diff
@@ -783,4 +783,16 @@ private Object writeReplace()
...
+    @Override
+    public ImmutableSet<T> asSet()
+    {
+        return Sets.immutable.empty();
+    }
+
+    @Override
+    public RichIterable<T> distinctView()
+    {
+        return this.asSet();
```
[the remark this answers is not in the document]
victornoel: I didn't think of it, I've just delegated to `asSet()` everytime `asSet()` has a simple implementation, should I use `this` or is it better to just use `Bags.immutable.empty()`?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableSetFromBagAdapter.java
```diff
@@ -0,0 +1,113 @@
...
+
+import org.eclipse.collections.api.bag.Bag;
+import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.set.ImmutableSet;
+import org.eclipse.collections.impl.set.immutable.AbstractImmutableSet;
+
+/**
+ * This class provides a {@link ImmutableSet} wrapper around any Bag interface instance.
+ *
+ * TODO {@link #asParallel(ExecutorService, int)}
```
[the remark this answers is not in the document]
victornoel: Actually that's not trivial, we will need to implement a whole class for this to be possible I think... it was one of the advantages of the first method (in #797) that we didn't have to implement such things. I will take a look at it to see what needs to be done.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableSetFromBagAdapter.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public int hashCode()
+    {
+        int result = 0;
+        for (T each : this.delegate.distinctView())
+        {
+            result += each.hashCode();
+        }
+        return result;
+    }
```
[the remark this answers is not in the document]
victornoel: I think it's because each of the subclass of `AbstractImmutableSet` have their own specific implementation for `equals` and `hashCode`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedSetFromBagAdapter.java
```diff
@@ -0,0 +1,158 @@
...
+    {
+        int result = 0;
+        for (T each : this.delegate.distinctView())
+        {
+            result += each.hashCode();
+        }
+
+        return result;
+    }
+}
```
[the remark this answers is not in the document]
victornoel: same thing as for `ImmutableSetFromBagAdapter`

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/AbstractSynchronizedRichIterable.java
```diff
@@ -105,6 +105,11 @@ protected Object getLock()
     @Override
     public boolean equals(Object obj)
     {
+        if (obj == this || obj == this.delegate)
+        {
+            return true;
+        }
```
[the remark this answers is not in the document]
victornoel: that's a good question, this is an optimization over the fact this class is applying synchronization, so it's a special case, I suppose nobody thought about it ^^

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/AbstractMultiReaderSet.java
victornoel: It was already there before, it got here because it was originally in `MultiReaderUnifiedSet` before I extracted this abstract class. Should I remove it? It does seems strange...

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/AbstractMultiReaderSet.java
victornoel: yes, I can remove it and just use `this.delegate`, should I?

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/AbstractMultiReaderSet.java
victornoel: yes

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/MultiReaderSetAdapter.java
victornoel: I did that to be similar to how `SetAdapter` was defined: the thing is the class must be public so that `MultiReaderSetAdapter.adapt()` can be called from anywhere. Should I make the constructor private maybe? Maybe in the case of SetAdapter, it is package private so that it is accessible for tests, but apart from that I don't know.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
victornoel: I homogenized it with the others implementation of `Set` where this happens, for example in `java.util.AbstractSet`. I think it is necessary in order to be coherent with the public contract of `java.util.Collection.containsAll(Collection<?>)` which can throw this exception.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnmodifiableMutableSet.java
victornoel: yes, you are right, it is a useless optimization here and in the other case where there is a delegate!

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/immutable/ImmutableTreeSet.java
victornoel: see answer above

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/ArrayListIterate.java
victornoel: since I introduced `Iterate.indexOf` needed by `ImmutableSortedSetFromBagAdapter`, I decided to be coherent and add it for the various iterate classes.

victornoel: note it is in a separate commit

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/Iterate.java
victornoel: while I was at it I thought it would be nice to do it :) should I remove it?

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/IterableIterate.java
victornoel: it is called only once in the newly introduced `ImmutableSortedSetFromBagAdapter` which has an object of type `RichIterable` only. But note that we are in a class inside an `internal` package, shouldn't that be alright? Also I thought that it would be non-breaking because `Iterable` is a supertype of `List`, but maybe I'm missing something about Java type compatibility in bytecode...

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnmodifiableMutableSet.java
victornoel: well except for synchronized implementation, that would avoid going through a lock

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/AbstractMultiReaderSet.java
victornoel: actually the `getDelegate()` method is an abstract method of the `AbstractMultiReaderMutableCollection` superclass, so it is needed.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/AbstractSynchronizedRichIterable.java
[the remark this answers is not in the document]
victornoel: same for checking `obj` is an instance of `Set`, this avoids going through locking if it's not a `Set`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableSetFromBagAdapter.java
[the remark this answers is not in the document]
victornoel: we can always starts by implementing it using`NonParallelUnsortedSetIterable` maybe as it is the case in `AbstractImmutableSortedSet`?

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/IterableIterate.java
donraab: I would be conservative and make this kind of change in a major version. In the near term I would inline the code the code in the one place it is used or add another method on IterableIterate. @motlin what do you think?

donraab: HI @victornoel I'd like to get this unblocked, as you and @motlin were making good progress through this review. Let's just add the code in the one place it is needed for now and we can update `IterableIterate` to take `Iterable` when we do a major release. Ok?

victornoel: @donraab sorry for the delay, so I was thinking: what about simply duplicating the method with `Iterable` as a parameter, I think that the overloading will only be taken into account at compile time, so it shouldn't have any impact on current uses of the old method. We can also deprecate the latest so that in the next major version with remove it?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableSetFromBagAdapter.java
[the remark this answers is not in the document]
victornoel: @motlin @donraab here I went with having `asParallel` implemented using `NonParallelUnsortedSetIterable` in `AbstractImmutableSet`, similar to how it was done in `AbstractImmutableSortedSet`. Then we can create another issue to solve this problem for all the subclasses of those two maybe? This seems orthogonal to the current PR.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/AbstractMultiReaderSet.java
victornoel: @motlin @donraab I left the `implements RandomAccess` that was already in `MultiReaderUnifiedSet` for a reason I ignore and we can create another issue to solve this mystery? I think it's orthogonal to this PR

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/AbstractMultiReaderSet.java
victornoel: @motlin @donraab I fixed this in the latest push

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnmodifiableMutableSet.java
victornoel: @motlin @donraab I fixed that in the latest push

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/IterableIterate.java
victornoel: @motlin @donraab I went with this solution for now, but I can change it for something more ad-hoc if it's not correct :)

victornoel:
@motlin @donraab I've pushed fixed for the raised issues. Please see my answers to all the comments above for details :)

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableSetFromBagAdapter.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public void each(Procedure<? super T> procedure)
+    {
+        this.delegate.distinctView().each(procedure);
+    }
+
+    @Override
+    public Iterator<T> iterator()
+    {
+        return this.delegate.distinctView().iterator();
```
motlin, line 60: Is there a test to make sure that this iterator's remove() method throws?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableSetFromBagAdapter.java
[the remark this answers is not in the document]
motlin: Right, it seems like the time to pull them up now.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/AbstractHashBag.java
```diff
@@ -327,6 +329,18 @@ public boolean contains(Object o)
         return this.groupByEach(function, HashBagMultimap.newMultimap());
     }
 
+    @Override
+    public MutableSet<T> asSet()
+    {
+        return UnmodifiableMutableSet.of(this.items.keySet());
```
motlin, line 335: Since this will throw, MutableSet is the wrong type IMO. Why not leave the type non-overridden?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/UnmodifiableBag.java
```diff
@@ -438,4 +440,16 @@ protected Object readResolve()
     {
         return this.getMutableBag().selectUnique();
     }
+
+    @Override
+    public MutableSet<T> asSet()
+    {
+        return UnmodifiableMutableSet.of(this.getMutableBag().asSet());
```
motlin, line 447: The current implementation will double wrap (two unmodifiable wrappers), no?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedSetFromBagAdapter.java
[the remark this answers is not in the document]
motlin: Right, but the implementations are identical, aren't they? Pulling them up would reduce duplicated code from the current review?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedSetFromBagAdapter.java
```diff
@@ -0,0 +1,155 @@
...
+        {
+            return this.containsAll(otherSet);
+        }
+        catch (ClassCastException ignored)
+        {
+            return false;
+        }
+        catch (NullPointerException ignored)
+        {
+            return false;
```
motlin, line 140: Are these `return false` lines covered by tests?

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/AbstractSynchronizedRichIterable.java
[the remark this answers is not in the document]
motlin: Should we remove it then? It seems unrelated to the task at hand.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/AbstractImmutableSet.java
```diff
@@ -472,6 +473,14 @@ public boolean isProperSubsetOf(SetIterable<? extends T> candidateSuperset)
...
-        return this.toSet().asParallel(executorService, batchSize);
+        if (executorService == null)
+        {
+            throw new NullPointerException();
+        }
+        if (batchSize < 1)
+        {
+            throw new IllegalArgumentException();
+        }
+        return new NonParallelUnsortedSetIterable<>(this);
```
motlin, line 484: What's the reason for this change? It's non-parallel before and after?

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/AbstractImmutableSet.java
```diff
@@ -472,6 +473,14 @@ public boolean isProperSubsetOf(SetIterable<? extends T> candidateSuperset)
     @Override
     public ParallelUnsortedSetIterable<T> asParallel(ExecutorService executorService, int batchSize)
     {
-        return this.toSet().asParallel(executorService, batchSize);
+        if (executorService == null)
+        {
+            throw new NullPointerException();
+        }
```
motlin, line 479: This is just Objects.requireNonNull(executorService)

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/AbstractImmutableSet.java
```diff
@@ -472,6 +473,14 @@ public boolean isProperSubsetOf(SetIterable<? extends T> candidateSuperset)
...
+        if (batchSize < 1)
+        {
+            throw new IllegalArgumentException();
```
motlin, line 482: The IllegalArgumentException should have some context, like the batchSize passed in.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/MultiReaderSetAdapter.java
motlin: It's to force the use of the adapt() method which has instanceof checks or something like that. Seems fine.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
motlin: Are the `return false` lines covered? If not, I think we should revert this change. It's unrelated and in one of the most heavily used classes.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/UnmodifiableBag.java
victornoel: @motlin yes maybe, but I thought it was safer like that. Maybe `UnmodifiableMutableSet.of` should check if it's not already an `UnmodifiableMutableSet`?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/AbstractHashBag.java
victornoel: @motlin just for coherency, even if Unmodifiable, it's still a `MutableSet` in terms of typing... but I can leave `SetIterable` here yes. I will let you decide

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableSetFromBagAdapter.java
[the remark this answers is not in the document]
victornoel: @motlin yes but each of the subclass of `AbstractImmutableSet` have different implementations, so we can't just move them up I think.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedSetFromBagAdapter.java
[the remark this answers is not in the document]
victornoel: @motlin No they are not identical, some of them are different

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/AbstractSynchronizedRichIterable.java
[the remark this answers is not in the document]
victornoel: @motlin ok

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
victornoel: @motlin ok

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedSetFromBagAdapter.java
victornoel: @motlin I will revert back to an implementation similar to the rest

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/AbstractImmutableSet.java
victornoel: @motlin before it was in O(n), now it's O(1) because before it was calling toSet.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/immutable/ImmutableSetFromBagAdapter.java
victornoel: @motlin good point, I added tests for the new classes (by extending the set test classes), it enabled me to find a few bugs too :)

victornoel:
@motlin see my latest push, I removed everything not related to the original ticket and took care of most of your comment except some where I answered you with questions :)

victornoel:
@motlin can you take a look? thanks :)

nikhilnanivadekar:
@motlin pinging here, can you please take a look? @victornoel is waiting on your feedback. Thanks!

donraab:
@motlin Looks like this has gone quiet, can you please take a look? I think @victornoel is still waiting on your feedback. Thanks!

@victornoel Apologies, this needs to be brought up to date and some issues resolved.

motlin:
I apologize, I've had very little time. My main feedback is the same as earlier. I've always been uncomfortable with the mutable views. I'll give some examples to make it more clear.

What does `mutableBag.asSet().remove()` do? If it's an unmodifiable view, it would throw of course. Otherwise, I think it would remove all occurrences and return true.

What about `mutableBag.asSet().add(something)`? I think if there are no occurrences in the bag, it would add one and return true. If there are any occurrences in the bag, it would leave it alone and return false.

So removing and adding the same element from this set would not return the bag to its original state. It would reduce the number of occurrences to 1. Is this ok? Is it the obviously correct behavior, and I'm making a big deal over nothing?

I don't want to add mutable, unmodifiable views and leave them unmodifiable forever. I'd want to make sure we have a path forward for making them mutable. If that's unlikely to ever happen, we should keep read-only types.

victornoel:
Closing this as this will most certainly never be merged and based on very outdated code

PR #877, opened 2020-04-30
Fix inspections violations.

motlin:
Merely refreshing the TeamCity settings caused it to either start working again, or use newer versions of inspections, or something like that. A whole bunch of real violations started showing up against master.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
@@ -77,6 +77,7 @@ import org.eclipse.collections.impl.map.mutable.UnifiedMap;
  * AbstractMultiReaderMutableCollection is a common abstraction that provides thread-safe collection behaviors.
  * Subclasses of this class must provide implementations of getDelegate() and getLock().
  */
+@SuppressWarnings({"unused", "TransientFieldInNonSerializableClass"})
 public abstract class AbstractMultiReaderMutableCollection<T> implements MutableCollection<T>
 {
     protected transient ReadWriteLock lock;
@@ -1981,13 +1982,11 @@ public abstract class AbstractMultiReaderMutableCollection<T> implements Mutable
 
     public static class ReadWriteLockWrapper
     {
-        private final ReadWriteLock readWriteLock;
         private final LockWrapper readLock;
         private final LockWrapper writeLock;
 
         public ReadWriteLockWrapper(ReadWriteLock readWriteLock)
         {
-            this.readWriteLock = readWriteLock;
             this.readLock = new LockWrapper(readWriteLock.readLock());
             this.writeLock = new LockWrapper(readWriteLock.writeLock());
         }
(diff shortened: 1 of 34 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
```diff
@@ -77,10 +77,11 @@
  * AbstractMultiReaderMutableCollection is a common abstraction that provides thread-safe collection behaviors.
  * Subclasses of this class must provide implementations of getDelegate() and getLock().
  */
+@SuppressWarnings("unused")
 public abstract class AbstractMultiReaderMutableCollection<T> implements MutableCollection<T>
 {
-    protected transient ReadWriteLock lock;
-    protected transient ReadWriteLockWrapper lockWrapper;
+    protected ReadWriteLock lock;
```
donraab, line 83: Why is it ok to remove transient here?

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
```diff
@@ -1981,13 +1982,11 @@ public void appendString(Appendable appendable, String start, String separator,
 
     public static class ReadWriteLockWrapper
     {
-        private final ReadWriteLock readWriteLock;
```
donraab, line 1984: Is readWriteLock is not used?

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
motlin: `transient` only affects `Serializable` classes. `AbstractMultiReaderMutableCollection` and its subclasses are not `Serializable`. Its subclasses are `Externalizable`

In theory, someone could write a new serializable subclass of AbstractMultiReaderMutableCollection. If you would prefer, I could bring back the keyword and suppress the warning.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
motlin: Correct, the field is not used. The parameter is still used in the constructor.

```java
    public static class ReadWriteLockWrapper
    {
        private final ReadWriteLock readWriteLock;
        private final LockWrapper readLock;
        private final LockWrapper writeLock;

        public ReadWriteLockWrapper(ReadWriteLock readWriteLock)
        {
            this.readWriteLock = readWriteLock;
            this.readLock = new LockWrapper(readWriteLock.readLock());
            this.writeLock = new LockWrapper(readWriteLock.writeLock());
        }
...
```

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
donraab: Yes, bring back the keyword and suppress the warning.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/mutable/AbstractMultiReaderMutableCollection.java
donraab: Ok

motlin:
@donraab done

PR #879, opened 2020-05-04
Add covariant return types to methods in MultiReaderList that return `this`.

motlin:


eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MultiReaderList.java
```diff
@@ -23,4 +36,65 @@
...
+    @Override
+    MultiReaderList<T> sortThisByLong(LongFunction<? super T> function);
+
+    @Override
+    MultiReaderList<T> sortThisByDouble(DoubleFunction<? super T> function);
+
+    @Override
+    default MultiReaderList<T> reverseThis()
+    {
+        Collections.reverse(this);
```
donraab, line 83: This needs to be protected by a call to `withWriteLockAndDelegate`.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MultiReaderList.java
```diff
@@ -23,4 +36,65 @@
...
+        return this;
+    }
+
+    @Override
+    default MultiReaderList<T> shuffleThis()
+    {
+        Collections.shuffle(this);
```
donraab, line 90: Same as above.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MultiReaderList.java
```diff
@@ -23,4 +36,65 @@
...
+        return this;
+    }
+
+    @Override
+    default MultiReaderList<T> shuffleThis(Random random)
+    {
+        Collections.shuffle(this, random);
```
donraab, line 97: Same as above.

mohrezaei:
As a more general rule to the "this needs to be protected by XYZ lock", I would suggest any delegating collection we have ought to always call the appropriate method on the delegate, instead of an external implementation, such as `Collections`. Said differently, not seeing the delegation should be a code smell.

motlin:
Sorry, this was meant to be a WIP PR. It didn't even compile.

There are no default methods now. This is a case where it's important for the overridden interface methods to be abstract.

We can reduce duplication more by pulling up default method bodies that delegate to withReadLockAndDelegate() and withWriteLockAndDelegate().

PR #881, opened 2020-05-07
Added to primitive iterables toArray() method which takes an array parameter

vmzakharov:
If the collection fits into the provided array it is used to store its elements and is returned from the method, otherwise a new array of the appropriate size is allocated and returned. If the iterable is empty, the target array is returned unchanged.

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
@@ -64,6 +64,16 @@ public interface <name>Iterable extends PrimitiveIterable
     <type>[] toArray();
 
     /**
+     * Converts the <name>Iterable to a primitive <type> array. If the collection fits into the provided array it is used
+     * to store its elements and is returned from the method, otherwise a new array of the appropriate size is allocated
+     * and returned. If the iterable is empty, the target array is returned unchanged.
+     */
+    default <type>[] toArray(<type>[] target)
+    {
+        return this.toList().toArray(target);
+    }
+
+    /**
      * Returns true if the value is contained in the <name>Iterable, and false if it is not.
      */
     boolean contains(<type> value);
--- a/eclipse-collections-code-generator/src/main/resources/impl/bag/mutable/primitiveHashBag.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/bag/mutable/primitiveHashBag.stg
@@ -746,6 +746,28 @@ public class <name>HashBag
     }
 
     @Override
+    public <type>[] toArray(<type>[] array)
+    {
+        if (array.length \< this.size())
+        {
+            array = new <type>[this.size()];
+        }
+
+        int[] index = {0};
+
+        <type>[] finalBypass = array;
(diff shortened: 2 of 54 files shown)


eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
```diff
@@ -63,6 +63,13 @@ public interface <name>Iterable extends PrimitiveIterable
      */
     <type>[] toArray();
 
+    /**
+     * Converts the <name>Iterable to a primitive <type> array. If the collection fits into the provided array it is used
+     * to store its elements and is returned from the method, otherwise a new array of the appropriate size is allocated
+     * and returned. If the iterable is empty, the target array is returned unchanged.
+     */
+    <type>[] toArray(<type>[] target);
```
donraab, line 71: If you implement this as a default method, then we can release in a minor version.

donraab:
Suggested one change which will make this PR able to be merged for a minor release.

eclipse-collections-code-generator/src/main/resources/impl/bag/mutable/primitiveHashBag.stg
```diff
@@ -745,6 +745,28 @@ public class <name>HashBag
...
+
+        <type>[] finalBypass = array;
+        this.forEachWithOccurrences((<type> each, int occurrences) ->
+        {
+            for (int i = 0; i \< occurrences; i++)
+            {
+                finalBypass[index[0]] = each;
+                index[0]++;
+            }
+        });
```
motlin, line 766: Nitpick: I don't see much advantage of using forEachWithOccurrences when the output is of size `size()`. Might as well delegate to forEachWithIndex or forEach?

eclipse-collections-code-generator/src/main/resources/impl/list/immutable/immutablePrimitiveArrayList.stg
```diff
@@ -92,6 +92,12 @@ final class Immutable<name>ArrayList
         return new Immutable<name>ArrayList(newArray);
     }
 
+    public static Immutable<name>ArrayList newListFrom(<name>List itemList)
+    {
+        <type>[] newArray = itemList.toArray(new <type>[itemList.size()]);
+        return new Immutable<name>ArrayList(newArray);
+    }
```
motlin, line 99: Is this an unrelated change that snuck in?

donraab: It's actually the reason for the change I believe.

donraab: This could be clarified in the commit message or be a separate commit to split the use case from the change.

vmzakharov: @motlin - a related change that snuck in. I will take it out from this PR.  @donraab is correct, there will be a separate change to take advantage of the toArray(target) in factory methods.

PR #885, opened 2020-05-16
Fixes #793: optimizes removeIf() implementation on BooleanArrayList

### LINKED ISSUE #793: Optimize removeIf() on BooleanArrayList.
ArrayList.removeIf() is O(n).
FastList.removeIf() is O(n^2). Same for primitive lists like BooleanArrayList.

vmzakharov:
Made removeIf() on BooleanArrayList follow the same implementation approach as on the other primitive lists and FastList

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/primitive/BooleanArrayList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/primitive/BooleanArrayList.java
@@ -351,18 +351,29 @@ public final class BooleanArrayList
     @Override
     public boolean removeIf(BooleanPredicate predicate)
     {
-        boolean changed = false;
+        int currentFilledIndex = 0;
         for (int i = 0; i < this.size; i++)
         {
             boolean item = this.items.get(i);
-            if (predicate.accept(item))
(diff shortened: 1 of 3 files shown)


mohrezaei:
I just realized that for stateless predicates, this method will return the number of `true` occurrences or `false` occurrences as a list of all `true`s or `false`s. E.g. if you start with a list that looks like `101101101`, the result will either be `111111` or `000`. There are much faster ways of counting bits, e.g. `BitSet.cardinality`. 

The wipe of the end seems necessary because BitSet doesn't seem to have a proper size/remove method if we ever want to rely on methods like `BitSet.cardinality`.

However, there is one thing about other primitive lists that is unoptimized: unlike arrays of Objects, primitives don't hold any extra memory, so there is no reason to wipe the end clear.

vmzakharov:
@mohrezaei no objections to the change though in the context of this issue?

mohrezaei:
No objection to the code as is. Further work that could be "good first issue" material:

1) Add fast boolean methods to `BooleanArrayList`: `cardinality` and `flip`.
2) Remove `wipeAndResetTheEnd` from primitive lists (not `BooleanArrayList`) after ensuring we never assume the unused end is all zeros.

PR #889, opened 2020-05-19
Closes #888: BooleanArrayStack is code generated from the common primitive stack template

### LINKED ISSUE #888: Code generate BooleanArrayStack from the common template
Currently `BooleanArrayStack` is the only primitive stack implementation that is not generated. Generating `BooleanArrayStack` from `primitveArrayStack.stg` will reduce the amount of duplicate code needed to be maintained in the framework. 

Unlike `BooleanArrayList`, which uses an underlying implementation different from other primitive lists (`BitSet` based rather than an array based), `BooleanArrayStack` implementation appears to be consistent with other primitive stacks. The only difference, inherited from the `BooleanIterable` API, is the absence of the `toSortedList()` method, which will be easy to exclude from the primitive array stack template just for the Boolean type. This approach to include/exclude certain behaviors is already implemented in other templates.

vmzakharov:
`BooleanArrayStack` is generated from `primitveArrayStack.stg`, which reduces the amount of duplicate code needed to be maintained in the framework. `BooleanArrayStack` implementation is consistent with other primitive stacks, the only difference (inherited from the `BooleanIterable` API) being the absence of the `toSortedList()` method. This change also makes `hashCode()` implementation on primitive stacks consistent with the one that was originally in `BooleanArrayStack` using a for-loop rather than an iterator.


eclipse-collections-code-generator/src/main/resources/impl/stack/abstractPrimitiveStack.stg
```diff
@@ -256,11 +256,9 @@ public abstract class Abstract<name>Stack implements <name>Stack
     public int hashCode()
     {
         int hashCode = 1;
-        <name>Iterable iterable = this.getDelegate().asReversed();
```
donraab, line 259: I think we should leave the current implementation in AbstractPrimitiveStack and add the new implementation override in the concrete ArrayStack implementations.

eclipse-collections-code-generator/src/main/resources/impl/stack/mutable/primitiveArrayStack.stg
```diff
@@ -151,12 +151,7 @@ public class <name>ArrayStack extends Abstract<name>Stack
         this.delegate.clear();
     }
 
-    @Override
```
donraab, line 154: This looks like an unrelated change. Please revert and create an issue if this is duplication.

vmzakharov: It is related. Please see the PR comment and the issue comment. BooleanIterable does not have toSortedList() so to have a single template for primitive stacks it needs to be excluded from code generation for boolean (just like sortThis() is already excluded from mutable primitive list generation for booleans).

eclipse-collections-code-generator/src/main/resources/impl/stack/abstractPrimitiveStack.stg
vmzakharov: Happy to make the change, just want to understand the reasoning behind it.  In `AbstactPrimitveStack` `getDelegate()` is defined as returning a `PrimitveArrayList` so the cost of random access of the underlying implementation should not be an issue. Or is it something else?

donraab: If someone extends the Abstract class, the implementation may not be RandomAccess.

eclipse-collections-code-generator/src/main/resources/impl/stack/mutable/primitiveArrayStack.stg
nikhilnanivadekar: Agree with you Vlad, however, such “baked in” changes make it difficult to write release notes. This is what I propose, break it up as an additional commit as a part of this same PR so that it reads as another commit.

donraab: I see, thanks for clarifying.

vmzakharov: Wouldn't describe it as a “baked in” change. The actual change is to have BooleanArrayStack generated and this is necessary to make it happen. The template change preserves APIs for the other primitive stacks and for the boolean array stack, so there is really no change in behavior. BooleanArrayStack never supported toSortedList, toSortedList is a part of the common template, so we exclude toSortedList generation for booleans in the template

nikhilnanivadekar:
Ok makes sense

PR #891, opened 2020-05-20
Closes #890: Implements Null Safe Comparators by Function

### LINKED ISSUE #890: Implement Null Safe Comparator By Function
Consider an object
```
public class Person
{
  private int age;
  private String firstName;
}

```
We have Comparators and Functions which are null-safe on their own, we need a comparator which brings null safety to the computed value.

Failing Test
```

MutableList<Person> people = Lists.mutable.empty();
Person p1 = new Person();
p1.setFirstName("A");
Person p2 = new Person();
p2.setFirstName("B");

people.sortThis(Comparators.byFunction(Person::getAge)); // Throws an NPE

We need something like
people.sortThis(Comparators.nullSafeValue(Person::getAge)) // the value returned after evaluating getAge() should be checked
```

vmzakharov:
Added two factory methods for byFunction comparators that handle null values. They use null safe comparators from java.util.Comparator and follow a similar naming convention.

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/block/factory/ComparatorsTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/block/factory/ComparatorsTest.java
@@ -29,6 +29,7 @@ import org.eclipse.collections.api.block.function.primitive.FloatFunction;
 import org.eclipse.collections.api.block.function.primitive.IntFunction;
 import org.eclipse.collections.api.block.function.primitive.LongFunction;
 import org.eclipse.collections.api.block.function.primitive.ShortFunction;
+import org.eclipse.collections.api.factory.Lists;
 import org.eclipse.collections.api.list.MutableList;
 import org.eclipse.collections.api.tuple.Pair;
 import org.eclipse.collections.impl.list.Interval;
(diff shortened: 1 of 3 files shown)


unit-tests/src/test/java/org/eclipse/collections/impl/block/factory/ComparatorsTest.java
```diff
@@ -196,6 +197,44 @@ public void fromFunctions()
...
+
+        Comparator<Person> personComparator = Comparators.byFunctionNullsLast(Person::getLastName);
+        Verify.assertNegative(personComparator.compare(alice, bob));
+        Verify.assertPositive(personComparator.compare(carol, alice));
+        Verify.assertNegative(personComparator.compare(alice, dan));
+        Verify.assertZero(personComparator.compare(bob, carol));
+
+        MutableList<Person> people = Lists.mutable.of(bob, dan, carol, alice);
+        people.sortThis(personComparator);
+        Verify.assertIterablesEqual(Lists.immutable.of(alice, dan, bob, carol), people);
```
donraab, line 216: Why not just use `Assert.assertEqual()`?

unit-tests/src/test/java/org/eclipse/collections/impl/block/factory/ComparatorsTest.java
```diff
@@ -196,6 +197,44 @@ public void fromFunctions()
...
+
+        Comparator<Person> personComparator = Comparators.byFunctionNullsFirst(Person::getLastName);
+        Verify.assertPositive(personComparator.compare(alice, bob));
+        Verify.assertNegative(personComparator.compare(carol, alice));
+        Verify.assertNegative(personComparator.compare(alice, dan));
+        Verify.assertZero(personComparator.compare(bob, carol));
+
+        MutableList<Person> people = Lists.mutable.of(bob, dan, carol, alice);
+        people.sortThis(personComparator);
+        Verify.assertIterablesEqual(Lists.immutable.of(bob, carol, alice, dan), people);
```
donraab, line 235: Just use `Assert.assertEqual()`.

vmzakharov: As discusses, also added serialization tests

nikhilnanivadekar:
LGTM

nikhilnanivadekar:
Thanks for your contribution!

PR #909, opened 2020-06-08
Closes #723 added LongInterval

### LINKED ISSUE #723: Implement LongInterval
Today we have IntInterval.  It would be useful to have LongInterval as well.  I don't know that we need more than just IntInterval and LongInterval, so I would hold off trying to code generate all of the primitive type versions of interval.

vmzakharov:
`LongInterval` supports elements of type 'long' but is otherwise consistent with other primitive lists and is limited to an int size.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/LongInterval.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/LongInterval.java
@@ -1,0 +1,1104 @@
+/*
+ * Copyright (c) 2019 Goldman Sachs and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompany this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.impl.list.primitive;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.Spliterator;
+import java.util.function.LongConsumer;
+
+import org.eclipse.collections.api.LazyLongIterable;
+import org.eclipse.collections.api.LongIterable;
+import org.eclipse.collections.api.RichIterable;
+import org.eclipse.collections.api.bag.primitive.MutableLongBag;
+import org.eclipse.collections.api.block.function.primitive.LongToObjectFunction;
+import org.eclipse.collections.api.block.function.primitive.ObjectLongIntToObjectFunction;
+import org.eclipse.collections.api.block.function.primitive.ObjectLongToObjectFunction;
+import org.eclipse.collections.api.block.predicate.primitive.LongPredicate;
+import org.eclipse.collections.api.block.procedure.primitive.LongIntProcedure;
+import org.eclipse.collections.api.block.procedure.primitive.LongLongProcedure;
+import org.eclipse.collections.api.block.procedure.primitive.LongProcedure;
+import org.eclipse.collections.api.factory.Lists;
+import org.eclipse.collections.api.iterator.LongIterator;
+import org.eclipse.collections.api.list.ImmutableList;
+import org.eclipse.collections.api.list.MutableList;
+import org.eclipse.collections.api.list.primitive.ImmutableLongList;
+import org.eclipse.collections.api.list.primitive.LongList;
+import org.eclipse.collections.api.list.primitive.MutableLongList;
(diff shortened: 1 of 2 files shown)


nikhilnanivadekar:
@mohrezaei can you do a review please? I did a rudimentary review. Would like your final review. Thanks!

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/LongInterval.java
```diff
@@ -0,0 +1,1092 @@
...
+        {
+            sum += this.get(i) * list.get(i);
+        }
+        return sum;
+    }
+
+    @Override
+    public boolean isEmpty()
+    {
+        return this.size() == 0;
```
mohrezaei, line 565: isn't this always false?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/LongInterval.java
```diff
@@ -0,0 +1,1092 @@
+/*
```
mohrezaei, line 1: [N.B. comment applies to whole class, so placed up here]
Why is the looping in this class so inconsistent? Sometimes it's fully expanded (e.g. `forEach`, `equals`, etc). Sometimes, it's using a for-loop with `this.get(i)`. Sometimes, it's using `forEach*` with a lambda. And yet other places, a `LongIterator` is used. It's like 4 different people wrote this class 😄

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/LongInterval.java
```diff
@@ -0,0 +1,1092 @@
...
+    @Override
+    public LazyLongIterable asReversed()
+    {
+        return ReverseLongIterable.adapt(this);
+    }
+
+    @Override
+    public long sum()
+    {
+        long sum = 0L;
```
mohrezaei, line 827: does the formula based implementation not work for longs? overflow?

vmzakharov: Should work, will fix

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/LongInterval.java
vmzakharov: Same issue in `IntInterval` (which I effectively cloned), can change implementation/add tests to both or have a separate PR to clean up this and other implementation issues. What do you suggest?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/LongInterval.java
vmzakharov: Same as above - I followed the implementation pattern of `IntInterval` - one reason didn't want to diverge because it would make it harder to turn it into a code generated template if we decide to do so in the future.   My preference is to leave it as it and have a separate PR to clean up both classes (and keep them consistent). Please let me know what you think.

mohrezaei: My preference would be to have a clean implementation here and another PR to clean up IntInterval. Others are welcome to chime in. @donraab @nikhilnanivadekar @motlin ?

nikhilnanivadekar: I am more of let’s get this one in first and then clean up one or both. I am waiting for these PRs to be closer to release 10.3.0. So the fastest way to that is more preferable for me.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/LongInterval.java
nikhilnanivadekar: For my understanding, is the proposal to return `false` instead of doing the check? Like Moh says it will always be `false` but this implementation is not incorrect or did I miss something? Interval always makes me think twice about these terminating methods because Interval is a mix of eager and lazy functionality.

mohrezaei: yes, the proposal is to return false. An implementation that computes PI to 700 digits and then returns `lastDigit == 0` would also work (the 700th digit of PI is 5) 😛 .

nikhilnanivadekar: LOL 😂

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/LongInterval.java
vmzakharov: I have fixed `sum()` and added more tests for it (specifically to address the concern about overflow during calculation), and have added a test for `isEmpty()`. My preference now would be to get `LongInterval` in as is and open a new issue for `LongInterval` and `IntInterval` code clean up (happy to work on it as the potential is certainly there!)

PR #912, opened 2020-06-11
Closes #908: Support sorting of primitive lists by indirect comparison

### LINKED ISSUE #908: Support sorting of primitive lists by indirect comparison.
The proposal is to add the following APIs on PrimitiveLists (in addition to the existing `sortThis()`):
- `PrimitiveList sortThis(PrimitiveComparator comparator)`
- `<T> PrimitiveList sortThisBy(PrimitiveFunction<T> keyExtractor) // sorts by the natural order of the keys`
- `<T> PrimitiveList sortThisBy(PrimitiveFunction<T> keyExtractor, Comparator<? super T> keyComparator)`

Where `PrimitiveComparator` is a Single Abstract Method interface that looks like
```
public interface PrimitiveComparator
{
    int compare(primitive value1, primitive value2);
}
```
To support this a new sort implementation will be required.

vmzakharov:
Adds indirect sorting methods to primitive collections, which are in addition to the existing `sortThis()`. Unlike `sortThis()`, which uses natural ordering of the primitives, the new methods take a function to generate sort keys with an optional key comparator,  or they can take a primitive comparator.

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/block/function/primitiveComparator.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/block/function/primitiveComparator.stg
@@ -1,0 +1,34 @@
+import "copyright.stg"
+
+targetPath() ::= "org/eclipse/collections/api/block/comparator/primitive"
+
+fileName(primitive) ::= "<primitive.name>Comparator"
+
+class(primitive) ::= <<
+<body(primitive.type, primitive.name, primitive.wrapperName)>
+>>
+
+body(type, name, wrapperName) ::= <<
+<copyright()>
+
+package org.eclipse.collections.api.block.comparator.primitive;
+
+import java.io.Serializable;
+
+/**
+ * <name>Comparator is a primitive Comparator that takes two arguments of type <type> and
+ * otherwise follows the contract of {@code java.util.Comparator}.
+ * This comparator can be used to sort primitive collections, which support indirect sorting,
+ * if a sort order other thant the natural one of the collection elements is required.
+ *
+ * This file was automatically generated from template file primitiveComparator.stg.
+ *
+ */
+@FunctionalInterface
+public interface <name>Comparator\<T>
+        extends Serializable
+{
+    int compare(<type> value1, <type> value2);
+}
+
+>>
--- a/eclipse-collections-code-generator/src/main/resources/api/list/mutablePrimitiveList.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/list/mutablePrimitiveList.stg
@@ -160,6 +160,32 @@ allMethods(type) ::= <<
 Mutable<name>List sortThis();
 
 /**
+ * Sorts the internal data structure of this list and returns the list itself as a convenience.
+ */
+default Mutable<name>List sortThis(<name>Comparator comparator)
+{
+    throw new UnsupportedOperationException("sortThis(<name>Comparator comparator) is not supported on " + this.getClass());
+}
+
+/**
+ * Sorts the internal data structure of this list based on the natural order of the key returned by {@code
+ * function}.
+ */
+default \<T> Mutable<name>List sortThisBy(<name>ToObjectFunction\<T> function)
+{
+    return sortThisBy(function, (Comparator\<? super T>) Comparator.naturalOrder());
+}
+
+/**
+ * Sorts the internal data structure of this list based on the key returned by {@code
+ * function} using the provided {@code comparator}.
+ */
+default \<T> Mutable<name>List sortThisBy(<name>ToObjectFunction\<T> function, Comparator\<? super T> comparator)
+{
+    return this.sortThis((i1, i2) -> comparator.compare(function.valueOf(i1), function.valueOf(i2)));
+}
+
+/**
  * Randomly permutes this list mutating its contents and returns the same list (this).
  *
  * Uses {@code java.util.Random} as the source of randomness.
@@ -203,7 +229,9 @@ auxiliaryImports ::= [
 ]
 
 allAuxiliaryImports(type) ::= <<
+import java.util.Comparator;
 import java.util.Random;
+import org.eclipse.collections.api.block.comparator.primitive.<name>Comparator;
 
 >>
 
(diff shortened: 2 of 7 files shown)


mohrezaei:
I looked at this a little while ago, saw the `QuickSort` implementation and went 🙈 and decided not to comment.

I don't understand the use case for this API. Maybe it's better to start there, for context.

vmzakharov:
One use case is sorting an ordered data structure that is impossible/too expensive to reorder. Then you can build an index into it by sorting an int list. I would imagine any other order scenario where sort order differs from the natural order applies – we do have this API on FastList.

Another evidence that there is interest in this pattern, see for example https://stackoverflow.com/questions/3699141/how-to-sort-an-array-of-ints-using-a-custom-comparator. 

The QuickSort implementation (taken straight out of TAOCP) certainly can be and should be optimized further. Right now, it is about 30% slower than the JDK one when applied to randomly shuffled sequences (tested on sizes 100 to 1 MM), but seems fast enough for most practical applications.

mohrezaei:
> One use case is sorting an ordered data structure that is impossible/too expensive to reorder. 

That would be a reasonable usecase in C++, but not in Java: Java doesn't have large, user definable structs (aka value type). You can try a simple benchmark for the function api: instead of filling a list with 0..n, and sorting by referencing the other list, just make a copy of the list and sort that.

Sorting random stuff is not the best benchmark for a sorting algorithm. Real life data is less random, more "patterny". The JDK algorithm recognizes a half-dozen common patterns and sorts them in `O(n)`. 

I'm still struggling with a usecase here. Is there a case where running a potentially `O(n^2)` algorithm worth it? Given the "unnatural" comparator, I'm worried these cases may occur more frequently. For example, try this comparator: `(a,b) -> (a & 1) - (b & 1)`

vmzakharov:
The data structure I had in mind would be, for example, a tabular data structure, something akin to a data frame rather than a list of structs. So, if that data structure is implemented as a columnar in-memory store, that would make it impractical to sort it directly. This kind of data structure is something I have come across in real life applications.

Good catch with the even/odd sorting comparator. The existing implementation does not handle it well (and there are other “special” cases that trip it). I am thinking adding an optimization to deal with it by switching to a different algorithm at a certain stack depth (or there may be other ways to fix it).

FWIW, even in its current state it does better on `int`s than `Arrays.sort(…)` on a randomly shuffled array of `Integer`s with the `Integer::compare` as the comparator.  And while it is certainly nowhere near heroically optimized, it should do better than the worst-case scenario in most realistic cases. 

Having said that, I think we should first agree on whether this interface is useful. If we decide it is, then we can decide what the minimal acceptable level of performance for the first iteration of its implementation is.

donraab:
FWIW, there are some related questions in StackOverflow:

https://stackoverflow.com/questions/3699141/how-to-sort-an-array-of-ints-using-a-custom-comparator
https://stackoverflow.com/questions/215271/sort-arrays-of-primitive-types-in-descending-order
https://stackoverflow.com/questions/7414299/sorting-int-array-in-descending-order
https://stackoverflow.com/questions/22666377/sort-an-array-of-primitives-with-a-custom-comparator-and-without-converting-to-o

Maybe we could make the sorting pluggable so if folks want to plug in their own sorting implementation or use another library like FastUtil which already has primitive array sorting with comparators.

http://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/ints/IntArrays.html

vmzakharov:
Meanwhile committed a change to the sort implementation that prevents performance degradation when there are sequences of same sort key values (including @mohrezaei's even/odd example)

mohrezaei:
I read through the stackoverflow questions and there are at least 2 real-world usecases that are poorly covered by the java libraries. 

- Reverse sort. Such a simple thing, and yet impossible without an expensive second pass.
- Sort by absolute value. Not a trivial case, but seems reasonable enough.

So I think I'm good with the usefulness/rationale. 

As for the algorithm, it seems we have multiple choices:

- Take a standard algorithm, like mildly optimized quicksort. Quicksort is troublesome because of the worst case behavior, and it's very hard to fix. Something like mergesort has better worst case behavior, but requires extra memory. So, this choice is not trivial.
- Use a third party implementation. I wouldn't like the new extra dependency. 
- Try to one-up the Java library sort algorithm. The work would be divided into pattern matching, and a sort algorithm, like [triple pivot quicksort](https://lusy.fri.uni-lj.si/sites/lusy.fri.uni-lj.si/files/publications/alopez2014-seminar-qsort.pdf), Timsort for the object-function case, etc.

Things we shouldn't do:

- Copy the code out of the JDK :grin:
- Pluggable: it seems like that would be next to impossible because there is no shared primitive comparator between the different libraries. Converting between different lambda types would be egregious.

mohrezaei:
Oh, more thing: the code should be organized very differently if we implement our own sort. It should be split into a (static) sort utility that just deals with arrays and the api on the list just calls that. The reason is it would be a value add for our users who need to sort arrays with a comparator.

vmzakharov:
A clean room mildly optimized implementation is exactly what I am going for here. 😀

A good point re separating the sort utility. I would make the existing inner sort classes into regular classes and change the code so that is is against arrays rather than lists. Then the implementation classes will use the sort utility, but the default methods on the primitive mutable list interfaces will do `toArray()`, use the sort utility to sort the array and then do `set(…)` on the list. The primitive sort APIs can be also added to `ArrayIterate` delegating to the utilities (if we decide to do it, would do it as a separate PR).

donraab:
@mohrezaei Good point on the issue with the pluggable approach. I forgot we'd have to convert the lambda types between the libraries.

There is another longer term option which is voting for or contributing the primitive array sort to OpenJDK. But it's doubtful this would be implemented before Valhalla is complete as it would involve adding new primitive comparator types which Valhalla is supposed to render unnecessary eventually.

I would go with mildly optimized implementation for now and either we optimize eventually or try and get the primitive array sorting into the JDK via a JEP. 

@vmzakharov I would make the default method throw `UnsupportedOperationException` for now as we can't reference the utility in the interfaces. Then implement the method calling the array utility classes in the primitive array lists.

vmzakharov:
@mohrezaei, @donraab  - implemented your suggestions, can you take a look please?

PR #913, opened 2020-06-11
Add keySet to MapIterable

### LINKED ISSUE #740: Add keySet to MapIterable and/or make API more consistent
I noticed that the primitive Maps have a `keySet` method, but the generic ones don't.

The immutable ones return an unmodifiable set (because a `MutableSet` is expected to be returned by the interfaces).

I think it would make sense to add it also to the generic Maps because there is only a `keyView` available and sometimes, it's useful to have a `Set` (or a `SetIterable` maybe?).

johnjstuart:
Signed-off-by: John Stuart <john.stuart@bnymellon.com>

closes #740

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/immutable/AbstractImmutableBiMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/immutable/AbstractImmutableBiMap.java
@@ -65,6 +65,7 @@ import org.eclipse.collections.impl.block.procedure.SelectInstancesOfProcedure;
 import org.eclipse.collections.impl.list.fixed.ArrayAdapter;
 import org.eclipse.collections.impl.multimap.set.UnifiedSetMultimap;
 import org.eclipse.collections.impl.partition.set.PartitionUnifiedSet;
+import org.eclipse.collections.impl.set.immutable.AbstractImmutableSet;
 import org.eclipse.collections.impl.set.mutable.UnifiedSet;
 import org.eclipse.collections.impl.utility.Iterate;
 import org.eclipse.collections.impl.utility.MapIterate;
@@ -186,9 +187,9 @@ public abstract class AbstractImmutableBiMap<K, V> extends AbstractBiMap<K, V> i
     }
 
     @Override
-    public Set<K> keySet()
+    public ImmutableKeySet<K> keySet()
     {
-        return this.delegate.castToMap().keySet();
+        return new ImmutableKeySet<>(this.delegate.keySet());
     }
 
     @Override
@@ -481,6 +482,75 @@ public abstract class AbstractImmutableBiMap<K, V> extends AbstractBiMap<K, V> i
         protected Object writeReplace()
         {
             return new ImmutableBiMapSerializationProxy<>(this);
+        }
+    }
+
+    protected static class ImmutableKeySet<K> extends AbstractImmutableSet<K>
+    {
+        private final ImmutableSet<K> original;
+
+        protected ImmutableKeySet(ImmutableSet<K> keySet)
+        {
+            this.original = keySet;
+        }
+
+        @Override
+        public int size()
+        {
+            return this.original.size();
+        }
+
+        @Override
+        public K getFirst()
+        {
+            return this.original.getFirst();
+        }
+
+        @Override
+        public K getLast()
+        {
+            return this.original.getLast();
+        }
+
+        @Override
+        public void each(Procedure<? super K> procedure)
+        {
+            this.original.each(procedure);
+        }
+
+        @Override
+        public boolean contains(Object object)
+        {
+            return this.original.contains(object);
+        }
+
+        @Override
+        public Iterator<K> iterator()
+        {
+            return this.original.iterator();
+        }
+
+        @Override
+        public boolean equals(Object object)
+        {
+            if (this.original == object)
+            {
+                return true;
+            }
+
+            if (!(object instanceof Set))
+            {
+                return false;
+            }
+
+            Set<?> other = (Set<?>) object;
+            return this.original.size() == other.size() && this.original.containsAll(other);
+        }
+
+        @Override
+        public int hashCode()
+        {
+            return this.original.hashCode();
         }
     }
 }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
@@ -20,6 +20,7 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutorService;
 
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.bag.MutableBag;
@@ -56,6 +57,7 @@ import org.eclipse.collections.api.multimap.set.MutableSetMultimap;
 import org.eclipse.collections.api.ordered.OrderedIterable;
 import org.eclipse.collections.api.partition.set.PartitionMutableSet;
 import org.eclipse.collections.api.set.MutableSet;
+import org.eclipse.collections.api.set.ParallelUnsortedSetIterable;
 import org.eclipse.collections.api.tuple.Pair;
 import org.eclipse.collections.impl.bimap.AbstractBiMap;
 import org.eclipse.collections.impl.block.factory.Predicates;
@@ -66,6 +68,7 @@ import org.eclipse.collections.impl.list.fixed.ArrayAdapter;
 import org.eclipse.collections.impl.map.mutable.UnifiedMap;
 import org.eclipse.collections.impl.multimap.set.UnifiedSetMultimap;
 import org.eclipse.collections.impl.partition.set.PartitionUnifiedSet;
+import org.eclipse.collections.impl.set.mutable.AbstractMutableSet;
 import org.eclipse.collections.impl.set.mutable.UnifiedSet;
 import org.eclipse.collections.impl.utility.Iterate;
 import org.eclipse.collections.impl.utility.MapIterate;
@@ -398,7 +401,7 @@ abstract class AbstractMutableBiMap<K, V> extends AbstractBiMap<K, V> implements
     }
 
     @Override
-    public Set<K> keySet()
+    public MutableSet<K> keySet()
     {
         return new KeySet();
     }
@@ -752,8 +755,10 @@ abstract class AbstractMutableBiMap<K, V> extends AbstractBiMap<K, V> implements
         }
     }
 
-    private class KeySet implements Set<K>, Serializable
+    private class KeySet extends AbstractMutableSet<K> implements Serializable
     {
+        private static final long serialVersionUID = 1L;
+
         @Override
         public boolean equals(Object obj)
         {
@@ -779,6 +784,26 @@ abstract class AbstractMutableBiMap<K, V> extends AbstractBiMap<K, V> implements
         }
 
         @Override
+        public K getFirst()
+        {
+            if (AbstractMutableBiMap.this.size() > 0)
+            {
+                return (K) this.toArray()[0];
+            }
+            return null;
+        }
+
+        @Override
+        public K getLast()
+        {
+            if (AbstractMutableBiMap.this.size() > 0)
+            {
+                return (K) this.toArray()[AbstractMutableBiMap.this.size() - 1];
+            }
+            return null;
+        }
+
+        @Override
         public boolean contains(Object key)
         {
             return AbstractMutableBiMap.this.delegate.containsKey(key);
@@ -794,12 +819,6 @@ abstract class AbstractMutableBiMap<K, V> extends AbstractBiMap<K, V> implements
         public <T> T[] toArray(T[] a)
         {
             return AbstractMutableBiMap.this.delegate.keySet().toArray(a);
-        }
-
-        @Override
-        public boolean add(K key)
-        {
-            throw new UnsupportedOperationException("Cannot call add() on " + this.getClass().getSimpleName());
         }
 
         @Override
@@ -821,6 +840,12 @@ abstract class AbstractMutableBiMap<K, V> extends AbstractBiMap<K, V> implements
                 }
             }
             return true;
+        }
+
+        @Override
+        public void each(Procedure<? super K> procedure)
+        {
+            AbstractMutableBiMap.this.delegate.keySet().each(procedure);
         }
 
         @Override
@@ -879,6 +904,13 @@ abstract class AbstractMutableBiMap<K, V> extends AbstractBiMap<K, V> implements
             MutableSet<K> replace = UnifiedSet.newSet(AbstractMutableBiMap.this.size());
             AbstractMutableBiMap.this.forEachKey(CollectionAddProcedure.on(replace));
             return replace;
+        }
+
+        @Override
+        public ParallelUnsortedSetIterable<K> asParallel(
+                ExecutorService executorService, int batchSize)
+        {
+            return AbstractMutableBiMap.this.delegate.keySet().asParallel(executorService, batchSize);
         }
     }
 
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/AbstractImmutableMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/AbstractImmutableMap.java
@@ -63,6 +63,7 @@ import org.eclipse.collections.api.partition.bag.PartitionImmutableBag;
 import org.eclipse.collections.api.partition.bag.PartitionMutableBag;
 import org.eclipse.collections.api.set.ImmutableSet;
 import org.eclipse.collections.api.set.MutableSet;
+import org.eclipse.collections.api.set.SetIterable;
 import org.eclipse.collections.api.tuple.Pair;
 import org.eclipse.collections.impl.bag.mutable.HashBag;
 import org.eclipse.collections.impl.bag.mutable.primitive.BooleanHashBag;
@@ -94,6 +95,7 @@ import org.eclipse.collections.impl.map.mutable.primitive.ObjectDoubleHashMap;
 import org.eclipse.collections.impl.map.mutable.primitive.ObjectLongHashMap;
 import org.eclipse.collections.impl.multimap.bag.HashBagMultimap;
 import org.eclipse.collections.impl.partition.bag.PartitionHashBag;
+import org.eclipse.collections.impl.set.immutable.AbstractImmutableSet;
 import org.eclipse.collections.impl.set.mutable.UnifiedSet;
 import org.eclipse.collections.impl.tuple.ImmutableEntry;
 import org.eclipse.collections.impl.utility.MapIterate;
@@ -200,6 +202,9 @@ public abstract class AbstractImmutableMap<K, V>
         }
         return map.toImmutable();
     }
+
+    @Override
+    public abstract AbstractImmutableSet<K> keySet();
 
     @Override
     public V put(K key, V value)
@@ -518,5 +523,45 @@ public abstract class AbstractImmutableMap<K, V>
     public <V1> ImmutableBag<V1> countByEach(Function<? super V, ? extends Iterable<V1>> function)
     {
         return this.countByEach(function, Bags.mutable.empty()).toImmutable();
+    }
+
+    public static class ImmutableKeySet<K> extends AbstractImmutableSet<K>
+    {
+        final SetIterable<K> original;
+
+        protected ImmutableKeySet(SetIterable<K> keySet)
+        {
+            this.original = keySet;
+        }
+
+        @Override
+        public int size()
+        {
+            return this.original.size();
+        }
+
+        @Override
+        public K getFirst()
+        {
+            return this.original.getFirst();
+        }
+
+        @Override
+        public K getLast()
+        {
+            return this.original.getLast();
+        }
+
+        @Override
+        public void each(Procedure<? super K> procedure)
+        {
+            this.original.each(procedure);
+        }
+
+        @Override
+        public Iterator<K> iterator()
+        {
+            return this.original.iterator();
+        }
     }
 }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableDoubletonMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableDoubletonMap.java
@@ -14,7 +14,6 @@ import java.io.Serializable;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.block.function.Function2;
@@ -102,9 +101,9 @@ final class ImmutableDoubletonMap<K, V>
     }
 
     @Override
-    public Set<K> keySet()
+    public ImmutableKeySet<K> keySet()
     {
-        return Sets.immutable.with(this.key1, this.key2).castToSet();
+        return new ImmutableKeySet<>(Sets.immutable.with(this.key1, this.key2));
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableUnifiedMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableUnifiedMap.java
@@ -13,7 +13,6 @@ package org.eclipse.collections.impl.map.immutable;
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.Map;
-import java.util.Set;
 
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.block.function.Function;
@@ -26,7 +25,6 @@ import org.eclipse.collections.api.tuple.Pair;
 import org.eclipse.collections.impl.collection.mutable.UnmodifiableMutableCollection;
 import org.eclipse.collections.impl.map.mutable.UnifiedMap;
 import org.eclipse.collections.impl.parallel.BatchIterable;
-import org.eclipse.collections.impl.set.mutable.UnmodifiableMutableSet;
 
 /**
  * @see ImmutableMap
@@ -120,9 +118,9 @@ public class ImmutableUnifiedMap<K, V>
     }
 
     @Override
-    public Set<K> keySet()
+    public ImmutableKeySet<K> keySet()
     {
-        return UnmodifiableMutableSet.of(this.delegate.keySet());
+        return new ImmutableKeySet<>(this.delegate.keySet());
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
@@ -44,8 +44,10 @@ import org.eclipse.collections.api.factory.Maps;
 import org.eclipse.collections.api.map.ConcurrentMutableMap;
 import org.eclipse.collections.api.map.ImmutableMap;
 import org.eclipse.collections.api.map.MutableMap;
+import org.eclipse.collections.api.set.MutableSet;
 import org.eclipse.collections.api.tuple.Pair;
 import org.eclipse.collections.impl.block.procedure.MapEntryToProcedure2;
+import org.eclipse.collections.impl.factory.Sets;
 import org.eclipse.collections.impl.list.mutable.FastList;
 import org.eclipse.collections.impl.utility.Iterate;
 import org.eclipse.collections.impl.utility.MapIterate;
@@ -965,9 +967,9 @@ public final class ConcurrentHashMap<K, V>
     }
 
     @Override
-    public Set<K> keySet()
+    public MutableSet<K> keySet()
     {
-        return new KeySet();
+        return Sets.adapt(new KeySet());
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMapUnsafe.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMapUnsafe.java
@@ -45,8 +45,10 @@ import org.eclipse.collections.api.factory.Maps;
 import org.eclipse.collections.api.map.ConcurrentMutableMap;
 import org.eclipse.collections.api.map.ImmutableMap;
 import org.eclipse.collections.api.map.MutableMap;
+import org.eclipse.collections.api.set.MutableSet;
 import org.eclipse.collections.api.tuple.Pair;
 import org.eclipse.collections.impl.block.procedure.MapEntryToProcedure2;
+import org.eclipse.collections.impl.factory.Sets;
 import org.eclipse.collections.impl.list.mutable.FastList;
 import org.eclipse.collections.impl.utility.Iterate;
 import org.eclipse.collections.impl.utility.MapIterate;
@@ -1082,9 +1084,9 @@ public class ConcurrentHashMapUnsafe<K, V>
     }
 
     @Override
-    public Set<K> keySet()
+    public MutableSet<K> keySet()
     {
-        return new KeySet();
+        return Sets.adapt(new KeySet());
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentMutableHashMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentMutableHashMap.java
@@ -28,8 +28,10 @@ import org.eclipse.collections.api.factory.Maps;
 import org.eclipse.collections.api.map.ConcurrentMutableMap;
 import org.eclipse.collections.api.map.ImmutableMap;
 import org.eclipse.collections.api.map.MutableMap;
+import org.eclipse.collections.api.set.MutableSet;
 import org.eclipse.collections.api.tuple.Pair;
 import org.eclipse.collections.impl.block.procedure.MapEntryToProcedure2;
+import org.eclipse.collections.impl.factory.Sets;
 import org.eclipse.collections.impl.utility.Iterate;
 import org.eclipse.collections.impl.utility.internal.IterableIterate;
 
@@ -164,9 +166,9 @@ public final class ConcurrentMutableHashMap<K, V>
     }
 
     @Override
-    public Set<K> keySet()
+    public MutableSet<K> keySet()
     {
-        return this.delegate.keySet();
+        return Sets.adapt(this.delegate.keySet());
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/MapAdapter.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/MapAdapter.java
@@ -21,6 +21,8 @@ import org.eclipse.collections.api.block.procedure.Procedure2;
 import org.eclipse.collections.api.factory.Maps;
 import org.eclipse.collections.api.map.ImmutableMap;
 import org.eclipse.collections.api.map.MutableMap;
+import org.eclipse.collections.api.set.MutableSet;
+import org.eclipse.collections.impl.factory.Sets;
 import org.eclipse.collections.impl.utility.Iterate;
 import org.eclipse.collections.impl.utility.MapIterate;
 
@@ -94,9 +96,9 @@ public class MapAdapter<K, V>
     }
 
     @Override
-    public Set<K> keySet()
+    public MutableSet<K> keySet()
     {
-        return this.delegate.keySet();
+        return Sets.adapt(this.delegate.keySet());
     }
 
     @Override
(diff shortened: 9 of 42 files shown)


donraab:
Hi @johnjstuart, thank you for the contribution! I feel like maybe we should find a middle ground between the readable `keysView()` and the mutable `keySet()`. @motlin @nikhilnanivadekar @victornoel What do you think about having a method named `unmodifiableKeySet()` which consistently returns an unmodifiable `Set`?

motlin:
Ideally, `MapIterable.keySet()` would return `SetIterable`. Sub-interfaces like `MutableMap.keySet()` would return `MutableSet` which also implements `Set`. One "trick" is that all `ImmutableSet` implementations implement `Set` even though `ImmutableSet` doesn't extend `Set` which is one reason this patch is so small.

So, I'm torn. Based on what I think is the ideal signature, I wouldn't have merged this PR. But I've been saying that for a long time now, and we haven't implemented the ideal signature. Maybe it would be better to land something like this and then enhance it (and break backwards compatibility) in a future major version.

victornoel:
IMHO, if having a `MapIterable.keySet` returning a `SetIterable` is too complex, at least I would like to have another named method (`keys`?) that would return a `SetIterable`.

I don't think it is a good idea to have `MapIterable.keySet` return `Set` because it breaks the whole iterable/mutable/immutable type hierarchy followed by eclipse-collections.

donraab:
Agreed. @johnjstuart can you look into how hard it would be to implement `keySet` on `MapIterable` to return `SetIterable`? In leaf interfaces that current return just `Set`, we would have to return a `MutableSet`. The most convenient solution in some cases may be to return a `SetAdapter` if necessary.

johnjstuart:
@donraab Will do.

johnjstuart:
@donraab and @motlin I think the biggest barrier with this approach is that `ImmutableSetIterable` does not extend `Set`. Since `AbstractImmutableMap` implements `java.util.Map` anything that inherits from `AbstractImmutableMap` will have to return a `Set` and `ImmutableSet` will clash with that.

In the mutable hierarchy, both map and set extend the `java.util` interfaces higher up in the chain (`MutableMapIterable` extends `Map` and `MutableSetIterable` extends `Set`). On the immutable side, `Map` and `Set` are only inherited at the `AbstractImmutable*` level. Does it make sense to push those upwards to make the immutable hierarchy consistent with mutable, or was it a design decision to have the JDK interfaces inherited at the abstract base class level?

nikhilnanivadekar:
Take a look at how subList() is implemented on the ImmutableList. It had a similar problem. It might give you an idea :)

victornoel:
> @donraab and @motlin I think the biggest barrier with this approach is that ImmutableSetIterable does not extend Set. Since AbstractImmutableMap implements java.util.Map anything that inherits from AbstractImmutableMap will have to return a Set and ImmutableSet will clash with that.

@johnjstuart an alternative would be not to use the `keySet` name for the method, because it's going to clash with existing method.

johnjstuart:
@nikhilnanivadekar Thanks for the tip!

@donraab and @motlin I just updated this PR with a working version of the change. I'm anticipating lots of feedback, so whenever you get a chance to take a look at it, it's ready. Thanks!

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/immutable/AbstractImmutableBiMap.java
```diff
@@ -483,4 +484,67 @@ protected Object writeReplace()
             return new ImmutableBiMapSerializationProxy<>(this);
         }
     }
+
+    protected static class ImmutableKeySet<K> extends AbstractImmutableSet<K>
```
victornoel, line 488: @johnjstuart don't forget to override contains too because it is implemented in `AbstractRichIterable` in the hierarchy!

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
```diff
@@ -796,12 +821,6 @@ public boolean contains(Object key)
             return AbstractMutableBiMap.this.delegate.keySet().toArray(a);
         }
 
-        @Override
-        public boolean add(K key)
-        {
-            throw new UnsupportedOperationException("Cannot call add() on " + this.getClass().getSimpleName());
```
victornoel, line 802: @johnjstuart why remove this but not `addAll`? (and maybe some others?)

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/AbstractImmutableMap.java
```diff
@@ -519,4 +523,44 @@ public ImmutableShortBag collectShort(ShortFunction<? super V> shortFunction)
     {
         return this.countByEach(function, Bags.mutable.empty()).toImmutable();
     }
+
+    public static class ImmutableKeySet<K> extends AbstractImmutableSet<K>
```
victornoel, line 527: @johnjstuart this class seems unused (and also useless, why wrap an `ImmutableSet` to make it Immutable?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/AbstractImmutableMap.java
```diff
@@ -201,6 +202,9 @@ public void clear()
         return map.toImmutable();
     }
 
+    @Override
+    public abstract AbstractImmutableSet<K> keySet();
```
victornoel, line 206: @johnjstuart what's the point of specifying that it should implement `AbstractImmutableSet`? I wouldn't specify anything in this class but only in subclasses IMHO.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableDoubletonMap.java
```diff
@@ -102,9 +101,9 @@ public V get(Object key)
     }
 
     @Override
-    public Set<K> keySet()
+    public ImmutableKeySet<K> keySet()
     {
-        return Sets.immutable.with(this.key1, this.key2).castToSet();
+        return new ImmutableKeySet<>(Sets.immutable.with(this.key1, this.key2));
```
victornoel, line 106: @johnjstuart what's the point of wrapping an already `ImmutableSet` to an `ImmutableSet`? (here and all the other subclasses of `AbstractImmutableMap`? Just return it.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
```diff
@@ -965,9 +967,9 @@ else if (o != null)
     }
 
     @Override
-    public Set<K> keySet()
+    public MutableSet<K> keySet()
     {
-        return new KeySet();
+        return SetAdapter.adapt(new KeySet());
```
victornoel, line 972: @johnjstuart can't `KeySet` directly implement `MutableSet` since it is implemented in the same file?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMapUnsafe.java
```diff
@@ -1082,9 +1084,9 @@ else if (o != null)
     }
 
     @Override
-    public Set<K> keySet()
+    public MutableSet<K> keySet()
     {
-        return new KeySet();
+        return SetAdapter.adapt(new KeySet());
```
victornoel, line 1089: @johnjstuart can't `KeySet` directly implement `MutableSet` since it is implemented in the same file?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -2404,6 +2428,13 @@ private void chainedAddToSet(Object[] chain, UnifiedSet<K> replace)
                 replace.add(UnifiedMap.this.nonSentinel(cur));
             }
         }
+
+        @Override
+        public ParallelUnsortedSetIterable<K> asParallel(
+                ExecutorService executorService, int batchSize)
+        {
+            return UnifiedMap.this.keysView().toSet().asParallel(executorService, batchSize);
```
victornoel, line 2436: @johnjstuart can't we avoid calling `toSet` that is not in O(1)?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnmodifiableMutableMap.java
```diff
@@ -390,6 +391,12 @@ public boolean equals(Object obj)
         return (MutableMap<K, V>) this.delegate;
     }
 
+    @Override
+    public MutableSet<K> keySet()
+    {
+        return SetAdapter.adapt(this.delegate.keySet()).asUnmodifiable();
```
victornoel, line 397: @johnjstuart you can write instead: `getMutableMap().asUnmodifiable()` I believe

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMap.java
```diff
@@ -149,9 +150,9 @@ public void clear()
     }
 
     @Override
-    public Set<K> keySet()
+    public UnmodifiableMutableSet<K> keySet()
     {
-        return Collections.unmodifiableSet(this.delegate.keySet());
+        return UnmodifiableMutableSet.of(this.delegate.keySet());
```
victornoel, line 155: @johnjstuart you should be able to write `this.delegate.keySet().asUnmodifiable()` I believe here

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/immutable/AbstractImmutableSortedMap.java
```diff
@@ -188,6 +189,9 @@ public void clear()
         return sortedMap.toImmutable();
     }
 
+    @Override
+    public abstract AbstractImmutableSortedSet<K> keySet();
```
victornoel, line 193: @johnjstuart I don't think this is needed at all and it constrains things too much IMHO for the subclasses

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/immutable/ImmutableEmptySortedMap.java
```diff
@@ -247,9 +252,9 @@ public V getIfAbsentValue(K key, V value)
     }
 
     @Override
-    public Set<K> keySet()
+    public ImmutableEmptySet<K> keySet()
     {
-        return Sets.immutable.<K>of().castToSet();
+        return new ImmutableEmptySet<>(Sets.immutable.<K>of());
```
victornoel, line 257: @johnjstuart you don't need to wrap it IMHO

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/immutable/ImmutableTreeMap.java
```diff
@@ -231,7 +235,7 @@ public K lastKey()
     }
 
     @Override
-    public Set<K> keySet()
+    public ImmutableSortedMapKeySet keySet()
```
victornoel, line 238: @johnjstuart I don't think it's a good idea to publicly expose a class internal to the file here, let's keep the interface, no?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/immutable/ImmutableEmptyMapWithHashingStrategy.java
```diff
@@ -92,9 +93,9 @@ public V get(Object key)
     }
 
     @Override
-    public Set<K> keySet()
+    public ImmutableKeySet<K> keySet()
     {
-        return Sets.immutable.<K>of().castToSet();
+        return new ImmutableKeySet<K>(Sets.immutable.of()){};
```
victornoel, line 98: @johnjstuart no need to wrap here

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
```diff
@@ -1751,12 +1750,38 @@ public boolean containsAll(Collection<?> collection)
...
+            }
+            return null;
+        }
+
+        @Override
+        public K getLast()
+        {
+            if (UnifiedMapWithHashingStrategy.this.size() > 0)
+            {
+                return (K) this.toArray()[UnifiedMapWithHashingStrategy.this.size() - 1];
```
victornoel, line 1780: @johnjstuart I don't think that convering to array just to get the last element is a good idea, it's too costly IMHO

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -2145,6 +2149,26 @@ public boolean isEmpty()
...
+            }
+            return null;
+        }
+
+        @Override
+        public K getLast()
+        {
+            if (UnifiedMap.this.size() > 0)
+            {
+                return (K) this.toArray()[UnifiedMap.this.size() - 1];
```
victornoel, line 2167: @johnjstuart I don't think that convering to array just to get the last element is a good idea, it's too costly IMHO

victornoel (changes requested):
@johnjstuart here is a bunch of comments. Mainly it seems you added extra wrappers in some place without reasons and also exposed internal types as return types of public method which IMHO is a bad idea.

Thanks :)

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/immutable/AbstractImmutableBiMap.java
victornoel: @johnjstuart no because it goes through the whole collection in O(n) while in this case, you can do it in O(1) since you are in a map implementation.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableDoubletonMap.java
victornoel: @johnjstuart but `ImmutableDoubletonMap` implements `ImmutableSet`, and `ImmutableSet.keySet()` is meant to return an `ImmutableSet`, not a Set. Note also that all `ImmutableSet` implementations already implements `Set`, see https://github.com/eclipse/eclipse-collections/blob/master/eclipse-collections-api/src/main/java/org/eclipse/collections/api/set/ImmutableSet.java#L42

victornoel: @johnjstuart hence I think it is useless to do all this wrapping

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/AbstractImmutableMap.java
victornoel: @johnjstuart yes I realised later that you use it in other classes and not here. But see my comments below about that usage, I don't think it is needed.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnmodifiableMutableMap.java
victornoel: @johnjstuart sorry, I meant `getMutableMap().keySet().asUnmodifiable()` (since now `MutableMap` has a `keySet()` that returns `MutableSet` I believe).

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMap.java
victornoel: @johnjstuart isn't `delegate.keySet` now returning `MutableMap` because of your changes? `delegate` is a `MutableOrderedMap`

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/immutable/AbstractImmutableSortedMap.java
victornoel: @johnjstuart I'm surprised this happens, worst case scenario, we should expose publicly an interface, not an internal implementation here if possible

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableUnifiedMap.java
```diff
@@ -120,9 +118,9 @@ public void forEachKeyValue(Procedure2<? super K, ? super V> procedure)
     }
 
     @Override
-    public Set<K> keySet()
+    public ImmutableKeySet<K> keySet()
     {
-        return UnmodifiableMutableSet.of(this.delegate.keySet());
+        return new ImmutableKeySet<>(this.delegate.keySet().toImmutable());
```
donraab, line 123: I wonder if this can be done by returning an ImmutableWrapper around the mutable keySet so it doesn't have to be copied.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/MapAdapter.java
```diff
@@ -94,9 +96,9 @@ public V remove(Object key)
     }
 
     @Override
-    public Set<K> keySet()
+    public MutableSet<K> keySet()
     {
-        return this.delegate.keySet();
+        return SetAdapter.adapt(this.delegate.keySet());
```
donraab, line 101: @johnjstuart you can use `Sets.adapt()` here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentMutableHashMap.java
```diff
@@ -164,9 +166,9 @@ public V remove(Object key)
     }
 
     @Override
-    public Set<K> keySet()
+    public MutableSet<K> keySet()
     {
-        return this.delegate.keySet();
+        return SetAdapter.adapt(this.delegate.keySet());
```
donraab, line 171: @johnjstuart you can use `Sets.adapt()` here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMapUnsafe.java
donraab: @johnjstuart you can use `Sets.adapt()` here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
donraab: @johnjstuart you can use `Sets.adapt()` here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -2145,6 +2149,26 @@ public boolean isEmpty()
             return UnifiedMap.this.isEmpty();
         }
 
+        @Override
+        public K getFirst()
+        {
+            if (UnifiedMap.this.size() > 0)
+            {
+                return (K) this.toArray()[0];
```
donraab, line 2157: @johnjstuart Just use `this.iterator().next()`?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
donraab: @johnjstuart use `IterableIterate.getLast(this)`?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnmodifiableMutableMap.java
donraab: @johnjstuart I agree with @victornoel recommendation here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/immutable/ImmutableTreeMap.java
```diff
@@ -351,12 +355,58 @@ public boolean containsAll(Collection<?> collection)
...
+        {
+            return 0;
+        }
+
+        @Override
+        public K getFirst()
+        {
+            if (this.size() > 0)
+            {
+                return (K) this.toArray()[0];
```
donraab, line 381: Use `IterableIterate.getFirst(this)`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/immutable/ImmutableTreeMap.java
```diff
@@ -351,12 +355,58 @@ public boolean containsAll(Collection<?> collection)
...
+            }
+            return null;
+        }
+
+        @Override
+        public K getLast()
+        {
+            if (this.size() > 0)
+            {
+                return (K) this.toArray()[0];
```
donraab, line 391: Use `IterableIterate.getLast(this)`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/immutable/ImmutableUnifiedMapWithHashingStrategy.java
```diff
@@ -137,9 +136,9 @@ public void forEachKeyValue(Procedure2<? super K, ? super V> procedure)
     }
 
     @Override
-    public Set<K> keySet()
+    public ImmutableKeySet<K> keySet()
     {
-        return UnmodifiableMutableSet.of(this.delegate.keySet());
+        return new ImmutableKeySet<K>(this.delegate.keySet().toImmutable()){};
```
donraab, line 141: @johnjstuart I wonder if we could wrap the mutable keySet here with an Immutable wrapper.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
```diff
@@ -1751,12 +1750,38 @@ public boolean containsAll(Collection<?> collection)
...
         {
             return UnifiedMapWithHashingStrategy.this.isEmpty();
         }
 
+        @Override
+        public K getFirst()
+        {
+            if (UnifiedMapWithHashingStrategy.this.size() > 0)
+            {
+                return (K) this.toArray()[0];
```
donraab, line 1770: Use `IterableIterate.getFirst(this)`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
donraab: Use `IterableIterate.getLast(this)`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
```diff
@@ -2018,6 +2037,13 @@ private void chainedAddToSet(Object[] chain, UnifiedSetWithHashingStrategy<K> re
                 replace.add(UnifiedMapWithHashingStrategy.this.nonSentinel(cur));
             }
         }
+
+        @Override
+        public ParallelUnsortedSetIterable<K> asParallel(
+                ExecutorService executorService, int batchSize)
+        {
+            return SetAdapter.adapt(this).asParallel(executorService, batchSize);
```
donraab, line 2045: @johnjstuart Use `Sets.adapt()`.

serialization-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMapSerializationTest.java
```diff
@@ -34,10 +34,8 @@ public void keySet()
         Verify.assertSerializedForm(
                 "rO0ABXNyAFhvcmcuZWNsaXBzZS5jb2xsZWN0aW9ucy5pbXBsLmNvbGxlY3Rpb24ubXV0YWJsZS5T\n"
                         + "eW5jaHJvbml6ZWRDb2xsZWN0aW9uU2VyaWFsaXphdGlvblByb3h5AAAAAAAAAAEMAAB4cHNyADNv\n"
-                        + "cmcuZWNsaXBzZS5jb2xsZWN0aW9ucy5pbXBsLnNldC5tdXRhYmxlLlNldEFkYXB0ZXIAAAAAAAAA\n"
-                        + "AQIAAUwACGRlbGVnYXRldAAPTGphdmEvdXRpbC9TZXQ7eHBzcgAzb3JnLmVjbGlwc2UuY29sbGVj\n"
-                        + "dGlvbnMuaW1wbC5zZXQubXV0YWJsZS5VbmlmaWVkU2V0AAAAAAAAAAEMAAB4cHcIAAAAAD9AAAB4\n"
-                        + "eA==",
+                        + "cmcuZWNsaXBzZS5jb2xsZWN0aW9ucy5pbXBsLnNldC5tdXRhYmxlLlVuaWZpZWRTZXQAAAAAAAAA\n"
```
donraab, line 37: Did you change serialVersionUID?

serialization-tests/src/test/java/org/eclipse/collections/impl/map/mutable/SynchronizedMutableMapSerializationTest.java
```diff
@@ -34,10 +34,8 @@ public void keySet()
         Verify.assertSerializedForm(
                 "rO0ABXNyAFhvcmcuZWNsaXBzZS5jb2xsZWN0aW9ucy5pbXBsLmNvbGxlY3Rpb24ubXV0YWJsZS5T\n"
                         + "eW5jaHJvbml6ZWRDb2xsZWN0aW9uU2VyaWFsaXphdGlvblByb3h5AAAAAAAAAAEMAAB4cHNyADNv\n"
-                        + "cmcuZWNsaXBzZS5jb2xsZWN0aW9ucy5pbXBsLnNldC5tdXRhYmxlLlNldEFkYXB0ZXIAAAAAAAAA\n"
-                        + "AQIAAUwACGRlbGVnYXRldAAPTGphdmEvdXRpbC9TZXQ7eHBzcgAzb3JnLmVjbGlwc2UuY29sbGVj\n"
-                        + "dGlvbnMuaW1wbC5zZXQubXV0YWJsZS5VbmlmaWVkU2V0AAAAAAAAAAEMAAB4cHcIAAAAAD9AAAB4\n"
-                        + "eA==",
+                        + "cmcuZWNsaXBzZS5jb2xsZWN0aW9ucy5pbXBsLnNldC5tdXRhYmxlLlVuaWZpZWRTZXQAAAAAAAAA\n"
```
donraab, line 37: Did you change serialVersionUID?

serialization-tests/src/test/java/org/eclipse/collections/impl/map/mutable/UnmodifiableMutableMapSerializationTest.java
```diff
@@ -33,11 +33,11 @@ public void serializedForm()
     public void keySet()
     {
         Verify.assertSerializedForm(
-                -9215047833775013803L,
-                "rO0ABXNyACVqYXZhLnV0aWwuQ29sbGVjdGlvbnMkVW5tb2RpZmlhYmxlU2V0gB2S0Y+bgFUCAAB4\n"
-                        + "cgAsamF2YS51dGlsLkNvbGxlY3Rpb25zJFVubW9kaWZpYWJsZUNvbGxlY3Rpb24ZQgCAy173HgIA\n"
-                        + "AUwAAWN0ABZMamF2YS91dGlsL0NvbGxlY3Rpb247eHBzcgAzb3JnLmVjbGlwc2UuY29sbGVjdGlv\n"
-                        + "bnMuaW1wbC5zZXQubXV0YWJsZS5VbmlmaWVkU2V0AAAAAAAAAAEMAAB4cHcIAAAAAD9AAAB4",
+                4749516097143016747L,
+                "rO0ABXNyAFhvcmcuZWNsaXBzZS5jb2xsZWN0aW9ucy5pbXBsLmNvbGxlY3Rpb24ubXV0YWJsZS5V\n"
```
donraab, line 37: Did you change serialVersionUID?

donraab (changes requested):
Wow, this is a big change @johnjstuart. Thank you for submitting this. I want to thank @victornoel for all fo the valuable review feedback, and for @johnjstuart explaining the thoughts behind each change.

I've proposed some changes inline.

I would like to have this reviewed by @motlin and @nikhilnanivadekar as this is a big and important change to the framework.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/immutable/AbstractImmutableBiMap.java
johnjstuart: Added a `contains` implementation.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableDoubletonMap.java
johnjstuart: Yes, that is my understanding as well.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
johnjstuart: Done 👍

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
johnjstuart: Done 👍

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/MapAdapter.java
johnjstuart: Done 👍

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentMutableHashMap.java
johnjstuart: Done 👍

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
johnjstuart: Done 👍

serialization-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMapSerializationTest.java
johnjstuart: I did change these. I reverted them in the current change, but the serialization tests fail.

donraab:
@johnjstuart @victornoel @motlin @nikhilnanivadekar Where are we with this PR?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
```diff
@@ -778,6 +783,26 @@ public boolean isEmpty()
             return AbstractMutableBiMap.this.isEmpty();
         }
 
+        @Override
+        public K getFirst()
+        {
+            if (AbstractMutableBiMap.this.size() > 0)
+            {
+                return (K) this.toArray()[0];
```
victornoel, line 791: @johnjstuart isn't really no other O(1) way to return this information? Frankly transforming the collection to an array is really a costly operation here… same with getLast below.
A solution would be to use `IterableIterate.getFirst` for example… careful with `IterableIterate.getLast` because it would iterate through the whole collection…

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
```diff
@@ -823,6 +842,12 @@ public boolean containsAll(Collection<?> source)
             return true;
         }
 
+        @Override
+        public void each(Procedure<? super K> procedure)
+        {
+            AbstractMutableBiMap.this.delegate.keySet().each(procedure);
```
victornoel, line 848: @johnjstuart why not directly use `AbstractMutableBiMap.this.forEachKey` here instead of reimplementing something inside this class? It think it would make things simpler and more consistent with the rest of the methods of this class

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
```diff
@@ -398,7 +401,7 @@ public V updateValue(K key, Function0<? extends V> factory, Function<? super V,
     }
 
     @Override
-    public Set<K> keySet()
+    public MutableSet<K> keySet()
     {
         return new KeySet();
```
victornoel, line 406: @johnjstuart do you know why this class does not simply return `delegate.keySet()` here?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/ImmutableDoubletonMap.java
victornoel: Ok, now I understand better, thank you :) it's a bit strange to expose internal classes like this to the public API. Should we introduce a dedicated public interface that extends both `Set` and `ImmutableSet` for example to represent this instead of leaking a class meant to be internal?

victornoel: For example, `ImmutableSet.castToSet()` would return such interface instead of just returning a `Set`!

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
victornoel: @johnjstuart or `IterableIterate.getFirst` as in `getLast`? It's really a pity to convert the iterable to an array just for this

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -2145,6 +2150,26 @@ public boolean isEmpty()
...
+            }
+            return null;
+        }
+
+        @Override
+        public K getLast()
+        {
+            if (UnifiedMap.this.size() > 0)
+            {
+                return IterableIterate.getLast(this);
```
victornoel, line 2168: @johnjstuart this is a bit dangerous in terms of performance because this iterates the whole collection, if we can we should find a better alternative I think

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/immutable/ImmutableTreeMap.java
victornoel: @donraab getLast will iterate over the whole collection, is that ok?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/immutable/AbstractImmutableSortedMap.java
victornoel: now it's clear :)

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/immutable/ImmutableUnifiedMapWithHashingStrategy.java
victornoel: @johnjstuart isn't `ImmutableKeySet` already behaving immutably? Why call `toImmutable` on top of this?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
```diff
@@ -2018,6 +2037,13 @@ private void chainedAddToSet(Object[] chain, UnifiedSetWithHashingStrategy<K> re
                 replace.add(UnifiedMapWithHashingStrategy.this.nonSentinel(cur));
             }
         }
+
+        @Override
+        public ParallelUnsortedSetIterable<K> asParallel(
+                ExecutorService executorService, int batchSize)
+        {
+            return org.eclipse.collections.impl.factory.Sets.adapt(this).asParallel(executorService, batchSize);
```
victornoel, line 2045: @johnjstuart do we really need the fully qualified name of `Sets`?

victornoel (changes requested):
I have added some more comment.

Now I better understand this diamond thing with the internal class. As I said in one of my comment, I wonder if we should introduce a dedicated interface extending `ImmutableSet` and `Set`, and this type would be returned by `ImmutableSet.castToSet()`.

This would avoid leaking internal classes on public methods and this would make the ImmutableSet contract maybe clearer in the code?

donraab:
Hi @johnjstuart, this PR is going on for a year now, and has been quiet for the past 6 months. I'm going to recommend we close it. Maybe we can consider pursuing a less challenging alternative, if and when there is a specific need for it.

PR #918, opened 2020-06-22
Add forEachInBoth to ListIterable.

donraab:

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/ListIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/ListIterable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Goldman Sachs and others.
+ * Copyright (c) 2020 Goldman Sachs and others.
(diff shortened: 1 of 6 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/ListIterable.java
```diff
@@ -303,4 +304,22 @@ default int binarySearch(T key)
...
+                this.forEachWithIndex((each, index) -> procedure.value(each, other.get(index)));
+            }
+            else
+            {
+                throw new IllegalArgumentException("Attempt to call forEachInBoth with two Lists of different sizes :"
+                        + this.size()
+                        + ':'
+                        + other.size());
+            }
+        }
```
motlin, line 323: The null check is uncovered by tests. Isn't it a bug? I can pass in forEachInBoth(null, procedure) and it silently does nothing?

donraab: Yes, I'll add `Objects.requireNonNull(other)`.

PR #922, opened 2020-06-29
Add withInitialCapacity() to primitive List and Set factories

fipro78:
Signed-off-by: Dirk Fauth <dirk.fauth@googlemail.com>

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/factory/list/mutablePrimitiveListFactory.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/factory/list/mutablePrimitiveListFactory.stg
@@ -38,6 +38,15 @@ public interface Mutable<name>ListFactory
     Mutable<name>List with();
 
     /**
+     * Same as {@link #empty()}. but takes in an initial capacity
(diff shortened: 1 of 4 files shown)


nikhilnanivadekar:
On another thought, if we want it in 10.3.0, @fipro78 can you add default implementation which throws

‘throw new UnsupportedOperationException(“Default method to ensure backwards compatibility”)’

I’ll leave it to @fipro78 to decide how to proceed.

fipro78:
You mean the interface should add the method as a default method?

nikhilnanivadekar:
Exactly, because 10.3.0 is a minor version, we cannot break API.

fipro78:
> Exactly, because 10.3.0 is a minor version, we cannot break API.

Sorry, totally missed the breaking change while working in the template. I updated the PR. Hope it is fine now.

donraab:
Hi @fipro78 , the changes look good to me. Please rebase and squash your commits and I will merge. Thanks!

donraab:
Merged. Thank you for the contribution @fipro78!

PR #927, opened 2020-06-30
Make primitive*HashMap.keySet() serializable

mohrezaei:
I believe this is strictly better than not having the keySet be serializable. I reasoned about it like this:

1) 80% use case: I have a PrimitiveSet in my hand. I need to serialize it.
Status before PR: always make a defensive copy, because if that happens to be a `keySet()`, it won't serialize and this might not even show up until production runtime!
Status after PR: no need to make a copy.

2) 19% use case: I have a Primitive*Map.keySet in my hand. I need to serialize it. Why doesn't it work??
Status before PR: because... reasons. Make a copy.
Status after PR: works now!

3) 1% use case: I have a complex serial blob that somewhere inside it has a Primitive*Map and somewhere else inside it has the keySet from that map. I want the keySet to be able to modify the original map after deserialization.
Status before PR: can't be done because the keySet is not serializable.
Status after PR: can't be done because the keySet is detached from the original map after deserialization.

### DIFF
--- a/checkstyle-suppressions.xml
+++ b/checkstyle-suppressions.xml
@@ -116,7 +116,15 @@
     <suppress checks="RedundantModifier" files="ImmutableIntArrayStack.java" />
     <suppress checks="RedundantModifier" files="ImmutableLongArrayStack.java" />
     <suppress checks="RedundantModifier" files="ImmutableShortArrayStack.java" />
-	
+    <suppress checks="RedundantModifier" files="AbstractMutableBooleanKeySet.java" />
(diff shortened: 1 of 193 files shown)


mohrezaei:
The Findbugs results are wrong (again!) because of the serialization replacement strategy. How do I turn that off?

nikhilnanivadekar:
Add it here: https://github.com/eclipse/eclipse-collections/blob/master/findbugs-exclude.xml

mohrezaei:
This is ready for a review. Any takers?

nikhilnanivadekar:
LGTM Thanks Moh!

PR #928, opened 2020-06-30
Add maven wrapper scripts.

motlin:


nikhilnanivadekar:
This looks good to me. A question about the copyright, since we are retaining the Apache Copyright notice, will that work fine with the Eclipse Foundation copyright stuff? I just want to double check that we are not missing anything. Can @motlin , @donraab or @mohrezaei any of you please confirm that it is fine?

donraab:
I think we need to ask the Eclipse Foundation for approval to include this.

nikhilnanivadekar:
Can one of you check with Eclipse Foundation please?

nikhilnanivadekar:
I think the initial setup steps which refer to maven wrapper needs to be fixed ASAP to use a normal maven import. They can be replaced with instructions for wrapper as a part of this PR. It’s a bad experience for folks setting up the project for the first time when the instructions don’t match reality. This has been voiced by multiple folks.

mohrezaei:
My $.02:

Typically, directly including source code requires either the same license or a weaker license. E.g. an Apache licensed project can include other Apache licensed stuff. A GPL project can include BSD code. But a BSD project can't include GPL code, etc.

EPL/EDL (our license) doesn't satisfy the above with respect to Apache. So at least at face value, we can't do this.

However, there is one mitigating factor here: this wrapper will not end up in the final distributed binaries or sources. It's only used for building, and optionally at that. So that potentially provides an escape hatch here: can we include the file in the repo, but somehow prominently note that it's not under EPL/EDL?

The other option is to go back to the owners of these scripts and ask them to re-license or dual-license them under something like BSD.

Honestly, this is way too much work for just a couple of scripts. I build using mvn, not mvnw. Changing the instructions instead is a lot easier.

motlin:
In addition, the maven wrappers have already been folded into the main maven project and are awaiting release in maven 3.7.0.

https://github.com/takari/maven-wrapper

This whole thing is temporary and probably not worth a lot of attention.

motlin:
Hm, I realized that the plugin is being folded into maven, so the instructions may change in the future, but this isn't going away. The point of the scripts is to be able to build the project without installing maven first, not even a future version like 3.7.0.

nikhilnanivadekar:
@motlin should we close this PR?

PR #929, opened 2020-06-30
Fix CheckStyle whitespace settings.

motlin:
Our original CheckStyle settings file was derived from CheckStyle's [own configuration](https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle_checks.xml). That file has evolved separately from ours over the years. Also, I've seen rule violations creep in that I thought were already banned. Rather than dig into what happened, I just synced up some rule settings from "upstream" and fixed the violations that I noticed.

### DIFF
--- a/.github/workflows/checkstyle.yml
+++ b/.github/workflows/checkstyle.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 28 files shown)


nikhilnanivadekar:
@motlin I have approved this PR. Feel free to merge after doing a rebase. Thanks!

PR #934, opened 2020-07-02
Remove implementations of OrderedIterable.toStack() that can use the default implementation instead.

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMap.java
```diff
@@ -1566,12 +1565,6 @@ public void forEachWithIndex(
         this.delegate.forEachWithIndex(fromIndex, toIndex, objectIntProcedure);
     }
 
-    @Override
```
donraab, line 1569: Why remove this implementation?

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/mutable/UnmodifiableSortedSet.java
```diff
@@ -452,12 +451,6 @@ public int compareTo(SortedSetIterable<T> o)
         return this.getSortedSet().compareTo(o);
     }
 
-    @Override
-    public MutableStack<T> toStack()
```
donraab, line 456: Why remove this implementation?

eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/immutable/ImmutableArrayStack.java
```diff
@@ -299,12 +297,6 @@ public boolean containsAllArguments(Object... elements)
         return this.delegate.asReversed().containsAllArguments(elements);
     }
 
-    @Override
-    public MutableStack<T> toStack()
```
donraab, line 303: Do you recall why we did this? Is this safe to replace with the default method?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMap.java
motlin: This is a philosophical thing where we should just decide/agree what we're going to do. Real mutations are always concrete. There are some composite methods that mutate, like MutableMapIterable.putPair() but they always delegate to another concrete method like put(). So it's always safe to delete methods from Unmodifiable wrappers; I've yet to see any counter-example. On the other hand, it would be super-consistent to delegate all read-only methods to the delegate. But we don't do that either. UnmodifiableMutableOrderedMap is missing overrides of all sorts of things.

* Our own mutating methods, like putPair
* JDK mutating methods, like replaceAll
* Our own read-only methods, like countByEach
* JDK read-only methods, like forEach(BiConsumer)

There are something like ~20 methods that you might expect to be overridden and are not. Given the current state of affairs, I think I've made things more consistent with this change, even if we choose to go in a different direction later.

eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/immutable/ImmutableArrayStack.java
motlin: The default implementation lives in StackIterable and is:
```java
    @Override
    default MutableStack<T> toStack()
    {
        return Stacks.mutable.withAllReversed(this);
    }
```

Which is implemented in `MutableStackFactoryImpl`
```java
    @Override
    public <T> MutableStack<T> withAllReversed(Iterable<? extends T> items)
    {
        return ArrayStack.newStackFromTopToBottom(items);
    }
```

So yes it's safe. I don't remember why we wrote this. Sometimes we've created patterns like this as an optimization, but I rarely want to do that. It would have to be for one of the most common iteration patterns and one of the most common collections to make it worth even performance testing.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMap.java
donraab: Thanks for the clarification. I'm good with this.

eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/immutable/ImmutableArrayStack.java
donraab: Great, didn't release it was overriding a default method on `StackIterable`. I'm good with this.

PR #938, opened 2020-07-02
Add default aggregateBy method to RichIterable that takes a target Ma…

donraab:
…p and remove some duplicate implementation code.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
@@ -2422,13 +2422,32 @@ public interface RichIterable<T>
             Function0<? extends V> zeroValueFactory,
             Function2<? super V, ? super T, ? extends V> nonMutatingAggregator)
     {
-        MutableMap<K, V> map = Maps.mutable.empty();
+        return this.aggregateBy(
+                groupBy,
+                zeroValueFactory,
+                nonMutatingAggregator,
+                Maps.mutable.empty());
+    }
+
+    /**
+     * Applies an aggregate function over the iterable grouping results into the target map based on the specific
+     * groupBy function. Aggregate results are allowed to be immutable as they will be replaced in place in the map. A
+     * second function specifies the initial "zero" aggregate value to work with (i.e. Integer.valueOf(0)).
+     *
+     * @since 10.3
+     */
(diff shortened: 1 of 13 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -2422,13 +2422,28 @@ default void appendString(Appendable appendable, String separator)
...
-        MutableMap<K, V> map = Maps.mutable.empty();
+        return this.aggregateBy(
+                groupBy,
+                zeroValueFactory,
+                nonMutatingAggregator,
+                Maps.mutable.empty());
+    }
+
+    /**
+     * @since 10.3
```
prathasirisha, line 2433: One request, can you add a method doc similar to the one above method?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/Bag.java
```diff
@@ -310,4 +313,25 @@ default DoubleSummaryStatistics summarizeDouble(DoubleFunction<? super T> functi
         this.forEachWithOccurrences((each, occurrences) -> target.add(function.valueOf(each, occurrences)));
         return target;
     }
+
+    /**
+     * @since 10.3
+     */
```
prathasirisha, line 319: same comment as above with doc.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
donraab: Done.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/Bag.java
donraab: Done.

donraab:
Hey @motlin can you take a look and confirm I got the generics correct on the new aggregateBy methods with target Maps? Thanks!

motlin:
I believe it should be `R extends MutableMapIterable<K, V>` for maximum flexibility. It would be great to show an example that aggregates into a MutableSortedMap.

Very nice, looks like it cuts about 100 lines of code.

donraab:
Good idea on `MutableMapIterable` @motlin!

donraab:
Ok, changed to use `MutableMapIterable`.

PR #940, opened 2020-07-06
Add OSGi metadata and correct repository build

fipro78:
Signed-off-by: Dirk Fauth <dirk.fauth@googlemail.com>

### DIFF
--- a/eclipse-collections/bnd.bnd
+++ b/eclipse-collections/bnd.bnd
@@ -1,0 +1,7 @@
+package-version=${version;===;${Bundle-Version}}
+
+Automatic-Module-Name: org.eclipse.collections.impl
+Bundle-SymbolicName: org.eclipse.collections.impl
+
+Export-Package: \
+	org.eclipse.collections.impl.*;version="${package-version}"
--- a/p2-feature/org.eclipse.collections.repository/category.xml
+++ b/p2-feature/org.eclipse.collections.repository/category.xml
@@ -1,7 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <site>
-   <feature url="features/org.eclipse.collections.feature.source_10.1.0.qualifier.jar" id="org.eclipse.collections.feature.source" version="10.1.0.qualifier">
-      <category name="org.eclipse.collections.source"/>
+   <feature url="features/org.eclipse.collections.feature_10.3.0.qualifier.jar" id="org.eclipse.collections.feature" version="10.3.0.qualifier">
+      <category name="category_org.eclipse.collections"/>
    </feature>
-   <category-def name="org.eclipse.collections" label="Eclipse Collections"/>
+   <feature url="features/org.eclipse.collections.feature.source_10.3.0.qualifier.jar" id="org.eclipse.collections.feature.source" version="10.3.0.qualifier">
+      <category name="category_org.eclipse.collections"/>
+   </feature>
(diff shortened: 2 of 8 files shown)


fipro78:
Don't merge, there is an issue with the packaging I need to solve first! Just noticed by looking into the JARs. Please wait for the update.

eclipse-collections/bnd.bnd
```diff
@@ -0,0 +1,11 @@
...
+
+Automatic-Module-Name: org.eclipse.collections.impl
+Bundle-SymbolicName: org.eclipse.collections.impl
+Export-Package: \
+ !about.html,!about_files, \
+ *.internal*;x-internal:=true;version="${package-version}", \
+ *;version="${package-version}"
+
+Import-Package: \
+	org.eclipse.collections.impl.*
```
motlin, line 11: I would expect impl to import api?

p2-feature/org.eclipse.collections.repository/category.xml
```diff
@@ -1,7 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <site>
-   <feature url="features/org.eclipse.collections.feature.source_10.1.0.qualifier.jar" id="org.eclipse.collections.feature.source" version="10.1.0.qualifier">
-      <category name="org.eclipse.collections.source"/>
+   <feature url="features/org.eclipse.collections.feature_10.3.0.qualifier.jar" id="org.eclipse.collections.feature" version="10.3.0.qualifier">
```
motlin, line 3: @nikhilnanivadekar did we miss this during 10.2.0?

pom.xml
```diff
@@ -546,6 +546,19 @@
                         </lifecycleMappingMetadata>
                     </configuration>
                 </plugin>
+
+                <plugin>
+                    <groupId>biz.aQute.bnd</groupId>
+                    <artifactId>bnd-maven-plugin</artifactId>
+					<version>5.1.1</version>
+					<executions>
```
motlin, line 554: Mix of tabs and spaces here

p2-feature/org.eclipse.collections.repository/category.xml
nikhilnanivadekar: yes possibly 😞 . I wish there was an automated way to do this during every release.

fipro78:
@motlin 
Anything I have to do here to make the patch work?

motlin:
> @motlin
> Anything I have to do here to make the patch work?

From what I understand about OSGi, I figured the impl bundle would need to declare some sort of import on api. No? That's my only open question.

fipro78:
> > @motlin
> > Anything I have to do here to make the patch work?
> 
> From what I understand about OSGi, I figured the impl bundle would need to declare some sort of import on api. No? That's my only open question.

The OSGi metadata contained in the MANIFEST.MF file needs to specify the required Import-Package statements on api. But that is generated by the bnd-maven-plugin by default. We only need to add the Import-Package instruction in the .bnd file if we want to restrict the imports or add some information like optional flags.

You can inspect the generated jar and have a look at the MANIFEST.MF to see what the bnd-maven-plugin generated.

motlin:
Great. I think this ought to be rebased on top of master and then it's good to go.

nikhilnanivadekar:
@fipro78 I will go ahead and merge this and create another milestone release to test out. Will ping back on this PR once milestone is pushed to p2. Thanks a ton for helping us fix these issues.

PR #945, opened 2020-07-11
Remove travis build configuration.

motlin:


prathasirisha:
Late comment - There are some references to travis builds in documentation and badges that would require updates I believe.

motlin:
Thanks for pointing this out. I raised #948 to address it.

PR #958, opened 2020-07-25
Implement MapIterable.getOrDefault()

nikhilnanivadekar:


nikhilnanivadekar:
@donraab I dont think the overrides are necessary. The ambiguous implementation issue is actually caught by the compiler and I _had_ to override the necessary interfaces. For example: `MutableMapIterable`, `ConcurrentMutableMap`, `AbstractBiMap`, `AbstractMapIterable`, 

Screenshot below:
image

nikhilnanivadekar:
Rebased on to upstream/master to make it current.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/AbstractMapIterable.java
```diff
@@ -55,6 +55,12 @@ protected boolean isAbsent(V result, K key)
         return result == null && !this.containsKey(key);
     }
 
+    @Override
```
donraab, line 58: Is this necessary?

nikhilnanivadekar: Yes for the Immutable side

nikhilnanivadekar: For transparency the default methods and implementations other than MapIterable are to fix compiler complaints, and I have put them at the highest place.

donraab: Not sure I understand. `ImmutableMap` extends `ImmutableMapIterable` which extends `MapIterable` so all `ImmutableMap` implementations should get the `default` implementation and the override in `AbstractMapIterable` should be unnecessary right?

anupyadavrpvv:
@nikhilnanivadekar  thanks for sharing the solution link.
when I was working on it all the test classes and interfaces in which you have implemented the getOrDefault() method was giving errors and was showing to implement this method in all these interfaces which I don't understand at that time.
Anyway, Thanks

nikhilnanivadekar:
No worries. Did you understand what the issue was? If not, I have explained it here: https://github.com/eclipse/eclipse-collections/pull/958#issuecomment-663919606

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/AbstractMapIterable.java
nikhilnanivadekar: @donraab It is necessary because `AbstractImmutableMap implements ImmutableMap, Map`. So the immutable side gets the ambiguous method issue. Screenshot below:
image

nikhilnanivadekar:
@donraab thanks for the approval. I have reviewed and commented on the review, that the method is not redundant as is needed. Merging this PR.

PR #960, opened 2020-07-28

motlin:


nikhilnanivadekar:
Does something need to be updated in inspections profiles to ensure these issues don’t come back?

motlin:
Yes, there's an inspection I want to enable called Statement Lambda may be Expression Lambda, in profile 3.

- The churn in the profile settings files seems related to the IntelliJ plugins one has installed. I have a lot of unrelated plugins installed that I'll disable to reduce churn.
- Anything out-of-date in the *.iml files seems like it can partially break TeamCity, and I need to double check these files are ok. Whenever these are broken for a while, I wind up fixing violations in batches.

nikhilnanivadekar:
Should we do this clean up after 10.3.0 is pushed? I am about start the final phase of the release. Do you have any PRs for functionality? I would want to first focus on functionalities that are pending.

motlin:
I think my PRs are all waiting for a major version then

donraab:
@motlin Do you want to update this now, and @nikhilnanivadekar are you ok with us merging this now?

donraab:
@motlin @nikhilnanivadekar Reviewing old PRs. Can we merge or close this?

nikhilnanivadekar:
Dont quite remember where we left this one off. But if it is not breaking,
then I think we should merge it.

motlin:
I refreshed it and got it ready to land

prathasirisha (approved):

PR #967, opened 2020-08-13
Test improvement: removed conditional test logic (test smell)

eas5:
This is a test improvement.

**Problem:**
Conditions within the test method will alter the behavior of the test and its expected output and would lead to situations where (I) the test fails to detect defects in the production method since test statements were not executed as a condition was not met, or (II) the test passes without executing the test steps.

**Solution:**
In this test class, we changed the conditional logic (which tests for a precondition) for an assumption, which will ignore the test execution if not met.

**Result (Sample):**
_Before:_
```
@Test
public void removeNullFromKeySet()
{
    if (this.newMap() instanceof ConcurrentMap || this.newMap() instanceof SortedMap)
    {
        return;
    }
    ...
}
```
_After:_
```
@Test
public void removeNullFromKeySet()
{
    Assume.assumeFalse(this.newMap() instanceof ConcurrentMap || this.newMap() instanceof SortedMap);
    ...
}
```

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
@@ -41,6 +41,7 @@ import org.eclipse.collections.impl.tuple.ImmutableEntry;
 import org.eclipse.collections.impl.tuple.Tuples;
 import org.eclipse.collections.impl.utility.Iterate;
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.Test;
 
 import static org.eclipse.collections.impl.factory.Iterables.iMap;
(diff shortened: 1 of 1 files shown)


motlin:
In my experience, failed assumptions are noisy. They show up in the IDE and maybe also in console output, as if they are TODOs (almost like ignored tests).

eas5:
A failed assumption indeed causes test execution to be ignored and may be shown in the IDE and console output. 
The idea is not having a passed test if this test was not executed, which is the current behavior. But if the current behavior is indeed the desired one, then verifying preconditions with assumptions may not be the case.

An alternative solution, but also noisy, is to use `assumeFalse(String message, boolean b)` passing, a "Test no applicable to this class" message to the AssumptionViolatedException. This way the execution log would not look like a TODO list. But again, this noisy behavior may not be the desired one.

donraab:
Interesting discussion. If we put assumptions such as these @motlin , they can act as reminders to override the behavior in the subclass tests for `ConcurrentMap` and `SortMap`. Once the overrides are in, the assumptions should never fail, correct?

Issue #968, opened 2020-08-14
Create Code Coverage Report


GracianoR:
A code coverage report is a nice addition to Eclipse Collections. It allows the community to find areas that need improvement and it's a good source of future work.
The report can be added in a similar way to the Mutation Test. We can leverage GitHub Actions to trigger based on a push to master (or another branch) or cron based (I don't think that's the best approach since tests may be executed on the same code).

GracianoR:
Can I take this one?

GracianoR:
How should we trigger the GH action? On push to master? Cron?

PR #971, opened 2020-08-14
Add GitHub action to generate code report #968

GracianoR:
Signed-off-by: GracianoR <rodrigo.graciano@bnymellon.com>

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -1,0 +1,35 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: Coverage Report
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'
+
+jobs:
+  applications:
+    name: Coverage Report ${{ matrix.java-version }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        java-version: [8]
+
+    runs-on: ${{ matrix.os }}
+
+    env:
+      JAVA_VERSION: ${{ matrix.java-version }}
+      
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set JDK ${{ matrix.java-version }}
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java-version }}
+      - name: Generate Coverage Report
+        run: mvn clean verify
+      - name: Archive Jacoco reports
+        uses: actions/upload-artifact@v1
+        with:
+          name: jacoco-reports
+          path: test-coverage/target/site/jacoco-aggregate
--- a/pom.xml
+++ b/pom.xml
@@ -442,9 +442,15 @@
                     <version>${jacoco.plugin.version}</version>
                     <executions>
                         <execution>
-                            <id>default-prepare-agent</id>
                             <goals>
                                 <goal>prepare-agent</goal>
+                            </goals>
+                        </execution>
+                        <execution>
+                            <id>report</id>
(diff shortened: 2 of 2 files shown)


mohrezaei:
I know there were some questions in the Issue, but now that there is a PR we can move the discussion here.

I would be in favor of doing this on every PR and integrating with [Coveralls](https://coveralls.io).

Output will look like [this](https://github.com/apache/commons-math/pull/156).

You can also look at that [project](https://github.com/apache/commons-math) to see how it's setup in pom.xml using this [plugin](https://github.com/trautonen/coveralls-maven-plugin).

GracianoR:
Thank you @mohrezaei. I also tried before using Codecov but I believe both of them would require creating an account at Coveralls/Codecov. This is an option as well and has the advantage of the badge ready to be used as well.

mohrezaei:
Creating an account shouldn't be a problem (I can do it for the project). I prefer coveralls based on the legalize.

mohrezaei:
Looks like coveralls/maven plugin [doesn't work](https://github.com/trautonen/coveralls-maven-plugin/issues/136) with github actions, so I guess we can forget all that ☹️

donraab:
I think we should just get the report generating to start so we can start identifying any coverage paths developers can work on. The build for that can stay scheduled to start. Eventually it would be great to have something like you linked to above with Coveralls and to build code coverage on every PR. I suspect Coveralls will eventually have support for GitHub Actions.

Thoughts?

.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
```
prathasirisha, line 14: Can remove extra blank lines? Line 12, Line 14.

.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 1.8
```
prathasirisha, line 17: In other GitHub action yaml files, we have java-version configured as a variable that gets used in line number 17 and 20. 
Though what you have works, can you make it consistent with other yaml files? This is just a  personal preference.

.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,27 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: Coverage Report
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'
```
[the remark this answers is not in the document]
GracianoR: I should have mentioned that

.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set JDK ${{ matrix.java-version }}
```
prathasirisha, line 15: @GracianoR I could be wrong, but I expected a variable definition like this for this to work. 
`matrix:`    
        `os: [ubuntu-latest]`  
        `java-version: [8]`   
See example [here](https://github.com/eclipse/eclipse-collections/blob/master/.github/workflows/acceptance.yml) - line 12-14

GracianoR:
@prathasirisha, changes were done and ready to be reviewed

donraab:
Thanks for the contribution @GracianoR and thank you for reviewing the PR @prathasirisha !

PR #973, opened 2020-08-16
Create JDK-EA builds and fix Compiler errors due to new default method in JDK-15

nikhilnanivadekar:
There will be failures in EBR Maven plugin. But that needs to be fixed in EBR Maven plugin. That said, the main library will be tested correctly.

### DIFF
--- a/.github/workflows/earlyaccess-unit.yml
+++ b/.github/workflows/earlyaccess-unit.yml
@@ -1,0 +1,39 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: JDK-EA Unit Tests
+
+on: [push, pull_request]
+
+jobs:
+  applications:
+    name: Unit Test EA
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        java-version: [15, 16]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Git Checkout
+        uses: actions/checkout@v2
+      - name: 'Download JDK'
+        uses: sormuras/download-jdk@v1
+        id: download-jdk
+        with:
+          feature: ${{ matrix.java-version }}
+      - name: 'Set up JDK'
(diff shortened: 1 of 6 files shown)


.github/workflows/earlyaccess-unit.yml
```diff
@@ -0,0 +1,39 @@
...
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ steps.download-jdk.outputs.version }}
+          jdkFile: "${{ steps.download-jdk.outputs.file }}"
+      - name: Set Maven Wrapper
+        run: mvn -N io.takari:maven:0.7.7:wrapper -Dmaven=3.6.3
+      - name: 'JDK Version'
+        run: java --version
+      - name: Unit-Test
+        run: ./mvnw verify --no-transfer-progress
```
mohrezaei, line 37: create a new profile in the top level pom.xml that excludes the p2 stuff (see this [example](https://stackoverflow.com/questions/4448599/maven-modules-optional-in-pom-xml)). Call it `ea-profile`. Then specify it on this line with `-P ea-profile`.

nikhilnanivadekar: We need to test the p2 stuff as well. Also the profile part is not working. If you look at earlyaccess-javadoc.yml in this PR, I specifically give the modules to build and still p2 repo is getting built. That is the whole deal with EBR plugin and I do not know how to solve that. I am working on fixing the problem with EBR plugin itself. PR raised there: https://github.com/eclipse/ebr/pull/40

nikhilnanivadekar:
@mohrezaei @donraab is this fine to merge?
The p2 failures are going to be painful, but we can solve that as step 2. I would rather merge this and cherry-pick the compiler error fixes and release 10.4.0.
JDK-15- Final RC will be released on 20th August, so would like to close out the release before that.
Thoughts?

nikhilnanivadekar:
After upgrading the EBR plugin to 1.3.0-SNAPSHOT the builds are working fine. My fix was merged here: https://github.com/eclipse/ebr/pull/40

Can this PR be merged?

PR #979, opened 2020-08-20
Prepare 10.4.0 release

nikhilnanivadekar:


nikhilnanivadekar:
@donraab are the release notes ok? I would like them to show that folks need to upgrade to 10.4.0 to use it with JDK-15. I hope it’s mentioned clearly enough. Suggestions welcome to make it better.

PR #980, opened 2020-08-20
Remove unsupported Javadoc option

nikhilnanivadekar:
The reason to remove this is because we need to run deploy builds with JDK 8.


nikhilnanivadekar:
Thanks @prathasirisha

PR #981, opened 2020-08-20
Deprecate ImmutableArrayStack. Closes #976

### LINKED ISSUE #976: Deprecate ImmutableArrayStack
Was looking around in the jacoco-report found [here](https://github.com/eclipse/eclipse-collections/actions/runs/210283622) and noticed that **ImmutableArrayStack** under *org.eclipse.collections.impl.stack.immutable* only has 3% code coverage.

This would be a great first ticket for OSS contribution in my opinion because test coverage is always welcome. 😁 

EDIT:
I don't seem to be able to add Labels to issues so if anyone can add "good first issue" and "help wanted" that would be greatly appreciated!

jjabamani-bny:
Deprecate ImmutableArrayStack. Closes #976
Signed-off-by: John Jabamani <john.jabamani@bnymellon.com>

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/immutable/ImmutableArrayStack.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/immutable/ImmutableArrayStack.java
@@ -112,7 +112,11 @@ import org.eclipse.collections.impl.utility.LazyIterate;
 
 /**
  * The immutable equivalent of ArrayStack. Wraps a FastList.
+ *
+ * @deprecated Replaced by {@link ImmutableNotEmptyStack}.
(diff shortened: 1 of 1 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/stack/immutable/ImmutableArrayStack.java
```diff
@@ -112,7 +112,11 @@
 
 /**
  * The immutable equivalent of ArrayStack. Wraps a FastList.
+ *
+ * Deprecated this class because it was replaced by singly-linked implementation of immutable stack.
+ * https://github.com/eclipse/eclipse-collections/pull/767
```
nikhilnanivadekar, line 117: No need to have a PR here. You should have a reference of what it was replaced by as a link to code.
Take a look at other `deprecate` examples in the project.

PR #993, opened 2020-09-02
Disable mutation tests from running on every fork

prathasirisha:

Found a couple of helpful resources such as [this](https://github.community/t/do-not-run-cron-workflows-in-forks/17636) outlining the solution.


nikhilnanivadekar:
we should do the same for test coverage as well.

nikhilnanivadekar (approved):

PR #998, opened 2020-09-10
Fix CollectIterable detect methods. Closes #994

### LINKED ISSUE #994: Function being called twice on LazyIterable findAny()
Hi,

I notice that with the following code, there is a duplicated function call.

```java
LazyIterable.collect(functions, func -> func.valueOf(someArg)).select(Objects::nonNull).getAny();
```
let say `functions` is an array of functions of this type:
```java
(someArg: String) -> {
  if ("CAR".equals(someArg)) return someArg;
 return null;
};
```

I notice that my function got called twice for the found value

This is not behaving like what I expected when I use `stream API`, for example:
```java
functions.stream().map(func -> func.valueOf(someArg)).filter(Objects::nonNull).findFirst().orElse(null);
```
here my function will only be called once for the found value.

Perhaps this has something to do with this [line](https://github.com/eclipse/eclipse-collections/blob/master/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/CollectIterable.java#L116) in the codebase?

prathasirisha:

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/CollectIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/lazy/CollectIterable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Goldman Sachs.
+ * Copyright (c) 2020 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


donraab (approved):
LGTM. Thanks for the contribution @prathasirisha! I will merge as soon as the builds pass.

PR #1003, opened 2020-09-15
Add jackson.md to README.md.

donraab:

### DIFF
--- a/docs/jackson.md
+++ b/docs/jackson.md
@@ -15,7 +15,7 @@ In this article we will cover the basics of Eclipse Collections, Jackson and tha
 
 ## Eclipse Collections
 
-As explained above, Eclipse Collections replaces many of the familiar container types from the standard library with new and improved versions. For this article, we will look at three variants of the `List` type:
+Eclipse Collections replaces many of the familiar container types from the standard library with new and improved versions. For this article, we will look at three variants of the `List` type:
 
 ```java
 import org.eclipse.collections.api.list.ImmutableList;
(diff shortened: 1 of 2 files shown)


docs/jackson.md
```diff
@@ -55,7 +55,7 @@ Or with Gradle:
 compile 'org.eclipse.collections:eclipse-collections:10.4.0'
```
nikhilnanivadekar, line 55: this should be `implementation`

docs/jackson.md
```diff
@@ -55,7 +55,7 @@ Or with Gradle:
...
 ```
 
-9.2.0 is the current Eclipse Collections version at time of writing. Check out [the maven repo](https://search.maven.org/search?q=g:org.eclipse.collections%20AND%20a:eclipse-collections&core=gav) for newer versions.
+10.4.0 is the current Eclipse Collections version at time of writing. Check out [the maven repo](https://search.maven.org/search?q=g:org.eclipse.collections%20AND%20a:eclipse-collections&core=gav) for newer versions.
```
nikhilnanivadekar, line 58: I would suggest deleting the version section altogether. It will make it 1 more thing to update for every release. Currently, there are too many things to remember to update, so I would not want to add more.

nikhilnanivadekar: Same comment about the Jackson version. We should just have a since Jackson version `so and so` Eclipse Collections is supported. It will be a nightmare to keep the doc synced with the Jackson versions.

nikhilnanivadekar (changes requested):
Requested changes for jackson.md

docs/jackson.md
donraab: I removed the version section for Eclipse Collections and referenced the GitHub releases for Maven and Gradle Coordinates. @yawkat @nikhilnanivadekar can you take a look and provide any additional feedback.

yawkat: Looks fine to me.

donraab:
Thanks @nikhilnanivadekar I will add a link

donraab:
@nikhilnanivadekar @yawkat I added Maven Central links for `jackson-databind` and `jackson-datatype-eclipse-collections`.

nikhilnanivadekar (approved):

Issue #988, opened 2020-08-28
Test Coverage: AtomicCountProcedure


johnjstuart:
`AtomicCountProcedure` currently has 0% test coverage. Add unit tests for this class

hongliu2020:
@donraab , can you please assign this issue to me? Thanks! Hong.

donraab:
Hi @hongliu2020, I have assigned the issue to you. Thanks for volunteering!

PR #1012, opened 2020-10-04
Issue# 988: Test Coverage for AtomicCountProcedure

hongliu2020:
@donraab 
Hi, Don
Can you please review the pull request for issue #988? 
A few questions: 
(1) I don't see mockito in test project as dependency. I'm asking because my first impression is to mock a Predicate to test AtomicCountPrecedure, and I don't see mockito used anywhere in eclipse collection test cases. 
(2) Is it a standard to add comments for unit test cases in the project?

Thanks!
Hong

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/block/procedure/AtomicCountProcedureTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/block/procedure/AtomicCountProcedureTest.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2020 The Bank of New York Mellon and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompany this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.impl.block.procedure;
+
+import org.eclipse.collections.impl.list.Interval;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class AtomicCountProcedureTest
+{
+    private static final int STRING_LENGTH = 5;
+    private static final int ZERO = 0;
+
+    @Test
+    public void getCount()
+    {
+        AtomicCountProcedure<String> atomicCountProcedure = new AtomicCountProcedure<>(each -> STRING_LENGTH < each.length());
+
+        atomicCountProcedure.value("word");
+        assertEquals(0,  atomicCountProcedure.getCount());
+
+        atomicCountProcedure.value("america");
+        assertEquals(1, atomicCountProcedure.getCount());
+    }
+
+    @Test(expected = ArithmeticException.class)
+    public void predicateException()
+    {
+        AtomicCountProcedure<String> atomicCountProcedure = new AtomicCountProcedure<>(each -> STRING_LENGTH < each.length() / ZERO);
+
+        atomicCountProcedure.value("word");
+    }
+
+    @Test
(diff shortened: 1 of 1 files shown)


unit-tests/src/test/java/org/eclipse/collections/impl/block/procedure/AtomicCountProcedureTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class AtomicCountProcedureTest
+{
+    private static final int STRING_LENGTH = 5;
+    private static final int ZERO = STRING_LENGTH - 5;
+
+    @Test
+    public void getCount()
+    {
+        AtomicCountProcedure<String> atomicCountProcedure = new AtomicCountProcedure<>((Predicate<String>) each ->
```
donraab, line 27: Hi @hongliu2020, I would expect the cast to predicate here to be unnecessary. Can you try removing it?

unit-tests/src/test/java/org/eclipse/collections/impl/block/procedure/AtomicCountProcedureTest.java
```diff
@@ -0,0 +1,45 @@
...
+        assertEquals(0,  atomicCountProcedure.getCount());
+
+        atomicCountProcedure.value("america");
+        assertEquals(1, atomicCountProcedure.getCount());
+    }
+
+    @Test(expected = ArithmeticException.class)
+    public void predicateException()
+    {
+        AtomicCountProcedure<String> atomicCountProcedure = new AtomicCountProcedure<>((Predicate<String>) each ->
```
donraab, line 40: Same comment as above on the cast.

unit-tests/src/test/java/org/eclipse/collections/impl/block/procedure/AtomicCountProcedureTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.eclipse.collections.api.block.predicate.Predicate;
+import org.junit.Assert;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class AtomicCountProcedureTest
+{
+    private static final int STRING_LENGTH = 5;
+    private static final int ZERO = STRING_LENGTH - 5;
```
donraab, line 22: I would recommend inlining 0 as it will be more self explanatory.

donraab (changes requested):
HI @hongliu2020, thank you for the contribution! I've added a few comments inline. On your question of Mockito, we have never had a need to mock anything like `Predicate`. It is easy enough to create inner classes or anonymous implementations to validate that the different function types are visited. In terms of commenting in tests, it is completely up to the developer, but generally we like the tests to be self-explanatory. Other than that, this test looks good to me.

donraab:
Hi @hongliu2020, thank you for the changes. These look good.

There is one test I think we are missing for `AtomicCountProcedure` which is a test for parallelism. If you create a List with 50,000 `Integers` and filter out all the evens using an `Executor` with 5 threads in a fixed thread pool, you should consistently get an answer of 25,000 from the count. This same code should fail using a `CountProcedure` which does not use an `AtomicInteger`.

hongliu2020:
Hi, @donraab,  it is very good point, I will add test for parallelism accordingly.

hongliu2020:
@donraab , Can you please review again? 
Thanks! 
Hong

donraab (changes requested):
Hi @hongliu2020, thank you for the contribution! The additions in the test look good, but it looks like some file deletions were included unintentionally in the commit. Please address and I will approve and merge. Thanks!

hongliu2020:
@donraab , I added .idea files back. Those 4 files were deleted automatically and updated with different versions by Intellij on my local. I'm wondering if the files under .idea/libraries should be specified in .gitignore? Thanks! Hong

donraab (approved):
LGTM. Thanks for the contribution @hongliu2020!

Please rebase your PR and I will merge.

PR #1016, opened 2020-10-18
Implement containsAny and containsNone on primitive iterables. Closes #1007.

### LINKED ISSUE #1007: Implement containsAny and containsNone on primitive iterables.

rinatigati:
Add default implementations of containsAny and containsNone on primitive iterables. This closes issue #1007.


nikhilnanivadekar:
I think there needs to be some special handling for doubles, floats and ints and longs. The way this is implemented downcasting will return false.

nikhilnanivadekar:
To add further clarification about downcasting, look at the code below:

```
LongList longList = LongLists.mutable.with(1L, 2L, 3L, 4L);
IntList intList = IntLists.mutable.with(1, 2, 3, 4);

assertTrue(intList.allSatisfy(longList::contains);
```

This code works. But because in the API the input is restricted to the same type we cannot use `LongList` and `IntList` interchangeably even where it is semantically correct. 


FWIW `assertTrue(longList.allSatisfy(intList::contains))` does not compile, and that is correct. I think we need to handle the above case as well.

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
```diff
@@ -79,17 +79,75 @@ public interface <name>Iterable extends PrimitiveIterable
...
+
+    /**
+     * Returns true if none of the values specified in the source array are contained
+     * in the <name>Iterable, and false if they are.
+     *
+     * @since 11.0
+     */
+    default boolean containsNone(<type>... source)
+    {
+        for (<type> item : source)
```
nikhilnanivadekar, line 130: change this to an indexed for loop

nikhilnanivadekar (changes requested):

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
mohrezaei: why? Arrays are the only types where the non-indexed for-loop has no overhead.

nikhilnanivadekar: I was thinking to keep it consistent in the entire codebase. Throughout we tend to prefer indexed for loops

mohrezaei:
The existing `containsAll` implementation is very badly implemented (it's `O(n^2)`). Should we fix that first before we add copy-paste analogous methods that further add to the bad implementation?

nikhilnanivadekar:
Makes sense to me. Especially since it’s a copy paste

mohrezaei:
`containsAll` algorithm:
if (this.size() <= 32 || iterable.size() < 4) use current algorithm
else create a set from `this` (if `this` is not a set already), then use that set in the algorithm

`containsAny` algorithm: use the fact that `a.containsAny(b) == b.containsAny(a)`
if (this.size() <= 32 || iterable.size() <=32) use current algorithm with the smaller one in the inner loop.
else, create a set from the smaller one (if not a set) and use it in the inner loop of algorithm.

`containsNone` same as `containsAny`

rinatigati:
Sounds good to me! I'll update the `containsAll`, `containsAny`, and `containsNone` implementations with the optimizations described above.

This might be me being dense, but what's the significance for using 32 in this scenario @mohrezaei? Curious since I'm sure there is something to it. 😅

mohrezaei:
32 is just a small number roughly corresponding to what should fit in an L1 cache of most CPUs. Because iterating is so fast over those, the loop can almost be considered `O(1)`.

We should probably write a benchmark to find the value empirically.

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
```diff
@@ -127,6 +127,140 @@ public interface <name>Iterable extends PrimitiveIterable
...
 
+    /**
+     * Returns true if any of the values specified in the source array are contained
+     * in the <name>Iterable, and false if they are not.
+     *
+     * @since 11.0
+     */
+    default boolean containsAny(<type>... source)
+    {
+        if (this.size() \<= 32 || source.length \<= 32)
```
mohrezaei, line 138: can be simplified to:
```java
<name>Iterable inside = this;
if (this.size() \> 32 && source.length \> 32 && !this instanceof <name>Set)
{
    inside = this.toSet();
}
for (<type> item : source)
{
        if (inside.contains(item))
        {
            return true;
        }
}
return false;
```
Other 3 methods can be simplified similarly.

mohrezaei (changes requested):

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
rinatigati: Good point! I simplified all 4 methods and updated the PR.

rinatigati:
Updated the PR with the optimizations for containsAny and containsNone. 

When I was adding more tests I noticed a bug with noneSatisfy on the ImmutableBooleanEmptySet. Both noneSatisfy and anySatisfy for that class returned false. I changed that and included it as a separate commit.

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
```diff
@@ -127,6 +127,96 @@ public interface <name>Iterable extends PrimitiveIterable
...
+    default boolean containsAny(<name>Iterable source)
+    {
+        <name>Iterable outside = this;
+        <name>Iterable inside = source;
+        if (this.size() \< source.size())
+        {
+            outside = source;
+            inside = this;
+        }
+        if (outside.size() > 32 && inside.size() > 32 && !(inside instanceof <name>Set))
```
mohrezaei, line 168: at this point `inside` is guaranteed to be the smaller of the two. Therefore, `inside.size() > 32` implies `outside.size() > 32`. We can drop `outside.size() > 32`. Same for `noneSatisfy`

mohrezaei (changes requested):

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
rinatigati: Updated. 😄

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
```diff
@@ -127,6 +127,96 @@ public interface <name>Iterable extends PrimitiveIterable
...
+        <name>Iterable inside = source;
+        if (this.size() \< source.size())
+        {
+            outside = source;
+            inside = this;
+        }
+        if (inside.size() > 32 && !(inside instanceof <name>Set))
+        {
+            inside = inside.toSet();
+        }
```
mohrezaei, line 171: I've been thinking about two special cases in the `containsAny/None` situation:
case 1:
this: large list, say 1M entries
other: large set, larger than this, say 2 M entries
Our logic will create a 1M entry set to do the operation. This is not necessary. 
case 2:
this: large set, say 1M entries
other: large set, larger than this, say 2 M entries
We won't create a set in this case, but the loop order is wrong: this should be on the outside.

Both can be fixed by changing this code to:
```java
        if (outside instance of <name>Set)
        {
            <name>Iterable tmp = outside;
            outside = inside;
            inside = tmp;
        }
        else if (inside.size() > 32 && !(inside instanceof <name>Set))
        {
            inside = inside.toSet();
        }
```

mohrezaei (changes requested):

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
donraab: Hi @rinatigati, any questions or updates on the suggestion from @mohrezaei above?

rinatigati: @donraab Nope! Sounded good to me. I updated the PR with @mohrezaei's suggestions.

Looks like there were artifact transfer failures/connection problems with a couple of the checks. Will have to re-run those somehow.

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
rinatigati: Changed to an indexed for loop.

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
```diff
@@ -127,6 +127,108 @@ public interface <name>Iterable extends PrimitiveIterable
...
 
+    /**
+     * Returns true if any of the values specified in the source array are contained
+     * in the <name>Iterable, and false if they are not.
+     *
+     * @since 11.0
+     */
+    default boolean containsAny(<type>... source)
+    {
+        <name>Iterable inside = this;
```
mohrezaei, line 138: Sorry, I just re-read this and I think it needs some adjustment, specifically, when `source.length == 32` (or bigger than maybe 2, less than 33). In those cases, the logic should essentially be like `Iterable` case. Can we write an acceptance test with source.length==32, this.size() == 1M, not a set (as a rough benchmark)? I think the following would do better:
```java
if (source.length \< 3 || this instanceof <name>Set) || this.size() \< 32) 
{
        for (int i = 0; i \< source.length; i++)
        {
            if (this.contains(source[i]))
            {
                return true;
            }
        }
        return false;
}
return this.containsAny(Arrays.asList(source)); // we have to create some garbage in this case, so let's not worry about the small wrapper
```

mohrezaei (changes requested):

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
rinatigati: No worries!

Added acceptance tests for containsAny and containsNone with source.length == 32 and this.size() == 1M. Runtime of those on my Macbook is ~50ms with the current implementation.

Tried changing it to what you provided above but containsAny can't take the output of Arrays.asList() and gives me the following error:

```
error: no suitable method found for containsAny(List<int[]>)
    method IntIterable.containsAny(int...) is not applicable
      (varargs mismatch; no instance(s) of type variable(s) T exist so that List<T> conforms to int)
    method IntIterable.containsAny(IntIterable) is not applicable
      (argument mismatch; no instance(s) of type variable(s) T exist so that List<T> conforms to IntIterable)
  where T is a type-variable:
    T extends Object declared in method <T>asList(T...)
```

I also can't use either IntSets or IntLists to make a primitive iterable since these are default methods in org.eclipse.collections.api and those are in org.eclipse.collections.impl. 🤷 Not sure where to go from here but I did push the acceptance tests I added.

nikhilnanivadekar: You can’t use the factories in API because of the packages and primitives are not onboarded to classloaders like the object side. You cannot use Arrays.asList() as you need to remain in primitive collections. 
An option which you have is to move this optimization down to implementations where you can use the factories and keep the unoptimized default method in the interface.

mohrezaei: before we go down that route, let's figure out the performance impact. Run the proposed new code and compare the timing with the older one.

donraab:
@rinatigati How is this progressing? Let me know if you need anything. Thanks!

donraab:
@rinatigati I re-ran the two failed builds and now all of the checks are passing. @mohrezaei, is there any other outstanding feedback to address?

mohrezaei:
@donraab yes, this [comment/thread](https://github.com/eclipse/eclipse-collections/pull/1016#discussion_r533853248) is still outstanding.

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
rinatigati: The older (current) code completes containsAny and containsNone with source of size 32 and the Iterable of size 1M in 50-60ms on my macbook.

To try out the proposed new code I overrode `containsAny(int... source)` and `containsNone(int.. source)` on AbstractIntIterable with it. I did change `Arrays.asList(source)` to `IntLists.mutable.with(source)` since I couldn't use Arrays.asList there.

With the same acceptance test (size.length == 32 and this.size() == 1M), the proposed new code completes containsAny and containsNone in 100-120ms on my macbook. It actually seems slower.

mohrezaei: ok, great. We're done then.

donraab:
@rinatigati Please rebase your PR. @mohrezaei is this ready to approve and merge otherwise?

mohrezaei (approved):

rinatigati:
@donraab Thanks for the heads up. Rebased my fork. The checks are running now.

Issue #310, opened 2017-08-10
Implement classic set operations on primitive sets.


donraab:
Implement the following set operations on primitive sets, to be symmetric with the same forms on SetIterable:

```
<type>Set union(<type>Set);
<type>Set intersect(<type>Set);
<type>Set difference(<type>Set);
<type>Set symmetricDifference(<type>Set);
boolean isSubSetOf(<type>Set);
boolean isProperSubsetOf(<type>Set);
LazyIterable<<type><type>Pair> cartesianProduct(<type>Set);
```

digsb:
@donraab Can I work on this issue?

donraab:
@digsb No one else is working on it currently that I am aware so please feel free if you'd like to work on it.  Thank you!

PR #1017, opened 2020-10-21
Implement intersect operation on primitive sets. Partially addresses issues #310. 

prathasirisha:

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/set/immutablePrimitiveSet.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/set/immutablePrimitiveSet.stg
@@ -59,6 +59,25 @@ public interface Immutable<name>Set extends Immutable<name>Collection, <name>Set
         }
     }
 
+    /**
+     * Returns the set of all objects that are members of both {@code this} and {@code set}. The intersection of
+     * [1, 2, 3] and [2, 3, 4] is the set [2, 3].
+     *
+     * @since 11.0.
+     */
+    @Override
+    default Immutable<name>Set intersect(<name>Set set)
+    {
+        if (this.size() \< set.size())
+        {
+            return this.select(set::contains);
+        }
+        else
+        {
+            return set.select(this::contains).toImmutable();
+        }
+    }
+
     <immutableAPI(fileName(primitive), type, name)>
 }
 
--- a/eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg
@@ -88,6 +88,25 @@ public interface Mutable<name>Set extends Mutable<name>Collection, <name>Set
             return set.toSet().withAll(this);
         }
     }
+
+    /**
+     * Returns the set of all objects that are members of both {@code this} and {@code set}. The intersection of
+     * [1, 2, 3] and [2, 3, 4] is the set [2, 3].
+     *
+     * @since 11.0.
+     */
+    @Override
+    default Mutable<name>Set intersect(<name>Set set)
+    {
+        if (this.size() \< set.size())
+        {
+            return this.select(set::contains);
+        }
+        else
+        {
+            return set.select(this::contains, this.newEmpty());
+        }
+    }
 }
 
 >>
--- a/eclipse-collections-code-generator/src/main/resources/api/set/primitiveSet.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/set/primitiveSet.stg
@@ -45,6 +45,14 @@ public interface <name>Set extends <name>Iterable
     <name>Set union(<name>Set set);
 
     /**
+     * Returns the set of all objects that are members of both {@code this} and {@code set}. The intersection of
+     * [1, 2, 3] and [2, 3, 4] is the set [2, 3].
+     *
+     * @since 11.0.
+     */
+    <name>Set intersect(<name>Set set);
+
+    /**
      * Follows the same general contract as {@link Set#equals(Object)}.
      */
     @Override
--- a/eclipse-collections-code-generator/src/main/resources/test/set/mutable/abstractImmutablePrimitiveSetTestCase.stg
+++ b/eclipse-collections-code-generator/src/main/resources/test/set/mutable/abstractImmutablePrimitiveSetTestCase.stg
@@ -303,6 +303,46 @@ public abstract class AbstractImmutable<name>HashSetTestCase extends AbstractImm
         Immutable<name>Set actual = set1.union(set2);
         Assert.assertEquals(expected, actual);
     }
+
+    @Test
+    public void intersect()
+    {
+        this.assertIntersect(
+                this.newWith(<["1", "2", "3"]:(literal.(type))(); separator=", ">),
+                this.newWith(<["3", "4", "5"]:(literal.(type))(); separator=", ">),
+                this.newWith(<["3"]:(literal.(type))(); separator=", ">));
+
+        this.assertIntersect(
+                this.newWith(<["1", "2", "3", "6"]:(literal.(type))(); separator=", ">),
+                this.newWith(<["3", "4", "5"]:(literal.(type))(); separator=", ">),
+                this.newWith(<["3"]:(literal.(type))(); separator=", ">));
+
+        this.assertIntersect(
+                this.newWith(<["1", "2", "3"]:(literal.(type))(); separator=", ">),
+                this.newWith(<["3", "4", "5" , "6"]:(literal.(type))(); separator=", ">),
+                this.newWith(<["3"]:(literal.(type))(); separator=", ">));
+
+        this.assertIntersect(
+                this.newWith(),
+                this.newWith(),
+                this.newWith());
+
+        this.assertIntersect(
+                this.newWith(),
+                this.newWith(<["3", "4", "5"]:(literal.(type))(); separator=", ">),
+                this.newWith());
+
+        this.assertIntersect(
+                this.newWith(<["1", "2", "3"]:(literal.(type))(); separator=", ">),
+                this.newWith(),
+                this.newWith());
+    }
+
+    private void assertIntersect(Immutable<name>Set set1, Immutable<name>Set set2, Immutable<name>Set expected)
+    {
+        Immutable<name>Set actual = set1.intersect(set2);
+        Assert.assertEquals(expected, actual);
+    }
 }
 
 >>
--- a/eclipse-collections-code-generator/src/main/resources/test/set/mutable/abstractPrimitiveSetTestCase.stg
+++ b/eclipse-collections-code-generator/src/main/resources/test/set/mutable/abstractPrimitiveSetTestCase.stg
@@ -497,6 +497,46 @@ public abstract class Abstract<name>SetTestCase extends AbstractMutable<name>Col
         Mutable<name>Set actual = set1.union(set2);
         Assert.assertEquals(expected, actual);
     }
+
+    @Test
+    public void intersect()
+    {
+        this.assertIntersect(
+                this.newWith(<["1", "2", "3"]:(literal.(type))(); separator=", ">),
+                this.newWith(<["3", "4", "5"]:(literal.(type))(); separator=", ">),
+                this.newWith(<["3"]:(literal.(type))(); separator=", ">));
+
+        this.assertIntersect(
+                this.newWith(<["1", "2", "3", "6"]:(literal.(type))(); separator=", ">),
+                this.newWith(<["3", "4", "5"]:(literal.(type))(); separator=", ">),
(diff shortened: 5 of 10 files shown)


donraab (approved):
LGTM. I've requested a review from @mohrezaei for any additional review feedback to make sure there aren't any optimization issues.

eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg
```diff
@@ -88,6 +88,29 @@ public interface Mutable<name>Set extends Mutable<name>Collection, <name>Set
...
+     *
+     * @since 11.0.
+     */
+    @Override
+    default Mutable<name>Set intersect(<name>Set set)
+    {
+        if (this.size() \< set.size())
+        {
+            Mutable<name>Set mutableSet = this.toSet();
+            mutableSet.retainAll(set);
```
mohrezaei, line 104: The `retainAll` implementation makes too many copies, which can be elided here. Inline this bit instead:
```java
        <name>Set retained = mutableSet.select(set::contains);
```

eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg
```diff
@@ -88,6 +88,29 @@ public interface Mutable<name>Set extends Mutable<name>Collection, <name>Set
...
+     * [1, 2, 3] and [2, 3, 4] is the set [2, 3].
+     *
+     * @since 11.0.
+     */
+    @Override
+    default Mutable<name>Set intersect(<name>Set set)
+    {
+        if (this.size() \< set.size())
+        {
+            Mutable<name>Set mutableSet = this.toSet();
```
mohrezaei, line 103: we're already a mutable set, no?

eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg
```diff
@@ -88,6 +88,29 @@ public interface Mutable<name>Set extends Mutable<name>Collection, <name>Set
...
+            mutableSet.retainAll(set);
+            return mutableSet;
+        }
+        else
+        {
+            Mutable<name>Set mutableSet = set.toSet();
```
mohrezaei, line 109: You won't need `toSet()` if you use `select`

eclipse-collections-code-generator/src/main/resources/api/set/immutablePrimitiveSet.stg
```diff
@@ -59,6 +59,29 @@ public interface Immutable<name>Set extends Immutable<name>Collection, <name>Set
...
+     *
+     * @since 11.0.
+     */
+    @Override
+    default Immutable<name>Set intersect(<name>Set set)
+    {
+        if (this.size() \< set.size())
+        {
+            Mutable<name>Set mutableSet = this.toSet();
+            mutableSet.retainAll(set);
```
mohrezaei, line 74: use `intersect` instead of `retainAll` (and below)

mohrezaei (changes requested):

eclipse-collections-code-generator/src/main/resources/api/set/immutablePrimitiveSet.stg
donraab: Won't this create two copies of the set then?

Actually, I think we can use the same approach as above... use `select` directly which will return an `ImmutableSet`.

eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg
prathasirisha: Updated this to use `select` instead of `retainAll`. Thank you for the suggestion!

eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg
prathasirisha: This is now simplified based on your suggestion to use to `select`.

eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg
prathasirisha: Done.

eclipse-collections-code-generator/src/main/resources/api/set/immutablePrimitiveSet.stg
prathasirisha: Used `select` here as well.

eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg
```diff
@@ -88,6 +88,25 @@ public interface Mutable<name>Set extends Mutable<name>Collection, <name>Set
...
+    @Override
+    default Mutable<name>Set intersect(<name>Set set)
+    {
+        if (this.size() \< set.size())
+        {
+            return this.select(set::contains);
+        }
+        else
+        {
+            return set.asLazy().select(this::contains).toSet();
```
mohrezaei, line 107: why `asLazy()`? You should avoid `toSet()` by using the two argument `select` with a new set.

mohrezaei (changes requested):

eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg
prathasirisha: Started out with the two argument approach like so - 

  `return set.select(this::contains, this.newEmpty());` 

Since the same won't work on the immutable side, I leaned towards symmetry. I will update the implementation on the mutable side to use select with a target empty collection and leave the lazy implementation on the immutable side. Let me know if that is good.

mohrezaei: why doesn't it work for the immutable?
``` set.select(this::contains, new <name>HashSet()).toImmutable()```

Comment (donraab) [reply eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg:107]: It's in a default method, so can't reference the `<name>HashSet`. This would have to be implemented in the concrete class.

Comment (nikhilnanivadekar) [reply eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg:107]: It’s a major version change, we can add it on implementations. I would rather do that as select() and toSet() on Immutable sides is doubly expensive.

Comment (donraab) [reply eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg:107]: Why is it doubly expensive? It's using lazy. It has an extra object creation for the primitive SelectIterable but we're creating a new primitive set anyway.

Comment (mohrezaei) [reply eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg]: The sequence `this.asLazy().select(that::contains).toSet().toImmutable()` on an immutable set first constructs some lazy stuff (cheap), then does the following:
- call `count` on `this` with the predicate being `that`. That's one loop through the whole thing.
- allocate an array of that size
- construct a `IfIntProcedure` object using `select` and `that::contains` as the predicate
- call `this.each(ifIntProc)`. This loops through the whole thing again.
- finally, the filled array is passed to the set constructor, where the result is allocated then filled by looping through the array.
- the contents of this is now copied to a second array.
- the second array is used to construct another set.
- this final set is wrapped as an immutable set.

so we compute the intersection twice and construct 3 temporary and 1 final place to put them (all the while recomputing hashes or copying stuff around).

It would take quite some effort to optimize that. Roughly, this would entail:
- stop using arrays -- use lists
- `toSet` and `toImmutable` should become lazy. this probably breaks everything, so the only realistic solution might be to have a `toImmutableSet`

PSA: don't use `asLazy` especially with immutables if memory, performance or sanity have any relevance to the code.

Comment (donraab) [reply eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg]: Wow, this is a great bit of research and discovery @mohrezaei! Thank you. It looks like @nikhilnanivadekar you were correct this is doubly expensive. It's clear we should just override `intersect` in the concrete classes for immutable for now. 

We have a bunch of positive work we can do here as a result in separate PRs. I have been toying with the idea of adding `toImmutableSet` and friends. I might explore this to see some of the possibilities.

Comment (prathasirisha) [reply eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg]: Thank you @mohrezaei for the detailed explanation, will bear in my mind some of the points you highlighted with `asLazy` usages. 

@mohrezaei, @nikhilnanivadekar , @donraab    - 
Upon revisiting the solution, noticed that `toImmutable()` is available for me to use in combination with `select` on the immutable side Like so - 

        `return set.select(this::contains).toImmutable();`

Hoping this is closest to what we want to achieve in terms of implementation without compromising on optimization. Please let me know if this works! 

Thank you for reviewing the code so far.

Comment (donraab) [reply eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg]: This looks good to me. The call `toImmutable() `will be a no-op if the set is already an `Immutable<name>Set`.

Comment (mohrezaei) [reply eclipse-collections-code-generator/src/main/resources/api/set/mutablePrimitiveSet.stg]: @donraab It's not already immutable. I think we should really fix `toImmutable()` on primitive sets.
```java
        IntHashSet mutableSet = IntHashSet.newSetWith(this.toArray());
```
that line is the worst offender.  If we properly use the `copyOnWrite` flag in the set, we can even avoid all copying, which will make this code as efficient as possible:
```java
    @Override
    public ImmutableIntSet toImmutable()
    {
        if (this.size() == 0)
        {
            return IntSets.immutable.with();
        }
        if (this.size() == 1)
        {
            return IntSets.immutable.with(this.intIterator().next());
        }
        IntHashSet mutableSet;
        if (this.table.length == smallestPowerOfTwoGreaterThan(this.size() << 1)) // table is not too big
        {
            mutableSet = this;
            this.copyOnWrite = true;
        }
        else
        {
            IntHashSet mutableSet = new IntHashSet(this.size()) ;
            mutableSet.addAll(this);
        }
        return new ImmutableIntHashSet(mutableSet.table, mutableSet.occupiedWithData, mutableSet.zeroToThirtyOne, mutableSet.zeroToThirtyOneOccupied);
    }
```

nikhilnanivadekar:
It’s not already an ImmutableSet.

donraab:
It may or may not be an ImmutableSet. It is IntSet, which means it may be either.

mohrezaei:
@donraab, select on an immutable set is already overriden to be immutable (so I'm not sure why we need an explicit `toImmutable()` in `intersect`), but the actual implementation does this:
```java
        public ImmutableIntSet select(IntPredicate predicate)
        {
            return this.select(predicate, new IntHashSet()).toImmutable();
        }
```
There is an inner mutable set, which because of the bad implementation of `toImmutable()` is copied 3 times. With the above code, that can be reduced to zero.

donraab:
@mohrezaei the reason we need `toImmutable` is that the type of the set is `IntSet`, so the return type for `select` is `IntSet`, even if the implementation returns `ImmutableIntSet`. So we have to call `toImmutable` to get back the `ImmutableIntSet`.

Your proposed fix to `toImmutable` is great and should be done, but won't change the code that @prathasirisha has proposed here for `intersect`.

mohrezaei:
@donraab oh, right, this is the else branch (with select on the argument, not `this`).

donraab:
@mohrezaei Can you submit a separate PR for the `toImmutable` fix? We can merge that first and then merge @prathasirisha code for intersect.

eclipse-collections-code-generator/src/main/resources/api/set/primitiveSet.stg
```diff
@@ -44,6 +44,14 @@ public interface <name>Set extends <name>Iterable
      */
     <name>Set union(<name>Set set);
 
+    /**
+     * Returns the set of all objects that are members of both {@code this} and {@code set}. The intersection of
+     * [1, 2, 3] and [2, 3, 4] is the set [2, 3].
+     *
+     * @since 11.0.
+     */
+    <name>Set intersect(<name>Set set);
```
mohrezaei, line 53: should this have an override in `immutableSet.stg` for the return type?

mohrezaei (changes requested):

mohrezaei:
@donraab #1019

eclipse-collections-code-generator/src/main/resources/api/set/primitiveSet.stg
prathasirisha: Do you mean `immutablePrimitiveSet.stg`? If yes, it is overridden there.

mohrezaei (approved):

PR #1019, opened 2020-10-23
Optimize toImmutable on primitive sets

mohrezaei:
Signed-off-by: Mohammad Rezaei <mohdev@rezaei.cc>

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/impl/set/mutable/primitiveHashSet.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/set/mutable/primitiveHashSet.stg
@@ -463,7 +463,17 @@ public class <name>HashSet extends Abstract<name>Set implements Mutable<name>Set
         {
             return <name>Sets.immutable.with(this.<type>Iterator().next());
         }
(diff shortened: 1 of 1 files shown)


mohrezaei:
what's the deal with the failing builds?

nikhilnanivadekar:
No idea about Unit Test build. You can retry and it generally clears off. It’s an annoying GH action problem.
Ignore the EA builds for now. I’ll raise a PR over the weekend to address those.

donraab (approved):
LGTM. Thanks @mohrezaei !

donraab:
Hi @mohrezaei, just curious on this code:

```
else
{
    mutableSet = new <name>HashSet(this.size());
    mutableSet.addAll(this);
}
```

Since this is in `IntHashSet`, would this be a better option?

```
else
{
    mutableSet = new <name>HashSet(this);
}

```

mohrezaei:
@donraab the constructor that takes a set blindly copies the backing array (including all extra space). E.g. a set with enough room with 1000 elements that contains just 50 will have a lot of extra space in the backing array. A presized set with `addAll` will have the right amount of space in the backing array.

PR #1021, opened 2020-10-24
Update developer details

nikhilnanivadekar:
@prathasirisha Please review this and if looks good do the honors of merging it. It will test your commit rights as well.

### DIFF
--- a/pom-template
+++ b/pom-template
@@ -55,5 +55,13 @@
         <developer>
             <name>Hiroshi Ito</name>
         </developer>
+
+        <developer>
+            <name>Nikhil J. Nanivadekar</name>
+        </developer>
+
+        <developer>
+            <name>Sirisha Pratha</name>
+        </developer>
     </developers>
 </project>
--- a/pom-template
+++ b/pom-template
@@ -42,22 +42,18 @@
     <developers>
         <developer>
             <name>Craig P. Motlin</name>
-            <email>craig.motlin@gs.com</email>
         </developer>
 
         <developer>
             <name>Donald Raab</name>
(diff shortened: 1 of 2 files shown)


prathasirisha (changes requested):
Hi @nikhilnanivadekar, thank you for updating the developer details information. 
A couple of comments and observations - 

1. copyright info needs to be updated on this file. 
2. It makes sense to remove the email addresses, as most of them were not up-to-date. We anyway get notifications via GitHub registered email. I see that you have left this is a separate commit, did you mean to squash them?
3. I also see bnd-maven plugin warnings in our builds like this - 

Warning: 561 [WARNING] Cannot consider developer in line '76' of file '/home/runner/work/eclipse-collections/eclipse-collections/pom.xml' for bundle header 'Bundle-Developers' as it does not contain the mandatory id.

I was wondering if `<id>` was left out intentionally? It is a warning, at the moment. But was just curious, if we should add this tag. 

4. I also noticed developer information is in this other file called pom-template. Can't tell how or where it is being used, don't think it is in use. I thought will let you know, in case you think it is important we change this too.

nikhilnanivadekar:
No idea about the bnd plug-in. Even if I were to add the id I don’t know what that’ll be. So I have been ignoring the warnings :)

I didn’t mean to squash them. They should be 2 separate commits.

Copyright update is an interesting problem. I plan to solve it by updating the entire project’s year to 2020. I can update it for this commit.

I’ll update the template as well.

nikhilnanivadekar:
@prathasirisha done. Please check.

prathasirisha (approved):
LGTM.

PR #1024, opened 2020-10-25
Optimize containsAll on primitive iterables.

rinatigati:

I moved up containsAll as a default method on primitive iterables. I was going to add containsAny and containsNone as default methods there so this way they are all in the same place.  The build and tests work fine with this change. If there is some reason to not do this, please let me know and I'll move it back to the abstract primitive iterable where it was before.

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
@@ -42,6 +42,7 @@ import org.eclipse.collections.api.collection.primitive.MutableLongCollection;
 import org.eclipse.collections.api.collection.primitive.MutableShortCollection;
 import org.eclipse.collections.api.iterator.<name>Iterator;
 import org.eclipse.collections.api.list.primitive.Mutable<name>List;
+import org.eclipse.collections.api.set.primitive.<name>Set;
 import org.eclipse.collections.api.set.primitive.Mutable<name>Set;
 
 /**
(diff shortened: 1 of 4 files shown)


mohrezaei:
Please add an acceptance test with 1 million element lists.  Something like this:
```java
        IntArrayList list1 = new IntArrayList(1_000_000);
        IntArrayList list2 = new IntArrayList(1_000_000);
        for(int i=0;i<1_000_000-1;i++) 
        {
            list1.add(i);
            list2.add(i);
        }
        list1.add(1_000_000);
        Assert.assertTrue(list1.containsAll(list2));
```
Old implementation takes forever; new implementation should take less than 10 seconds.

rinatigati:
Good point! @mohrezaei I'll make sure to add that and update the PR.

rinatigati:
@mohrezaei Added the acceptance test and updated the PR. The test with the new implementation took ~200 ms on my laptop while the test with the old implementation took 110 seconds. Much much faster. 👍

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
```diff
@@ -79,16 +80,52 @@ public interface <name>Iterable extends PrimitiveIterable
...
 
     /**
-     * Returns true if the all of the values specified in the source array are contained
+     * Returns true if all of the values specified in the source array are contained
      * in the <name>Iterable, and false if they are not.
      */
-    boolean containsAll(<type>... source);
+    default boolean containsAll(<type>... source)
+    {
+        if (this.size() \<= 32 || source.length \< 4)
```
donraab, line 88: This could be simplified, with a small cost.

```
default boolean containsAll(<type>... source)
{
    return this.containsAll(new <name>ArrayList(source));
}

```

mohrezaei (approved):

eclipse-collections-code-generator/src/main/resources/api/primitiveIterable.stg
mohrezaei: That's not a good tradeoff for small arrays. I'd rather keep the code as is.

donraab: This is one of those places where having primitive array utility classes with some basic patterns would have helped. Something for another day perhaps. Please feel free to merge.

PR #1028, opened 2020-11-03
Add new methods withMapIterable to Mutable Maps and newWithMap, newWithMapIterable to Immutable Maps. Closes #875

### LINKED ISSUE #875: ImmutableMap newWithAll() method that takes a MapIterable
First of all, thank you for this really cool library :D.

I need to merge `ImmutableMap`s and I couldn't find a method to do it directly. 
Currently, I use this:
`myImmutableMap.newWithAllKeyValues(myOtherImmutableMap.keyValuesView())` 
But if feels like a workaround. Is there a more direct approach?
What I was looking for is something like:
`myImmutableMap.newWithAll(myOtherImmutableMap)`. 

What do you think?

An example:
```
var myImmutableMap = Maps.immutable.of("firstKey", "firstValue");
var myOtherImmutableMap = Maps.immutable.of("secondKey", "secondValue");

var result1 = myImmutableMap.newWithAllKeyValues(myOtherImmutableMap.keyValuesView());
// Could be:
var result2 = myImmutableMap.newWithAll(myOtherImmutableMap);
// AND/OR:
var result3 = Maps.immutable.ofAll(myImmutableMap, myOtherImmutableMap);

var expectedResult = Maps.immutable.of("firstKey", "firstValue", "secondKey", "secondValue");

assertEquals(result1, expectedResult);
assertEquals(result2, expectedResult);
assertEquals(result3, expectedResult);

// Note: if two keys are identical in the two maps, I would expect the same behavior as newWithAllKeyValues
```

nsardana-bny:
Added two methods withMap, withMapIterable to Mutable Maps and newWithMap, newWithMapIterable to Immutable Maps. Closes #875

Signed-off-by: nsardana-bny <neha.sardana@bnymellon.com>

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/FixedSizeMap.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/FixedSizeMap.java
@@ -74,6 +74,12 @@ public interface FixedSizeMap<K, V>
     }
 
     @Override
+    default FixedSizeMap<K, V> withMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
+    {
+        throw new UnsupportedOperationException("Cannot call withMapIterable() on " + this.getClass().getSimpleName());
+    }
+
+    @Override
     default void putAllMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
     {
         throw new UnsupportedOperationException("Cannot call putAllMapIterable() on " + this.getClass().getSimpleName());
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/FixedSizeMap.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/FixedSizeMap.java
@@ -72,4 +72,10 @@ public interface FixedSizeMap<K, V>
     {
         throw new UnsupportedOperationException("Cannot call withMap() on " + this.getClass().getSimpleName());
     }
+
+    @Override
+    default void putAllMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
+    {
+        throw new UnsupportedOperationException("Cannot call putAllMapIterable() on " + this.getClass().getSimpleName());
+    }
 }
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/FixedSizeMap.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/FixedSizeMap.java
@@ -70,7 +70,6 @@ public interface FixedSizeMap<K, V>
     @Override
     default FixedSizeMap<K, V> withMap(Map<? extends K, ? extends V> map)
     {
-        this.putAll(map);
-        return this;
+        throw new UnsupportedOperationException("Cannot call withMap() on " + this.getClass().getSimpleName());
     }
 }
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ImmutableMap.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ImmutableMap.java
@@ -10,6 +10,8 @@
 
 package org.eclipse.collections.api.map;
 
+import java.util.Map;
+
 import org.eclipse.collections.api.bag.ImmutableBag;
 import org.eclipse.collections.api.bag.primitive.ImmutableBooleanBag;
 import org.eclipse.collections.api.bag.primitive.ImmutableByteBag;
@@ -54,6 +56,12 @@ public interface ImmutableMap<K, V>
 
     @Override
     ImmutableMap<K, V> newWithAllKeyValues(Iterable<? extends Pair<? extends K, ? extends V>> keyValues);
+
+    @Override
+    ImmutableMap<K, V> newWithMap(Map<? extends K, ? extends V> map);
+
+    @Override
+    ImmutableMap<K, V> newWithMapIterable(MapIterable<? extends K, ? extends V> mapIterable);
 
     @Override
     ImmutableMap<K, V> newWithAllKeyValueArguments(Pair<? extends K, ? extends V>... keyValuePairs);
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
@@ -172,6 +172,11 @@ public interface MutableMapIterable<K, V> extends MapIterable<K, V>, Map<K, V>
         return this;
     }
 
+    default void putAllMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
+    {
+        mapIterable.forEachKeyValue(this::put);
+    }
+
     /**
      * This method allows mutable, fixed size, and immutable maps the ability to add elements to their existing
      * elements. In order to support fixed size maps, a new instance of a map would have to be returned including the
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
@@ -172,6 +172,12 @@ public interface MutableMapIterable<K, V> extends MapIterable<K, V>, Map<K, V>
         return this;
     }
 
+    default MutableMapIterable<K, V> withMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
+    {
+        this.putAllMapIterable(mapIterable);
+        return this;
+    }
+
     default void putAllMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
     {
         mapIterable.forEachKeyValue(this::put);
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
@@ -23,6 +23,7 @@ import org.eclipse.collections.api.block.function.Function2;
 import org.eclipse.collections.api.block.predicate.Predicate;
 import org.eclipse.collections.api.block.predicate.Predicate2;
 import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.map.MapIterable;
 import org.eclipse.collections.api.multimap.set.MutableSetMultimap;
 import org.eclipse.collections.api.ordered.OrderedIterable;
 import org.eclipse.collections.api.partition.set.PartitionMutableSet;
@@ -299,6 +300,15 @@ public class SynchronizedBiMap<K, V> extends AbstractSynchronizedMapIterable<K, 
         {
             this.putAll(map);
             return this;
+        }
+    }
+
+    @Override
+    public void putAllMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
+    {
+        synchronized (this.lock)
+        {
+            mapIterable.forEachKeyValue(this.getDelegate()::put);
         }
     }
 
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
@@ -304,6 +304,13 @@ public class SynchronizedBiMap<K, V> extends AbstractSynchronizedMapIterable<K, 
     }
 
     @Override
+    public MutableBiMap<K, V> withMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
+    {
+        this.putAllMapIterable(mapIterable);
+        return this;
+    }
+
+    @Override
     public void putAllMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
     {
         synchronized (this.lock)
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/AbstractImmutableMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/AbstractImmutableMap.java
@@ -51,6 +51,7 @@ import org.eclipse.collections.api.block.procedure.Procedure2;
 import org.eclipse.collections.api.factory.Bags;
 import org.eclipse.collections.api.factory.Maps;
 import org.eclipse.collections.api.map.ImmutableMap;
+import org.eclipse.collections.api.map.MapIterable;
 import org.eclipse.collections.api.map.MutableMap;
 import org.eclipse.collections.api.map.primitive.ImmutableObjectDoubleMap;
 import org.eclipse.collections.api.map.primitive.ImmutableObjectLongMap;
@@ -169,6 +170,22 @@ public abstract class AbstractImmutableMap<K, V>
             map.put(keyValuePair.getOne(), keyValuePair.getTwo());
         }
         return map.toImmutable();
+    }
+
+    @Override
+    public ImmutableMap<K, V> newWithMap(Map<? extends K, ? extends V> map)
+    {
+        MutableMap<K, V> mutableMap = Maps.mutable.ofMapIterable(this);
+        mutableMap.putAll(map);
+        return mutableMap.toImmutable();
+    }
+
+    @Override
+    public ImmutableMap<K, V> newWithMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
+    {
+        MutableMap<K, V> mutableMap = Maps.mutable.ofMapIterable(this);
+        mapIterable.forEachKeyValue(mutableMap::put);
+        return mutableMap.toImmutable();
     }
 
     @Override
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/SynchronizedMutableMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/SynchronizedMutableMap.java
@@ -126,6 +126,13 @@ public class SynchronizedMutableMap<K, V>
     }
 
     @Override
+    public MutableMap<K, V> withMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
+    {
+        this.putAllMapIterable(mapIterable);
+        return this;
+    }
+
+    @Override
     public void putAllMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
     {
         synchronized (this.lock)
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/SynchronizedMutableMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/SynchronizedMutableMap.java
@@ -41,6 +41,7 @@ import org.eclipse.collections.api.block.predicate.Predicate2;
 import org.eclipse.collections.api.block.procedure.Procedure;
 import org.eclipse.collections.api.factory.Maps;
 import org.eclipse.collections.api.map.ImmutableMap;
+import org.eclipse.collections.api.map.MapIterable;
 import org.eclipse.collections.api.map.MutableMap;
 import org.eclipse.collections.api.multimap.bag.MutableBagMultimap;
 import org.eclipse.collections.api.multimap.set.MutableSetMultimap;
@@ -121,6 +122,15 @@ public class SynchronizedMutableMap<K, V>
         {
             this.putAll(map);
             return this;
+        }
+    }
+
+    @Override
+    public void putAllMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
+    {
+        synchronized (this.lock)
+        {
+            mapIterable.forEachKeyValue(this.getDelegate()::put);
         }
     }
 
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/SynchronizedSortedMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/SynchronizedSortedMap.java
@@ -135,6 +135,13 @@ public class SynchronizedSortedMap<K, V>
     }
 
     @Override
+    public MutableSortedMap<K, V> withMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
+    {
+        this.putAllMapIterable(mapIterable);
+        return this;
+    }
+
+    @Override
     public void putAllMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
     {
         synchronized (this.lock)
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/SynchronizedSortedMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/SynchronizedSortedMap.java
@@ -45,6 +45,7 @@ import org.eclipse.collections.api.list.primitive.MutableFloatList;
 import org.eclipse.collections.api.list.primitive.MutableIntList;
 import org.eclipse.collections.api.list.primitive.MutableLongList;
 import org.eclipse.collections.api.list.primitive.MutableShortList;
+import org.eclipse.collections.api.map.MapIterable;
 import org.eclipse.collections.api.map.MutableMap;
 import org.eclipse.collections.api.map.MutableMapIterable;
 import org.eclipse.collections.api.map.sorted.ImmutableSortedMap;
@@ -130,6 +131,15 @@ public class SynchronizedSortedMap<K, V>
         {
             this.putAll(map);
             return this;
+        }
+    }
+
+    @Override
+    public void putAllMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
+    {
+        synchronized (this.lock)
+        {
+            mapIterable.forEachKeyValue(this.getDelegate()::put);
         }
     }
 
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/UnmodifiableBiMapTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/UnmodifiableBiMapTest.java
@@ -198,6 +198,41 @@ public class UnmodifiableBiMapTest extends AbstractMutableBiMapTestCase
 
     @Override
     @Test
+    public void withMapIterable()
+    {
+        Assert.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMapIterable(Maps.mutable.with(1, Character.valueOf('a'))));
+    }
+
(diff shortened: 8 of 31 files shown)


nikhilnanivadekar:
@nsardana-bny thanks for PR. Looking at this PR I wonder if we need the 2 APIs withMap() and withMapIterable(). The withMap() is nothing but a putAll() and we can add a putAll() which takes a MapIterable. Thoughts?

nikhilnanivadekar:
Ignore my previous comment, didnt realize it was on ImmutableMaps. I missed that on my small phone screen 😄

nsardana-bny:
Good question @nikhilnanivadekar! 
Just a clarification withMap() already existed with Mutable Maps. 
The reason for adding withMapIterable() into Mutable Maps is to create the symmetry with the newly created newWithMapIterable() in Immutable Maps. This feature on Immutable Map was requested by the user.
Also, I added a newWithMap() method to Immutable Maps which follows the already existing withMap() method in Mutable Map.
Hope the explanation helps!

nikhilnanivadekar:
Understood. Can you break the commits please? It is super difficult to review everything with a 1 big commit. Additionally it will be a nightmare while creating the release notes. I will review it once before you split the commits.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/FixedSizeMap.java
```diff
@@ -70,7 +70,12 @@
     @Override
     default FixedSizeMap<K, V> withMap(Map<? extends K, ? extends V> map)
     {
-        this.putAll(map);
-        return this;
```
nikhilnanivadekar, line 74: This is a breaking change albeit a correct change. But I would want it to be a part of a separate commit to show it on the Release notes.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ImmutableMap.java
```diff
@@ -145,7 +153,9 @@
      * @since 9.2
      */
     @Override
-    default <P, R> ImmutableBag<R> flatCollectWith(Function2<? super V, ? super P, ? extends Iterable<R>> function, P parameter)
+    default <P, R> ImmutableBag<R> flatCollectWith(
```
nikhilnanivadekar, line 156: please revert. This is an unrelated change.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ImmutableMap.java
```diff
@@ -216,7 +226,11 @@
     {
         MutableMap<K1, V2> map = Maps.mutable.empty();
         this.forEachKeyValue((key, value) -> {
-            map.updateValueWith(keyFunction.valueOf(key), zeroValueFactory, nonMutatingAggregator, valueFunction.valueOf(value));
```
nikhilnanivadekar, line 219: please revert. This is an unrelated change.

nsardana-bny:
Would you like me to break the commits in two? One for Immutable Maps and one for Mutable Maps? 
or four commits one for each method?

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
```diff
@@ -35,7 +36,8 @@
 import org.eclipse.collections.impl.set.mutable.SynchronizedMutableSet;
 import org.eclipse.collections.impl.utility.LazyIterate;
 
-public class SynchronizedBiMap<K, V> extends AbstractSynchronizedMapIterable<K, V> implements MutableBiMap<K, V>, Serializable
+public class SynchronizedBiMap<K, V> extends AbstractSynchronizedMapIterable<K, V>
+        implements MutableBiMap<K, V>, Serializable
```
nikhilnanivadekar, line 40: please revert. This is an unrelated change.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
```diff
@@ -85,7 +87,8 @@ public V forcePut(K key, V value)
     @Override
     public MutableBiMap<K, V> asUnmodifiable()
     {
-        throw new UnsupportedOperationException(this.getClass().getSimpleName() + ".asUnmodifiable() not implemented yet");
+        throw new UnsupportedOperationException(this.getClass().getSimpleName()
```
nikhilnanivadekar, line 90: please revert. This is an unrelated change.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/AbstractImmutableMap.java
```diff
@@ -171,6 +172,22 @@ public void clear()
         return map.toImmutable();
     }
 
+    @Override
+    public ImmutableMap<K, V> newWithMap(Map<? extends K, ? extends V> map)
+    {
+        UnifiedMap<K, V> unifiedMap = UnifiedMap.newMap(this);
```
nikhilnanivadekar, line 178: Effective Java principles state that interface should be on the left hand side. Also, this should ideally be replaced by Maps.mutable.with().

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/immutable/AbstractImmutableMap.java
```diff
@@ -171,6 +172,22 @@ public void clear()
...
+        unifiedMap.putAll(map);
+        return unifiedMap.toImmutable();
+    }
+
+    @Override
+    public ImmutableMap<K, V> newWithMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
+    {
+        UnifiedMap<K, V> unifiedMap = UnifiedMap.newMap(this);
```
nikhilnanivadekar, line 186: same comment as above

unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/UnmodifiableBiMapTest.java
```diff
@@ -73,7 +80,15 @@
     }
 
     @Override
-    protected <K, V> MutableBiMap<K, V> newMapWithKeysValues(K key1, V value1, K key2, V value2, K key3, V value3, K key4, V value4)
+    protected <K, V> MutableBiMap<K, V> newMapWithKeysValues(
```
nikhilnanivadekar, line 83: please revert. This is an unrelated change.

unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/UnmodifiableBiMapTest.java
```diff
@@ -165,28 +180,36 @@ public void withKeyValue()
     @Test
     public void withMap()
     {
-        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(Maps.mutable.with(1, Character.valueOf('a'))));
```
nikhilnanivadekar, line 168: please revert. This is an unrelated change.

unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/UnmodifiableBiMapTest.java
```diff
@@ -165,28 +180,36 @@ public void withKeyValue()
...
+        Verify.assertThrows(
+                UnsupportedOperationException.class,
+                () -> this.newMapWithKeyValue(1, 'a').withMap(Maps.mutable.with(1, Character.valueOf('a'))));
     }
 
     @Override
     @Test
     public void withMapEmpty()
     {
-        Verify.assertThrows(UnsupportedOperationException.class, () -> this.newMapWithKeyValue(1, 'a').withMap(Maps.mutable.empty()));
```
nikhilnanivadekar, line 175: please revert. This is an unrelated change.

nikhilnanivadekar (changes requested):

nikhilnanivadekar:
I skimmed through as much as I could briefly but there are a lot of indentation only and unrelated changes. That needs to be reverted first to reduce the churn and make an easier review.
Regarding breaking up the commits, I need you to break at least this 1 change in a separate commit:
FixedSizeMap bug fix where `withMap` now throws. Clearly elaborate that it is a breaking change.

Regarding the other changes, I understand what the user asks, but that should be restricted only to Immutable side. The reason for that is for Immutable side we cannot mutate the data structure, but we need to create a copy and return a new one.

The mutable side has a different signature and having that in the same commit is a bit confusing.

I hope that helps.

prathasirisha:
@nsardana-bny, re-ran that failing job, all checks/builds passed now. Let us know if this is ready for review! Thank you for the contribution :)

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
```diff
@@ -302,6 +303,16 @@ public V forcePut(K key, V value)
         }
     }
 
+    @Override
+    public MutableBiMap<K, V> withMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
```
nikhilnanivadekar, line 307: Shouldn't this be in `AbstractSynchronizedMapIterable` ?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/SynchronizedMutableMap.java
```diff
@@ -124,6 +125,16 @@ public SynchronizedMutableMap(MutableMap<K, V> newMap, Object newLock)
         }
     }
 
+    @Override
+    public MutableMap<K, V> withMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
```
nikhilnanivadekar, line 129: Same comment as `SynchronizedBiMap`

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/SynchronizedSortedMap.java
```diff
@@ -133,6 +134,16 @@ public SynchronizedSortedMap(MutableSortedMap<K, V> newMap, Object lock)
         }
     }
 
+    @Override
+    public MutableSortedMap<K, V> withMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
```
nikhilnanivadekar, line 138: Same comment as `SynchronizedBiMap`

unit-tests/src/test/java/org/eclipse/collections/impl/bimap/mutable/UnmodifiableBiMapTest.java
```diff
@@ -60,6 +60,13 @@
         return UnmodifiableBiMap.of(Maps.mutable.of(key, value));
     }
 
+    @Test
```
nikhilnanivadekar, line 63: Please reorder this to move it around the other tests of method.

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
```diff
@@ -574,6 +575,17 @@ public void withMap()
         Verify.assertMapsEqual(UnifiedMap.newWithKeysValues("A", 1, "B", 22, "C", 3), mapWith);
     }
 
+    @Test
```
nikhilnanivadekar, line 578: Reorder the tests please to bunch all of them together.

nikhilnanivadekar (changes requested):

nikhilnanivadekar:
@nsardana-bny thanks for the PR and splitting the commits. It is much easier to review. I have reviewed 2 of the 3 commits and both of them have some feedback to address. Lastly, the commit message should follow the directions from the [Contribution Guide](https://github.com/eclipse/eclipse-collections/blob/master/CONTRIBUTING.md#commit-messages). The commit message should use imperative mode as well as demonstrate the intent of the change. Can you please reword it correctly. Thanks!

Lastly, as a part of this PR, I think it is worth to add another API `putAllMapIterable()` and then `withMapIterable` can delegate to `putAllMapIterable()`. You have done about 90% of the work already. This API should be a different commit as well and when you rebase, you can rearrange the commits to follow a nice chronology. If you want, happy to have you raise a different PR, but that PR will need to be merged before this PR. Thanks again!

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
nsardana-bny: Thanks for your comments.
Are you expecting this method to be like this below?
public MutableBiMap<K, V> withMapIterable(AbstractSynchronizedMapIterable <? extends K, ? extends V> mapIterable)
I used MapIterable instead of AbstractSynchronizedMapIterable as it is overriding default method from MutableMap. Did I miss something here? Please correct me if I am wrong.

nsardana-bny:
Thanks @nikhilnanivadekar for all your comments. I am adding a new test to FixedSizeMapFactoryTest.java. I am also changing the wording on the commit message as per the contribution guidelines.  Just need a clarification on the comment above and I will update the PR together with all the changes.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
nikhilnanivadekar: Sorry, let me try to rephrase it. If you check where withMap() method is implemented, this method should also be implemented in the same class. I would have expected this method to be added in the abstract class. But there might be something I might be missing, so please take a look.

prathasirisha (commented):
Thank you, Neha! LGTM for the changes you made on the Immutable Side.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
nsardana-bny: Thanks for reviewing! 
SynchronizedSortedMap, SynchronizedMutableMap and SynchronizedBiMap, all have custom implementations of withMap() method, so I decided to add withMapIterable() to each of these three classes. Also, AbstractSynchronizedMapIterable, the parent class, is inheriting withMap() method from MutableMapIterable but not overriding in the abstract class itself. In MutableMapIterable, we do have withMap() implemented so I added withMapIterable() to the class. Hope it makes sense. Thanks again!

nsardana-bny:
Hi @nikhilnanivadekar, for putAllMapIterable() method, do you agree with the final look of withMapIterable() method?

for example:-
default MutableMapIterable<K, V> withMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
    {
        this.putAllMapIterable(mapIterable);
        return this;
    }

default void putAllMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
    {
        mapIterable.forEachKeyValue(this::put);
    }

Please confirm if the method signature looks good. Thanks!

nikhilnanivadekar:
The method signature looks good. I am fine to accept it as is for this PR as I would like to close it out soon. But we should add covariant overrides as a separate commit.

nikhilnanivadekar (approved):
LGTM for FixedSizeMap commit

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -172,6 +172,11 @@ default V getOrDefault(Object key, V defaultValue)
         return this;
     }
 
+    default void putAllMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
```
nikhilnanivadekar, line 181: @donraab the return type here is a `void` should we return a `MutableMapIterable` instead of void, i.e. return a `this`?
The way I look at it is that, if we return a `MutableMapIterable` it never hurts, but if we return a void, reassigning in user code is impossible. 
I prefer returning `this`. Thoughts?
Any case, we will take this in as `void` and fix it in a next PR.

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
```diff
@@ -608,6 +608,45 @@ public void withMapNull()
         Assert.assertThrows(NullPointerException.class, () -> this.newMap().withMap(null));
     }
 
+    @Test
+    public void putAllMapIterable()
+    {
+        MutableMapIterable<String, Integer> map = this.newMapWithKeysValues("A", 1, "B", 2);
+        MutableMapIterable<String, Integer> simpleMap = Maps.mutable.with("B", 22, "C", 3);
+        map.putAllMapIterable(simpleMap);
+        Verify.assertMapsEqual(UnifiedMap.newWithKeysValues("A", 1, "B", 22, "C", 3), map);
```
nikhilnanivadekar, line 661: Can be replaced with `Maps.mutable.with()` instead of `UnifiedMap`

nikhilnanivadekar (changes requested):
Please fix the assertion and then this one is good to merge.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -172,6 +172,12 @@ default V getOrDefault(Object key, V defaultValue)
         return this;
     }
 
+    default MutableMapIterable<K, V> withMapIterable(MapIterable<? extends K, ? extends V> mapIterable)
```
nikhilnanivadekar, line 175: I dont think you need default method here because all the implementing interfaces should have a default method. It is easy to check by removing this method, or you can check the coverage and this method will be uncovered in this interface. If it is uncovered you can delete it.
That said, it is good to have default methods to help users with custom implementations, so I would say, you can add a test to cover this method. It can be done separately as a different PR.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
nikhilnanivadekar: Revisiting my point from earlier, I think let us have `putAllMapIterable` return a `void` because we are adding `withMapIterable` which is returns `this`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/SynchronizedBiMap.java
nikhilnanivadekar: The actual reason why you cannot move this implementation to the abstract class is because of the covariant return type. So, you have to add it in the implementation itself.

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
```diff
@@ -608,6 +608,50 @@ public void withMapNull()
...
 
+    @Test
+    public void withMapIterable()
+    {
+        MutableMapIterable<String, Integer> map = this.newMapWithKeysValues("A", 1, "B", 2);
+        MutableMapIterable<String, Integer> simpleMap = Maps.mutable.with("B", 22, "C", 3);
+        map.putAll(simpleMap);
+        MutableMapIterable<String, Integer> mapWith = map.withMapIterable(simpleMap);
+        Assert.assertSame(map, mapWith);
+        Verify.assertMapsEqual(UnifiedMap.newWithKeysValues("A", 1, "B", 22, "C", 3), mapWith);
```
nikhilnanivadekar, line 619: Same comment as the previous commit, change to `Maps.mutable.with()`

unit-tests/src/test/java/org/eclipse/collections/impl/map/immutable/ImmutableMapIterableTestCase.java
```diff
@@ -383,6 +383,81 @@ public void newWithKeyValue()
...
+    public void newWithMapIterable()
+    {
+        ImmutableMapIterable<Integer, String> immutable = this.classUnderTest();
+        ImmutableMapIterable<Integer, String> immutable2 =
+                immutable.newWithMapIterable(Maps.immutable.of(
+                        Integer.MAX_VALUE,
+                        Integer.toString(Integer.MAX_VALUE),
+                        Integer.MIN_VALUE,
+                        Integer.toString(Integer.MIN_VALUE)));
+        Verify.assertSize(immutable.size() + 2, immutable2);
```
nikhilnanivadekar, line 435: I think we need better assertions here. It should be possible to assert with a `MutableMap` no?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
donraab: I think `putAllMapIterable` should be symmetric with `putAll`. What does `putAll` return?

nikhilnanivadekar: `void`

unit-tests/src/test/java/org/eclipse/collections/impl/map/immutable/ImmutableMapIterableTestCase.java
nsardana-bny: Updated the code to handle assert with a MutableMap using Verify.assertMapsEqual() method.

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
nsardana-bny: Updated to use Maps.mutable.with() instead of UnifiedMap methods. Updated the same for putallmapiterable test cases too.

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/MutableMapIterableTestCase.java
nsardana-bny: Updated to use Maps.mutable.with() instead of UnifiedMap methods. Updated the same for putallmapiterable test cases too.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
nsardana-bny: Thanks @nikhilnanivadekar for reviewing. If you don't mind, can I do this on another PR? Thanks again for your comments!

nikhilnanivadekar: Sure!

nikhilnanivadekar (approved):
LGTM! Thanks for your contribution.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/FixedSizeMap.java
```diff
@@ -70,7 +70,18 @@
...
```
motlin, line 74: This is a strange inconsistency we have. `withKeyValue` on a fixed-size map is capable of replacing a value at a key without growing the map, but put() always throws.

mohrezaei: `put()` has to return boolean on a `V` (old value), so it's not possible to have a proper implementation on `FixedSizedMap`. On the other hand, anything that returns a map is entirely implementable. I don't see why we wouldn't implement these...

motlin: I'm not sure I understand this point. Just to focus on put(), I'm wondering why the implementation in SingletonMap isn't something like:

```java
    @Override
    public V put(K key, V value)
    {
        if (Objects.equals(key, this.key1))
        {
            V result = this.value1;
            this.value1 = value;
            return result;
        }
        throw new UnsupportedOperationException("Cannot call put() on " + this.getClass().getSimpleName());
    }
```

The actual implementation in the superclass is:
```java
    @Override
    public V put(K key, V value)
    {
        throw new UnsupportedOperationException("Cannot call put() on " + this.getClass().getSimpleName());
    }
```

mohrezaei: I don't see an issue with changing `put` to allow mutation of the value for the same key. (My point was that `put` can't work for the general case, i.e., an arbitrary key). The other methods work just fine for arbitrary parameters, so throwing is particularly strange there.

For some reason, I thought the small maps were immutable, probably remembering really ancient code? On the mutable side, mutations an be allowed as long as the map size doesn't change.

motlin: Tagging @nsardana-bny @nikhilnanivadekar @donraab for attention.

donraab: These classes are ancient and not consistent with their mutable or immutable counterparts. The small maps are immutable. The small lists are not. I took a look at `FixedSizeList` and `FixedSizeSet` to see how their `withAll` methods are implemented. I think we should make the `withMapIterable` method consistent and return a new `Map`, except maybe in the case where the map passed in is empty. I'm not keen on the empty handling below, but might be kind of hard to change this behavior if it is depended on anywhere.

FixedSizeSet:

```
public MutableSet<T> withAll(Iterable<? extends T> elements)
{
    if (Iterate.isEmpty(elements))
    {
        return this;
    }
    return Sets.fixedSize.ofAll(this.toList().withAll(elements));
}
```
FixedSizeList:

```
public MutableList<T> withAll(Iterable<? extends T> elements)
{
    if (Iterate.isEmpty(elements))
    {
        return this;
    }
    return Lists.fixedSize.ofAll(this.toList().withAll(elements));
}
```

We should also change the result type for both of these methods (and the other with method) to be `FixedSizeList` or `FixedSizeSet`.

mohrezaei: > The small maps are immutable.

That's what I vaguely remembered, but I then checked the implementation. Not only does `SingletonMap` extend mutable super classes, but already has a method that mutates its value:

```java
    @Override
    public MutableMap<K, V> withKeyValue(K addKey, V addValue)
    {
        // Map behavior specifies that if you put in a duplicate key, you replace the value
        if (Objects.equals(this.key1, addKey))
        {
            this.value1 = addValue;
            return this;
        }
        return new DoubletonMap<>(this.key1, this.value1, addKey, addValue);
    }
```
(I just realized @motlin pointed out the same thing above).
Given that these are in the mutable side and are already mutable, I would suggest we should make them as consistent with the rest of the mutable implementations as possible.
The mutable map `with` methods mutate the instance, so I have no issue with `withAll` empty checks.

PR #1030, opened 2020-11-08
Update version numbers and add links to 10.x release notes in README.md compatibility matrix.

donraab:

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Eclipse Collections is compatible with Java 8+. Eclipse Collections is a part of
 
 ## Eclipse Collections and JDK Compatibility Matrix
 
-|EC     | 7.x.x | 8.x.x | 9.x.x |10.0.0 |10.0.1 |10.0.2 |10.0.3 |10.0.4 |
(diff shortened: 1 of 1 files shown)


nikhilnanivadekar:
The purpose of this commit is to: Is it update version numbers or it is link to release blogs?

nikhilnanivadekar:
I don’t think these links are going to be discoverable while reading the ReadMe. Instead I think we should update the release notes to link to the blog.

donraab:
Why not do both (this table and release notes)? We could just limit it to 10.x in the table today. I added the links as it made more sense to add information to the 10.x versions as they were being called out individually but only showing check marks added no value.

nikhilnanivadekar:
I would rather then update the link to direct to release notes than the blog. The reason for that is it remains static and we can update the readme during the release prep and always have an easily derivable link. It’s a pretty common workflow for developers to refer to Releases tab when they are looking for details of a specific release.

donraab:
Ok, I will change the link to the release notes.

nikhilnanivadekar:
Thank you! Please update the commit message as well.

donraab:
Should be ready to merge.

nikhilnanivadekar (approved):

PR #1032, opened 2020-11-14
Port optimizations for primitive set operations to SetIterables.

donraab:

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/SetIterables.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/SetIterables.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 Goldman Sachs and others.
+ * Copyright (c) 2020 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -18,7 +18,6 @@ import org.eclipse.collections.api.set.ImmutableSet;
 import org.eclipse.collections.api.set.MutableSet;
 import org.eclipse.collections.api.set.SetIterable;
 import org.eclipse.collections.api.tuple.Pair;
-import org.eclipse.collections.impl.set.mutable.SetAdapter;
 import org.eclipse.collections.impl.set.mutable.UnifiedSet;
 import org.eclipse.collections.impl.set.strategy.mutable.UnifiedSetWithHashingStrategy;
 import org.eclipse.collections.impl.tuple.Tuples;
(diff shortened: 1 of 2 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/SetIterables.java
```diff
@@ -51,8 +52,16 @@ private SetIterables()
             SetIterable<? extends E> setB,
             R targetSet)
     {
-        Iterate.addAllIterable(setA, targetSet);
-        Iterate.addAllIterable(setB, targetSet);
+        if (setA.size() > setB.size())
```
mohrezaei, line 55: the order doesn't seem to make any difference here...

mohrezaei (changes requested):

mohrezaei:
For optimization code, it's good to add an acceptance test.

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/SetIterables.java
donraab: Yes, I changed back to optimizing `union` so I can use `UnifiedSet.newSet()`, and reverted `unionInto`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/SetIterables.java
```diff
@@ -43,7 +42,13 @@ private SetIterables()
             SetIterable<? extends E> setA,
             SetIterable<? extends E> setB)
     {
-        return SetIterables.unionInto(setA, setB, UnifiedSet.newSet());
+        if (setA.size() > setB.size())
+        {
+            UnifiedSet<E> result = UnifiedSet.newSet(setA);
```
mohrezaei, line 47: The following code should reduce bytecode size and increase inlining chance:
```java
if (setA.size() > setB.size())
{
    SetIterable<? extends E> tmp = setA;
    setA = setB;
    setB = tmp;
}
```
(and below)

mohrezaei (changes requested):

prathasirisha:
Hi @donraab, 

1. Noticed that the javadoc for `intersect` is a bit inaccurate. With the changes you are making, this line is not valid - `The output will contain instances from {@code this}, not {@code set}.`

2. copyright information needs to be updated.

donraab:
Hi @prathasirisha, I wil fix the Javadoc in `SetIterable` in a separate PR. I'm only focusing changes in the `SetIterables` utility in this PR. I'm adding a test for the class as well since there was no dedicated test class. I've ported the tests you added for the primitive union, intersect and difference, and added a test for validating the performance improvement of `union` over `unionInto`.

mohrezaei (approved):

donraab:
The TeamCity inspections 2 build is failing for a seemingly unrelated change.

PR #1035, opened 2020-11-26
Add toImmutableList/Set/Bag to RichIterable interface.

donraab:

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Goldman Sachs and others.
+ * Copyright (c) 2021 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -26,6 +26,7 @@ import java.util.stream.Collector;
 import java.util.stream.Stream;
 
 import org.eclipse.collections.api.bag.Bag;
+import org.eclipse.collections.api.bag.ImmutableBag;
 import org.eclipse.collections.api.bag.MutableBag;
 import org.eclipse.collections.api.bag.MutableBagIterable;
 import org.eclipse.collections.api.bag.sorted.MutableSortedBag;
@@ -58,7 +59,10 @@ import org.eclipse.collections.api.collection.primitive.MutableIntCollection;
 import org.eclipse.collections.api.collection.primitive.MutableLongCollection;
 import org.eclipse.collections.api.collection.primitive.MutableShortCollection;
 import org.eclipse.collections.api.factory.Bags;
+import org.eclipse.collections.api.factory.Lists;
 import org.eclipse.collections.api.factory.Maps;
+import org.eclipse.collections.api.factory.Sets;
+import org.eclipse.collections.api.list.ImmutableList;
 import org.eclipse.collections.api.list.MutableList;
 import org.eclipse.collections.api.map.MapIterable;
 import org.eclipse.collections.api.map.MutableMap;
@@ -70,6 +74,7 @@ import org.eclipse.collections.api.multimap.Multimap;
 import org.eclipse.collections.api.multimap.MutableMultimap;
 import org.eclipse.collections.api.ordered.OrderedIterable;
 import org.eclipse.collections.api.partition.PartitionIterable;
+import org.eclipse.collections.api.set.ImmutableSet;
 import org.eclipse.collections.api.set.MutableSet;
 import org.eclipse.collections.api.set.sorted.MutableSortedSet;
 import org.eclipse.collections.api.tuple.Pair;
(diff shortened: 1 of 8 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -2487,4 +2490,34 @@ default void appendString(Appendable appendable, String separator)
...
+
+    /**
+     * Converts the RichIterable to the default ImmutableBag implementation.
+     *
+     * @since 11.0
+     */
+    default ImmutableBag<T> toImmutableBag()
+    {
+        return this.toBag().toImmutable();
+    }
```
motlin, line 2522: These should live with the other to*() methods, probably right after toBiMap().

The other to*() methods are not default - why the difference?

motlin: toImmutableSortedSet(), toImmutableSortedBag(), toImmutableMap(), toImmutableSortedMap()? Separate PR?

motlin: Why prefer `this.to<type>().toImmutable()` over `<type>s.immutable.withAll(this)`?

mohrezaei: `to<type>()` implementations appear to have various overrides, so the code is not the same.

donraab: @motlin Yes, I was planning for separate PRs for the other to methods since they have more overrides. I wanted to see if these made sense first.

donraab: > Why prefer this.to<type>().toImmutable() over <type>s.immutable.withAll(this)?

My gut feel is that this made the most sense for the default case, but happy to change the default implementation. I'm also curious if there might be opportunities for optimizations in various implementations.

motlin: Both have different opportunities for optimization. My thought is that if both are as optimized as possible, the `this.to<type>().toImmutable()` form will still make one more copy.

It kind of doesn't matter. Even if these methods are "slow" I'd still use them. It's just hard to see in a code review where these methods ought to be overridden imagining if they were not default methods.

donraab: > Both have different opportunities for optimization. My thought is that if both are as optimized as possible, the `this.to<type>().toImmutable()` form will still make one more copy.
> 
> It kind of doesn't matter. Even if these methods are "slow" I'd still use them. It's just hard to see in a code review where these methods ought to be overridden imagining if they were not default methods.

For this PR I was really just looking for confirmation that these method names and signatures make sense. We can choose to optimize them by reducing copying. This will further illustrate their value and support the cost of adding them.

donraab: @motlin @mohrezaei I'm going to change the code to use Craig's suggested approach. Another benefit is that all of the factory methods are optimized for empty.

mohrezaei: `withAll` creates an intermediate array. It would be best to get rid of that first. There is some optimization around small sets that appears to be buggy and should be fixed at the same time. Instead of creating an array, the iterable should be used to construct a UnifiedSet. If the set size is small, then it should be converted to a small immutable set, otherwise wrapped to immutable.
bug: pass in a list of size 1000 that has the same item. Instead of a singleton immutable set, the current code creates an wrapped unified set with a single entry.

donraab: @mohrezaei These changes definitely make sense. I think we need to revisit the `withAll` implementations for the immutable factories across the board. I will create issues for these.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/ImmutableBag.java
```diff
@@ -219,4 +219,15 @@
...
+    /**
+     * Overrides toImmutableBag in RichIterable to return this.
+     *
+     * @since 11.0
+     */
+    @Override
+    default ImmutableBag<T> toImmutableBag()
+    {
+        return this;
+    }
```
motlin, line 232: Overrides like this don't hurt, but aren't really necessary, since withAll() has the same optimization.

```java
    @Override
    public <T> ImmutableBag<T> withAll(Iterable<? extends T> items)
    {
        if (items instanceof ImmutableBag<?>)
        {
            return (ImmutableBag<T>) items;
        }
        if (items instanceof Bag<?>)
        {
            Bag<T> bag = (Bag<T>) items;
            if (bag.isEmpty())
            {
                return this.with();
            }
            if (bag.size() == 1)
            {
                return this.with(bag.getFirst());
            }
            if (bag.sizeDistinct() < ImmutableArrayBag.MAXIMUM_USEFUL_ARRAY_BAG_SIZE)
            {
                return ImmutableArrayBag.copyFrom(bag);
            }
            return new ImmutableHashBag<>(bag);
        }
        return this.of((T[]) Iterate.toArray(items));
    }
```

donraab: @motlin I prefer the clarity of the direct return.

donraab:
@motlin @mohrezaei @prathasirisha @nikhilnanivadekar I believe there was general agreement on the addition of these APIs. Is there anything in this PR that still needs to be addressed?

mohrezaei (approved):

prathasirisha (approved):

PR #1052, opened 2021-01-05
Closes #1044: Adds toSortedList(Comparator) and toSortedListBy(Function) to primitive Iterables

### LINKED ISSUE #1044: Add toSortedList(primitive Comparator) and toSortedListBy(primitive Function) to primitive Iterables
We have `sortThis(primitive Comparator)` and `sortThisBy(primitive Function)` which mutate the underlying list but do not have the corresponding `toSortedList` and and `toSortedListBy` methods that return a new list.

vmzakharov:
Supports `toSortedListBy` implementations on primitive Itrerables that accept one of the following:
- a primitive comparator
- a primitive to object function that takes the elements of the iterable
- a primitive to object function that takes iterable elements and a comparator applied to the results of the function
These methods convert the collection to a corresponding primitive Mutable List implementation, which is sorted based on the primitive comparator , the natural order of the values returned by the function, or the comparator applied to the results returned by the function respectively.

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/block/function/primitiveComparator.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/block/function/primitiveComparator.stg
@@ -25,7 +25,7 @@ import java.io.Serializable;
  *
  */
 @FunctionalInterface
-public interface <name>Comparator\<T>
+public interface <name>Comparator
         extends Serializable
 {
     int compare(<type> value1, <type> value2);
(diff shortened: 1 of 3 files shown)


eclipse-collections-code-generator/src/main/resources/api/block/function/primitiveComparator.stg
```diff
@@ -25,7 +25,7 @@ import java.io.Serializable;
  *
  */
 @FunctionalInterface
-public interface <name>Comparator\<T>
+public interface <name>Comparator
```
donraab, line 28: Interesting, I guess we missed the unused generics on the initial commit. Good Catch.

donraab (approved):
LGTM. Commented on the removal of unused generics on primitive comparator interface, but this should be fine. Thanks for the contribution!

PR #1056, opened 2021-01-26
Add missing overrides of aggregateBy().

motlin:


eclipse-collections-api/src/main/java/org/eclipse/collections/api/bimap/ImmutableBiMap.java
```diff
@@ -134,6 +138,36 @@
         return this.groupByUniqueKey(function, target).toImmutable();
     }
 
+    @Override
```
donraab, line 141: Hi @motlin, what do you think about adding since tags to these methods so we can easily see when they were added?

donraab (changes requested):
LGTM. One minor question on adding since tags so we know when these were added.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bimap/ImmutableBiMap.java
motlin: Done

donraab (approved):
LGTM.

PR #1057, opened 2021-01-27
Add missing overrides of aggregateInPlaceBy().

motlin:


eclipse-collections-api/src/main/java/org/eclipse/collections/api/bimap/ImmutableBiMap.java
```diff
@@ -134,6 +139,22 @@
         return this.groupByUniqueKey(function, target).toImmutable();
     }
 
+    @Override
+    default <KK, VV> ImmutableMap<KK, VV> aggregateInPlaceBy(
```
donraab, line 143: Hi @motlin, please add since tags as you did for aggregateBy.

donraab (changes requested):

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bimap/ImmutableBiMap.java
motlin: done

donraab (approved):
LGTM.

PR #1058, opened 2021-02-03
Use Maven Locations for p2 update site build

fipro78:
With this pull request I would like to simplify the build of the p2 update site.

Right now there is a p2-repository module in the main build that uses Eclipse Bundle Recipes (EBR) to re-bundle the Eclipse Collections artifacts (api and impl) into a single bundle and adds the proper OSGi meta-data. In a second step the p2-feature build consumes that locally build update site and creates a feature and another update site with that feature. The resulting update site from this build is then published.

With the new Maven Location types that are supported since Tycho 2.2 this process can be simplified and the result is more consistent to the original deployment.

1. The OSGi meta-data is already included in the original artifacts by using the bnd-maven-plugin. This way the artifacts in Maven Central contain the correct OSGi meta-data. Re-bundling is therefore not necessary anymore. The p2-repository module and the usage of EBR can be dropped.
2. The p2-feature build can be separated from any local build results. It was already executed after the main build, and that needs to be kept. For this the following modifications are done:
    * introduce a new module org.eclipse.collections.target that uses the new Maven Location to consume the original Eclipse Collections artifacts from Maven
    * the org.eclipse.collections.feature is updated to contain the original api and impl modules instead of the re-bundled artifact created by EBR
    * the target-platform-configuration is now configured to use the target definition instead of repositories

To make this work the Tycho version needs to be updated to 2.2, which then needs a Java 11 for the build execution. But as the p2-feature build is separated this should be ok as it does not affect the main build.

With this change the usage of the p2-repository module result can be dropped. Actually the whole p2-repository module with the EBR usage can be dropped. Unfortunately the builds verification fails if I remove it. So this obviously needs to be done afterwards.

The implication of this change is that the org.eclipse.collections bundle produced by EBR is not available anymore. So users that used Require-Bundle on that bundle will get errors and need to update. But IMHO people should use Import-Package to avoid such issues, and it will be more consistent that there are separate api and impl bundles to be consumed instead of an additional special bundle.

With the Maven Location it is of course also possible for consumers in the future to directly consume from Maven Central, but IMHO providing a p2 update site with a feature is still useful for Eclipse application developers. So providing it further is still useful.

### DIFF
--- a/.github/workflows/earlyaccess-javadoc.yml
+++ b/.github/workflows/earlyaccess-javadoc.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        java-version: [22]
+        java-version: [23]
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/earlyaccess-unit.yml
+++ b/.github/workflows/earlyaccess-unit.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        java-version: [22]
+        java-version: [23]
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/unit-no-p2.yml
+++ b/.github/workflows/unit-no-p2.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        java-version: [17, 21]
+        java-version: [17, 21, 22]
 
     runs-on: ${{ matrix.os }}
 


nikhilnanivadekar:
@prathasirisha can you please take a look at this and close it out?

nikhilnanivadekar:
Hi @fipro78 couple of questions:
1) instead of creating a module in this repo since the builds are all separate would it make sense to create a completely new repo for p2 stuff?
2) do you have an example build or any build configurations which I would have to do presumably on Eclipse CI to do the separation.

To be honest, I am not an expert in anything related to p2 and last time I did this @ujhelyiz helper me. So I’ll need your help to understand what needs to change on Eclipse CI.
Lastly can @ujhelyiz or @guw review the PR since I would like it to be reviewed by someone who knows/understands p2.

Thanks for your patience and contribution!

fipro78:
> 1. instead of creating a module in this repo since the builds are all separate would it make sense to create a completely new repo for p2 stuff?

That depends on how you want to manage the project. With this change the build of the p2 repository is completely separated from the artifacts build. There are no local file dependencies anymore, so you could do this. You only need to take care of the build sequence, as you first need to build and deploy the artifacts to a Maven repository, then you call the p2 build that consumes those artifacts. I think it should be possible to create some pipeline build to ensure the sequence. That should even work with separate repositories. But to be honest, I am not sure if it is worth creating a new repository just for the build of the p2 update site. I think this is more a project management decision on the code structure. If you move it out to another repository, IMHO it will increase the amount of work on your side, as you need to manage two repositories for updating versions. Right now you have some scripts that go over the whole code base right?

> 2. do you have an example build or any build configurations which I would have to do presumably on Eclipse CI to do the separation.

Unfortunately not. You are the first project I encountered that first builds the Maven artifacts, and then at the same time create a p2 update site. The most Maven centric projects typically don't care about p2 update sites, which then causes the create of EBR projects and deployment in Eclipse Orbit, so they can be consumed by Eclipse RCP projects. And Eclipse centric projects typically only create p2 update sites and don't deploy on Maven Central. And projects that also deploy to Maven Central still have that in one build process (build artifacts -> build p2 -> deploy to Maven Central).
Eclipse Collections would be some sort of pioneer in that area, that could inspire other projects. ;-)

> To be honest, I am not an expert in anything related to p2 and last time I did this @ujhelyiz helper me. So I’ll need your help to understand what needs to change on Eclipse CI.

Well I don't have access to your builds. But I suppose you already have two separate builds on the Eclipse CI. One for building the artifacts and one for building the p2 site. At least from looking into the pom.xml files I assume this, as the p2 build was not part of the main build (not listed in the modules). So probably you don't have to change much. The only thing is that the p2 build needs to be executed with Java 11 because Tycho 2.2 needs it. But that should not bother the main artifacts build. But you can drop the p2-repository module and therefore the whole Tycho dependency via EBR in the main artifacts build.

> Lastly if possible, can @ujhelyiz or @guw please review the PR since I would like it to be reviewed by someone who knows/understands p2.

Hope that helps. IMHO the change should be unintrusive related to your current build structure on the Eclipse CI. If you want to split the sources in different repositories, well then you need to manage that.

nikhilnanivadekar:
Thanks for the clarification. Now it makes it a bit more clear for my inexperienced brain 😄 

How can I give you access to our CI? If there is a way happy to give you access. If that’s not possible, I am happy to have a live working session once things calm down a little bit on the work front. We can create a milestone release and fix issues if any during the release. 
thoughts?

Lastly, should we rename it from target to something else? May be can we reuse the existing p2 modules which might be left obsolete after this change? Or I got this completely wrong? Thanks for your patience and cooperation!

fipro78:
I am not sure if you can simply grant me access to the whole Eclipse Collections Jenkins instance. But IIRC you can grant me access on the jobs itself. Not sure if typically it is possible to inspect the build configuration even as anonymous user. Could be that the read access for anonymous users are enabled by default. If you post the link to your Eclipse CI server I can have a look what I am able to see.

Why do you want to rename the target module? It contains the Target Definition that is used as base for the Tycho build. That is specific to Tycho/p2, but the naming is kind of best practice. This way it is clear that this module contains the target definition. Via a target definition you specify the artifacts that should be consumed by the build in a more granular way. Compared with the specification of repositories you can restrict for example the versions that should be consumed. And in our specific case we can even directly consume artifacts from Maven Central and do not need a p2 repository in between.

ujhelyiz:
Sorry for the late response - I have read about the Maven Location feature, but I am not familiar with it, so I don't know how well it works. If it can be used to replace the EBR-generated bundle, and simply provide the feature-based site that works it would be very helpful.

For now, the feature-based build is entirely separate from the normal build, as Tycho required the bundle to already exist when the build was started; however, I don't see how to create a build of the feature with the latest snapshot (very useful for testing; the latest snapshot is almost certainly not added to Maven central too early).

Finally, if you have an eclipse.org user, you could use that to log in to the server itself - I am not sure whether you have to be a committer for at least a single project or you only need to have an eclipse.org user. If you are only interested in the current configuration, that should be enough.

fipro78:
Well it consumes artifacts from a maven repository. So if you build the snapshot via mvn install, a follow up build on the same machine should be able to consume the artifacts from the local repository.

nikhilnanivadekar:
Hi, any consensus on this? I am happy to leave it as it is.

fipro78:
@nikhilnanivadekar 
Is there anything I can do here? Not sure about the state of this discussion?

PR #1059, opened 2021-02-05
Fix types on Multimap.*MultiValues().

motlin:
It occurred to me that the type `C` in `AbstractMultimap<K, V, C extends RichIterable<V>>` extends RichIterable rather than Iterable. Unless we want to leave room for future Multimaps where the value types aren't ours, we can be more specific with our types.


prathasirisha (approved):

prathasirisha:
Hi @motlin - if you rebase this PR, I am happy to merge this in!

motlin:
@prathasirisha done!

PR #1063, opened 2021-02-11
Add badges pointing to javadoc.io.

motlin:


motlin:
Any interest in badges like this? They point to the latest version at all times - so right now they point at milestone releases while the README points at a stable release.

prathasirisha:
I like it! My only comment would be to distinguish between the names of the two javadoc badges. 
Something like this, perhaps - https://javadoc.io/badge2/org.eclipse.collections/eclipse-collections-api/javadoc_eclipse_collections_api.svg? 

I tested it out in my IDE real quick - 

image

motlin:
Good idea, thanks for the suggestion @prathasirisha. Made the change.

prathasirisha (approved):

Issue #1046, opened 2020-12-23
Add toSortedMap and toSortedMapBy to Collectors2


donraab:
The methods `toSortedMap` and `toSortedMapBy` exist on `RichIterable` but there is no equivalent methods on the `Collectors2` utility class.

rajivsethumadhavan:
@donraab  - I'd like to work on this. Can you please assign it to me?

nikhilnanivadekar:
Done @rajivsethumadhavan

PR #1069, opened 2021-02-26
Add toSortedMap and toSortedMapBy to Collectors2 #1046

jeshet:
Signed-off-by: Jacob Eshet <jeshet123@gmail.com>


donraab (changes requested):
Hi @jeshet, thank you for the contribution! Please add tests for these methods in the corresponding `Collectors2Test` class.

jeshet:
Amazing. I've failed to remove a single line after 7 commits. I'm having trouble signing off on merge commits and I think I'll stop now before we end up with 50. This was a rather small contribution so feel free to do anything with the code I provided.

PR #1071, opened 2021-03-03
Add toImmutableSortedBag/List/Set with Comparator and toImmutableSort…

donraab:
…edBagBy/ListBy/SetBy with Function to RichIterable.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
@@ -1793,6 +1793,27 @@ public interface RichIterable<T>
     }
 
     /**
+     * Converts the collection to an ImmutableList implementation and sorts it using the specified comparator.
+     *
+     * @since 11.0
+     */
+    default ImmutableList<T> toImmutableSortedList(Comparator<? super T> comparator)
+    {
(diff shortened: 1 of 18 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -1792,6 +1792,27 @@ default T detectIfNone(Predicate<? super T> predicate, Function0<? extends T> fu
...
+     *
+     * @since 11.0
+     */
+    default ImmutableList<T> toImmutableSortedList(Comparator<? super T> comparator)
+    {
+        return Lists.immutable.withAllSorted(comparator, this);
+    }
+
+    /**
+     * Converts the collection to a MutableList implementation and sorts it based on the natural order of the
```
prathasirisha, line 1806: Minor comment - This should be ImmutableList in javadoc.

prathasirisha (changes requested):

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
donraab: Good catch, thank you! Will fix now.

prathasirisha (approved):
LGTM

PR #1090, opened 2021-05-01
change signature of pops methods in ImmutableStack

### LINKED ISSUE #1048: ImmutableStack should return Pair for pop and pop(int)
The primary purpose of `pop` is to remove and return the thing(s) you want from the top of a stack.

Today the signature of `pop` and `pop(int)` on ImmutableStack are as follows:

```
ImmutableStack<T> pop();

ImmutableStack<T> pop(int count);
```
A new ImmutableStack instance is returned with the element(s) removed, but they element(s) are not returned. .

I believe the signatures should be:
```
Pair<T, ImmutableStack<T>> pop();

Pair<ListIterable<T>, ImmutableStack<T>> pop(int count);
```
Today, these APIs are inconsistent with `peek` and `peek(int)`.

raffopiteco:
Closes #1048

Signed-off-by: Roberto Falzarano <robertofalzarano@gmail.com>


raffopiteco:
I wrote the signature in the wrong format

PR #1091, opened 2021-05-01
change signature of pops methods in ImmutableStack

### LINKED ISSUE #1048: ImmutableStack should return Pair for pop and pop(int)

raffopiteco:
Closes #1048

Signed-off-by: Roberto Falzarano <robertofalzarano@gmail.com>

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/stack/ImmutableStack.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/stack/ImmutableStack.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs and others.
+ * Copyright (c) 2022 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -28,6 +28,7 @@ import org.eclipse.collections.api.block.predicate.Predicate2;
 import org.eclipse.collections.api.block.procedure.Procedure;
 import org.eclipse.collections.api.block.procedure.Procedure2;
 import org.eclipse.collections.api.factory.Maps;
+import org.eclipse.collections.api.list.ListIterable;
 import org.eclipse.collections.api.map.ImmutableMap;
 import org.eclipse.collections.api.map.MutableMap;
 import org.eclipse.collections.api.map.primitive.ImmutableObjectDoubleMap;
@@ -51,6 +52,22 @@ public interface ImmutableStack<T> extends StackIterable<T>
     ImmutableStack<T> pop();
 
     ImmutableStack<T> pop(int count);
+
+    /**
+     * @since 11.1
+     */
+    default Pair<T, ImmutableStack<T>> peekAndPop()
+    {
+        throw new UnsupportedOperationException("Default method to ensure backwards compatibility");
+    }
+
+    /**
+     * @since 11.1
+     */
+    default Pair<ListIterable<T>, ImmutableStack<T>> peekAndPop(int count)
+    {
+        throw new UnsupportedOperationException("Default method to ensure backwards compatibility");
+    }
 
     @Override
     ImmutableStack<T> takeWhile(Predicate<? super T> predicate);
--- a/unit-tests-java8/src/test/java/org/eclipse/collections/test/stack/immutable/ImmutableStackTestCase.java
+++ b/unit-tests-java8/src/test/java/org/eclipse/collections/test/stack/immutable/ImmutableStackTestCase.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs.
+ * Copyright (c) 2022 Goldman Sachs.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -15,6 +15,7 @@ import java.util.Iterator;
 
 import org.eclipse.collections.api.factory.Stacks;
 import org.eclipse.collections.api.stack.ImmutableStack;
+import org.eclipse.collections.api.tuple.Pair;
 import org.eclipse.collections.test.stack.StackIterableTestCase;
 import org.junit.Test;
 
@@ -51,6 +52,31 @@ public interface ImmutableStackTestCase extends StackIterableTestCase
     {
         ImmutableStack<Integer> immutableStack = this.newWith(5, 1, 4, 2, 3);
         ImmutableStack<Integer> emptyStack = immutableStack.pop().pop().pop().pop().pop();
+        assertSame(Stacks.immutable.with(), emptyStack);
+        assertThrows(EmptyStackException.class, emptyStack::pop);
+    }
+
+    @Test
+    default void MutableStack_peekAndPop()
+    {
+        ImmutableStack<Integer> immutableStack = this.newWith(5, 1, 4, 2, 3);
+        Pair<Integer, ImmutableStack<Integer>> elementAndStack = immutableStack.peekAndPop();
+        Integer poppedElement = elementAndStack.getOne();
+        ImmutableStack<Integer> poppedStack = elementAndStack.getTwo();
+        assertEquals(5, poppedElement);
+        assertEquals(Stacks.immutable.withReversed(1, 4, 2, 3), poppedStack);
+        assertEquals(Stacks.immutable.withReversed(5, 1, 4, 2, 3), immutableStack);
+    }
+
+    @Test
+    default void ImmutableStack_peekAndPop_throws()
+    {
+        ImmutableStack<Integer> immutableStack = this.newWith(5, 1, 4, 2, 3);
+        ImmutableStack<Integer> emptyStack = immutableStack.peekAndPop().getTwo()
+                .peekAndPop().getTwo()
+                .peekAndPop().getTwo()
+                .peekAndPop().getTwo()
+                .peekAndPop().getTwo();
         assertSame(Stacks.immutable.with(), emptyStack);
         assertThrows(EmptyStackException.class, emptyStack::pop);
     }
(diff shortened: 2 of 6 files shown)


raffopiteco:
Hi to everyone, this my first contribution to the project. I have some trouble with the ECA. As you can see, the error message says that I'm not covered by necessary legal agreements even if my account on the Eclipse Foundation platform as the green mark for the ECA. I signed the ECA version 3.1.0.

mohrezaei:
That's because the commit data has this:
From: Roberto Falzarano <roberto.falzarano@intelligentia.it>
which I presume is not the right email that you registered with Eclipse with.

You have to set your git email properly before committing. Also, it's not necessary to close and open new PRs just to change such things. Use the proper git commands locally to set it right and then force push that to your fork. The PR will update automatically.

raffopiteco:
Thank you for your suggestion. I didn't realize that the commit author was wrong.

motlin:

The changes to `pop(length)` have a nice efficiency benefit - we only traverse the linked list once.

The changes to `pop()` aren't as clear to me.

Before:
```java
var top = stack.peek();
stack = stack.pop();
```

After:
```java
var pair = stack.pop();
var top = pair.getOne();
stack = top.getTwo();
```

So the second example isn't a clear readability win. It's also not a clear efficiency win, since it creates a pair. Note that the first example isn't subject to race conditions or anything like that since the stack is immutable. So I think I prefer `pop()` without a length as-is. Thoughts?

raffopiteco:
Hi @motlin, your reasoning tells me that also the change to `pop(length)` is not so good. 

Before:
```java
var topItems = stack.peek(length);
stack = stack.pop(length);
```
After:
```java
var pair = stack.pop(length);
var topItems = pair.getOne();
stack = pair.getTwo();
```

Also in this case the second example should have the same drawbacks. I think that the goal is to have the two methods that almost have the same behavior.

donraab:
@raffopiteco @motlin, I discussed with @nikhilnanivadekar today and we think it is better to add these as new APIs and to document the current API and usage pattern described by @motlin for the current APIs in the JavaDoc. We can also deprecate one or more of the old methods if we thing they are suboptimal. This is a safer way to the evolve API than changing the existing API.

Current:

```
var top = stack.peek();
stack = stack.pop();

var topItems = stack.peek(length);
stack = stack.pop(length);
```
New:

```
var pair = stack.pop2();
var top = pair.getOne();
stack = top.getTwo();

var pair = stack.pop2(length);
var topItems = pair.getOne();
stack = pair.getTwo();

```
Thoughts? Naming is hard, so definitely open to a better name than `pop2`.

raffopiteco:
This is the best arrangement you can find.  
Since `pop()` returns an instance of `ImmutableStack`, the new one could be named `popAndGet()` to emphasize that it also returns the elements at the top.

donraab:
@raffopiteco I think `popAndGet` might be fine to use as the name. @motlin @nikhilnanivadekar thoughts?

So we would have two new methods:

```
Pair<T, ImmutableStack<T>> popAndGet();
Pair<ListIterable<T>, ImmutableStack<T>> popAndGet(int count);
```

motlin:
Or peekAndPop

raffopiteco:
Just to sum up so I'll start working on the changes:

1. revert the `pop()` and `pop(length)` implementation;
2. implements the following methods for the `ImmutableStack` interface: `Pair<T, ImmutableStack<T>> peekAndPop();` and `Pair<ListIterable<T>, ImmutableStack<T>> peekAndPop(int count);`;
2.1. In `ImmutableStack` we can have a default implementation that makes use of underlying `peek()` and `pop()` methods;
2.2. `ImmutableArrayStack` is deprecated, so here we can throw `UnsupportedOperationException`;
2.3. `ImmutableNotEmptyStack` will have it own implementation;
2.4. In `ImmutableEmptyStack` the `EmptyStackException` will be thrown. But here I have a doubt, does `peekAndPop(int count)` always throw the `EmptyStackException`? Or should have the same behavior of `pop(length)`? Because in this case the stack is always empty, so could be reasonable to throw the exception in any case;
3. The current API and usage pattern described by @motlin, should be documented in this issue? Or a different one will be created? If it should be documented in this issue, could the JavaDoc comments be applied on the interface?
4. Have you thought about which methods should be deprecated?

Analyzing the code I see that `ImmutableEmptyStack.pop(int count)` returns the instance of the object. Shouldn't it always throw `EmptyStackException` as `ImmutableEmptyStack.pop()` does?

I wait for a feedback before I start working on it.

motlin:
My opinions... feel free to debate:

peekAndPop() methods should be default methods, where the default
implementation actually calls peek and pop, and wraps the two results in a
pair.

peekAndPop() should be overridden in the linked-list implementation for
efficiency. By usually delegating to peek() and pop(), we’ll get the same
exceptions thrown in the same situations. However, if the error message
says something like “cannot call peek on an empty stack” then we might want
to override just to correct the method name.

No methods should be deprecated.

The documentation with usage patterns is nice to have in any PR. It doesn’t
have to be the same PR. The javadoc belongs on the interface.

RELEASE_NOTE_DRAFT.md
```diff
@@ -31,6 +31,7 @@ The Eclipse Collections team gives a huge thank you to everyone who participated
 * Added toSortedMap and toSortedMapBy to Collectors2.
 * Added Norwegian translation for the Eclipse Collections website.
 * Added ClassComparer utility. 
+* Added peekAndPop to ImmutableStack
```
donraab, line 34: I think this can be removed for this PR as we groom these during release process from the commits.

unit-tests-java8/src/test/java/org/eclipse/collections/test/stack/immutable/ImmutableStackTestCase.java
```diff
@@ -14,6 +14,7 @@
 import java.util.Iterator;
 
 import org.eclipse.collections.api.stack.ImmutableStack;
+import org.eclipse.collections.api.tuple.Pair;
```
donraab, line 17: Change copyright to "2022 Goldman Sachs and others."

donraab (changes requested):
Hi @raffopiteco, this looks mostly good to me. Two minor changes suggested. Please update the copyright notices with current year in all of the files as well.

RELEASE_NOTE_DRAFT.md
raffopiteco: Done

donraab (changes requested):
Hi @raffopiteco, thank you for the pushing the latest updates. Since we've already released 11.0, we have two options. Option 1, changes the since tags to 12.0. Option 2, implement the methods on the interfaces as default methods and change the since tag to 11.1. I'm not sure how hard it would be to provide a default implementation of the interface additions. It might be hard because of Pair as a return types. One option is to throw `UnsupportedOperationException` in the default methods.

raffopiteco:
Hi @donraab, looking at `ImmutableStack` interface it seems not possible to return a `Pair` without coupling the api with the implementation.
Following your suggestion I can change the since tag to 11.1 and write a default implementation that throws `UnsupportedOperationException`.

RELEASE_NOTE_DRAFT.md
```diff
@@ -30,7 +30,7 @@ The Eclipse Collections team gives a huge thank you to everyone who participated
 * Added toImmutableSortedMap and toImmutableSortedMapBy to Collectors2.
 * Added toSortedMap and toSortedMapBy to Collectors2.
 * Added Norwegian translation for the Eclipse Collections website.
-* Added ClassComparer utility. 
+* Added ClassComparer utility.
```
donraab, line 33: Please revert this change.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/stack/ImmutableStack.java
```diff
@@ -52,6 +53,20 @@
 
     ImmutableStack<T> pop(int count);
 
+    /**
+     * @since 11.1
+     */
+    default Pair<T, ImmutableStack<T>> peekAndPop() {
```
donraab, line 59: The code formatting is not following the project standard here. Beginning curly brace should be on next line.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/stack/ImmutableStack.java
```diff
@@ -52,6 +53,20 @@
...
+        throw new UnsupportedOperationException("Not implemented yet!");
```
donraab, line 60: Change the message from "Not implemented yet!" to "Concrete class responsible for implementation" since we won't be able to implement this without making it possible to create Pairs in the api module.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/stack/ImmutableStack.java
```diff
@@ -52,6 +53,20 @@
...
+    }
+
+    /**
+     * @since 11.1
+     */
+    default Pair<ListIterable<T>, ImmutableStack<T>> peekAndPop(int count) {
```
donraab, line 66: Same comment on code formatting.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/stack/ImmutableStack.java
```diff
@@ -52,6 +53,20 @@
...
+        throw new UnsupportedOperationException("Not implemented yet!");
```
donraab, line 67: Same comment on message.

donraab (changes requested):
Thank you for the update @raffopiteco! I've requested a few minor changes which after they are made I will approve and merge the PR for 11.1.

donraab:
@raffopiteco We're planning on releasing 11.1 soon so if you can address the minor changes I've requested that would be great. Thanks!

donraab:
Hi @raffopiteco, thanks for updating. Please rebase your PR, and I will approve and merge.

donraab (approved):
LGTM. Thanks for the contribution @raffopiteco!

raffopiteco:
Thanks @donraab for supporting me in this first issue!

PR #1100, opened 2021-05-26
Replace HashingStrategies.longHashCode with java 8 Long.hashCode

yawkat:
The helper method is not necessary anymore now that EC is java 8+ only.


donraab (approved):
LGTM. Thanks for the contribution @yawkat !

PR #1103, opened 2021-06-02
Add converters from Pair and Triple to List types.

donraab:


mohrezaei:
Philosophical question: shouldn't `Twin` and `Triplet` implement unmodifiable list? That would largely obviate this change.

donraab:
It's a great question @mohrezaei, and `Twin` and `Triplet` certainly could implement `List` or `ImmutableList`. Unfortunately that would be a heavier lift than the static converter methods, and wouldn't work with `Pair` or `Triple`, which is what I was looking for in this particular case. I had some code that used `zip` which returns a `Pair`, and I needed to convert it to a `List`. The types happen to be the same, but unfortunately there isn't a `zip` that returns `Twin`. Not sure that would be worth adding.

prathasirisha (approved):

Issue #1104, opened 2021-06-06
Add specialized forEachKey(ValueType) methods to Multimap leaf interfaces


donraab:
There is a method `forEachKeyMultiValues` on `Multimap` with the following signature:

`void forEachKeyMultiValues(Procedure2<? super K, ? super RichIterable<V>> procedure);`

I would like to propose that we add the following methods with specific types to the leaf interfaces in the Multimap hierarchy.

`MutableListMultimap` -> `void forEachKeyMutableList(Procedure2<? super K, ? super MutableList<V>> procedure);`
`ImmutableListMultimap` -> `void forEachKeyImmutableList(Procedure2<? super K, ? super ImmutableList<V>> procedure);`

And so on and so forth for the other leaf interfaces. For the `MutableMultimap` types, the procedure should be passed an unmodifiable wrapper to the collection as is done today with `get()`.

donraab:
Thoughts @motlin @prathasirisha @nikhilnanivadekar? I've started implementing these, and I think it has been a type specific method we have been missing for a long time. It is also safer than `forEachKeyMultiValues` because that method does not wrap the underlying collection in an unmodifiable wrapper (because it doesn't know the type), which means folks can modify the `List` or `Set` directly if they cast it, which we know folks do.

nikhilnanivadekar:
IMO this might be a bit of an overkill. Casting is generally fine. It’s annoying but I think it’s ok.
Curious to hear what @motlin , @mohrezaei , @prathasirisha think.

mohrezaei:
What do the implementations look like? do they have to re-implement every time?

donraab:
AbstractMutableListMultimap:

```
    @Override
    public void forEachKeyMutableList(Procedure2<? super K, ? super MutableList<V>> procedure)
    {
        this.getMap().forEachKeyValue((key, value) -> procedure.value(key, value.asUnmodifiable()));
    }

```
ImmutableListMultimapImpl:

```
    @Override
    public void forEachKeyImmutableList(Procedure2<? super K, ? super ImmutableList<V>> procedure)
    {
        this.getMap().forEachKeyValue(procedure);
    }

```

mohrezaei:
Since the impls are simple (hey, that rhymes :smile:), I think it's fine.

motlin:
[...]

PR #1106, opened 2021-06-13
Add specialized forEachKey(ValueType) methods to List/Set/BagMultimap…

donraab:
… leaf interfaces. Partially addresses #1104.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/multimap/bag/ImmutableBagMultimap.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/multimap/bag/ImmutableBagMultimap.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs.
+ * Copyright (c) 2021 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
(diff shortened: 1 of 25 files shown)


motlin:
It's possible to put these methods on quite a few more interfaces, like ListMultimap.

The tests ought to use static types that match the value collections, rather than `Iterable`.

I still think it makes more sense to use read-only types (like ListIterable) rather than mutable types that will throw if you call mutating methods (MutableList wrapped in unmodifiable wrapper).

WDYT?

donraab:
I think both are useful. The question I have with the more abstract one is whether it is ok to not wrap the ListIterable in an unmodifiable collection and allow direct manipulation of the List in the Multimap by casting. Thoughts? If we think it is ok to cast and modify, then maybe wrapping the MutableList in an unmodifiable collection is unnecessary. 

But this is inconsistent with what we do in get today.

motlin:
It’s not ok to cast and modify, because there’s other state that can get
out of sync, and it’s also possible to violate invariants, like the backing
map never contains empty value collections.

It would be ok to use mutable wrappers that support mutation by preserving
those invariants, but it’s a lot of work.

donraab:
Ok, so we agree that my approach of returning an unmodifiable wrapper in the case of mutable collection returns make sense. I am currently working on adding higher level methods in this PR. It is a lot more work, and I am only doing this for ListIterable, SetIterable, Bag for now since I had already done the work for the mutable and immutable versions of these types. If we like the approach, then I will work in a separate PR to finish the others.

motlin:
sounds good!

donraab:
@motlin, I backed out my changes for the abstract methods. I was up to 40 files changed and felt the change was too large and complicated to review. The problem with adding the more abstract methods is that they start intersecting with sorted and unsorted varieties of `Multimap`. I would suggest this be done as a separate PR as it will be easier to review and find issues. I made the changes you suggested in the tests using concrete types instead of `Iterable`.

motlin (approved):
Makes sense, LGTM!

donraab:
Thank you for the review @motlin. I just updated a few of the copyrights and will wait for the builds to pass and then I'll merge.

PR #1116, opened 2021-07-01
Add selectWithIndex and rejectWithIndex to OrderedIterable and ListIt…

donraab:
…erable.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/ImmutableList.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/ImmutableList.java
@@ -26,6 +26,7 @@ import org.eclipse.collections.api.block.function.primitive.ObjectIntToObjectFun
 import org.eclipse.collections.api.block.function.primitive.ShortFunction;
 import org.eclipse.collections.api.block.predicate.Predicate;
 import org.eclipse.collections.api.block.predicate.Predicate2;
+import org.eclipse.collections.api.block.predicate.primitive.ObjectIntPredicate;
 import org.eclipse.collections.api.block.procedure.Procedure;
 import org.eclipse.collections.api.collection.ImmutableCollection;
 import org.eclipse.collections.api.list.primitive.ImmutableBooleanList;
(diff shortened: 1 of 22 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/ImmutableList.java
```diff
@@ -98,6 +99,24 @@
         return this.collect(each -> function.valueOf(each, index[0]++));
     }
 
+    /**
+     * @since 11.0
```
prathasirisha, line 103: Missing Javadoc.  Seems like it is missing from the rest of the files too but I am not adding individual comments :)

unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableListTestCase.java
```diff
@@ -667,6 +668,58 @@ public void collectWithIndexWithTarget()
                 setOfPairs.collect(ObjectIntPair::getOne, Sets.mutable.empty()));
     }
 
+    /**
+     * @since 11.0.
```
prathasirisha, line 672: Do we need @ since tag in test methods?

prathasirisha (changes requested):

unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/AbstractImmutableListTestCase.java
donraab: They're not strictly necessary but also don't hurt to have IMO.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/ImmutableList.java
donraab: Good point. I will write some javadoc for the interface methods and overrides.

donraab: Added Javadoc for several interfaces and one utility class @prathasirisha. Thanks for pointing out.

prathasirisha (approved):
LGTM

Issue #1117, opened 2021-07-12
IntIntHashMap NullPointerException invoking remove on values collection


johnou:
eclipse-collections 10.4.0

IntIntHashMap usage

```java
    private final IntIntHashMap lockedSpaces = new IntIntHashMap();

    boolean lockSpaceIfCan(int playerId, int lockIndex) {
        return lockedSpaces.getIfAbsentPut(lockIndex, playerId) == playerId;
    }

    boolean releaseLockedSpaceById(int playerId, int lockIndex) {
        if (lockedSpaces.get(lockIndex) == playerId) {
            lockedSpaces.removeKey(lockIndex);
            return true;
        }
        return false;
    }

    private void releaseLockedSpaceByPlayerId(int playerId) {
        lockedSpaces.values().removeAll(playerId);
    }
```

Stacktrace

```
java.lang.NullPointerException: null
	at org.eclipse.collections.impl.map.mutable.primitive.IntIntHashMap$ValuesCollection.remove(IntIntHashMap.java:1572) ~[eclipse-collections-10.4.0.jar:?]
	at org.eclipse.collections.impl.map.mutable.primitive.AbstractMutableIntValuesMap$AbstractIntValuesCollection.removeAll(AbstractMutableIntValuesMap.java:845) ~[eclipse-collections-10.4.0.jar:?]
```

Hi, my initial conclusion is that the stacektrace indicates the map was accessed by multiple threads concurrently,  however upon checking the code paths the map object in question should be single threaded (actor).. I was hoping to confirm whether or not the behaviour of IntIntHashMap should mimic HashMap, specifically https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#values-- "The set is backed by the map, so changes to the map are reflected in the set, and vice-versa.".

mohrezaei:
This is a bug.

mohrezaei:
PR #1118 has a fix for this. This happens when remove is invoked for a value mapped to the special key zero. If you want to avoid this issue in your code while this is fixed/released, you can do something like:

```java
    private void releaseLockedSpaceByPlayerId(int playerId) {
        if (lockedSpaces.get(0) == playerId) {
            lockedSpaces.remove(0);
        } else {
            lockedSpaces.values().removeAll(playerId);
        }
    }
```

johnou:
@mohrezaei thanks for the quick response!

nikhilnanivadekar:
[...]

PR #1118, opened 2021-07-12
fix primitive hash map values collections remove for special key zero

mohrezaei:


mohrezaei:
fixes #1117

nikhilnanivadekar:
@mohrezaei do we need to have the same check for `MutableIntSet` as well?

mohrezaei:
@nikhilnanivadekar sets don't have a values collection, so this shouldn't be an issue there.

Issue #420, opened 2017-12-17
Optimize primitive collect methods for primitive Bags when collecting into a primitive Bag.


donraab:
This is a performance bug.

shreyateeza:
I will try to solve this issue. Please assign it to me. @donraab

nikhilnanivadekar:
@shreyateeza thank you for volunteering to contribute. Consider it assigned to you.
Its a long pending feature request to GitHub to allow new collaborators to be assigned. Once you contribute to Eclipse Collections I think future issues can be assigned to you.
Thanks for understanding!

nolequen:
@donraab, but the collect method signature of primitive iterables doesn't fit the requirements. For example IntIterable:
`<V, R extends Collection<V>> R collect(IntToObjectFunction<? extends V> function, R target)`
MutableIntBag doesn't extend the Collection class.

Is it neccessary to add a new collect method with MutableIntBag as target parameter?

donraab:
@nolequen First, sorry for the delay in responding. 

No, the existing signatures are fine.  In the case of an MutableIntBag with the collect signature as you have above, the method can be overridden as follows:

```
default <V, R extends Collection<V>> R collect(IntToObjectFunction<? extends V> function, R target)
{
    if (target instanceof MutableBagIterable<?>)
    {
        MutableBagIterable<V> targetBag = (MutableBagIterable<V>) target;

        this.forEachWithOccurrences((each, occurrences) ->
                targetBag.addOccurrences(function.valueOf(each), occurrences));
    }
    else
    {
        this.forEachWithOccurrences((each, occurrences) -> {
            V value = function.valueOf(each);
            for (int i = 0; i < occurrences; i++)
            {
                target.add(value);
            }
        });
    }
    return target;
}

```

This could be applied similarly across all of the other collect primitive variants (collectInt, collectLong, etc.).

nolequen:
@donraab thanks! Now it is clear.

nikhilnanivadekar:
Hi @shreyateeza following up, will you be working on this?

donraab:
Hi @shreyateeza, any update on this?  Let me know if you will be working on it.

mohrezaei:
I don't believe this is a valid optimization because of the function involved. Here is an example function that will break if the code is changed:

```
String prettyPrintBag(int item)
{
    int prev = -1;
    int count = 0;

    if (item == prev)
    {
         count++;
    }
    else
    {
        count = 0;
    }
[...]

PR #1131, opened 2021-08-28
Optimize primitive collect methods for primitive Bags

LuizFrra:
https://github.com/eclipse/eclipse-collections/issues/420

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/bag/immutablePrimitiveBag.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/bag/immutablePrimitiveBag.stg
@@ -14,11 +14,16 @@ body(type, name) ::= <<
 
 package org.eclipse.collections.api.bag.primitive;
 
+import java.util.Collection;
+import java.util.Set;
+
 import org.eclipse.collections.api.<name>Iterable;
 import org.eclipse.collections.api.bag.ImmutableBag;
+import org.eclipse.collections.api.bag.MutableBagIterable;
 import org.eclipse.collections.api.block.function.primitive.<name>ToObjectFunction;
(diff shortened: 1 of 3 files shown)


mohrezaei:
Since this is a generic `collect` method, should we also optimize for when the target is a `Set`?

LuizFrra:
I can add a check for when target is a Set add only once, What you think about ?

mohrezaei:
@LuizFrra Sounds good.

LuizFrra:
Done, I also add the optimization for the others primitive collect methods.

eclipse-collections-code-generator/src/main/resources/api/bag/immutablePrimitiveBag.stg
```diff
@@ -14,11 +14,16 @@ body(type, name) ::= <<
 
 package org.eclipse.collections.api.bag.primitive;
 
+import java.util.Collection;
+import java.util.Set;
```
mohrezaei, line 18: I was thinking about the primitive `Set`, not `java.util.Set` (primitive sets don't implement that).

mohrezaei (changes requested):

mohrezaei:
Woops, sorry, I didn't realize the signature contained a to-object function. 

I don't think this optimization is valid because of the function, which may have state (and do things like counting).

LuizFrra:
I can't figure out why the optimization isnt't valid. should I close de PR or just remove the optimization for the ImmutablePrimitiveBag ? Could You elaborate more please ?

Sorry, English isn't my main language.

mohrezaei:
@LuizFrra the issue is subtle. I commented on the original issue #420 with an example function that will break with the optimization.

PR #1142, opened 2021-09-21
Add JDK 17 to standard builds and update EA builds to JDK 18.

donraab:

### DIFF
--- a/.github/workflows/unit-no-p2.yml
+++ b/.github/workflows/unit-no-p2.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        java-version: [15, 16]
+        java-version: [17]
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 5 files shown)


nikhilnanivadekar:
We should move checkstyle to 17.

.github/workflows/unit-no-p2.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        java-version: [15, 16]
+        java-version: [15, 16, 17]
```
nikhilnanivadekar, line 14: We should probably only have it as 17

nikhilnanivadekar:
This one should only run on 8 and 11 : https://github.com/eclipse/eclipse-collections/blob/master/.github/workflows/unit.yml

nikhilnanivadekar (changes requested):

donraab:
> We should move checkstyle to 17.

I tried and it failed. I rolled back to 11.

donraab:
> This one should only run on 8 and 11 : https://github.com/eclipse/eclipse-collections/blob/master/.github/workflows/unit.yml

Done.

.github/workflows/unit-no-p2.yml
donraab: Done

donraab:
Ready to merge.

nikhilnanivadekar (approved):

Issue #1053, opened 2021-01-07
Implement subList() on primitive lists


vmzakharov:
Currently `subList()` on primitive lists is stubbed with
```
throw new UnsupportedOperationException("subList not yet implemented!");
```
The new behavior should be consistent with the behaviors currently defined in `FastList` for the mutable lists and `AbstractImmutableList` for the immutable ones.

gaurangkudale:
Hi @vmzakharov,
Can I work on this issue? if yes can please direct me to the code line

sneznaj:
i am already working on this

sneznaj:
i am already working on this

gaurangkudale:
Hi, @sneznaj Do you need any help? we can collaborate

sneznaj:
@vmzakharov FastList does not contain a sublist method or class

hugo-we:
any progress?

knowledge-graphlet:
```
public ImmutableIntList subList(int fromIndex, int toIndex) {
    if (fromIndex < 0 || toIndex > this.size() || fromIndex > toIndex) {
        throw new IndexOutOfBoundsException("From: " + fromIndex + ", To: " + toIndex + ", Size: " + this.size());
    }

    if (fromIndex == toIndex) {
        return IntLists.immutable.empty();
    }
    
    if (fromIndex == 0 && toIndex == this.size()) {
        return this;
    }

    MutableIntList subList = IntLists.mutable.withInitialCapacity(toIndex - fromIndex);
    for (int i = fromIndex; i < toIndex; i++) {
        subList.add(this.get(i));
    }
    return subList.toImmutable();
}

```

Is this an acceptable implementation?

mohrezaei:
No, it's not. Sub-lists in java are not copies.

knowledge-graphlet:
For mutable lists, I can see the distinction... Presuming you want the sublist to be a projection of the "backing" list, so that updates are reflected in the sublist. 

But if the list is immutable, and there is no object identity in the values, and there is no ability to update elements (so no notion of updates from the "backing" list). So, I'm not sure what you would be trying to accomplish that an immutable list that copies the range doesn't provide.  It seems you could have a pretty simple set of implementations for the primitive immutable types.

mohrezaei:
[...]

PR #1155, opened 2021-11-16
mplement subList() on primitive immutable lists

sneznaj:
part of #1053

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/primitive/ImmutableBooleanArrayList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/primitive/ImmutableBooleanArrayList.java
@@ -640,7 +640,7 @@ final class ImmutableBooleanArrayList
     @Override
     public ImmutableBooleanList subList(int fromIndex, int toIndex)
     {
-        throw new UnsupportedOperationException("subList not yet implemented!");
+        return new ImmutableSubList<>(this, fromIndex, toIndex);
     }
 
     private Object writeReplace()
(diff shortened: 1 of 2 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/LongInterval.java
```diff
@@ -532,7 +532,7 @@ public ImmutableLongList distinct()
     @Override
     public ImmutableLongList subList(int fromIndex, int toIndex)
     {
-        throw new UnsupportedOperationException("subList not yet implemented!");
+        return new LongIntervalSubList<>(this, fromIndex, toIndex);
```
mohrezaei, line 535: This can be replaced with just (another) LongInterval, so there is no need for the inner class.

mohrezaei (changes requested):

sneznaj:
#1160

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/primitive/ImmutableBooleanArrayList.java
```diff
@@ -718,4 +718,111 @@ public boolean next()
...
+
+        @Override
+        public ImmutableBooleanList newWith(boolean element)
+        {
+            BitSet newItems = (BitSet) this.original.items.clone();
+            if (element)
+            {
+                newItems.set(this.size);
+            }
+            return new ImmutableBooleanArrayList(newItems, this.size + 1);
```
motlin, line 770: This doesn't look like it uses the from offset.

PR #1160, opened 2021-11-23
implement subList() on primitive immutable lists

sneznaj:
implement subList() on primitive immutable lists

part of #1053

AbstractImmutableList defines  subList() as returning an object of an inner class
but here it is implemented as returning an object of the class itself

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/primitive/ImmutableBooleanArrayList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/primitive/ImmutableBooleanArrayList.java
@@ -638,9 +638,9 @@ final class ImmutableBooleanArrayList
     }
 
     @Override
-    public ImmutableBooleanList subList(int fromIndex, int toIndex)
+    public ImmutableBooleanArrayList subList(int fromIndex, int toIndex)
     {
-        throw new UnsupportedOperationException("subList not yet implemented!");
+        return new ImmutableBooleanArrayList(this.BitSet.get(fromIndex, toIndex), toIndex - fromIndex)
     }
 
     private Object writeReplace()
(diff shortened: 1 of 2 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/primitive/ImmutableBooleanArrayList.java
```diff
@@ -638,9 +638,9 @@ public void appendString(
     }
 
     @Override
-    public ImmutableBooleanList subList(int fromIndex, int toIndex)
+    public ImmutableBooleanArrayList subList(int fromIndex, int toIndex)
     {
-        throw new UnsupportedOperationException("subList not yet implemented!");
+        return new ImmutableBooleanArrayList(this.BitSet.get(fromIndex, toIndex), toIndex - fromIndex)
```
mohrezaei, line 643: copying the bitset is too expensive. Use a subclass instead.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/primitive/LongInterval.java
```diff
@@ -530,9 +530,9 @@ public ImmutableLongList distinct()
     }
 
     @Override
-    public ImmutableLongList subList(int fromIndex, int toIndex)
+    public LongInterval subList(int fromIndex, int toIndex)
     {
-        throw new UnsupportedOperationException("subList not yet implemented!");
+        return new LongInterval(fromIndex, toIndex, this.step);
```
mohrezaei, line 535: this looks wrong. The fromIndex/toIndex of the list is not the same as the from/to of the interval.

mohrezaei (changes requested):

mohrezaei:
Sorry, when I asked for not using a subclass, I only meant `LongInterval`.
Two more things:

- Please don't create new PRs for the same work. Just push more code to your branch. You're using master, which is a bit unusual, but it should work the same.
- You must add tests for all the new code.

donraab:
Closing this as two PRs with the same name could be a single PR.

PR #1163, opened 2021-12-05
Add superset tests for FastList, UnifiedMap and UnifiedSet.

donraab:


unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/FastListTest.java
```diff
@@ -126,6 +128,16 @@ public void wrapCopy()
         Assert.assertEquals(expected, actual);
     }
 
+    @Test
+    public void fastListProperSuperSetOfArrayList()
```
mohrezaei, line 132: Duplicated code in the 3 classes: please introduce a simple static method that takes two classes as parameter.

mohrezaei (changes requested):

mohrezaei (approved):

Issue #1166, opened 2021-12-13
`nCopies` variation of `ImmutableList`


dmlloyd:
It is sometimes useful to have a memory-efficient list comprising a fixed number of copies of a given input.  I would propose two variants: one which returns `n` copies of a single item (similar to `java.util.Collections#nCopies`), and one which returns `n` copies of an inner `ImmutableList` (if the inner list is empty, then the empty list can be used since `n × 0 = 0`).

PR #1168, opened 2021-12-16
Introduce `ImmutableListFactory.ofNCopiesOf` method with implementation

dmlloyd:
Partially addresses #1166; does not include primitive support yet


dmlloyd:
Also I'm not quite sure where/how to add unit tests... this project is surprisingly complex. Any pointers are appreciated.

prathasirisha:
> Also I'm not quite sure where/how to add unit tests... this project is surprisingly complex. Any pointers are appreciated.

Hi @dmlloyd - Apologies for the late response.  I took a quick look at your PR to help you with this question. 

For the changes, you are making in `ImmutableListFactoryImpl` you will add tests in the module unit-tests and the class would be `ListsTest`. 

For the changes you are making on the `ImmutableNCopiesList`, you will add a new test in 
1. module unit-tests and package `org.eclipse.collections.impl.list.immutable` 
2. module serialization-tests and same package as above `org.eclipse.collections.impl.list.immutable`
Other test classes in this package will be a good starting point for your new test class.  

Hope this helps. Please reach back, if you have any more questions. Thank you for your contribution!

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableNCopiesList.java
```diff
@@ -0,0 +1,173 @@
...
+
+import org.eclipse.collections.api.block.predicate.Predicate;
+import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.factory.Lists;
+import org.eclipse.collections.api.list.ImmutableList;
+
+/**
+ * This is a single element immutable List which is created by calling
+ * Immutable.newListWith(one) method.
+ */
```
motlin, line 26: This Javadoc looks pasted from a singleton list.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/list/ImmutableListFactory.java
```diff
@@ -144,4 +144,9 @@
         MutableList<T> list = (MutableList<T>) items.toSortedList(comparator);
         return list.toImmutable();
     }
+
+    /**
+     * @since 12.0
+     */
+    <T> ImmutableList<T> ofNCopiesOf(int copies, T item);
```
motlin, line 151: We have a similar method MutableListFactory#withNValues(). I think it's awkward to have "of" in the name twice. I'm also not sure if "copies" is appropriate in the name when these are all the same item, but then again, that's true of java.util.Collections#nCopies where the name comes from.

Right now I'm leaning toward also using `withNValues` as the name but I'd like to hear opinions.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableNCopiesList.java
```diff
@@ -0,0 +1,173 @@
...
+        implements Serializable, RandomAccess
+{
+    private static final long serialVersionUID = 1L;
+
+    private final T item;
+    private final int copies;
+
+    ImmutableNCopiesList(T item, int copies)
+    {
+        assert copies > 1;
```
motlin, line 38: The assertion is fine, but we shouldn't use the assert keyword.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableNCopiesList.java
```diff
@@ -0,0 +1,173 @@
...
+    public int count(Predicate<? super T> predicate)
+    {
+        return predicate.accept(this.item) ? this.copies : 0;
+    }
+
+    @Override
+    public int indexOf(Object object)
+    {
+        return Objects.equals(this.item, object) ? 0 : -1;
+    }
```
motlin, line 115: Could delegate to contains() here

motlin (commented):
Needs some test coverage too

PR #1174, opened 2022-01-04
Add fused method for collect + makeString

Desislav-Petrov:


eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -2298,6 +2298,17 @@ default String makeString(String start, String separator, String end)
...
 
+    /**
+     * Returns a string representation of the collection, created by applying the function supplied to each element,
+     * with the elements separated by the specified separator and enclosed between the start and end strings.
+     *
+     * @return a string representation of the mapped collection
+     */
+    default String makeString(Function<? super T, String> function, String start, String separator, String end)
+    {
+        return this.collect(function).makeString(start, separator, end);
```
donraab, line 2309: This should be `this.asLazy().collect(function).makeString(start, separator, end)`. The call to `collect` here will create a temporary collection otherwise in most cases except for `LazyIterable`.

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
```diff
@@ -2025,6 +2025,10 @@ default void RichIterable_makeString_appendString()
                 iterable.makeString("[", "/", "]"),
                 iterable.reduceInPlace(Collectors2.makeString("[", "/", "]")));
 
+        assertEquals(
+                iterable.makeString(Integer::toUnsignedString, "[", "/", "]"),
+                iterable.reduceInPlace(Collectors2.makeString("[", "/", "]")));
```
donraab, line 2030: There should be a new test for this method with a difference name, and the assertion here should switch the calls (`expected`, `actual`). Also `reduceInPlace` with `Collectors2.makeString()` does not do the same thing as `makeString` with a function. Replace the current call to reduceInPlace with `iterable.asLazy().collect(Integer::toUnsignedString).makeString("[", "/", "]")`.

donraab (changes requested):
Thank you for the contribution! I've requested some changes. Also, there needs to be a test added to `AbstractRichIterableTestCase` for this new method. This is a current pain with a legacy split we have in our unit test suite.

One other thing is that this method should be overridden in Synchronized and MultiReader types.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
Desislav-Petrov: Sorted

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
Desislav-Petrov: done

README_EXAMPLES.md
```diff
@@ -37,8 +37,7 @@ MutableList<Person> people = Lists.mutable.with(
 ```
 Then we will `collect` their last names into a new `MutableList`, and finally output the names to a comma delimited String using `makeString`. 
 ```java
-MutableList<String> lastNames = people.collect(person -> person.getLastName());
-Assert.assertEquals("Smith, Watson, Williams", lastNames.makeString());
+Assert.assertEquals("Smith, Watson, Williams", lastNames.makeString(Person::getLastName, "", ",", ""));
```
donraab, line 40: While this code works, it doesn't fit well with the example as it was meant to demonstrate `collect`. This could be a further refactoring of the example, after the lambda is converted to a method reference.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -2298,6 +2298,17 @@ default String makeString(String start, String separator, String end)
         return stringBuilder.toString();
     }
 
+    /**
+     * Returns a string representation of the collection, created by applying the function supplied to each element,
+     * with the elements separated by the specified separator and enclosed between the start and end strings.
+     *
+     * @return a string representation of the mapped collection
+     */
+    default String makeString(Function<? super T, String> function, String start, String separator, String end)
```
donraab, line 2307: I'm thinking it would be better to have the `function` return `Object` as the type and let `makeString` apply `String.valueOf()` as it does currently. I was originally thinking `? extends CharSequence` to allow for more "String-like" implementations, but there is no need to be restrictive here as `makeString` works with any object currently.

donraab (changes requested):
Thank you for the updates. It's looking good. I have a recommendation on relaxing the return type of the `Function` to `Object`. This should just work, and will make it consistent with `collect().makeString()`, which does not currently force the result of `collect()` to be a collection of `String`.

README_EXAMPLES.md
Desislav-Petrov: I was thinking the same actually.. i'll revert this bit and will leave it as a later enhancement.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
Desislav-Petrov: done

unit-tests/src/test/java/org/eclipse/collections/impl/AbstractRichIterableTestCase.java
```diff
@@ -1841,6 +1841,13 @@ public void makeStringWithSeparatorAndStartAndEnd()
         Assert.assertEquals(collection.toString(), collection.makeString("[", ", ", "]"));
     }
 
+    @Test
+    public void fusedCollectMakeString()
+    {
+        RichIterable<Integer> collection = this.newWith(1, 2, 3);
+        Assert.assertEquals(collection.toString(), collection.makeString(Integer::toUnsignedString, "[", ", ", "]"));
```
donraab, line 1848: Please make this consistent with the `RichIterableTestCase` test above using `asLazy().collect().makeString()`.

donraab (changes requested):
Looking good. One minor change. Thank you!

unit-tests/src/test/java/org/eclipse/collections/impl/AbstractRichIterableTestCase.java
Desislav-Petrov: Ops - my bad, fixing

Desislav-Petrov:
All changes done, merged master as well

donraab:
Hi @Desislav-Petrov, please read the guidelines on pull requests in the last section of the CONTRIBUTING.md.

https://github.com/eclipse/eclipse-collections/blob/master/CONTRIBUTING.md

You should squash your commits to a single commit and rebase your PR on master.

Thank you!

Desislav-Petrov:
Hi @donraab , thx for the feedback - I believe it should all be good now.

donraab (approved):
LGTM. Thanks for the contribution @Desislav-Petrov!

PR #1175, opened 2022-01-12
Generate primitive List factories in API using ServiceLoader.

donraab:

### DIFF
--- a/eclipse-collections-code-generator-maven-plugin/src/main/java/org/eclipse/collections/codegenerator/maven/GenerateMojo.java
+++ b/eclipse-collections-code-generator-maven-plugin/src/main/java/org/eclipse/collections/codegenerator/maven/GenerateMojo.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs.
+ * Copyright (c) 2021 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -63,19 +63,15 @@ public class GenerateMojo extends AbstractMojo
         List<URL> urls = Arrays.asList(((URLClassLoader) GenerateMojo.class.getClassLoader()).getURLs());
 
         boolean[] error = new boolean[1];
-        EclipseCollectionsCodeGenerator.ErrorListener errorListener = new EclipseCollectionsCodeGenerator.ErrorListener()
-        {
-            public void error(String string)
-            {
-                GenerateMojo.this.getLog().error(string);
-                error[0] = true;
-            }
+        EclipseCollectionsCodeGenerator.ErrorListener errorListener = string -> {
+            this.getLog().error(string);
+            error[0] = true;
         };
-        EclipseCollectionsCodeGenerator gsCollectionsCodeGenerator =
+        EclipseCollectionsCodeGenerator codeGenerator =
                 new EclipseCollectionsCodeGenerator(this.templateDirectory, this.project.getBasedir(), urls, errorListener);
         if (!this.skipCodeGen)
         {
-            int numFilesWritten = gsCollectionsCodeGenerator.generateFiles();
+            int numFilesWritten = codeGenerator.generateFiles();
             this.getLog().info("Generated " + numFilesWritten + " files");
         }
         if (error[0])
@@ -83,7 +79,7 @@ public class GenerateMojo extends AbstractMojo
             throw new MojoExecutionException("Error(s) during code generation.");
         }
 
-        if (gsCollectionsCodeGenerator.isTest())
+        if (codeGenerator.isTest())
         {
             this.project.addTestCompileSourceRoot(EclipseCollectionsCodeGenerator.GENERATED_TEST_SOURCES_LOCATION);
         }
--- a/eclipse-collections-code-generator/src/main/java/org/eclipse/collections/codegenerator/EclipseCollectionsCodeGenerator.java
+++ b/eclipse-collections-code-generator/src/main/java/org/eclipse/collections/codegenerator/EclipseCollectionsCodeGenerator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs.
+ * Copyright (c) 2021 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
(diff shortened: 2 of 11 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/primitive/ServiceLoaderUtils.java
```diff
@@ -0,0 +1,64 @@
...
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Proxy;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+
+import org.eclipse.collections.api.factory.ThrowingInvocationHandler;
+
+final class ServiceLoaderUtils
```
motlin, line 22: I think it would be fine to make org.eclipse.collections.api.factory.ServiceLoaderUtils public, instead of making a second package private copy.

eclipse-collections-code-generator-maven-plugin/src/main/java/org/eclipse/collections/codegenerator/maven/GenerateMojo.java
```diff
@@ -63,27 +63,23 @@ public void execute() throws MojoExecutionException
...
         boolean[] error = new boolean[1];
-        EclipseCollectionsCodeGenerator.ErrorListener errorListener = new EclipseCollectionsCodeGenerator.ErrorListener()
-        {
-            public void error(String string)
-            {
-                GenerateMojo.this.getLog().error(string);
-                error[0] = true;
-            }
+        EclipseCollectionsCodeGenerator.ErrorListener errorListener = string -> {
+            GenerateMojo.this.getLog().error(string);
```
motlin, line 67: There should be a second inspection firing here about the redundant prefix to `this`

eclipse-collections-code-generator/src/main/java/org/eclipse/collections/codegenerator/EclipseCollectionsCodeGenerator.java
```diff
@@ -30,10 +30,12 @@
 {
     public static final String GENERATED_TEST_SOURCES_LOCATION = "target/generated-test-sources/java/";
     public static final String GENERATED_SOURCES_LOCATION = "target/generated-sources/java/";
+    public static final String GENERATED_SERVICES_LOCATION = "target/generated-sources/resources/META-INF/services";
```
motlin, line 33: I was expecting to see `target/classes/META-INF/services` here. These files aren't sources (they don't get compiled).

eclipse-collections-code-generator/src/main/java/org/eclipse/collections/codegenerator/EclipseCollectionsCodeGenerator.java
```diff
@@ -188,16 +196,28 @@ private ST findTemplate(String templateName)
...
-                : new File(this.moduleBaseDir, GENERATED_SOURCES_LOCATION + targetPath.render());
+        if (this.isTest)
+        {
+            return new File(this.moduleBaseDir, GENERATED_TEST_SOURCES_LOCATION + targetPath.render());
+        }
+        if (this.isService)
+        {
+            return new File(this.moduleBaseDir, GENERATED_SERVICES_LOCATION + targetPath.render());
+        }
+        return new File(this.moduleBaseDir, GENERATED_SOURCES_LOCATION + targetPath.render());
```
motlin, line 220: This if-statement makes it look like there ought to be some sort of assertion that a template is a test or service, but not both.

eclipse-collections/pom.xml
```diff
@@ -46,6 +46,18 @@
...
+            <resource>
+                <directory>target/generated-sources/resources</directory>
+                <excludes>
+                    <exclude>**/*.crc</exclude>
+                </excludes>
+            </resource>
+            <resource>
+                <directory>src/main/resources</directory>
+            </resource>
+        </resources>
```
motlin, line 59: Once the output path is corrected, you won't need this resources config. If that turns out to be impossible, you should change the name from "sources" to "resources", meaning `target/generated-resources`

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/primitive/ServiceLoaderUtils.java
donraab: Done

eclipse-collections-code-generator-maven-plugin/src/main/java/org/eclipse/collections/codegenerator/maven/GenerateMojo.java
donraab: Fixed.

eclipse-collections-code-generator/src/main/java/org/eclipse/collections/codegenerator/EclipseCollectionsCodeGenerator.java
donraab: I tried generating here, but it gets wiped out in the build process, so I decided to mirror the checked in files in src\main\resources.

eclipse-collections-code-generator/src/main/java/org/eclipse/collections/codegenerator/EclipseCollectionsCodeGenerator.java
donraab: This was intentional.

eclipse-collections/pom.xml
donraab: I could probably do this and change to target/generated-resources, but I was mirroring the way src exists with src\main\resources and src\main\java.

eclipse-collections-code-generator/src/main/java/org/eclipse/collections/codegenerator/EclipseCollectionsCodeGenerator.java
motlin: The reason is that the code generator plugin currently only has a single target, with a single phase, generate-sources.

```java
/**
 * @goal generate
 * @phase generate-sources
 * @requiresDependencyResolution compile
 */
public class GenerateMojo extends AbstractMojo
{
```

So it's running too late to generate resources.

eclipse-collections-code-generator-maven-plugin/src/main/java/org/eclipse/collections/codegenerator/maven/GenerateMojo.java
```diff
@@ -63,27 +63,23 @@ public void execute() throws MojoExecutionException
...
         if (error[0])
         {
             throw new MojoExecutionException("Error(s) during code generation.");
         }
 
-        if (gsCollectionsCodeGenerator.isTest())
+        if (codeGenerator.isTest())
         {
             this.project.addTestCompileSourceRoot(EclipseCollectionsCodeGenerator.GENERATED_TEST_SOURCES_LOCATION);
         }
```
motlin, line 85: And here's the place that would need to change so you don't have to manually set up resources directories. Something like:

```
        Resource resource = new Resource();
        resource.setDirectory(outputDirectory);
        this.project.addResource(resource);
```

This stuff is tricky, I can probably help with another PR on top.

donraab: I originally tried something similar to this and couldn't get it to work. Please feel free to merge this PR and then submit another on top with the changes. I'll be happy to review and merge.

donraab: @motlin Do you want to merge and apply changes in a separate PR?

PR #1183, opened 2022-01-31
Reduce code duplication between code generators.

motlin:


donraab (approved):
LGTM. Thanks for the refactoring @motlin!

PR #1190, opened 2022-02-08
Add tests for Collection.removeAll() to the trait-based test suite.

motlin:


unit-tests-java8/src/test/java/org/eclipse/collections/test/CollectionTestCase.java
```diff
@@ -93,7 +94,7 @@ default void Collection_add()
     }
 
     @Test
-    default void Collection_remove()
+    default void Collection_remove_removeAll()
```
motlin, line 97: At first I wrote this as a separate test method. But then I realized the two methods are declared on the same interface, and they were only overridden together and only once, in the unmodifiable subclass.

unit-tests-java8/src/test/java/org/eclipse/collections/test/UnmodifiableCollectionTestCase.java
```diff
@@ -25,9 +26,11 @@ default void Collection_add()
 
     @Override
     @Test
-    default void Collection_remove()
+    default void Collection_remove_removeAll()
     {
         assertThrows(UnsupportedOperationException.class, () -> this.newWith(1, 2, 3).remove(3));
+
+        assertThrows(UnsupportedOperationException.class, () -> this.newWith(1, 2, 3).removeAll(Lists.mutable.with(3, 2)));
```
motlin, line 33: Here's the one and only override.

unit-tests-java8/src/test/java/org/eclipse/collections/test/set/SetTestCase.java
```diff
@@ -105,6 +108,7 @@ default void Collection_add()
     @Test
     default void Collection_size()
     {
+        CollectionTestCase.super.Collection_size();
```
motlin, line 111: The three missing calls to super were old bugs. I only noticed this because I was looking around for overrides of remove. The assertions from the superclass do apply and do pass.

motlin (commented):
Overall, you can see 44 lines added, 115 removed, and still more assertions are now applied to more types.

prathasirisha:
LGTM. I approved the changes. I see one job failed but it's equivalent actions build passed (checkstyle) and we can't re-run the teamcity job.  I am okay to merge it, @donraab please feel free to do so it if you approve this PR as well :)

PR #1191, opened 2022-02-09
Add test coverage for ArrayAdapter. Split out assertions for fixed size and unmodifiable lists.

motlin:


unit-tests-java8/src/test/java/org/eclipse/collections/test/list/UnmodifiableListTestCase.java
```diff
@@ -10,34 +10,21 @@
...
-    }
-
-    @Override
-    @Test(expected = UnsupportedOperationException.class)
-    default void List_set_out_of_bounds()
-    {
-        List<Integer> list = this.newWith(1, 2, 3);
-        list.set(4, 4);
+        assertThrows(UnsupportedOperationException.class, () -> this.newWith().set(-1, 4));
+        assertThrows(UnsupportedOperationException.class, () -> this.newWith(1, 2, 3).set(4, 4));
```
motlin, line 28: Combined the tests here too. Collections that are unmodifiable tend to throw UnsupportedOperationException no matter what, even if the method would have failed for other reasons like index out of bounds, and even if the method wouldn't have performed a mutation.

unit-tests-java8/src/test/java/org/eclipse/collections/test/list/mutable/ArrayAdapterTest.java
```diff
@@ -0,0 +1,28 @@
...
+@RunWith(Java8Runner.class)
+public class ArrayAdapterTest implements FixedSizeListTestCase
+{
+    @SafeVarargs
+    @Override
+    public final <T> MutableList<T> newWith(T... elements)
+    {
+        return ArrayAdapter.adapt(elements);
+    }
+}
```
motlin, line 28: This is what I had set out to do. Add test coverage for ArrayAdapter. I found that the existing unmodifiable test didn't work for it, because mutations that don't change the size are supported. In other words, it's a fixed sized list, not an unmodifiable one.

unit-tests-java8/src/test/java/org/eclipse/collections/test/list/mutable/ArrayListTest.java
```diff
@@ -0,0 +1,30 @@
...
+@RunWith(Java8Runner.class)
+public class ArrayListTest implements ListTestCase
+{
+    @SafeVarargs
+    @Override
+    public final <T> List<T> newWith(T... elements)
+    {
+        return new ArrayList<>(Arrays.asList(elements));
+    }
+}
```
motlin, line 30: It's really useful to apply these trait tests to build-in JCF types. It helps validate that our contracts are identical. Anything we assert about our own java.util.List implementations also applies to ArrayList.

unit-tests-java8/src/test/java/org/eclipse/collections/test/list/mutable/ArraysAsListTest.java
```diff
@@ -0,0 +1,71 @@
...
+@RunWith(Java8Runner.class)
+public class ArraysAsListTest
+        implements FixedSizeListTestCase
+{
+    @SafeVarargs
+    @Override
+    public final <T> List<T> newWith(T... elements)
+    {
+        return Arrays.asList(elements);
+    }
```
motlin, line 33: Here I wanted to do something similar as the ArrayListTest. I figured that Arrays.asList(...) is very similar to ArrayAdapter and I could see if the two collections pass the same tests. It turns out they do not because our fixed-size collections throw immediately if you try to remove from them, and the Arrays.asList() implementation iterates through looking for the item to remove, and only throws if it finds the item. This is a rare case where our contract deviates from the JDK and yet is obviously better. I kept the test though, with the overrides showing the different behavior.

unit-tests-java8/src/test/java/org/eclipse/collections/test/list/mutable/FixedSizeListTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+@RunWith(Java8Runner.class)
+public class FixedSizeListTest implements FixedSizeListTestCase
+{
+    @SafeVarargs
+    @Override
+    public final <T> MutableList<T> newWith(T... elements)
+    {
+        return Lists.fixedSize.with(elements);
+    }
```
motlin, line 27: By creating this test, I also manage to add a little coverage for the singleton/doubleton/tripleton lists. The assertions are the same ones being applied to ArrayAdapter. The assertions are not written in a way that's careful to use every possible list size, so the coverage from this will be spotty. Still, it's strictly better than not having this test at all.

unit-tests-java8/src/test/java/org/eclipse/collections/test/list/mutable/UnmodifiableListTest.java
```diff
@@ -0,0 +1,33 @@
...
+public class UnmodifiableListTest
+        implements UnmodifiableListTestCase
+{
+    @SafeVarargs
+    @Override
+    public final <T> List<T> newWith(T... elements)
+    {
+        List<T> result = new ArrayList<>(Arrays.asList(elements));
+        return Collections.unmodifiableList(result);
+    }
```
motlin, line 32: Here we show that Collections.unmodifiableList() and our unmodifiable lists pass the same assertions and have the same contract. I wasn't sure this would be true for subtle interactions discussed above, like throwing unsupported vs index out of bound, or throwing vs ignoring when attempting to perform mutations that don't mutate like clearing an empty list.

prathasirisha (approved):

Issue #1196, opened 2022-02-19
Use Guava's testlib for increased test coverage


ben-manes:
Guava provides a convenient [testlib](https://github.com/google/guava/tree/master/guava-testlib) for their collection tests. This can be used for other collection implementations, making a handy second set of tests to catch oversights. For example it found a few [small bugs](https://github.com/vigna/fastutil/issues/269) in FastUtil. In the test case below the `Map.entrySet().toString()` is missing, so it does not pretty-print the contents (this can be ignored using `suppressing(Method)`).

<details>
<summary>Maps.mutable tests</summary>

```java
import java.util.Map;
import java.util.function.Supplier;

import org.eclipse.collections.api.factory.Maps;

import com.google.common.collect.testing.NavigableMapTestSuiteBuilder;
import com.google.common.collect.testing.TestStringMapGenerator;
import com.google.common.collect.testing.features.CollectionFeature;
import com.google.common.collect.testing.features.CollectionSize;
import com.google.common.collect.testing.features.MapFeature;

import junit.framework.Test;
import junit.framework.TestCase;

/**
 * @author ben.manes@gmail.com (Ben Manes)
 */
public class EclipseMapTest extends TestCase {

  public static Test suite() {
    return suite("Maps.mutable", Maps.mutable::empty);
  }

  public static Test suite(String name, Supplier<Map<String, String>> factory) {
    return NavigableMapTestSuiteBuilder
        .using(new TestStringMapGenerator() {
          @Override protected Map<String, String> create(Map.Entry<String, String>[] entries) {
            var map = factory.get();
            for (var entry : entries) {
              map.put(entry.getKey(), entry.getValue());
            }
            return map;
          }
        })
        .named(name)
        .withFeatures(
            CollectionSize.ANY,
            MapFeature.GENERAL_PURPOSE,
            MapFeature.ALLOWS_NULL_KEYS,
            MapFeature.ALLOWS_NULL_VALUES,
            MapFeature.ALLOWS_NULL_ENTRY_QUERIES,
            CollectionFeature.SUPPORTS_ITERATOR_REMOVE)
        .createTestSuite();
  }
}
```

```
junit.framework.ComparisonFailure: emptyCollection.toString should return [] expected:<[[]]> but was:<[org.eclipse.collections.impl.map.mutable.UnifiedMap$EntrySet@0]>
	at junit.framework.Assert.assertEquals(Assert.java:100)
[...]

PR #1199, opened 2022-02-20
Fix issue mentioned in #1196

mehmandarov:
Adding missing toString() method in UnifiedMap.EntrySet. This should solve the bug mentioned in issue #1196. However, there are currently no tests covering this case in the EC code. Apart from that, the code should be good to go.

The way I see it, there are two options with regard to tests:

1. Add `guava-testlib` dependency and the test code similar to the code provided in the issue, or
2. Write a separate test that covers this case.

Would love to hear your thoughts on this. 

Signed-off-by: Rustam Mehmandarov <mehmandarov@gmail.com>


motlin:
This is a bug fix so we should add a test, ideally together with the fix.

motlin:
I started writing a test for this, and I found a few more bugs. I'll put up a PR shortly.

motlin:
Addressed in https://github.com/eclipse/eclipse-collections/pull/1200

donraab:
@motlin @mehmandarov Assuming we can close this PR now.

PR #1209, opened 2022-02-25
Primitive inject into

victornoel:
This is a PR for #639 :)


victornoel:
In any case, I've added tests :)

donraab:
> @donraab I just realized that `RichIterable` has `injectInto` only for `int`/`float`/`double`/`long`, should we do the same here or is it best to cover all primitive types?

I was wondering the same. I think it is probably fine to just cover them all. This does make the testing more challenging for things like boolean (haven't seen your tests yet so curious what you came up with). I want to deprecate the four methods on the object side because they are overloads, but we need to add the properly named versions at the same time.

donraab (approved):
LGTM. Thanks for the contribution @victornoel !

PR #1210, opened 2022-02-26
Add Bag.distinctView() for #726

victornoel:
A simple implementation for #726 after the failed attempts of #797, #861 and #869 :)

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/Bag.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/Bag.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs and others.
+ * Copyright (c) 2022 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -213,6 +213,17 @@ public interface Bag<T>
     String toStringOfItemToCount();
 
     ImmutableBagIterable<T> toImmutable();
+
+    /**
+     * Returns an unmodifiable view on the distinct elements with the same complexity as the Bag implementation.
+     *
+     * @return an unmodifiable view on the distinct elements of the Bag.
+     * @since 11.1
+     */
+    default RichIterable<T> distinctView()
+    {
+        throw new UnsupportedOperationException(this.getClass().getSimpleName() + ".distinctView() not implemented yet");
+    }
 
     /**
      * @since 8.0
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/MultiReaderHashBag.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/MultiReaderHashBag.java
@@ -662,6 +662,14 @@ public final class MultiReaderHashBag<T>
         }
     }
 
+    @Override
+    public RichIterable<T> distinctView()
+    {
+        throw new UnsupportedOperationException(
+                "distinctView is not supported directly on MultiReader collections.  "
+                        + "If you would like to use such as view, you must either use withReadLockAndDelegate() or withWriteLockAndDelegate().");
+    }
+
     //Exposed for testing
 
     static final class UntouchableMutableBag<T>
@@ -723,6 +731,12 @@ public final class MultiReaderHashBag<T>
 
         @Override
         public MutableBag<T> asUnmodifiable()
+        {
+            throw new UnsupportedOperationException("cannot wrap an UntouchableMutableBag");
+        }
+
+        @Override
+        public RichIterable<T> distinctView()
         {
             throw new UnsupportedOperationException("cannot wrap an UntouchableMutableBag");
         }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/SynchronizedBag.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/SynchronizedBag.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs and others.
+ * Copyright (c) 2022 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -14,6 +14,7 @@ import java.io.Serializable;
 import java.util.Collection;
 import java.util.Collections;
 
+import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.bag.ImmutableBag;
 import org.eclipse.collections.api.bag.MutableBag;
 import org.eclipse.collections.api.bag.primitive.MutableBooleanBag;
@@ -50,6 +51,7 @@ import org.eclipse.collections.api.partition.bag.PartitionMutableBag;
 import org.eclipse.collections.api.set.MutableSet;
 import org.eclipse.collections.api.tuple.Pair;
 import org.eclipse.collections.api.tuple.primitive.ObjectIntPair;
+import org.eclipse.collections.impl.SynchronizedRichIterable;
 import org.eclipse.collections.impl.collection.mutable.AbstractSynchronizedMutableCollection;
 import org.eclipse.collections.impl.collection.mutable.SynchronizedCollectionSerializationProxy;
 
@@ -450,6 +452,15 @@ public class SynchronizedBag<T>
         synchronized (this.getLock())
         {
             return this.getDelegate().selectUnique();
+        }
+    }
+
+    @Override
+    public RichIterable<T> distinctView()
+    {
+        synchronized (this.getLock())
+        {
+            return SynchronizedRichIterable.of(this.getDelegate().distinctView(), this.lock);
         }
     }
 }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedBagImpl.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedBagImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs and others.
+ * Copyright (c) 2022 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -21,6 +21,7 @@ import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.RandomAccess;
 
+import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.bag.Bag;
 import org.eclipse.collections.api.bag.sorted.ImmutableSortedBag;
 import org.eclipse.collections.api.bag.sorted.MutableSortedBag;
@@ -44,6 +45,7 @@ import org.eclipse.collections.api.set.sorted.MutableSortedSet;
 import org.eclipse.collections.impl.Counter;
 import org.eclipse.collections.impl.bag.sorted.mutable.TreeBag;
 import org.eclipse.collections.impl.block.factory.Comparators;
+import org.eclipse.collections.impl.list.fixed.ArrayAdapter;
 import org.eclipse.collections.impl.list.mutable.FastList;
 import org.eclipse.collections.impl.map.sorted.mutable.TreeSortedMap;
 import org.eclipse.collections.impl.partition.bag.sorted.PartitionImmutableSortedBagImpl;
@@ -790,6 +792,13 @@ class ImmutableSortedBagImpl<T>
         Counter counter = new Counter();
         this.forEachWithOccurrences((each, count) -> counter.add((each == null ? 0 : each.hashCode()) ^ count));
         return counter.getCount();
+    }
+
+    @Override
+    public RichIterable<T> distinctView()
+    {
+        // TODO use an implementation taking advantage of the sorting of the array
+        return ArrayAdapter.adapt(this.elements).asUnmodifiable();
     }
 
     private class InternalIterator implements Iterator<T>
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/mutable/SynchronizedSortedBag.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/mutable/SynchronizedSortedBag.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs and others.
+ * Copyright (c) 2022 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -18,6 +18,7 @@ import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 
 import org.eclipse.collections.api.LazyIterable;
+import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.bag.ParallelBag;
 import org.eclipse.collections.api.bag.sorted.ImmutableSortedBag;
 import org.eclipse.collections.api.bag.sorted.MutableSortedBag;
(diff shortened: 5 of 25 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/SynchronizedBag.java
```diff
@@ -452,4 +453,13 @@ public MutableShortBag collectShort(ShortFunction<? super T> shortFunction)
...
     }
+
+    @Override
+    public RichIterable<T> distinctView()
+    {
+        synchronized (this.getLock())
+        {
+            return this.getDelegate().distinctView();
+        }
+    }
```
motlin, line 464: Same here, the view needs a lock.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/mutable/SynchronizedSortedBag.java
```diff
@@ -743,4 +744,13 @@ public int detectLastIndex(Predicate<? super T> predicate)
...
     }
+
+    @Override
+    public RichIterable<T> distinctView()
+    {
+        synchronized (this.getLock())
+        {
+            return this.getDelegate().distinctView();
+        }
+    }
```
motlin, line 755: Needs a lock

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedBagImpl.java
```diff
@@ -792,6 +794,13 @@ public int hashCode()
         return counter.getCount();
     }
 
+    @Override
+    public RichIterable<T> distinctView()
+    {
+        // TODO ideally we would want to reuse ImmutableTreeSet implementation instead
+        return ArrayAdapter.adapt(this.elements).asUnmodifiable();
```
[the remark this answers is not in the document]
victornoel: I've added `ImmutableTreeSet.wrap` in the same manner as the similar and already existing `FastList.newListWith`

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/MultiReaderHashBag.java
```diff
@@ -678,6 +678,15 @@ public void readExternal(ObjectInput in) throws IOException, ClassNotFoundExcept
...
     }
 
+    @Override
+    public RichIterable<T> distinctView()
+    {
+        try (LockWrapper wrapper = this.lockWrapper.acquireReadLock())
+        {
+            return this.getDelegate().distinctView();
+        }
+    }
```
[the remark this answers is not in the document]
victornoel: I've introduced `MultiReaderRichIterable` (in a separate commit, with tests)

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/SynchronizedBag.java
victornoel: done

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/mutable/SynchronizedSortedBag.java
victornoel: done

victornoel:
I've added a few tests for the `Synchronized` and `MultiReader` implementations but since `RichIterable` wrappers are not meant to implement `equals`/`hashCode` as would be expected with concrete classes such as `Set`, I didn't add tests checking the iterables returned by `distinctView()`. We can add some, but they will be very coupled to the internal implementation of the method...

donraab:
@motlin would appreciate if you can get through finishing the review on this one with @victornoel so we can keep making progress on clearing some of our PR queue. Thanks!

motlin (changes requested):
I believe that `getDelegate()` already existed but you elevated its visibility from private to public.

This feels like a little bit of a silly example, but this is now possible.

```java
MultiReaderFastList<Integer> integers = this.newWith(7, 5, 4, 3, 2, 1);
MutableList<Integer>[] delegate = new MutableList[1];
integers.withWriteLockAndDelegate(each -> delegate[0] = ((MultiReaderFastList.UntouchableMutableList<Integer>) each).getDelegate());
delegate[0].add(8);
assertEquals(integers, this.newWith(7, 5, 4, 3, 2, 1, 8));
```

Would it be possible to change delegate to be a protected field (ideally) or a protected method at least?

victornoel:
> Would it be possible to change delegate to be a protected field (ideally) or a protected method at least?

You are correct, it was of course a mistake ^^ they should have been protected as they implement a protected abstract method in the parent class.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/MultiReaderHashBag.java
```diff
@@ -1138,15 +1027,16 @@ public String toStringOfItemToCount()
...
-            return (MutableBag<T>) this.delegate;
+            return this.getDelegate().selectUnique();
         }
 
         @Override
-        public MutableSet<T> selectUnique()
+        public RichIterable<T> distinctView()
         {
-            return this.getDelegate().selectUnique();
+            throw new UnsupportedOperationException("cannot wrap an UntouchableMutableBag");
```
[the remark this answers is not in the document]
victornoel: done

victornoel:
@donraab @motlin thx for the feedbacks, I think it should be good now :) there were a few open questions such as:
- should we introduce `MultiReaderRichIterable` interface and have it the return type of `MultiReaderBag.distinctView()`?
- should actually `MultiReaderHashBag.UntouchableMutableBag.distinctView()` throw an exception?

But I feel like that it can always be done in a backward compatible way later if anybody asks for it with a relevant use case, so maybe not needed for now :)

donraab (commented):
Hi @victornoel, I started reviewing to see if I could help move this along to completion and discovered quickly that this PR was more about moving behavior to an abstract class than about adding distinctView. I would recommend in the future splitting the refactoring like this into a separate PR as it will make the review easier to complete. I like where you headed and wound up with the refactoring, but it is a lot of changes to have to compare and unfortunately slows down the review of what you ultimately wanted which was distinctView. The separate commits definitely help. I'm going to let @motlin to complete the review on this one as I don't want to create unnecessary noise on stuff he has already reviewed and confirmed. I will triple check the refactoring portion for anything potentially missing after the PR is merged. I appreciate your patience. The distinctView portion of the code looks good to me.

donraab:
> * should we introduce `MultiReaderRichIterable` interface and have it as the return type of `MultiReaderBag.distinctView()`?

Interesting question. Would there be anything difference between `RichIterable` and `MultiReaderRichIterable`?

donraab:
> @donraab I can totally break this into two PR, don't hesitate to ask if it eases you work :)
> 

Thanks @victornoel. I'll leave this up to @motlin as he had given most of the feedback on this review already. It seems like this is already pretty close so would be great to get it completed.

donraab:
> Basically, the implementation returned (a `MultiReaderRichIterable`) will throw some exception for example when calling `iterator()`. Since right now we return a `RichIterable`, the user won't be able to follow the recommendation from the exception (to use one of the `with{Read,Write}LockAndDelegate()` methods).
> 
> I suspect the same question could be asked for https://github.com/eclipse/eclipse-collections/blob/master/eclipse-collections%2Fsrc%2Fmain%2Fjava%2Forg%2Feclipse%2Fcollections%2Fimpl%2Flist%2Fmutable%2FMultiReaderFastList.java#L570 for example.

Ok, understood. The only problem I see is that the `with{Read,Write}LockAndDelegate()` probably won't be able to be abstracted to this type because they take the more specific delegate type (`MutableList`, `MutableSet`, `MutableBag`).

victornoel:
@donraab I would have said that in the same manner as https://github.com/eclipse/eclipse-collections/blob/master/eclipse-collections-api%2Fsrc%2Fmain%2Fjava%2Forg%2Feclipse%2Fcollections%2Fapi%2Flist%2FMultiReaderList.java takes a `MutableList` (the assumption is that it's a multi reader on a mutable list), then `MultiReaderRichIterable.with{Read,Write}LockAndDelegate()` would take a `RichIterable` (the assumption is that it's actually mutable even if we don't have access to mutable methods, well, except `iterator()` actually, so we do).
So it's a bit like `UnmodifiableRichIterable` takes a `RichIterable` with the assumption it's actually a mutable one.

victornoel:
@donraab @motlin any news for this PR? :)

donraab:
Hi @victornoel, let's break this into two separate PRs. I like what you're doing with `MultiReaderRichIterable`, but that's just going to take me longer to reason about (for instance, to decide if it requires a major release or not), and I'd like help get `distinctView` to done. I'll try to finish the review on both.

victornoel:
@donraab ok, I will do this soon. But note that we need `MultiReaderRichIterable` to be able to implement `distinctView` next.

donraab:
> But note that we need `MultiReaderRichIterable` to be able to implement `distinctView` next.

Not sure I understand. Do you mean `MultiReaderRichIterable` needs to have `distinctView`, or `distinctView` needs to have `MultiReaderRichIterable` in order to be implemented? If it is the latter, then I must have missed the dependency.

donraab:
@victornoel Never mind my last comment. I think I found the depedency.

```
    @Override
    public RichIterable<T> distinctView()
    {
        try (LockWrapper wrapper = this.lockWrapper.acquireReadLock())
        {
            return new MultiReaderRichIterable<>(this.getDelegate().distinctView(), this.lock);
        }
    }
```

I think this should just throw UnsupportedOperationException. If someone wants to use this method, they will need to take a read or write lock beforehand.

victornoel:
@donraab so like `asSynchronized` and `asUnmodifiable` behave in this class basically?

Sounds good to me, it makes things much simpler to implement (i.e., I will get rid of `MultiReaderRichIterable`).

motlin:
Would you mind if I do the refactoring of creating AbstractMultiReaderMutableRichIterable to rebase on top of?

victornoel:
@motlin from what I understand by @donraab comment, we don't need MultiReaderRichIterable anymore. So yeah, feel free to tackle it in another MR if you want, but I won't use it once I apply @donraab feedbacks (except if you think differently of course).

motlin:
@victornoel @donraab whatever you prefer. I addressed the problems here: https://github.com/eclipse/eclipse-collections/pull/1255

donraab:
@motlin @victornoel Let's keep the changes separate, and leave only the `distinctView` changes in this PR. I would still leave `distinctView` throw UOE on `MultiReaderBag` as well. The number of places I have seen anyone use `MultiReaderBag` is small.

victornoel:
@donraab @motlin I've applied the changes as discussed and now I realize one more problem: we were expecting user of `MultiReaderBag.distinctView` to now rely on `with{Read,Write}LockAndDelegate()` if they needed to access the view, but as it is currently implemented, the object available through those methods (`UntouchableMutableBag`) follows the behaviour of the other view (unmodifiable and synchronized), i.e., throwing exception.
So now we have no way to access a distinct view on a `MultiReaderBag` :) I'm not not sure what to do with this situation, but maybe it's good enough for now and can find a solution when the need really arises?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/Bag.java
```diff
@@ -214,6 +214,14 @@
 
     ImmutableBagIterable<T> toImmutable();
 
+    /**
+     * Returns an unmodifiable view on the distinct elements with the same complexity as the Bag implementation.
+     *
+     * @return an unmodifiable view on the distinct elements of the Bag.
+     * @since 11.1
+     */
+    RichIterable<T> distinctView();
```
donraab, line 223: Hi @victornoel, this method will need a default implementation in order to be included in the 11.1 release. The simplest option would be to throw an `UnsupportedOperationException` with a message that this must be implemented in subclasses.

unit-tests/src/test/java/org/eclipse/collections/impl/bag/mutable/SynchronizedBagTest.java
```diff
@@ -302,4 +303,11 @@ public void selectUnique()
         MutableSet<String> actual = bag.selectUnique();
         Assert.assertEquals(expected, actual);
     }
+
+    @Test
+    public void distinctViewSynchronized()
+    {
+        RichIterable<Integer> actual = this.newWith(1, 1, 2, 3, 4, 4).distinctView();
+        Verify.assertInstanceOf(SynchronizedRichIterable.class, actual);
```
donraab, line 311: Please add an assertion that the distinct elements are `1, 2, 3, 4` as well.

donraab (changes requested):
Hi @victornoel, a couple of minor changes. I think this is looking good and should be able to be merged soon.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedBagImpl.java
```diff
@@ -792,6 +794,12 @@ public int hashCode()
         return counter.getCount();
     }
 
+    @Override
+    public RichIterable<T> distinctView()
+    {
+        return ImmutableTreeSet.wrap(this.elements, this.comparator);
```
motlin, line 800: Here, you could use `return ArrayAdapter.adapt(this.elements).asUnmodifiable();`

I realize that methods on the view like contains() would perform better if it could use a comparator. But if we need that, we should create a new array-as-set view rather than change an immutable type into a public/mutable one.

victornoel: It does not make it mutable: a user of the returned object cannot mutate it.
It's just that, like with `FastList.newListWith`, it is unsafe to modify the originally passed array to the wrap method, so it only concerns the creator of the object.

If you think anyway that it's best to introduce a new array-as-set class, then I will indeed go with ArrayAdapter and we can maybe create a ticket for improving thing in a later iteration.

victornoel: I just remembered I already had it implemented in my previous tentative to solve this issue, I will just do that then!

unit-tests/src/test/java/org/eclipse/collections/impl/bag/mutable/SynchronizedBagTest.java
victornoel: I added more tests in multiple places and this class should be covered as well :)

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedBagImpl.java
victornoel: done, can you check the naming of the new classes (one for unsorted and one for sorted) and their implementations? I just brought back what I had done for #797.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedBagImpl.java
```diff
@@ -792,6 +793,12 @@ public int hashCode()
         return counter.getCount();
     }
 
+    @Override
+    public RichIterable<T> distinctView()
+    {
+        return new ImmutableSortedSetFromArrayBagAdapter(this.comparator, this.elements);
```
donraab, line 799: IntelliJ inspections 4 build is failing on Raw use of parameterized class here

victornoel: Yep my bad. Anyway, I'm now convinced it's a bad idea to add those new classes since one is just going to reimplement the same thing as `ImmutableTreeSet` and the other is no better than just `ArrayAdapter`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedBagImpl.java
victornoel: Actually, sorry about the back and forth @motlin, but as I was sharing in https://github.com/eclipse/eclipse-collections/pull/1210#discussion_r841204941, I believe it's a bad idea to bring those classes. I will just go with the `ArrayAdapter` in both case and create a ticket to improve the situation later on. IMO the solution is really to use `ImmutableTreeSet`, but there is no rush to decide or find an alternative :)

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/sorted/immutable/ImmutableSortedBagImpl.java
donraab: @victornoel Makes sense.

donraab (changes requested):
@victornoel I'm on a mission to get this merged. I think it's ready and we can sort out any additional changes after the merge. Please update the Copyright notices in the files with 2022 and add "and others" where appropriate to the Copyright holder names. Thanks!

victornoel:
@donraab it should be good now, thx!

donraab (approved):
LGTM. Thank you for the contribution and your patience @victornoel !

motlin:
Thanks all. My availability has been hit or miss. I’m glad you got this
over the finish line in my absence.

victornoel:
Hehe, thank you all for all the feedback, it wasn't easy, but it's was mostly due to the nature of the problem!

PR #1212, opened 2022-03-02
Add primitive Set factories in API. Closes #1205.

### LINKED ISSUE #1205: Generate primitive Set factories in API using ServiceLoader.
This should follow the same pattern as the primitive List factories that are now generated in the API module.

rinatigati:


donraab (approved):
LGTM! Thanks for the contribution @rinatigati !

PR #1219, opened 2022-03-06
Remove UnifiedSet dependencies on primitive HashSet implementation...

donraab:
…classes.


rinatigati (approved):
LGTM! Javadoc17 build is not passing though.

I guess I missed BooleanSets in my commit because it's one that isn't actually generated. Doh!

The changes in UnifiedSet are similar to changes we'll have to make elsewhere to break dependencies, correct?

donraab:
Correct @rinatigati . We'll want to replace all instances of the primitive implementation constructors with the api factories.

PR #1223, opened 2022-03-08
Remove dependencies on primitive collection implementations.

donraab:

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/SetAdapter.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/SetAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs and others.
+ * Copyright (c) 2022 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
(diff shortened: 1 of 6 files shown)


emilie-robichaud (approved):
LGTM! 👍

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/SetAdapter.java
```diff
@@ -272,63 +272,63 @@ public int hashCode()
...
+        MutableBooleanSet result = BooleanSets.mutable.empty();
         this.forEach(new CollectBooleanProcedure<>(booleanFunction, result));
         return result;
     }
 
     @Override
     public MutableByteSet collectByte(ByteFunction<? super T> byteFunction)
     {
-        ByteHashSet result = new ByteHashSet(this.size());
+        MutableByteSet result = ByteSets.mutable.empty();
```
emilie-robichaud, line 283: why did you choose to initialize as empty here instead of using the set size?

donraab: It seemed odd to me that we were pre-sizing any collect methods for sets. It was even more odd for byte, which has at  most 256 values in a Set. I noticed some of the collectByte implementations didn't pre-size. We could pre-size with other types, but since we are collecting, we have no idea how to predict whether the source size will be equal to the target size anyway, so I opted for being consistent with our object collect for UnifiedSet which does not pre-size.

emilie-robichaud: Makes sense 👍

donraab:
There seems to be some issue with the Java 18 Early Access builds not being able to download in GitHub actions.

PR #1224, opened 2022-03-08
Add primitive Bag factories in API. Closes #1206.

### LINKED ISSUE #1206: Generate primitive Bag factories in API using ServiceLoader.
This should follow the same pattern as the primitive List factories that are now generated in the API module.

emilie-robichaud:


goldbal330 (approved):
LGTM

PR #1240, opened 2022-03-20
Migrate primitive -> object maps to service loader

Desislav-Petrov:
@donraab happy to hear any feedback - thanks

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/test/factory/primitiveObjectMapsTest.stg
+++ b/eclipse-collections-code-generator/src/main/resources/test/factory/primitiveObjectMapsTest.stg
@@ -1,0 +1,66 @@
+import "copyright.stg"
+import "primitiveLiteral.stg"
+
+isTest() ::= "true"
+
+targetPath() ::= "org/eclipse/collections/impl/factory/primitive"
+
+fileName(primitive) ::= "<primitive.name>ObjectMapFactoryTest"
+
+skipBoolean() ::= "true"
+
+class(primitive) ::= <<
+<body(primitive.type, primitive.name, primitive.wrapperName, primitive.charPrimitive)>
+>>
+
+body(type, name, wrapperName, charPrimitive) ::= <<
+<copyright()>
+
+package org.eclipse.collections.impl.factory.primitive;
+
+import org.eclipse.collections.api.factory.map.primitive.Immutable<name>ObjectMapFactory;
+import org.eclipse.collections.api.factory.map.primitive.Mutable<name>ObjectMapFactory;
+import org.eclipse.collections.impl.map.mutable.primitive.<name>ObjectHashMap;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * This file was automatically generated from template file primitiveObjectMapsTest.stg.
+ */
+public class <name>ObjectMapFactoryTest
+{
+    @Test
+    public void immutables()
+    {
+        this.assertImmutable<name>ObjectMapFactory(<name>ObjectMaps.immutable);
+        this.assertImmutable<name>ObjectMapFactory(org.eclipse.collections.api.factory.primitive.<name>ObjectMaps.immutable);
+    }
+
+    private void assertImmutable<name>ObjectMapFactory(Immutable<name>ObjectMapFactory mapFactory)
+    {
+        Assert.assertSame(<name>ObjectHashMap.newMap().toImmutable(), mapFactory.with());
+        Assert.assertSame(<name>ObjectHashMap.newMap().toImmutable(), mapFactory.of());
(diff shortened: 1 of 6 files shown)


eclipse-collections-code-generator/src/main/resources/test/factory/primitiveObjectMapsTest.stg
```diff
@@ -0,0 +1,69 @@
...
+
+package org.eclipse.collections.impl.map.primitive;
+
+import org.eclipse.collections.impl.factory.primitive.<name>ObjectMaps;
+import org.eclipse.collections.impl.map.mutable.primitive.<name>ObjectHashMap;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+/**mvn
```
rinatigati, line 27: Looks like there is a stray `mvn` here in this block comment.

eclipse-collections-code-generator/src/main/resources/test/factory/primitiveObjectMapsTest.stg
```diff
@@ -0,0 +1,69 @@
...
+skipBoolean() ::= "true"
+
+class(primitive) ::= <<
+<body(primitive.type, primitive.name, primitive.wrapperName, primitive.charPrimitive)>
+>>
+
+body(type, name, wrapperName, charPrimitive) ::= <<
+<copyright()>
+
+package org.eclipse.collections.impl.map.primitive;
```
rinatigati, line 19: Package should be `factory` instead of `map` since we're testing the factories.

eclipse-collections-code-generator/src/main/resources/test/factory/primitiveObjectMapsTest.stg
```diff
@@ -0,0 +1,69 @@
...
+
+import org.eclipse.collections.impl.factory.primitive.<name>ObjectMaps;
```
rinatigati, line 21: We don't want to be importing the `impl` ObjectMaps here since we're generating new ObjectMaps factories under the `org.eclipse.collections.api.factory.primitive` package in this PR. Want to be testing those factories.

There are some examples of testing factories in the other template files under this package like: https://github.com/eclipse/eclipse-collections/blob/master/eclipse-collections-code-generator/src/main/resources/test/factory/primitiveStacksTest.stg

You can take a look there to see how the factories are tested as well as take a look at the various assertions used. There are a few factory testing template files under `src/main/resources/test/factory` to check out.

rinatigati (changes requested):

eclipse-collections-code-generator/src/main/resources/test/factory/primitiveObjectMapsTest.stg
donraab: Thanks for reviewing @rinatigati ... @Desislav-Petrov the file linked above shows the strategy we have used for testing both the api and impl versions of the factory classes for primitive Stacks. We've used the same strategy across other factory classes as well.

eclipse-collections-code-generator/src/main/resources/test/factory/primitiveObjectMapsTest.stg
Desislav-Petrov: ops - my bad, fixed

eclipse-collections-code-generator/src/main/resources/test/factory/primitiveObjectMapsTest.stg
Desislav-Petrov: agreed

eclipse-collections-code-generator/src/main/resources/test/factory/primitiveObjectMapsTest.stg
rinatigati: I don't see it reflected in the PR or your fork yet, so it might not have been pushed yet. I'll resolve this as soon as I see it though! 👍

eclipse-collections-code-generator/src/main/resources/test/factory/primitiveObjectMapsTest.stg
Desislav-Petrov: Ah sry - think I finally got my head around it. So the idea here is that we get the factories obtained via the api module which will be calling the ServiceLoader.. which makes sense as this is how these will be obtained anyway. In fact, after re-working the test it caught a bug:) Thanks for the explanations

eclipse-collections-code-generator/src/main/resources/test/factory/primitiveObjectMapsTest.stg
Desislav-Petrov: @rinatigati hey apologies - just pushed the changes the, the ack for the issue was only reflected on my local branch as I was still not ready to repush

donraab (approved):
LGTM. Thanks for the contribution @Desislav-Petrov !

PR #1242, opened 2022-03-21
Add MultiReader to Sets factory in API module. Closes #1238

### LINKED ISSUE #1238: Add multiReader to Sets factory in API module
The Sets factory in the impl module has a multiReader version, that should be added to the API module version of Sets.

API module version:
https://github.com/eclipse/eclipse-collections/blob/master/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/Sets.java

Impl module version:
https://github.com/eclipse/eclipse-collections/blob/master/eclipse-collections/src/main/java/org/eclipse/collections/impl/factory/Sets.java

emilie-robichaud:


donraab (approved):

PR #1245, opened 2022-03-22
Add object -> primitive map factory to API module.

johnjstuart:


donraab (approved):
LGTM. Thanks for the contribution @johnjstuart !

PR #1250, opened 2022-03-23
Remove dependencies on primitive collection implementations

johnjstuart:


donraab (approved):
LGTM. Thanks for the contribution @johnjstuart !

rinatigati (commented):
I really think that the commit message needs to be more specific. This will be the 6th commit recently that will have the same exact commit message. There is no way to tell what is being changed at a glance in each individual commit/PR without looking into it.

johnjstuart:
> I really think that the commit message needs to be more specific. This will be the 6th commit recently that will have the same exact commit message. There is no way to tell what is being changed at a glance in each individual commit/PR without looking into it.

Yeah, good point. Updated.

prathasirisha:
@donraab it looks like the pending teamcity chekstyle build [passed ](https://teamcity.jetbrains.com/buildConfiguration/OpenSourceProjects_EclipseCollections_CheckStyle/3819433?showRootCauses=false&expandBuildChangesSection=true) but is not reflecting here. In any case, checkstyle GitHub actions build passed, we should be good to merge this PR. 

I will work with @nikhilnanivadekar on removing these redundant teamcity jobs.

donraab:
@johnjstuart @prathasirisha The checkstyle build is passing on TC. I kicked it off again to make sure. Not sure why the PR is not getting updated. I am going to merge anyway.

PR #1254, opened 2022-03-24
Add injectIntoKeyValue to MapIterable. Closes #1244.

### LINKED ISSUE #1244: Add injectIntoKeyValue to MapIterable
`MapIterable` currently has `injectInto` which takes the values of the map. There is no equivalent that takes both key and value. This will provide better symmetry with `forEachKeyValue` which does exist.

The signature for this method should be as follows:

`<IV> IV injectIntoKeyValue(IV injectedValue, Function3<? super IV, ? super K, ? super V, ? extends IV> function);
`

It will need to be implemented as a default method on `MapIterable`, and overridden as necessary.

emilie-robichaud:

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MapIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MapIterable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs.
+ * Copyright (c) 2022 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -21,6 +21,7 @@ import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.block.function.Function;
 import org.eclipse.collections.api.block.function.Function0;
 import org.eclipse.collections.api.block.function.Function2;
+import org.eclipse.collections.api.block.function.Function3;
 import org.eclipse.collections.api.block.predicate.Predicate2;
 import org.eclipse.collections.api.block.procedure.Procedure;
 import org.eclipse.collections.api.block.procedure.Procedure2;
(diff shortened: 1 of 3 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MapIterable.java
```diff
@@ -95,6 +98,23 @@
      */
     void forEachKeyValue(Procedure2<? super K, ? super V> procedure);
 
+    /**
+     * Calls {@code injectInto} with each <em>key-value</em> pair of the map.
```
donraab, line 102: I would rephrase this slightly to "Implements the injectInto pattern with each key-value..." since it doesn't actually call `injectInto`.

donraab (changes requested):
Hi @emilie-robichaud, thanks for the contribution. I requested a couple of changes inline. If you can also update copyright year and add "and others" if necessary. Thanks!

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MapIterable.java
```diff
@@ -95,6 +96,25 @@
      */
     void forEachKeyValue(Procedure2<? super K, ? super V> procedure);
 
+    /**
+     * Implements the {@code injectInto} pattern with each <em>key-value</em> pair of the map.
+     * <pre>
+     *     MapIterable&lt;Integer, Integer&gt; map1 = this.newWithKeysValues(3, 3, 2, 2, 1, 1);
```
donraab, line 102: Please replace `this.newWithKeysValues(3, 3, 2, 2, 1, 1)` in the example with `Maps.immutable.with(3, 3, 2, 2, 1, 1)`.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/MapIterableTestCase.java
```diff
@@ -183,6 +184,18 @@ default void MapIterable_forEachKeyValue()
...
     }
 
+    @Test
+    default void MapIterable_injectIntoKeyValue()
+    {
+        MapIterable<Integer, Integer> map1 = this.newWithKeysValues(3, 3, 2, 2, 1, 1);
+        Integer sum1 = map1.injectIntoKeyValue(0, (sum, key, value) -> sum + key + value);
+        assertEquals(12, sum1);
+
+        MutableMap<Integer, String> result = map1.injectIntoKeyValue(Maps.mutable.empty(), (map, key, value)
```
donraab, line 194: Please move the lambda parameter list to the next line so it is aligned with the separator and the expression.

donraab (approved):

PR #1273, opened 2022-04-02
A few improvements on Set implementations

victornoel:
Hi,

Here are a few commits that were laying around as left overs of my abandoned work on #797 and #869 :)

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/AbstractImmutableSet.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/AbstractImmutableSet.java
@@ -12,6 +12,7 @@ package org.eclipse.collections.impl.set.immutable;
 
 import java.util.Collection;
 import java.util.Iterator;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
 
@@ -81,6 +82,7 @@ import org.eclipse.collections.impl.block.procedure.RejectProcedure;
 import org.eclipse.collections.impl.block.procedure.SelectInstancesOfProcedure;
 import org.eclipse.collections.impl.block.procedure.SelectProcedure;
 import org.eclipse.collections.impl.collection.immutable.AbstractImmutableCollection;
+import org.eclipse.collections.impl.lazy.parallel.set.NonParallelUnsortedSetIterable;
 import org.eclipse.collections.impl.list.mutable.FastList;
 import org.eclipse.collections.impl.multimap.set.UnifiedSetMultimap;
 import org.eclipse.collections.impl.partition.set.PartitionUnifiedSet;
@@ -456,6 +458,11 @@ public abstract class AbstractImmutableSet<T> extends AbstractImmutableCollectio
     @Override
     public ParallelUnsortedSetIterable<T> asParallel(ExecutorService executorService, int batchSize)
     {
-        return this.toSet().asParallel(executorService, batchSize);
+        Objects.requireNonNull(executorService);
+        if (batchSize < 1)
+        {
+            throw new IllegalArgumentException("batchSize must be greater than zero, but was: " + batchSize);
+        }
+        return new NonParallelUnsortedSetIterable<>(this);
     }
 }
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/ImmutableUnifiedSet.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/ImmutableUnifiedSet.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs.
+ * Copyright (c) 2022 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -43,6 +43,11 @@ final class ImmutableUnifiedSet<T>
     @Override
     public boolean equals(Object other)
     {
+        if (other == this)
+        {
+            return true;
+        }
+
         return this.delegate.equals(other);
     }
 
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/MultiReaderUnifiedSet.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/MultiReaderUnifiedSet.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs and others.
+ * Copyright (c) 2022 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -426,6 +426,16 @@ public final class MultiReaderUnifiedSet<T>
     @Override
     public boolean equals(Object o)
     {
+        if (o == this || o == this.delegate)
+        {
+            return true;
+        }
+
+        if (!(o instanceof Set))
+        {
+            return false;
+        }
+
         try (LockWrapper wrapper = this.lockWrapper.acquireReadLock())
         {
             return this.delegate.equals(o);
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/SetAdapter.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/SetAdapter.java
@@ -144,6 +144,11 @@ public final class SetAdapter<T>
     @Override
     public boolean equals(Object obj)
     {
+        if (obj == this)
+        {
+            return true;
+        }
+
         return this.delegate.equals(obj);
     }
 
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnmodifiableMutableSet.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnmodifiableMutableSet.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs.
+ * Copyright (c) 2022 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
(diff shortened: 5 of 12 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
```diff
@@ -1282,8 +1282,20 @@ public boolean equals(Object object)
             return false;
         }
 
-        Set<?> other = (Set<?>) object;
-        return this.size() == other.size() && this.containsAll(other);
+        Set<?> s = (Set<?>) object;
```
donraab, line 1285: Hi @victornoel, this confused me at first but after digging around, I think I understand why you added the try catch. It's possible for two instances of `UnifiedSet` of different types to be compared, and it's possible `contains` can throw a `ClassCastException` in a `ChainedBucket`. I think this is possible a bug in `contains` and if it can be replicated in a test, should be fixed there. @mohrezaei @motlin @nikhilnanivadekar @prathasirisha What do you think?

victornoel: Actually, I added those try/catch because the javadoc from `Collection.containsAll` says they can optionally be thrown :)

donraab: Well, would you look at that. I guess since it's part of the contract, there is nothing for us to fix in `contains` or `containsAll`.

mohrezaei: I don't think catching the exceptions is actually sensible. `this.containsAll` is not some random method on an unknown set implementation. The implementation is right here. It resolves to:
```
        if (iterable instanceof MutableCollection)
        {
            return ((MutableCollection<T>) iterable).allSatisfyWith(predicate, parameter);
        }
        if (iterable instanceof ArrayList)
        {
            return ArrayListIterate.allSatisfyWith((ArrayList<T>) iterable, predicate, parameter);
        }
        if (iterable instanceof RandomAccess)
        {
            return RandomAccessListIterate.allSatisfyWith((List<T>) iterable, predicate, parameter);
        }
        if (iterable != null)
        {
            return IterableIterate.allSatisfyWith(iterable, predicate, parameter);
        }
```
notice how it never actually calls `containsAll` (on the unknown collection). So unless those exceptions are possible from `allSatisfyWith`, I'd suggest reverting this bit.

mohrezaei: Also, the doc is very specific about when those exceptions should be thrown. The NPE is clearly not allowed to be thrown by `UnifiedSet`, because we allow null elements. The CCE is a bit peculiar, as any `Object` can fit in a `Set`, especially at runtime when generic types are erased. I suppose a badly implemented `equals` method could throw a CCE, but that would be because it's a badly implemented `equals` method, not because the logic in the set has determined that something doesn't fit.

victornoel: I've removed this change and the other two similar after verifying they all were indeed calling `this.containsAll` that delegates to `allSatisfyWith`. Thx!

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/ImmutableUnifiedSet.java
```diff
@@ -43,6 +44,16 @@ public int size()
...
+        if (other == this)
+        {
+            return true;
+        }
+
+        if (!(other instanceof Set))
+        {
+            return false;
+        }
+
```
motlin, line 56: While this is fine, the delegate is a Set and is going to immediately perform the same check (whether Other is a Set)

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/AbstractImmutableSet.java
```diff
@@ -456,6 +458,11 @@ public boolean isProperSubsetOf(SetIterable<? extends T> candidateSuperset)
     @Override
     public ParallelUnsortedSetIterable<T> asParallel(ExecutorService executorService, int batchSize)
     {
-        return this.toSet().asParallel(executorService, batchSize);
+        Objects.requireNonNull(executorService);
+        if (batchSize < 1)
+        {
+            throw new IllegalArgumentException("batchSize must be greater than zero, but was: " + batchSize);
+        }
+        return new NonParallelUnsortedSetIterable<>(this);
```
motlin, line 466: Is this overridden? There a possible efficient implementation for ImmutableUnifiedSet.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/SetAdapter.java
```diff
@@ -144,6 +144,16 @@ public boolean containsAll(Collection<?> collection)
...
     {
+        if (obj == this)
+        {
+            return true;
+        }
+
+        if (!(obj instanceof Set))
+        {
+            return false;
+        }
```
motlin, line 155: Same here. The delegate is a Set, so it's contractually obligated to perform the same check.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnmodifiableMutableSet.java
```diff
@@ -104,6 +104,16 @@
...
     {
+        if (obj == this)
+        {
+            return true;
+        }
+
+        if (!(obj instanceof Set))
+        {
+            return false;
+        }
```
motlin, line 115: Same here

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/immutable/AbstractImmutableSortedSet.java
```diff
@@ -488,13 +489,10 @@ public boolean isProperSubsetOf(SetIterable<? extends T> candidateSuperset)
...
-        {
-            throw new NullPointerException();
-        }
+        Objects.requireNonNull(executorService);
         if (batchSize < 1)
         {
-            throw new IllegalArgumentException();
+            throw new IllegalArgumentException("batchSize must be greater than zero, but was: " + batchSize);
         }
         return new NonParallelSortedSetIterable<>(this);
```
motlin, line 499: Is this overridden? I can't think of a good way to make this efficient with the building blocks we have.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/mutable/SortedSetAdapter.java
```diff
@@ -157,6 +157,16 @@ public boolean containsAll(Collection<?> collection)
...
     {
+        if (obj == this)
+        {
+            return true;
+        }
+
+        if (!(obj instanceof Set))
+        {
+            return false;
+        }
```
motlin, line 168: Same here

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/MultiReaderUnifiedSet.java
```diff
@@ -426,6 +426,16 @@ public MutableShortSet collectShort(ShortFunction<? super T> shortFunction)
...
     {
+        if (o == this || o == this.delegate)
+        {
+            return true;
+        }
+
+        if (!(o instanceof Set))
+        {
+            return false;
+        }
```
motlin, line 437: This one is worth keeping because it avoids grabbing the lock. Is it covered by tests?

victornoel: I'm not sure, but I updated that test because it wasn't passing anymore with the change: `MultiReaderUnifiedSetScalaTest`

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/immutable/AbstractImmutableSortedSet.java
victornoel: Yep, just in `ImmutableTreeSet` but not all the methods are implemented with parallelization

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/AbstractImmutableSet.java
victornoel: Yes indeed, there are two overrides for this in `ImmutableUnifiedSet` and `ImmutableUnifiedSetWithHashingStrategy` (delegating to the wrapped `UnfiedSet`).

I haven't looked into other subclasses, the purpose here was really to avoid the costly call to `toSet()`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/immutable/ImmutableUnifiedSet.java
victornoel: right! a bit of over-engineering ^^ thx!

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/mutable/TreeSortedSet.java
```diff
@@ -253,13 +253,7 @@ public boolean equals(Object object)
...
         }
 
-        if (!(object instanceof Set))
-        {
-            return false;
-        }
-
-        Set<?> other = (Set<?>) object;
-        return other.equals(this.treeSet);
+        return this.treeSet.equals(object);
```
motlin, line 256: I'm not opposed to this change, but it was a weird sort of optimization. What is the speed of treeSet.equals(unifiedSet)? It could be linear or O(n log n). You iterate over one and call contains on the other.

motlin (approved):

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/mutable/TreeSortedSet.java
victornoel: `equals` in object-oriented APIs has always felt wrong to me because of this kind of things, ideally `a.equals(b)` should be the same as `b.equals(a)`, but in reality, there could be some situation like here where one way is maybe better than the other (until it's not because you don't know the concrete implementation of `object` :)

thx for the review in any case!

PR #1277, opened 2022-04-04
Remove unnecessary implementations of toSortedList/Set/Bag/MapBy.

donraab:

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/AbstractSynchronizedRichIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/AbstractSynchronizedRichIterable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs and others.
+ * Copyright (c) 2022 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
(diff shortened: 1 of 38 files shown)


prathasirisha (approved):

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/AbstractSynchronizedRichIterable.java
```diff
@@ -949,15 +930,6 @@ public double injectInto(double injectedValue, DoubleObjectToDoubleFunction<? su
...
 
-    @Override
-    public <V extends Comparable<? super V>> MutableSortedBag<T> toSortedBagBy(Function<? super T, ? extends V> function)
-    {
-        synchronized (this.lock)
-        {
-            return this.delegate.toSortedBag(Comparators.byFunction(function));
-        }
-    }
-
```
motlin, line 960: Some of these methods, specifically the ones on Synchronized wrappers, are ones that we've left in the past even if perhaps redundant because it's easier to reason about and ensure that everything is synchronized if every method is implemented this way.

On the other hand, if we're ok with removing methods like this, there are so many more that could be removed.

I'm not sure I have a strong preference but this seems inconsistent.

donraab: @motlin I agree. I think if we can get better at ensuring thread safety for the entire API in tests, then we can remove more composite methods like this.

PR #1294, opened 2022-04-10
Update ParallelIterate and FJIterate JavaDoc to use lambdas.

donraab:


donraab:
@motlin When you get a chance if you can review, would be appreciated. Thanks!

motlin (approved):

PR #1298, opened 2022-04-11
Use org.eclipse.collections.api.factory instead of org.eclipse.collections.impl.factory wherever possible.

motlin:


donraab (approved):
LGTM

PR #1300, opened 2022-04-11
Improve and combine the flatCollect primitive tests in unit-tests-java8.

motlin:


motlin:
In addition to combining tests like in the other two PRs, these assertions for flatCollect are now consistent with collect. Before, they were not overridden in RichIterableUniqueTestCase.java and SortedNaturalOrderTestCase.java which means the assertions must not have covered ordering and duplicates.

donraab (approved):

PR #1302, opened 2022-04-11
Combine tests RichIterable_detect and RichIterable_detectOptional.

motlin:


donraab (commented):
My preference would be to see these test split up more, not combined.

motlin:
One of the reasons that `unit-tests-java8` is so much faster than `unit-tests` is because it doesn't split up tests the same way. The idea in `unit-tests-java8` is that if the methods appears in the same places (on the same interfaces, overridden for efficiency in the same classes, etc) then they ought to be tested together.

Conversely, one of the reasons that tests in `unit-tests` are so slow is that they are so small that they can take longer to instantiate than to run. This is especially noticeable in the IDE, and I generally won't run the tests in `unit-tests` on battery power on my laptop.

On my machine, I see `unit-tests-java8` run in 4.735s. With the changes I'm working on, I see 8,597 tests. The tests are so compact that it's practical to run them with coverage gathering on. On my laptop, on battery power, I can run them with coverage in 7.07s.

donraab (approved):

PR #1305, opened 2022-04-11
Replace FastList.newList(iterable) with Lists.mutable.withAll().

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/CompositeFastList.java
```diff
@@ -238,7 +239,7 @@ public boolean addAll(Collection<? extends E> collection)
             return false;
         }
         Collection<? extends E> collectionToAdd =
-                collection instanceof FastList ? collection : FastList.newList(collection);
+                collection instanceof FastList ? collection : Lists.mutable.withAll(collection);
```
donraab, line 242: I think this needs to be left as `FastList` as the implementation must be a `FastList`.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
```diff
@@ -126,7 +126,7 @@ public static <T> MultiReaderFastList<T> newList(int capacity)
 
     public static <T> MultiReaderFastList<T> newList(Iterable<? extends T> iterable)
     {
-        return new MultiReaderFastList<>(FastList.newList(iterable));
```
donraab, line 129: I think this should be left as a `FastList` as this is a `MultiReaderFastList`.

donraab (changes requested):
A few suggestions inline.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/CompositeFastList.java
motlin: Discarded this change.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/MultiReaderFastList.java
motlin: Discarded this change to MultiReaderFastList.

donraab (approved):
LGTM. I'll merge once builds complete. Thanks for the contribution @motlin !

PR #1309, opened 2022-04-13
Disambiguate and deprecate primitive injectInto methods on RichIterable.

donraab:


donraab:
@motlin I've wanted to deprecate these methods and add non-ambiguous ones for a while. These methods were added as overloads before lambdas in Java 8, and we discovered after they can have ambiguity issues when used with lambdas.

motlin (approved):
You could consider changing "use injectIntoInt instead" to use a @link but it's not super important.

PR #1310, opened 2022-04-13
Replace UnifiedSet.newSet(iterable) with Sets.mutable.withAll().

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/SetIterables.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs and others.
+ * Copyright (c) 2022 Goldman Sachs and others.
```
donraab, line 2: I think you meant to replace UnifiedSet with Sets in here.

donraab (changes requested):

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/SetIterables.java
motlin: I reverted the change and forgot to revert the copyright. I'll take care of this shortly.

donraab (approved):
LGTM. @motlin I am fine if you want to merge the copyright year-update only change on the one file.

PR #1313, opened 2022-04-13
Replace UnifiedSet.newSet(initialCapacity) with Sets.mutable.withInitialCapacity().

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/MultiReaderUnifiedSet.java
```diff
@@ -109,7 +109,7 @@ public static <T> MultiReaderUnifiedSet<T> newSet()
 
     public static <T> MultiReaderUnifiedSet<T> newSet(int capacity)
     {
-        return new MultiReaderUnifiedSet<>(UnifiedSet.newSet(capacity));
+        return new MultiReaderUnifiedSet<>(Sets.mutable.withInitialCapacity(capacity));
```
donraab, line 112: This should remain `UnifiedSet`.

donraab (changes requested):
One change requested.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/MultiReaderUnifiedSet.java
motlin: Oops, fixed.

donraab (approved):
LGTM

PR #1322, opened 2022-04-16
add unit test for EclipseCollectionsCodeGenerator

sullis:
Add unit test for EclipseCollectionsCodeGenerator. 

EclipseCollectionsCodeGeneratorTest validates the generated Java source files for:
- api
- impl
- test

This PR uses the [classgraph](https://github.com/classgraph/classgraph) library to load resources from the classpath.

When I started this PR, I tried casting the system class loader to URLClassLoader. This approach worked on Java 8 but did not work on later versions of Java.

Reference:  https://stackoverflow.com/questions/54618721/migration-to-java11-classcastexception


motlin:
I wasn't familiar with the ClassGraph library before. I like how using it makes it possible to delete so much code from FileUtils. To me, it seems like introducing ClassGraph is the essence of this change. In addition, there are more changes to production code and templates. However, the PR and commit message only mentions adding unit tests.

sullis:
> I wasn't familiar with the ClassGraph library before. I like how using it makes it possible to delete so much code from FileUtils. To me, it seems like introducing ClassGraph is the essence of this change. In addition, there are more changes to production code and templates. However, the PR and commit message only mentions adding unit tests.

I have updated the PR title and PR description.  The description now contains a better explanation of my work.

motlin:
EclipseCollectionsCodeGenerator.generate() delegates to `EclipseCollectionsCodeGenerator.sourceFileExists(outputFile)` before rendering a file. So if a hand-written boolean version exists, we won't generate code for the template anyway.

eclipse-collections-code-generator/src/test/java/org/eclipse/collections/codegenerator/EclipseCollectionsCodeGeneratorTest.java
```diff
@@ -0,0 +1,83 @@
...
+    public void validateImplTemplates() throws Exception
+    {
+        validateTemplates("impl", 1469);
+    }
+
+    @Test
+    public void validateTestTemplates() throws Exception
+    {
+        validateTemplates("test", 2342);
+    }
```
motlin, line 46: I'd prefer not to assert an exact expected number of files, or people will be forced to change it all the time. Asserting it's not an empty collection would be fine IMO.

eclipse-collections-code-generator/src/test/java/org/eclipse/collections/codegenerator/EclipseCollectionsCodeGeneratorTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private static void validateTemplates(String templateDirectory, int expectedFileCount) throws Exception
+    {
+        BasicErrorListener errorListener = new BasicErrorListener();
+        File outputDirectory = Files.createTempDirectory("temp-").toFile();
```
motlin, line 51: We can use JUnit's TemporaryFolder rule so that the files are deleted during tearDown of the test rather than the JVM.

eclipse-collections-code-generator/src/test/java/org/eclipse/collections/codegenerator/EclipseCollectionsCodeGeneratorTest.java
```diff
@@ -0,0 +1,83 @@
...
+        assertTrue(file.canRead());
+        assertTrue(file.length() > 0);
+        ParserConfiguration config = new ParserConfiguration();
+        config.setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_8);
+        JavaParser parser = new JavaParser(config);
+        String code = new String(Files.readAllBytes(file.toPath()), StandardCharsets.UTF_8);
+        ParseResult<CompilationUnit> parseResult = parser.parse(code);
+        assertTrue(String.valueOf(parseResult.getProblems()) + "\n" + code,
+                parseResult.isSuccessful());
+    }
```
motlin, line 82: I think it's kind of cool to have a test assertion like this that asserts a string is valid Java code. But I don't think we really need it either. Our builds generate code and compile it with javac from Java 8, so if this assertion were to fail our build would already be broken. Thoughts?

There are other assertions I could imagine that would be pretty interesting. Are there any places we declare skipBoolean = true but the generated source code matches the hand-written source code? So we don't need to skip boolean generation?

donraab:
@motlin @sullis I think it makes sense to break this into separate PRs. The skipBoolean change looks straightforward and we should review that separately. If this means we are removing currently generated classes/interfaces then we'll have to think about this, but since 12.0 is a major release I don't have as many concerns about this. On the Classgraph front, looks like it should be ok, but would ask @nikhilnanivadekar if he things this will need Eclipse Foundation approval. It's a build time dependency only, and MIT license, so should be ok I think.

PR #1331, opened 2022-04-19
Add injectIntoKeyValue to objectPrimitiveMap. Closes #1217.

### LINKED ISSUE #1217: Add injectIntoKeyValue to primitive Maps
Today we can iterate over the key / value combinations in primitive maps using `forEachkeyValue`. It would be nice to have a method which injects some mutable result into the iteration and returns it. The following is possible today for instance for transferring a primitive map's data to an object map.

```
IntObjectMap source;
MutableMap<?, ?> target = Maps.mutable.empty();
source.forEachKeyValue(target::put);
```

If we add `injectIntoKeyValue` to primitive maps, we can write the following:

`MutableMap<?, ?> target = source.injectIntoKeyValue(Maps.mutable.empty(), target::withKeyValue);
`

Today, we don't have any easier way to copy or transform data from a primitive map to an object map.

This issue came up in a question on SO today. The OP was looking for a Map view of the primitive maps so they could pass it to another API. 

https://stackoverflow.com/questions/71349112/how-convert-adapt-byteobjecthashmap-to-a-jdk-map

emilie-robichaud:

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/block/function/objectObjectPrimitiveToObjectFunction.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/block/function/objectObjectPrimitiveToObjectFunction.stg
@@ -1,0 +1,30 @@
+import "copyrightAndOthers.stg"
+
+targetPath() ::= "org/eclipse/collections/api/block/function/primitive"
+
+fileName(primitive) ::= "ObjectObject<primitive.name>ToObjectFunction"
+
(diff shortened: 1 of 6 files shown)


donraab (changes requested):
Thank you for the contribution @emilie-robichaud. Code looks good.  Please add the copyrighAndOthers in all of the templates you've changed and I will approve merge.

donraab (approved):

Issue #1323, opened 2022-04-16
Turn on additional IntelliJ inspections and fix violations.


motlin:
It's been a while since we turned on additional IntelliJ inspections. There are many newly created inspections that are worth adding. Older inspections that were buggy have been fixed or got new options so we can configure them to have no false positives.

When turning on inspections, we currently have to make changes in two profiles, which correspond with two config files.

1.  Turn on the new inspection in `.idea/inspectionProfiles/IDE.xml` at the INSPECTION or WARNING severity. This makes the warnings visible in the text editor.
2. Turn on the new inspection in one of `.idea/inspectionProfiles/{1_Severe.xml,2_Inconsistent_Constructs.xml,3_Consistent_Style.xml,4_No_Auto_fix.xml}` at the ERROR severity. This makes the error appear in one of the four relevant TeamCity builds.

**1_Severe.xml**
This profile contains a few severe inspections.

**2_Inconsistent_Constructs.xml**

This profile contains inspections which flag constructs that are pointless, or a little bit wrong.

Add:

* Method is identical to its super method
* Constant on wrong side of comparison (constants go on the right)
* Unused assignment
* Redundant type cast
* Collection declared by class, not interface
* Type may be weakened (configured with our interfaces as stop classes)
* Interface may be annotated as ‘@FunctionalInterface’
* ‘size() == 0’ can be replaced with ‘isEmpty()’

**3_Consistent_Style.xml**

This profile contains inspections which flag constructs that are redundant or inconsistent. There is plenty over overlap with profile (2). When in doubt, inspections that affect the bytecode go into (2) and inspections that don't go into (3).

Add:

* Explicit type argument T can be replaced with <>
* Missing ‘@Override’ annotation
* Commented out code
* ‘if’ statement can be replaced with conditional or boolean expression
* Redundant ‘if’ statement

**4_No_Auto_fix.xml**

This profile contains inspections which have warnings but no auto-fixes. These are more difficult to fix and must be used sparingly. Examples include naming conventions, where an author has to stop and pick a name.

goldbal330:
I could work on it.

PR #1336, opened 2022-04-25
Turn on additional IntelliJ inspections and fix violations. #1323

goldbal330:
In this PR I enabled two new inspections:

Interface may be annotated as ‘@FunctionalInterface’
‘size() == 0’ can be replaced with ‘isEmpty()’

I kept size() == 0 check in some methods in Verify class since it seemed like those calls were used to validate that size() method is working as expected.

Signed-off-by: Alexander Goldberg <alexander.goldberg@bnymellon.com>

### DIFF
--- a/.idea/inspectionProfiles/IDE.xml
+++ b/.idea/inspectionProfiles/IDE.xml
@@ -764,6 +764,7 @@
     <inspection_tool class="StaticCollection" enabled="true" level="INSPECTION" enabled_by_default="true">
       <option name="m_ignoreWeakCollections" value="false" />
     </inspection_tool>
+    <inspection_tool class="InterfaceMayBeAnnotatedFunctional" enabled="true" level="WARNING" enabled_by_default="true" />
     <inspection_tool class="StaticFieldReferenceOnSubclass" enabled="true" level="INSPECTION" enabled_by_default="true" />
     <inspection_tool class="StaticGuardedByInstance" enabled="true" level="INSPECTION" enabled_by_default="true" />
     <inspection_tool class="StaticImport" enabled="true" level="WARNING" enabled_by_default="true">
--- a/jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/IntIntMapSmallStressTest.java
+++ b/jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/IntIntMapSmallStressTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs.
+ * Copyright (c) 2022 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -319,7 +319,7 @@ public class IntIntMapSmallStressTest extends AbstractJMHTestRunner
             {
                 newMap.remove(this.ecIntKeysForMap[i]);
             }
-            if (newMap.size() != 0)
+            if (newMap.notEmpty())
             {
                 throw new AssertionError("size is " + newMap.size());
             }
@@ -336,7 +336,7 @@ public class IntIntMapSmallStressTest extends AbstractJMHTestRunner
             {
                 newMap.remove(this.jdkIntKeysForMap[i]);
             }
-            if (newMap.size() != 0)
+            if (!newMap.isEmpty())
             {
                 throw new AssertionError("size is " + newMap.size());
             }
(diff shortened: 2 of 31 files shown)


jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/IntIntMapSmallStressTest.java
```diff
@@ -336,7 +336,7 @@ public void jdkRemove()
             {
                 newMap.remove(this.jdkIntKeysForMap[i]);
             }
-            if (newMap.size() != 0)
+            if (!newMap.isEmpty())
```
donraab, line 339: Could be `notEmpty` like above?

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/IntIntMapSmallStressTest.java
```diff
@@ -353,7 +353,7 @@ public void kolobokeRemove()
             {
                 newMap.remove(this.kolobokeIntKeysForMap[i]);
             }
-            if (newMap.size() != 0)
+            if (!newMap.isEmpty())
```
donraab, line 356: `notEmpty()`?

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/LongLongMapLargeStressTest.java
```diff
@@ -246,7 +246,7 @@ public void kolobokeRemove()
             {
                 newMap.remove(this.kolobokeLongKeysForMap[i]);
             }
-            if (newMap.size() != 0)
+            if (!newMap.isEmpty())
```
donraab, line 249: `notEmpty()`?

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/LongLongMapSmallStressTest.java
```diff
@@ -252,7 +252,7 @@ public void kolobokeRemove()
             {
                 newMap.remove(this.kolobokeLongKeysForMap[i]);
             }
-            if (newMap.size() != 0)
+            if (!newMap.isEmpty())
```
donraab, line 255: `notEmpty()`?

donraab (changes requested):
Looks mostly good @goldbal330. Thanks for contribution. A few questions inline.

.idea/inspectionProfiles/IDE.xml
```diff
@@ -764,6 +764,7 @@
     <inspection_tool class="StaticCollection" enabled="true" level="INSPECTION" enabled_by_default="true">
       <option name="m_ignoreWeakCollections" value="false" />
     </inspection_tool>
+    <inspection_tool class="InterfaceMayBeAnnotatedFunctional" enabled="true" level="WARNING" enabled_by_default="true" />
```
motlin, line 767: Here you turn on InterfaceMayBeAnnotatedFunctional, but alphabetically this looks like the position for SizeReplaceableByIsEmpty maybe? Is that one also turned on elsewhere?

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/IntIntMapSmallStressTest.java
goldbal330: newMap is of type IntIntMap. IntIntMap interface does not have a notEmpty() method.

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/LongLongMapLargeStressTest.java
goldbal330: The class is coming from net.openhft.koloboke.collect.map.hash package and does not offer a notEmpty() method.

jmh-tests/src/main/java/org/eclipse/collections/impl/jmh/LongLongMapSmallStressTest.java
goldbal330: The class is coming from net.openhft.koloboke.collect.map.hash package and does not offer a notEmpty() method

.idea/inspectionProfiles/IDE.xml
goldbal330: It has already been turned on in IDE.xml (line 742)

goldbal330: <inspection_tool class="SizeReplaceableByIsEmpty" enabled="true" level="WARNING" enabled_by_default="true" />

donraab (approved):

PR #1341, opened 2022-05-01
Add tests for containsAny/containsNone to RichIterableTestCase

donraab:

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/AbstractMapIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/AbstractMapIterable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs.
+ * Copyright (c) 2022 Goldman Sachs.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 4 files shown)


mohrezaei (changes requested):
`this.containsValue` is `O(n)`, which makes these methods `O(m*n)` all the time. 

If the incoming collection has a fast `contains` (because it's tiny or `O(1)`), the order should be reversed.

donraab:
> `this.containsValue` is `O(n)`, which makes these methods `O(m*n)` all the time.
> 
> If the incoming collection has a fast `contains` (because it's tiny or `O(1)`), the order should be reversed.

Yes, agreed. The `containsAll` implementation is currently suffering from this as well. I wanted to get the patterns working first without iterators as `containsAll` is. Then I figured I would try and tackle optimizing all three patterns on `UnifiedMap`.

donraab:
@mohrezaei I discovered there are some open TODOs to optimize `UnifiedMapWithHashingStrategy` for `anySatisfy`/`allSatisfy`/`noneSatisfy` to not use iterator. I have reverted my changes as `UnifiedMap` was already optimal, and I've overridden the failing tests for `UnifiedMapWithHashingStrategy`.

The optimizations of `anySatisfy`/`allSatisfy`/`noneSatisfy` can be done in a separate PR.

mohrezaei (approved):

Issue #1328, opened 2022-04-18
Add SortedMaps.mutable.empty(Comparator) and SortedSets.mutable.empty(Comparator).


motlin:
We're missing symmetry. This exists and works:

```java
SortedBags.mutable.empty(comparator);
```

But these don't exist.

```java
SortedMaps.mutable.empty(comparator);
SortedSets.mutable.empty(comparator);
```

Desislav-Petrov:
I can take that if u don't mind

PR #1345, opened 2022-05-03
Add missing empty with comparator methods

Desislav-Petrov:


unit-tests/src/test/java/org/eclipse/collections/impl/factory/SortedMapsTest.java
```diff
@@ -0,0 +1,26 @@
...
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.impl.factory;
+
+import org.eclipse.collections.api.factory.map.sorted.MutableSortedMapFactory;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class SortedMapsTest
```
Desislav-Petrov, line 17: Looks like we didn't have this - is there a reason for that? If not, i'll expand the test cases on a subsequent PR

Desislav-Petrov:
hi @donraab - would appreciate a PR if u have a moment

motlin:
Could you link to the issue this fixes, with something like `Fixes #1328` in the commit message?

Desislav-Petrov:
@motlin Sure - done

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/map/sorted/MutableSortedMapFactory.java
```diff
@@ -22,6 +22,8 @@
      */
     <K, V> MutableSortedMap<K, V> empty();
 
+    <K, V> MutableSortedMap<K, V> empty(Comparator<? super K> comparator);
```
donraab, line 25: This will require a `since` tag of 12.0 since it is an API change without a default implementation.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/set/sorted/MutableSortedSetFactory.java
```diff
@@ -21,6 +21,8 @@
      */
     <T> MutableSortedSet<T> empty();
 
+    <T> MutableSortedSet<T> empty(Comparator<? super T> comparator);
```
donraab, line 24: This will require a `since` tag of 12.0 since it is an API change without a default implementation.

unit-tests/src/test/java/org/eclipse/collections/impl/factory/SortedMapsTest.java
donraab: No reason that I can think of. It may have just been missing. Thank you for adding it.

donraab (changes requested):
Hi @Desislav-Petrov, thanks for the contribution. This will have to wait for a major release as there are API changes with no default implementations.

donraab:
Hi @Desislav-Petrov, took a look at this and should be simple enough to review and merge now since next planned release is a major release.

donraab (approved):
LGTM, just needs to be rebased.

Desislav-Petrov:
hi @donraab - rebased now - thanks

Desislav-Petrov:
hi @donraab - rebased now - thanks

donraab:
Hi @Desislav-Petrov, I think you need to fetch and rebase again. It looks like you are behind a few commits.

Desislav-Petrov:
hi @donraab  - checked history, looks like it should be good now

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/mutable/MutableSortedMapFactoryImpl.java
```diff
@@ -27,6 +27,12 @@ public <K, V> MutableSortedMap<K, V> empty()
         return TreeSortedMap.newMap();
     }
 
+    @Override
+    public <K, V> MutableSortedMap<K, V> empty(Comparator<? super K> comparator)
+    {
```
donraab, line 32: @Desislav-Petrov Please move the code from `with(Comparator)` to here with the null check for comparator and change with(Comparator) to call empty().

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/mutable/MutableSortedSetFactoryImpl.java
```diff
@@ -25,6 +25,12 @@ public <T> MutableSortedSet<T> empty()
         return TreeSortedSet.newSet();
     }
 
+    @Override
+    public <T> MutableSortedSet<T> empty(Comparator<? super T> comparator)
+    {
+        return TreeSortedSet.newSet(comparator);
```
donraab, line 31: @Desislav-Petrov Please move the code from with(Comparator) to here and change with(Comparator) to call empty().

donraab (changes requested):
Hi @Desislav-Petrov, I noticed a small duplication that should be fixed by changing with calls with comparator to call empty with comparator.

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/mutable/MutableSortedSetFactoryImpl.java
Desislav-Petrov: Good spot @donraab  - addressed

eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/mutable/MutableSortedSetFactoryImpl.java
```diff
@@ -25,6 +25,16 @@ public <T> MutableSortedSet<T> empty()
         return TreeSortedSet.newSet();
     }
 
+    @Override
+    public <T> MutableSortedSet<T> empty(Comparator<? super T> comparator)
+    {
+        if (comparator == null)
```
donraab, line 31: Hi @Desislav-Petrov, I went and looked and was curious why we didn't have the similar null handling logic for Sets. I would prefer to keep the logic the same as it looks like TreeSet itself has the null handling logic.

Desislav-Petrov: Got u - I was trying to be consistent with what we do in the other place but I guess for sets is different. Will fix.

donraab: Thanks!

Desislav-Petrov: hi @donraab - the null check for sortedset is removed now

donraab (approved):
LGTM! Thanks for the contribution!

PR #1346, opened 2022-05-11
Make noneSatisfy a default method on primitiveIterable.stg.

donraab:

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/impl/lazy/collectPrimitiveIterable.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/lazy/collectPrimitiveIterable.stg
@@ -119,12 +119,6 @@ public class Collect<name>Iterable\<T>
     }
 
     @Override
-    public boolean noneSatisfy(final <name>Predicate predicate)
-    {
-        return this.iterable.allSatisfy((T each) -> !predicate.accept(this.function.<type>ValueOf(each)));
-    }
-
-    @Override
     public <type>[] toArray()
     {
         final <type>[] array = new <type>[this.size()];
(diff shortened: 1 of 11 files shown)


donraab:
@motlin When you get a chance if you can review. This felt like a decent change to make to reduce some code duplication on the primitive side. We can probably replicate on the object collection side as well eventually if this makes sense to you.

eclipse-collections-code-generator/src/main/resources/impl/lazy/collectPrimitiveIterable.stg
```diff
@@ -118,12 +118,6 @@ public class Collect<name>Iterable\<T>
         return this.iterable.allSatisfy((T each) -> predicate.accept(this.function.<type>ValueOf(each)));
     }
 
-    @Override
-    public boolean noneSatisfy(final <name>Predicate predicate)
-    {
-        return this.iterable.allSatisfy((T each) -> !predicate.accept(this.function.<type>ValueOf(each)));
-    }
-
```
motlin, line 126: All other implementations of none are !any(predicate). This one is all(!predicate). That implies any and all can be written in terms of each other too.

any --> !all(!predicate)
all --> !any(!predicate)

Is that right?

motlin (approved):

eclipse-collections-code-generator/src/main/resources/impl/lazy/collectPrimitiveIterable.stg
donraab: This makes sense. The difference is that none can be defined with a single ! on the outside when using any. Quite a while ago we changed the basic implementations of all three to use a common `shortCircuit` method which took multiple `boolean` parameters.

```
public static <T> boolean shortCircuit(
        T[] array,
        int size,
        Predicate<? super T> predicate,
        boolean expected,
        boolean onShortCircuit,
        boolean atEnd)
{
    for (int i = 0; i < size; i++)
    {
        T each = array[i];
        if (predicate.accept(each) == expected)
        {
            return onShortCircuit;
        }
    }
    return atEnd;
}
```

Issue #1344, opened 2022-05-02
Optimize containsAll on RichIterable


donraab:
There are now optimized versions of `containsAny` and `containsNone` on `RichIterable`. Similar optimizations could be implemented for `containsAll`, as they have been for primitive collections.

It is more work for `containsAll`, as this is an abstract method on `Collection` and `List`, so needs to be overridden in multiple places in addition to defining a default method on `RichIterable`. The `containsAny` and `containsNone` implementations were able to be done primarily through the default method implementations on `RichIterable`.

znight1020:
Hi! @donraab, I’d like to work on this, if it is still available.

My current direction is to optimize `containsAll` for large non-set `RichIterable` instances and update the relevant `Collection` / `List` overrides so those types delegate to the optimized implementation.

PR #1353, opened 2022-05-21
Optimize containsAll on RichIterable. #1344

youssef-abdallah:
Signed-off-by: YoussefAbdallah <youssefabdullah@hotmail.com>


eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -295,17 +295,38 @@ else if (inside.size() > 32 && !(inside instanceof SetIterable))
     /**
      * Returns true if all elements in source are contained in this collection.
      *
-     * @since 1.0
+     * @since 11.1
```
motlin, line 298: We shouldn't change since tags. This method really was here since version 1.0.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -295,17 +295,38 @@ else if (inside.size() > 32 && !(inside instanceof SetIterable))
...
+    default boolean containsAllIterable(Iterable<?> source)
+    {
+        if (source instanceof RichIterable)
+        {
+            RichIterable<?> outside = this;
+            RichIterable<?> inside = (RichIterable<?>) source;
+            if (this.size() < inside.size())
+            {
+                return false;
+            }
```
motlin, line 309: On line 306, we've already called size() on both collections. If the sizes are unequal, we can return false. In other words, replace `<` with `!=`.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -295,17 +295,38 @@ else if (inside.size() > 32 && !(inside instanceof SetIterable))
...
      *
-     * @since 1.0
+     * @since 11.1
      */
-    boolean containsAllIterable(Iterable<?> source);
+    default boolean containsAllIterable(Iterable<?> source)
+    {
+        if (source instanceof RichIterable)
+        {
+            RichIterable<?> outside = this;
```
motlin, line 304: We already know that `this` is a RichIterable, because we're inside the interface. No need to assign it to another variable.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -295,17 +295,38 @@ else if (inside.size() > 32 && !(inside instanceof SetIterable))
...
+            RichIterable<?> inside = (RichIterable<?>) source;
+            if (this.size() < inside.size())
+            {
+                return false;
+            }
+            if (outside.size() > 32 && !(outside instanceof SetIterable))
+            {
+                outside = outside.toSet();
+            }
```
motlin, line 313: How did you pick 32?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -295,17 +295,38 @@ else if (inside.size() > 32 && !(inside instanceof SetIterable))
...
+            return inside.allSatisfy(outside::contains);
```
motlin, line 314: If you use `Iterate.allSatisfy()` then you don't need to check whether `source` is a RichIterable.

motlin (changes requested):

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
youssef-abdallah: Would this be correct? `A.containsAll(B)` should return `true` iff all the elements of `B` are contained in `A`, so `A` and `B` don't necessarily have the same size.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
youssef-abdallah: To make the implementation consistent with `RichIterable#containsNoneIterable`.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
donraab: @motlin `Iterate.allSatisfy()` can't be used in RichIterable as it is an implementation class.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
motlin: If A and B don't have the same size then they are definitely not equal.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -297,15 +297,23 @@ else if (inside.size() > 32 && !(inside instanceof SetIterable))
      *
      * @since 1.0
      */
-    boolean containsAllIterable(Iterable<?> source);
+    default boolean containsAllIterable(Iterable<?> source)
+    {
+        RichIterable<?> outside = this.toSet();
+        return StreamSupport.stream(source.spliterator(), false).allMatch(outside::contains);
+    }
```
motlin, line 304: If this is a set, then calling toSet() is unnecessary, and this implementation is slower than necessary. I think we should leave this abstract on RichIterable.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -297,15 +297,23 @@ else if (inside.size() > 32 && !(inside instanceof SetIterable))
...
      *
      * @see Collection#containsAll(Collection)
      * @since 1.0
      */
-    boolean containsAll(Collection<?> source);
+    default boolean containsAll(Collection<?> source)
+    {
+        return source instanceof RichIterable ? this.containsAllIterable(source)
+                : source.stream().allMatch(this::contains);
+    }
```
motlin, line 316: I don't understand the conditional here. What does it matter if `source` is a `RichIterable`? It's being passed to a method that takes `Iterable`.

In addition, containsAllIterable() may convert `this` to a set, but if the collection is a plain one, not a RichIterable, then we don't convert anything to a set. Why the difference? If it's for performance, this inconsistency doesn't make sense to me.

motlin (changes requested):
I'm skeptical that this is an optimization, and suspect it's slower in some cases. There are no JMH tests or other performance tests or commentary explaining why it's faster. The purpose of this change seems to be purely to optimize code since it is more verbose than the previous implementation. I don't think we ought to merge this until we can be confident it's an optimization.

PR #1363, opened 2022-06-14
Refactor to use internal array iterate - closes #1350

### LINKED ISSUE #1350: Move collectPrimitive code to InternalArrayIterate from FastList and ArrayIterate
`FastList` and `ArrayIterate` should have their implementations replaced with delegation to `InternalArrayIterate` for the following primitive `collect` methods.

```
collectBoolean
collectByte
collectChar
collectDouble
collectInt
collectFloat
collectLong
collectShort
```

Desislav-Petrov:


eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/InternalArrayIterate.java
```diff
@@ -1214,4 +1226,108 @@ public static <R extends Collection<T>, T> R rejectWithIndex(
         }
         return target;
     }
+
+    public static <R extends MutableBooleanCollection, T> R collectBoolean(
```
Desislav-Petrov, line 1230: I wonder if there's any point in passing the size explicitly - looks like we can deduct it from items.length directly? In the appropriate places we have a null check so we can reduce the param list?

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
```diff
@@ -803,11 +799,7 @@ public MutableByteList collectByte(ByteFunction<? super T> byteFunction)
     @Override
     public <R extends MutableByteCollection> R collectByte(ByteFunction<? super T> byteFunction, R target)
     {
-        for (int i = 0; i < this.size; i++)
-        {
-            target.add(byteFunction.byteValueOf(this.items[i]));
-        }
-        return target;
+        return InternalArrayIterate.collectByte(this.items, this.size, byteFunction, target);
```
Desislav-Petrov, line 802: I have left the flatCollectX untouched but it seems reasonable to push them in the internal array iterable as well - i know atm they are only used in one place but seems it will be more consistent for all of these to come from internal array iterate. Further, if in future we do need to reuse this functionality it would already be moved to a common util class. Any thoughts?

Desislav-Petrov:
hi @donraab  - left a few comments - keep to hear what you think

donraab:
> hi @donraab - left a few comments - keen to hear what you think

@Desislav-Petrov I noticed we don't have a test class for `InternalArrayIterate`. We should create one.

eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/FastList.java
Desislav-Petrov: Makes sense - thx

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/InternalArrayIterate.java
Desislav-Petrov: got u - my bad, thx both for the explanation

Desislav-Petrov:
@donraab - test added, thx for the comment

donraab (approved):
Hi @Desislav-Petrov, the code looks good to me. It would be great to add "since" tags on each of the new methods with 12.0 on InternalArrayIterate since they are new public API. You'll need to rebase as well.

prathasirisha:
> hi @donraab - changes made, believe the failing checks aren't related to my PR

Restarted the failed builds. More [details](https://youtrack.jetbrains.com/issue/TW-76846/TeamCity-Inspection-builds-timeouts) about why the builds were failing.

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/InternalArrayIterate.java
```diff
@@ -1214,4 +1226,156 @@ public static <R extends Collection<T>, T> R rejectWithIndex(
         }
         return target;
     }
+
+    /**
+     * Adds all array elements to the target MutableBooleanCollection that return false when evaluating the specified predicate which is
```
donraab, line 1231: @Desislav-Petrov The javadoc is incorrect. This looks like the description for reject, not collect.

donraab (changes requested):
Hi @Desislav-Petrov, the JavaDoc is incorrect for the collect methods. It looks like the description for reject, not collect.

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/InternalArrayIterate.java
Desislav-Petrov: Apologies, my bad - sorted

unit-tests/src/test/java/org/eclipse/collections/impl/utility/internal/InternalArrayIterateTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+public class InternalArrayIterateTest
+{
+    @Test
+    public void collectBoolean()
+    {
+        Integer[] items = {1, 2, 3};
+
+        MutableBooleanList result = InternalArrayIterate.collectBoolean(
+                items, 3, (BooleanFunction<Integer>) anObject -> anObject % 2 == 0, BooleanLists.mutable.withInitialCapacity(3));
```
donraab, line 48: Hi @Desislav-Petrov, I noticed you have casts for the functions in here. Are they necessary?

Desislav-Petrov: Good point - thx - not needed so removed in latest version

prathasirisha:
Re-ran the failing builds. Looks like one of them went to success and the rest are running now.

donraab (approved):
LGTM

PR #1367, opened 2022-06-24
Reflection fallback for serviceloader

mohrezaei:
This is an experimental patch. Don't merge immediately.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/ServiceLoaderUtils.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/ServiceLoaderUtils.java
@@ -13,12 +13,274 @@ package org.eclipse.collections.api.factory;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Proxy;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.ServiceLoader;
 import java.util.stream.Collectors;
 
 public final class ServiceLoaderUtils
 {
+    private static final Map<String, String> FACTORY_IMPL = new HashMap<>();
+
+    static
+    {
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.bag.ImmutableBagFactory", "org.eclipse.collections.impl.bag.immutable.ImmutableBagFactoryImpl");
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.bag.MultiReaderBagFactory", "org.eclipse.collections.impl.bag.mutable.MultiReaderMutableBagFactory");
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.bag.MutableBagFactory", "org.eclipse.collections.impl.bag.mutable.MutableBagFactoryImpl");
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.bag.primitive.ImmutableBooleanBagFactory", "org.eclipse.collections.impl.bag.immutable.primitive.ImmutableBooleanBagFactoryImpl");
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.bag.primitive.ImmutableByteBagFactory", "org.eclipse.collections.impl.bag.immutable.primitive.ImmutableByteBagFactoryImpl");
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.bag.primitive.ImmutableCharBagFactory", "org.eclipse.collections.impl.bag.immutable.primitive.ImmutableCharBagFactoryImpl");
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.bag.primitive.ImmutableDoubleBagFactory", "org.eclipse.collections.impl.bag.immutable.primitive.ImmutableDoubleBagFactoryImpl");
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.bag.primitive.ImmutableFloatBagFactory", "org.eclipse.collections.impl.bag.immutable.primitive.ImmutableFloatBagFactoryImpl");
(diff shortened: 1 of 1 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/factory/ServiceLoaderUtils.java
```diff
@@ -13,28 +13,315 @@
...
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.map.primitive.MutableBooleanCharMapFactory", "org.eclipse.collections.impl.map.mutable.primitive.MutableBooleanCharMapFactoryImpl");
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.map.primitive.MutableBooleanDoubleMapFactory", "org.eclipse.collections.impl.map.mutable.primitive.MutableBooleanDoubleMapFactoryImpl");
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.map.primitive.MutableBooleanFloatMapFactory", "org.eclipse.collections.impl.map.mutable.primitive.MutableBooleanFloatMapFactoryImpl");
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.map.primitive.MutableBooleanIntMapFactory", "org.eclipse.collections.impl.map.mutable.primitive.MutableBooleanIntMapFactoryImpl");
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.map.primitive.MutableBooleanLongMapFactory", "org.eclipse.collections.impl.map.mutable.primitive.MutableBooleanLongMapFactoryImpl");
+        FACTORY_IMPL.put("org.eclipse.collections.api.factory.map.primitive.MutableBooleanShortMapFactory", "org.eclipse.collections.impl.map.mutable.primitive.MutableBooleanShortMapFactoryImpl");
```
prathasirisha, line 160: I dont think we need these combinations. 

We are skipping code generation for these - Mutable/ImmutableBoolean(Byte|Char|Doube|Float|Int|Long|Short)MapFactory and Mutable/ImmutableBoolean(Byte|Char|Doube|Float|Int|Long|Short)MapFactoryImpl

donraab (approved):
Since the extra factories that are included are also showing up in the services directory, I think this is good to go. We can clear up the unused factories later, as they will never be instantiated anyways since there should be no actual corresponding factory classes.

prathasirisha:
> for the record, this is the code used to create the static stuff:
> 
> ```
> import java.io.File;
> import java.io.FileInputStream;
> import java.io.FileReader;
> import java.io.LineNumberReader;
> 
> public class PrintIt
> {
>     public static void main(String[] args) throws Exception
>     {
>         File dir = new File("/temp/p17");
>         File[] files = dir.listFiles();
>         for(File f: files)
>         {
>             LineNumberReader lineNumberReader = new LineNumberReader(new FileReader(f));
>             String impl = lineNumberReader.readLine();
>             System.out.println("FACTORY_IMPL.put(\""+f.getName()+"\", \""+impl+"\");");
>         }
>     }
> }
> ```
> 
> Just unzip the service directory somewhere (`/temp/p17` in the code above) and run this code.

Thank you for sharing this.

prathasirisha (approved):

prathasirisha:
Please rebase your MR @mohrezaei. 

Also waiting to hear from @motlin and @nikhilnanivadekar on these changes.

mohrezaei:
@prathasirisha If your tests pass, we can commit with the two approvals, per our usual policy. I don't think the code is extraordinary enough to not fall under normal policy.

motlin:
I was pinged for feedback, so I'll share my thoughts. I know this landed already and I think it's good. Just recording some more thoughts for later.

I looked through slf4j and logback code to see how they deal with this problem. It turns out, they don't! LoggerFactory calls `ServiceLoader.load(SLF4JServiceProvider.class)` and there's no fallback. `org.slf4j.LoggerFactory#bind` prints out errors if one or many implementations are found, which I think I used for inspiration at the time. And it's not like those errors are rare, people see them all the time. Anyway, lots of time had passed and I figured maybe slf4j added a fallback, but they have not.

The many calls to FACTORY_IMPL.put are necessary as long as this code lives in the api module. If it moved into the impl module, it could be type-safe. Then, the loadByReflection() implementation would change so that there's a single call to Class.forName(), and it would always be the FQCN of the ServiceLoaderUtils class.

```
public final class ServiceLoaderUtils
    implements Supplier<Map<Class<?>, Class<?>>>
{
    private static final Map<Class<?>, Class<?>> FACTORY_IMPL = new HashMap<>();

    static
    {
        FACTORY_IMPL.put(
                ImmutableBagFactory.class,
                ImmutableBagFactoryImpl.class);
        FACTORY_IMPL.put(
                MultiReaderBagFactory.class,
                MultiReaderMutableBagFactory.class);
        // ...
    }

    @Override
    public Map<Class<?>, Class<?>> get()
    {
        return FACTORY_IMPL;
    }
}
```

mohrezaei:
I was honestly surprised to see how inconsistent our api/impl packaging is. The loader can be cleaned up with a bit of consistency: `org.eclipse.api.foo.bar.GnatFactory` should have an implementation at `org.eclipse.impl.foo.bar.GnatFactoryImpl`. All the contortions with moving these things around makes more of a mess than trying to pick a super duper perfect package for the impl. This change can be done with backward compatibility: 
    - leave the existing factory where it is.
    - mark it deprecated
    - remove all content and extend the factory in the proper package.

Then the loader code turns into: 
```java
String implClass = apiClassName.replace(".api.", ".impl.")+"Impl";
```

nikhilnanivadekar:
I would caution against depreciating and moving stuff. Traditionally just like the JDK, we tend to deprecate but never delete. If 12.0.0 is planned to be a massively breaking release then I am all up for it. We should then at least plan to remove it for 13.0.0 or 14.0.0.

mohrezaei:
Actually, I like deprecating and not removing (for a very long time). An empty class never hurt anybody.

PR #1382, opened 2022-08-24
fix : rename parameter

ibrahim-5:
rename the 'list' parameter in org.eclipse.collections.impl.factory.Sets.adapt to 'set'


donraab (changes requested):
Hi @ibrahim-5, the change looks good. Thank you for contributing. One small request to update your commit message to include closes with the issue number. We use these when creating the release notes. Thanks!

donraab (approved):
LGTM! Thank you for the contribution @ibrahim-5!

Issue #1396, opened 2022-09-26
Expose compact in all primitive maps


victornoel:
It seems some primitive map implementation (e.g., `ObjectIntHashMap`) have a compact method while some others (e.g., `ObjectBooleanHashMap`) do not.

It would be great to have it implemented everywhere.

victornoel:
Closing as this should be available in 12.0.0.M1

PR #1397, opened 2022-09-26
[11.2] Expose compact in ObjectBooleanHashMap and ObjectBooleanHashMapWithHashingStrategy

victornoel:
For https://github.com/eclipse/eclipse-collections/issues/1396.

I didn't find any other case but if you see some other places I could do this, please share :)

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/primitive/ObjectBooleanHashMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/primitive/ObjectBooleanHashMap.java
@@ -1109,6 +1109,14 @@ public class ObjectBooleanHashMap<K> implements MutableObjectBooleanMap<K>, Exte
         return Math.min(capacity - 1, capacity / OCCUPIED_DATA_RATIO);
     }
 
+    /**
+     * Rehashes every element in the set into a new backing table of the smallest possible size and eliminating removed sentinels.
+     */
+    public void compact()
+    {
+        this.rehash(this.smallestPowerOfTwoGreaterThan(this.size()));
+    }
+
     private void rehashAndGrow()
     {
         this.rehash(this.keys.length << 1);
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/primitive/ObjectBooleanHashMapWithHashingStrategy.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/primitive/ObjectBooleanHashMapWithHashingStrategy.java
@@ -1132,6 +1132,14 @@ public class ObjectBooleanHashMapWithHashingStrategy<K> implements MutableObject
         return Math.min(capacity - 1, capacity / OCCUPIED_DATA_RATIO);
     }
 
+    /**
+     * Rehashes every element in the set into a new backing table of the smallest possible size and eliminating removed sentinels.
+     */
+    public void compact()
+    {
+        this.rehash(this.smallestPowerOfTwoGreaterThan(this.size()));
+    }
+
     private void rehashAndGrow()
     {
         this.rehash(this.keys.length << 1);


eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/primitive/ObjectBooleanHashMap.java
```diff
@@ -1109,6 +1109,14 @@ private int maxOccupiedWithData()
         return Math.min(capacity - 1, capacity / OCCUPIED_DATA_RATIO);
     }
 
+    /**
+     * Rehashes every element in the set into a new backing table of the smallest possible size and eliminating removed sentinels.
+     */
+    public void compact()
```
motlin, line 1115: I'd suggest the name `trimToSize()` for consistency with UnifiedMap and ArrayList.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/primitive/ObjectBooleanHashMap.java
```diff
@@ -1109,6 +1109,14 @@ private int maxOccupiedWithData()
...
+    {
+        this.rehash(this.smallestPowerOfTwoGreaterThan(this.size()));
```
motlin, line 1117: There ought to be a check that the new capacity is different than the old capacity. Otherwise, we spend O(n) time and space for no reason.

motlin: I'm just realizing that your changes are consistent with the code in primitive-primitive maps. I guess we can land this but we ought to fix these problems later.

victornoel:
@motlin thx for the feedbacks, I think it would be a good idea to take this opportunity to solve:
> There ought to be a check that the new capacity is different than the old capacity. Otherwise, we spend O(n) time and space for no reason.

As for the naming of the method, I don't think I will remove `compact`, but maybe should we add `trimToSize` and optionally mark `compact` as deprecated? WDYT?

victornoel:
or maybe `trimToSize` should do the check, while `compact` should not? I guess there could be good reason to call `compact` anyway, for example to "eliminate removed sentinels" as specified in the javadoc of the `compact` method?

motlin (approved):

nikhilnanivadekar:
Since it is a new method, I would rename it to `trimToSize` to make it consistent. We should deprecate the other `compact` methods and add `trimToSize` . That way we will have symmetry between object and primitive side.

victornoel:
@motlin @nikhilnanivadekar so we merge this as-is, and then we introduce `trimToSize` in another PR (I can do it) that conditionally do what it's in `compact` as well as deprecates `compact` but keep it implemented as it is now. Is that correct?

nikhilnanivadekar:
Yup! I vote for that. Thoughts @motlin ?

victornoel:
another question: any change to get this in 11.2 or is 12.0 close to release?

nikhilnanivadekar:
12.0 is no where close. But if you can get this, the deprecation sorted out
I can cut the 11.2. quickly . I can work on optimizing it with size check
on the side.
How does that sound?

victornoel:
I've just pushed: https://github.com/eclipse/eclipse-collections/pull/1403 :)

if both could go to 11.2 it would be really cool! I am not in a rush, but it would be nice to have it in a few weeks for example (or more if you need to).

nikhilnanivadekar:
Releasing a minor version is quick. So I can trigger it sometime in the next couple of weeks. Let’s close out as many PRs as we can.
I’ll wait to see if @donraab , @motlin have anything they need for 11.2 in their pipeline

nikhilnanivadekar:
Is this ready to go?

victornoel:
@nikhilnanivadekar from my point of view it is :)

nikhilnanivadekar:
Sorry I had not followed your question correctly. I would say that this needs to be named as `trimToSize()`. We should not introduce a new method only to deprecate it.

Thoughts @motlin

victornoel:
@nikhilnanivadekar trimToSize is introduced in a separate PR: https://github.com/eclipse/eclipse-collections/pull/1403

victornoel:
oh right ,sorry, I misread what you said, I will just merge the two PR and not introduce compact here!

victornoel:
Closing this, superseded by https://github.com/eclipse/eclipse-collections/pull/1403

Issue #500, opened 2018-04-08
Override Java 8 default methods for correctness and efficiency.


motlin:
There are several default methods added to Java 8 collections interfaces that are not overridden in Eclipse Collections. Default methods are often composites that delegate to several other methods. For example, Map.compute() delegates to get(), containsKey(), put(), and remove(). For synchronized/concurrent collections, it may be necessary to override for correctness. For others, it can be more efficient to override (to only call hashcode once, for example).

```
java.util.Collection#removeIf
java.util.List#replaceAll
java.util.List#sort
java.util.Map#getOrDefault
java.util.Map#forEach
java.util.Map#replaceAll
java.util.Map#computeIfAbsent
java.util.Map#computeIfPresent
java.util.Map#compute
java.util.Map#merge
```

List.sort() is especially surprising. We did all this work to rename `sort` to `sortThis` to not interfere. Now we're in a state where FastList.asSynchronized().sort() is not atomic.

donraab:
I am working on List#sort and List#replaceAll.

ylexus:
Can this be revived? `UnifiedMap.computeIfAbsent()`, for example, looks like a simple override and delegation to `getIfAbsentPutWithKey`

donraab:
Thank @ylexus , yes this can be worked on now as next release is a major version.

PR #1401, opened 2022-09-28
Adding default override for Map.computeIfPresent - closes #1400

### LINKED ISSUE #1400: Override Java 8 default method Map#computeIfPresent for MutableMaps
It seems like issue #500 has not been resolved yet. 

The default `Map#computeIfPresent` has not been overridden and delegated to other Eclipse Collection methods. I wanted to get started on working on this issue.

rzrobin213:
For #500 and #1400. 

From my tests, the override has the same functionality as the default `Map#computeIfPresent` implementation. Let me know if there's any concerns with it!

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ConcurrentMutableMap.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ConcurrentMutableMap.java
@@ -12,6 +12,7 @@ package org.eclipse.collections.api.map;
 
 import java.util.Map;
 import java.util.concurrent.ConcurrentMap;
+import java.util.function.BiFunction;
 
 import org.eclipse.collections.api.block.procedure.Procedure;
 
@@ -42,5 +43,12 @@ public interface ConcurrentMutableMap<K, V>
     {
         this.putAllMapIterable(mapIterable);
         return this;
+    }
+
+    @Override
+    default V computeIfPresent(K key,
+            BiFunction<? super K,? super V,? extends V> remappingFunction)
+    {
+        return ConcurrentMap.super.computeIfPresent(key, remappingFunction);
     }
 }
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
@@ -13,6 +13,7 @@ package org.eclipse.collections.api.map;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.BiFunction;
 
 import org.eclipse.collections.api.bag.MutableBag;
 import org.eclipse.collections.api.block.function.Function;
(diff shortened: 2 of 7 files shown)


unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MutableMapTestCase.java
```diff
@@ -53,4 +61,36 @@ default void Iterable_remove()
                 this.getExpectedFiltered(3, 3, 3, 2, 1),
                 this.getExpectedFiltered(3, 3, 2, 2, 1)));
     }
+
+    @Test
+    default void Iterable_computeIfPresent()
```
motlin, line 66: This test should move up to MutableMapIterableTestCase, and be named Map_computeIfPresent()

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MutableMapTestCase.java
```diff
@@ -53,4 +61,36 @@ default void Iterable_remove()
...
+    {
+        MutableMap<Integer, Integer> map = this.newWithKeysValues(1, 1, 2, 2, 3, 3);
+        Map<Integer, Integer> hashMap = new LinkedHashMap<>();
+        hashMap.put(1, 1);
+        hashMap.put(2, 2);
+        hashMap.put(3, 3);
+        if (map instanceof UnmodifiableMutableMap)
```
motlin, line 73: Instead of an instanceof, we can use inheritance. This ought to be similar to how MutableMapIterableTestCase#Map_clear is overridden in UnmodifiableMutableMapTest#Map_clear

motlin (commented):
I think this is an improvement but I admit I was expecting an entirely different implementation. Basically, I'd expect that the classes that have an implementation of MutableMapIterable#getIfAbsentPut would also have similar implementations for getOrDefault, putIfAbsent, computeIfAbsent, computeIfPresent, compute, and merge. The problem with all of those default methods is that they are not atomic and they compute the hashcode several times by delegating to more than one of get, containsKey, put, and remove. The new default methods in this review are likely atomic, but they continue to compute the hashcode twice.

rzrobin213:
Thank you for the feedback! I'll take a look at the classes that have `MutableMapIterable#getIfAbsentPut` implemented and consider overrides for `computeIfPresent` that only call hashcode once.

mohrezaei:
The concurrent map implementation is seriously broken. 
- There is an infinite loop when `o == RESIZED`. 
- Observing `o != null` is not sufficient to decide we have to compute, because `o` can be a variety of things (other keys, or special sentinels). 

Testing this implementation is highly non-trivial.

The JDK's implementation (`java.util.ConcurrentHashMap`) locks for this function, so two threads calling with the same key won't both compute. However, that's not the contract specified in the interface (`java.util.ConcurrentMap`): 

> The default implementation makes no guarantees about synchronization or atomicity properties of this method. Any implementation providing atomicity guarantees must override this method and document its concurrency properties. In particular, all implementations of subinterface java.util.concurrent.ConcurrentMap must document whether the function is applied once atomically only if the value is not present.

I can't think of any sort of atomicity guarantee we could implement here. Fixing the issues above and testing this is a lot of work, requiring a deep understanding of both how this map is implemented and how the concurrency semantics work in Java. Given all that, calling hashcode twice seems like a more practical solution.

PR #1406, opened 2022-10-04
Optimize any/all/noneSatisfy on UnifiedMapWithHashingStrategy - closes #1342

### LINKED ISSUE #1342: Optimize any/all/NoneSatisfy on UnifiedMapWithHashingStrategy
The methods `anySatisfy`, `allSatisfy` and `noneSatisfy` should be optimized on `UnifiedMapWithHashingStrategy` similar to how they are optimized on `UnifiedMap`.

Desislav-Petrov:
PR  - would appreciate a review @donraab 

Looks like existing tests already cover this refactoring

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/UnifiedMapWithHashingStrategy.java
@@ -28,6 +28,8 @@ import org.eclipse.collections.api.block.HashingStrategy;
 import org.eclipse.collections.api.block.function.Function;
 import org.eclipse.collections.api.block.function.Function0;
 import org.eclipse.collections.api.block.function.Function2;
+import org.eclipse.collections.api.block.predicate.Predicate;
+import org.eclipse.collections.api.block.predicate.Predicate2;
 import org.eclipse.collections.api.block.procedure.Procedure;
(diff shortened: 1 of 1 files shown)


donraab (changes requested):
Thank you for the contribution @Desislav-Petrov. I think it would be better to duplicate the functionality in the shortCircuit/shortCircuitWith methods than to externalize the functionality to MapIterate. MapIterate is intended to iterate over Maps, and this iterates over an array, requiring MapIterate to understand how the internals of UnifiedMap and UnifiedMapWithHashingStrategy are internally implemented.

Desislav-Petrov:
Thanks for the review @donraab - fair points, made the changes & rebased

motlin (changes requested):
To review this, I checked it out locally and diff'd UnifiedMap with UnifiedMapWithHashingStrategy. I found that the new methods were added in a different location, so diff'ing didn't work. I made the fix so that I could continue the review, but you'll have to apply it as well.

https://github.com/eclipse/eclipse-collections/commit/d4018e6e47b8d680d7a97e48267eea16dd317417

Once I did this I could see that the new implementation and the old are identical, and are good to go. Please move the methods to the same positions and then we ought to merge this.

Desislav-Petrov:
Thx for the feedback @motlin - changes made

motlin:
@Desislav-Petrov could you also rebase onto upstream/master and then I'll merge?

Desislav-Petrov:
Done @motlin

Issue #1418, opened 2022-12-29
Tricky NullPointerException in IntObjectMap.forEachKeyValue


magicprinc:
If you remove key 0 (int) in forEachKeyValue - use will enjoy NullPointerException 

`@Test public void testBugRemoveNPE (){
    try {
      var m = IntObjectMaps.mutable.of(0, 'a');
      m.forEachKeyValue((k,v)->m.removeKey(k));
      fail();
    } catch (NullPointerException e){
      //assertEquals("Cannot read field \"containsOneKey\" because \"this.sentinelValues\" is null", e.getMessage());
    }
  }`

Problem is here:
`org.eclipse.collections.impl.map.mutable.primitive.IntObjectHashMap#forEachKeyValue`

`if (this.sentinelValues.containsOneKey)`

after removing key 0 aka EMPTY_KEY - sentinelValues can be null

The problem is easy to fix:

`

if (this.sentinelValues != null && this.sentinelValues.containsZeroKey)
{
  procedure.value(EMPTY_KEY, this.sentinelValues.zeroValue);
}
if (this.sentinelValues != null && this.sentinelValues.containsOneKey)
{
  procedure.value(REMOVED_KEY, this.sentinelValues.oneValue);
}

`

mohrezaei:
While the above is true, I don't believe we support map modifications within `forEachKeyValue`. `remove` is "safe" by accident (and not safe in the case of the zero key). `put` would cause all manner of problems.

Perhaps that needs to be better documented.

Additionally, we could add `removeIf` to primitive maps (I'm assuming this is how you ran into this, as the code sample to reproduce this is not likely your real code).

magicprinc:
It is absolutely safe (even if by accident). Only this case with key==0 is a problem.

I didn't want to call remove in forEachKeyValue in the first place, BUT

keyValuesView iterator doesn't support remove (throws exception)

mohrezaei:
The above code fits well within a `removeIf` implementation.

We have two choices here:
1) Document `forEachKeyValue` as not allowing modifications. This allows us to change the map implementation internals much easier. For example, if we implement auto-shrinking, `remove` will no longer be safe, even by accident. Additionally, add `removeIf`. Maybe implement `keyValuesView.iterator().remove()`.
2) Document `forEachKeyValue` as allowing `remove` (but not `put`!) and test all scenarios with this and fix various issues.

Relying on an accident, instead of an intent, is neither good for your code nor for the library.

[...]

PR #1419, opened 2023-01-01
add removeIf to primitiveObject maps

mohrezaei:
See #1418. This is just for primitiveObject maps. Implementations for objectPrimitive and primitivePrimitive are still needed.


donraab (approved):
LGTM

PR #1425, opened 2023-01-09
Override Java 8 default method Map#forEach - closes #1423

### LINKED ISSUE #1423: Override Java 8 default method Map#forEach
As referenced in #500, the default Map implementation of the forEach method is not overridden in the EC Map implementations. 

This can be delegated to the `MapIterable#forEachKeyValue` which would give a performance benefit as there is no call to `entrySet` and there is no need to generate `Map.Entry` objects. This is referenced in this StackOverflow [question](https://stackoverflow.com/questions/46898/how-do-i-efficiently-iterate-over-each-entry-in-a-java-map/13943066#13943066).

I would like to work on this issue if possible.

rzrobin213:
Signed-off-by: Robin Zheng <robin.zheng@bnymellon.com>

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ConcurrentMutableMap.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ConcurrentMutableMap.java
@@ -12,6 +12,7 @@ package org.eclipse.collections.api.map;
 
 import java.util.Map;
 import java.util.concurrent.ConcurrentMap;
+import java.util.function.BiConsumer;
 
 import org.eclipse.collections.api.block.procedure.Procedure;
 
@@ -42,5 +43,11 @@ public interface ConcurrentMutableMap<K, V>
     {
         this.putAllMapIterable(mapIterable);
         return this;
+    }
+
+    @Override
+    default void forEach(BiConsumer<? super K, ? super V> action)
+    {
+        MutableMap.super.forEach(action);
     }
 }
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
@@ -13,6 +13,7 @@ package org.eclipse.collections.api.map;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.BiConsumer;
 
 import org.eclipse.collections.api.bag.MutableBag;
 import org.eclipse.collections.api.block.function.Function;
@@ -432,5 +433,11 @@ public interface MutableMapIterable<K, V> extends MapIterable<K, V>, Map<K, V>
                 nonMutatingAggregator,
                 valueFunction.valueOf(value)));
         return map;
+    }
+
+    @Override
+    default void forEach(BiConsumer<? super K, ? super V> action)
+    {
+        MapIterable.super.forEach(action);
     }
 }
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MapIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MapIterable.java
@@ -11,9 +11,11 @@
 package org.eclipse.collections.api.map;
 
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Spliterator;
 import java.util.Spliterators;
+import java.util.function.BiConsumer;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
@@ -348,5 +350,11 @@ public interface MapIterable<K, V> extends RichIterable<V>
                 nonMutatingAggregator,
                 valueFunction.valueOf(value)));
         return map;
+    }
+
+    default void forEach(BiConsumer<? super K, ? super V> action)
+    {
+        Objects.requireNonNull(action);
+        this.forEachKeyValue(action::accept);
     }
 }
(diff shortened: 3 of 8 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
```diff
@@ -433,4 +434,11 @@ default <K1, V1, V2> MutableMapIterable<K1, V2> aggregateBy(
                 valueFunction.valueOf(value)));
         return map;
     }
+
+    @Override
+    default void forEach(BiConsumer<? super K, ? super V> action)
+    {
+        Objects.requireNonNull(action);
+        this.forEachKeyValue(action::accept);
+    }
```
motlin, line 443: In my opinion, this implementation should be pulled up to MapIterable. So the version here would be:

```java
    @Override
    default void forEach(BiConsumer<? super K, ? super V> action)
    {
        MapIterable.super.forEach(action);
    }
```

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ConcurrentMutableMap.java
```diff
@@ -43,4 +44,10 @@ default ConcurrentMutableMap<K, V> withMapIterable(MapIterable<? extends K, ? ex
         this.putAllMapIterable(mapIterable);
         return this;
     }
+
+    @Override
+    default void forEach(BiConsumer<? super K, ? super V> action)
+    {
+        this.forEachKeyValue(action::accept);
+    }
```
motlin, line 52: Does this really need an override? Wouldn't the version inherited from MapIterable or MutableMapIterable be fine?

If this is an inheritance problem, then it should refer to a superclass.

```java
    @Override
    default void forEach(BiConsumer<? super K, ? super V> action)
    {
        MutableMap.super.forEach(action);
    }
```

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MutableMapIterableTestCase.java
```diff
@@ -277,4 +279,18 @@ default void MutableMapIterable_updateValue()
...
+        MutableMapIterable<Object, Integer> map = this.newWith(3, 3, 3, 2, 2, 1);
+        MutableCollection<Integer> forEach = this.newMutableForFilter();
+        map.forEach((key, value) -> forEach.add(value + 10));
+        assertEquals(this.newMutableForFilter(13, 13, 13, 12, 12, 11), forEach);
+
+        MutableMapIterable<Integer, String> map2 = this.newWithKeysValues(3, "Three", 2, "Two", 1, "Three");
+        MutableCollection<String> forEach2 = this.newMutableForFilter();
+        map2.forEach((key, value) -> forEach2.add(key + value));
+        assertEquals(this.newMutableForFilter("3Three", "2Two", "1Three"), forEach2);
+    }
```
motlin, line 295: The prefix matches the class. Putting together the other feedback, this belongs in org.eclipse.collections.test.map.MapIterableTestCase and should be called MapIterable_forEach. We ought to have this in the class MapTestCase, but it looks like we don't have that class yet! Even though we do have org.eclipse.collections.test.list.ListTestCase.

motlin (changes requested):

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java
donraab: This could be done in a separate PR. `MapIterable` does not extend `Map`. By putting this as a default method on `MapIterable`, it might create a challenge similar to what we faced with [`isEmpty` default methods on `CharSequence`](https://stuartmarks.wordpress.com/2020/09/22/incompatibilities-with-jdk-15-charsequence-isempty/). We understand this particular problem now, so we can fix anything that arises, but I think it would be better to do this in small steps. This PR is meant to fix the performance issue currently faced with `Map` `forEach` delegating to `entrySet` in the default implementation.

motlin: Won't this create an entrySet()?

`Maps.immutable.empty().castToMap().forEach((k, v) -> {})`

motlin: I stepped through in a debugger, it creates an empty UnifiedSet for the entrySet.

donraab: Yes it will. Good point. Ok, I agree then it should be moved up to `MapIterable`.

rzrobin213: Thanks for the feedback @motlin and @donraab!

I was thinking about moving it up to `MapIterable` but had the same concerns as @donraab about it not being a subclass of the Map class. I agree that it does make more sense that `forEach` would have the same issues for `ImmutableMap`. I will work on moving it up.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/ConcurrentMutableMap.java
rzrobin213: Yup, it is an inheritance issue. Thanks!

donraab:
@rzrobin213 I approved the builds to run. Looks like you have a checkstyle failuire due to unused imports.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MutableMapIterableTestCase.java
rzrobin213: For another PR, would it be better to refactor the test cases to have a `MapTestCase` in the style of `ListTestCase`?

motlin: Yes, it would be great to have MapTestCase.

motlin (approved):

PR #1426, opened 2023-01-12
Implement Comparators.fromPredicate(), to create a Comparator from a Predicate2 that can answer isBefore.

motlin:
I have a use-case for sorting where I can answer whether one object should come before another object, but I don't have a Comparator. I looked around for a function that convert a BiPredicate to a Comparator and couldn't find one, so here it is. I looked on Comparator, Comparators, and on BiPredicate.

The unit test sorts Instants using the isBefore() method. This is a good example of using a single-arg, non-static method.

```java
ImmutableList<Instant> orderedInstants = Lists.immutable.with(
        Instant.parse("2000-01-01T00:00:00Z"),
        Instant.parse("2001-01-01T00:00:00Z"),
        Instant.parse("2002-01-01T00:00:00Z"),
        Instant.parse("2003-01-01T00:00:00Z"));
MutableList<Instant> shuffledInstants = orderedInstants.toList().shuffleThis();
shuffledInstants.sortThis(Comparators.fromPredicate(Instant::isBefore));
```

In my real use-case, I have a multimap of dependencies. If a key/value appears in the multimap, then the key is a dependent of the value and should come before. This is a good example of using a two-arg method.

Anonymized, but realistic example:

```java
MutableSetMultimap<MyObject, MyObject> transitiveDependencies = getTransitiveDependencies(directDependencies);

Comparator<MyObject> dependencyComparator = fromPredicate(transitiveDependencies::containsKeyAndValue);

Comparator<MyContainingObject> comparator = Comparator
        .comparing(MyContainingObject::getDependencyKey, dependencyComparator)
        .thenComparing(MyContainingObject::getName);

MutableList<MyContainingObject> sortedObjects = SetAdapter
        .adapt(objects)
        .toSortedList(comparator);
```

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/block/factory/Comparators.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/block/factory/Comparators.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Goldman Sachs and others.
+ * Copyright (c) 2023 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -12,6 +12,7 @@ package org.eclipse.collections.impl.block.factory;
 
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.function.BiPredicate;
 
 import org.eclipse.collections.api.block.SerializableComparator;
 import org.eclipse.collections.api.block.factory.SerializableComparators;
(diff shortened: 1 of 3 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/block/factory/Comparators.java
```diff
@@ -553,4 +563,30 @@ public int compare(T one, T two)
...
+            {
+                return -1;
+            }
+            if (this.isBeforePredicate.test(o2, o1))
+            {
+                return 1;
+            }
+            return 0;
+        }
+    }
```
motlin, line 591: Maybe I should called it isLessThan instead of isBefore?

motlin:
@donraab rebased

eclipse-collections/src/main/java/org/eclipse/collections/impl/block/factory/Comparators.java
donraab: @motlin isLessThanPredicate might make more sense here. I usually think of isBefore in terms of time.

donraab (commented):
One comment responding to your question.

donraab (approved):
LGTM.

PR #1438, opened 2023-02-16
Update EA versions to 19, 20, 21

aqsa505:
Signed-off-by: Aqsa Malik [aqsa.malik@bnymellon.com](mailto:aqsa.malik@bnymellon.com)

### DIFF
--- a/eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/ClassComparer.java
+++ b/eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/ClassComparer.java
@@ -12,6 +12,8 @@ package org.eclipse.collections.impl.test;
 
 import java.io.IOException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
 
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.factory.SortedSets;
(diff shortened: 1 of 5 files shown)


donraab:
Hi @aqsa505, the failures of the tests here are real. Two new static methods were added to `HashSet` and `HashMap` in JDK 19. We may want to either update the tests to exclude these methods, or change ClassComparer to have a new methods that excludes static methods or something.

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/ClassComparer.java
```diff
@@ -179,6 +192,12 @@ public MutableSortedSet<String> getMethodNames(Class<?> classOne)
         return ArrayIterate.collectIf(classOne.getMethods(), this::includeMethod, this::methodName, SortedSets.mutable.empty());
     }
 
+    public MutableSortedSet<String> getInstanceMethodNames(Class<?> classOne)
+    {
+        Method[] staticMethods = Arrays.stream(classOne.getMethods()).filter(method -> Modifier.isStatic(method.getModifiers())).toArray(Method[]::new);
```
donraab, line 197: I think should either use reject from EC or use filter( !Modifier.isStatic). It would then make more sense for the variable to be named instanceMethods instead of staticMethods.

donraab (changes requested):
I think the filter needs to be reversed.

donraab (approved):
LGTM! Thanks for the contribution @aqsa505!

Issue #1455, opened 2023-04-16
Create primitiveStream methods for IntBag, LongBag and DoubleBag


donraab:
Since 10.0, we have had primitiveStream on `IntList`, `LongList` and `DoubleList`. This makes the following possible:

```
List<String> list = IntLists.immutable.of(1, 2, 3)
        .primitiveStream()
        .mapToObj(Integer::toString)
        .toList();

Assertions.assertEquals(List.of("1", "2", "3"), list);
```
There is no equivalent for `IntBag`, `LongBag`, and `DoubleBag` today. This will additionally require creating spliterator() for each of these.

Desislav-Petrov:
hi @donraab - happy to take a look at this

donraab:
Thanks for volunteering @Desislav-Petrov! I have assigned the issue to you.

Tushar1619:
Is this issue still open @donraab ?

ahsanilya-s:
hey @donraab assign mehappy to take a look at this.

PR #1459, opened 2023-05-12
Create primitiveStream methods for Int/Long/DoubleBag - #1455

Desislav-Petrov:


Desislav-Petrov:
hi @donraab - happy to hear any feedback, the JDK21 fail is unrelated to the commit from what I see

donraab:
Hi @Desislav-Petrov, yes the JDK 21 failure looks to be a problem with the new SequencedCollections JEP - https://openjdk.org/jeps/431 that has been pushed to the JDK 21 EA build.

eclipse-collections-code-generator/src/main/resources/impl/bag/immutable/immutablePrimitiveHashBag.stg
```diff
@@ -472,6 +476,14 @@ final class Immutable<name>HashBag implements Immutable<name>Bag, Serializable
             return this.bag.toImmutable();
         }
     }
+    <if(primitive.specializedStream)>
+
+    @Override
+    public Spliterator.Of<name> spliterator()
+    {
+        return Spliterators.spliterator(this.delegate.toArray(), Spliterator.IMMUTABLE);
```
donraab, line 484: This is a very costly implementation of a spliterator. This should not create a copy of the primitive HashBag into an array. It can delegate to the spliterator of the primitive HashBag.

eclipse-collections-code-generator/src/main/resources/impl/bag/mutable/primitiveHashBag.stg
```diff
@@ -881,6 +885,14 @@ public class <name>HashBag
             this.canRemove = false;
         }
     }
+<if(primitive.specializedStream)>
+
+    @Override
+    public Spliterator.Of<name> spliterator()
+    {
+        return Spliterators.spliterator(this.items.keySet().toArray(), Spliterator.SIZED);
```
donraab, line 893: This is a very costly implementation of a spliterator and I also believe it is incorrect for a Bag. By using only the keySet, items in the bag will be missed, as the counts could be more than 1.

donraab (changes requested):
Hi @Desislav-Petrov, the implementations of spliterator are too costly by requiring copies to be made with toArray. In one instance to call to toArray is done on the keySet, which would leave out elements in the bag if they have more than one count. Please come up with an approach for the spliterator implementations to do not require creating a copy of the elements. Thanks!

eclipse-collections-code-generator/src/main/resources/impl/bag/immutable/immutablePrimitiveHashBag.stg
Desislav-Petrov: Smart - makes sense, thx

eclipse-collections-code-generator/src/main/resources/impl/bag/mutable/primitiveHashBag.stg
donraab: Without custom spliterators, this PR will not be worthwhile. It would be providing an API that would never be desirable to use because of the performance cost.

Desislav-Petrov: hi @donraab  - i've been scratching my head of how this spliterator will look like and here are some thoughts/observations:

1) In all the implementations i've come across you rely on some type of an array - the HashMap spliterator relies on access to the Nodes[]. The primitive hashmaps (intint for example) that back the bag have keyValues[] structure that provides the storage backing but there's no easy 
2) Java Hashmap also comes with KeySpliterator and ValueSpliterator but I don't think these by themselves will be useful to us in this case because we care about both keys + vals. I was thinking we can possibly get away with doing the spliterator on the key but that's not entirely accurate because i might have 100s of elements in the bag that represent the same thing (100s of 1s for example).
3) I'm wondering if we can do the spliterator in the primitive hashmap, then from the hashbag we will delegate to it?

Happy to explore any other venues you might have in mind too.

donraab: Hi @Desislav-Petrov, I think maybe we just close this PR for now and wait for someone to request this feature for an actual use case. I think the asymmetry for primitive bags having primitive Stream support will be ok for now.

PR #1461, opened 2023-06-12
Add default overrides for getFirst and getLast in MutableList and Mut…

donraab:
…ableSortedSet to fix JDK 21 EA build failure

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Goldman Sachs and others.
+ * Copyright (c) 2023 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -60,6 +60,26 @@ import org.eclipse.collections.api.tuple.Pair;
 public interface MutableList<T>
         extends MutableCollection<T>, List<T>, Cloneable, ListIterable<T>
 {
+    /**
+     * This default override exists because java.util.List added a default getFirst() method in Java 21.
+     * @since 12.0
+     */
+    @Override
+    default T getFirst()
+    {
+        return this.isEmpty() ? null : this.get(0);
+    }
+
+    /**
+     * This default override exists because java.util.List added a default getLast() method in Java 21.
+     * @since 12.0
+     */
+    @Override
+    default T getLast()
+    {
+        return this.isEmpty() ? null : this.get(this.size() - 1);
+    }
+
     @Override
     default MutableList<T> with(T element)
     {
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/set/sorted/MutableSortedSet.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/set/sorted/MutableSortedSet.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Goldman Sachs and others.
+ * Copyright (c) 2023 Goldman Sachs and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
@@ -50,6 +50,28 @@ import org.eclipse.collections.api.tuple.Pair;
 public interface MutableSortedSet<T>
         extends MutableSetIterable<T>, SortedSetIterable<T>, SortedSet<T>, Cloneable
 {
+    /**
+     * This default override exists because java.util.SortedSet added a default getFirst() method in Java 21.
+     *
+     * @since 12.0
+     */
+    @Override
+    default T getFirst()
+    {
+        return this.first();
+    }
+
+    /**
+     * This default override exists because java.util.SortedSet added a default getLast() method in Java 21.
+     *
+     * @since 12.0
+     */
+    @Override
+    default T getLast()
+    {
+        return this.last();
+    }
+
     @Override
     default MutableSortedSet<T> with(T element)
     {
@@ -160,7 +182,9 @@ public interface MutableSortedSet<T>
      * @since 9.2
      */
     @Override
-    default <P, V> MutableList<V> flatCollectWith(Function2<? super T, ? super P, ? extends Iterable<V>> function, P parameter)
+    default <P, V> MutableList<V> flatCollectWith(
+            Function2<? super T, ? super P, ? extends Iterable<V>> function,
+            P parameter)
     {
         return this.flatCollect(each -> function.apply(each, parameter));
     }
--- a/serialization-tests/src/test/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMapSerializationTest.java
+++ b/serialization-tests/src/test/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMapSerializationTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Two Sigma.
+ * Copyright (c) 2023 Two Sigma and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * and Eclipse Distribution License v. 1.0 which accompany this distribution.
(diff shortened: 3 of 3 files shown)


serialization-tests/src/test/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMapSerializationTest.java
```diff
@@ -21,16 +21,31 @@
...
+                            + "ABAAAAAAeAA=",
+                    UnmodifiableMutableOrderedMap.of(OrderedMaps.adapt(new LinkedHashMap<>())));
+        }
+    }
+
+    private boolean isBeforeJava21()
+    {
+        String javaVersion = System.getProperty("java.version").replace("-ea", "");
+        String[] javaVersionParts = javaVersion.split("\\.");
+        if ("1".equals(javaVersionParts[0]))
```
Bananeweizen, line 44: the if is superfluous. the same will happen with the parseInt code.

donraab: Hi @Bananeweizen , yes, the if is superfluous here, and I was debating on changing it. Something is bothering me about changing it, even though it would work.

serialization-tests/src/test/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMapSerializationTest.java
```diff
@@ -21,16 +21,31 @@
...
+                    "rO0ABXNyAE5vcmcuZWNsaXBzZS5jb2xsZWN0aW9ucy5pbXBsLm1hcC5vcmRlcmVkLm11dGFibGUu\n"
+                            + "VW5tb2RpZmlhYmxlTXV0YWJsZU9yZGVyZWRNYXAAAAAAAAAAAQIAAUwACGRlbGVnYXRldAAzTG9y\n"
+                            + "Zy9lY2xpcHNlL2NvbGxlY3Rpb25zL2FwaS9tYXAvTXV0YWJsZU9yZGVyZWRNYXA7eHBzcgBCb3Jn\n"
+                            + "LmVjbGlwc2UuY29sbGVjdGlvbnMuaW1wbC5tYXAub3JkZXJlZC5tdXRhYmxlLk9yZGVyZWRNYXBB\n"
+                            + "ZGFwdGVyAAAAAAAAAAECAAFMAAhkZWxlZ2F0ZXQAD0xqYXZhL3V0aWwvTWFwO3hwc3IAF2phdmEu\n"
+                            + "dXRpbC5MaW5rZWRIYXNoTWFwNMBOXBBswPsCAAFaAAthY2Nlc3NPcmRlcnhyABFqYXZhLnV0aWwu\n"
+                            + "SGFzaE1hcAUH2sHDFmDRAwACRgAKbG9hZEZhY3RvckkACXRocmVzaG9sZHhwP0AAAAAAAAB3CAAA\n"
+                            + "ABAAAAAAeAA=",
+                    UnmodifiableMutableOrderedMap.of(OrderedMaps.adapt(new LinkedHashMap<>())));
+        }
```
motlin, line 37: The fact that there's no else block makes me wonder what happens in the new version. Is the serialized form changed? Does serialization now throw? Why?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
```diff
@@ -60,6 +60,26 @@
...
+    /**
+     * This default override exists because java.util.List added a default getLast() method in Java 21.
+     * @since 12.0
+     */
+    @Override
+    default T getLast()
+    {
+        return this.isEmpty() ? null : this.get(this.size() - 1);
+    }
+
```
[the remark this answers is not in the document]
donraab: > This is my first time looking at the changes coming in the SDK in version 21. The new getFirst() and getLast() methods have a different contract than ours. They throw, rather than return null, on empty collections. This is arguably a better choice, but it's a breaking change that we would never make if we didn't have to. I know this code review is focused on compiler errors and I'm not suggesting any code changes. Just wanted to start the conversation.
> 
> ```
> /**
>  * {@inheritDoc}
>  *
>  * @implSpec
>  * If this List is not empty, the implementation in this interface returns the result
>  * of calling {@code get(0)}. Otherwise, it throws {@code NoSuchElementException}.
>  *
>  * @throws NoSuchElementException {@inheritDoc}
>  * @since 21
>  */
> ```

I discovered this is the behavior for `MutableSortedSet` because it delegates to `first` / `last` methods which also throw `NoSuchElementException`. Agreed, we probably will never change it.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/set/sorted/MutableSortedSet.java
```diff
@@ -50,6 +50,28 @@
...
+    /**
+     * This default override exists because java.util.SortedSet added a default getLast() method in Java 21.
+     *
+     * @since 12.0
+     */
+    @Override
+    default T getLast()
+    {
+        return this.last();
+    }
```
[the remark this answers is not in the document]
donraab: This is not possible as `getFirst` does not exist on `SortedSet` until JDK 21 and we are still compiling with JDK 8.

serialization-tests/src/test/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMapSerializationTest.java
donraab: Yes the serialized form has changed. Something has changed in regards to serialization for `LinkedHashMap` but I do not have an easy way of determining what changed. I wanted to fix the builds for now and decide amongst the committers how best to handle the else situation, or remove the jdk version test altogether. We are not responsible for guaranteeing `LinkedHashMap` serialization, so we may want to rethink this test to guarantee our own serialization, not the JDKs.

donraab: @motlin I'm going to add a temporary serialization test just for LinkedHashMap so we can see the build failure and confirm it is unrelated to our types.

serialization-tests/src/test/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMapSerializationTest.java
donraab: Ok, I've removed the if statement in the latest push. Thanks for the feedback @Bananeweizen.

serialization-tests/src/test/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMapSerializationTest.java
donraab: @motlin As I expected, the JDK 21 build fails on a new serialization test for an empty `LinkedHashMap`.

> Error: 9:951 [ERROR]   UnmodifiableMutableOrderedMapSerializationTest.serializedFormLinkedHashMap:43 Serialization was broken. expected:<...zaE1hcDTATlwQbMD7AgA[BWgALYWNjZXNzT3JkZXJ4
> cgARamF2YS51dGlsLkhhc2hNYXAFB9rBwxZg0QMAAkYACmxvYWRGYWN0b3JJAAl0aHJlc2hvbGR4
> cD9AAAAAAAAAdwgAAAAQAAAAAHgA]> but was:<...zaE1hcDTATlwQbMD7AgA[CWgALYWNjZXNzT3JkZXJJ
> AAdwdXRNb2RleHIAEWphdmEudXRpbC5IYXNoTWFwBQfawcMWYNEDAAJGAApsb2FkRmFjdG9ySQAJ
> dGhyZXNob2xkeHA/QAAAAAAAAHcIAAAAEAAAAAB4AAAAAAA=]>

motlin: Interesting! The [only recent change](https://github.com/openjdk/jdk/commits/3ce1240ca1b4139980444c171e317f4bfeff9314/src/java.base/share/classes/java/util/LinkedHashMap.java) to LinkedHashMap was to make it a SequencedCollection, and [they did not change its serialVersionUID](https://github.com/openjdk/jdk/blame/3ce1240ca1b4139980444c171e317f4bfeff9314/src/java.base/share/classes/java/util/LinkedHashMap.java#L213), so it's possible this is a bug upstream.

mohrezaei: > it's possible this is a bug upstream.

I agree. Someone should report this.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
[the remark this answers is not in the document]
motlin: I'm not advocating for any changes, just thinking out loud with you since this is all new to me. Just want to point out that `public interface List<E> extends SequencedCollection<E>` in Java 21, which means that LinkedList is a SequencedCollection that's not random access. That means the default implementation _in the JDK_ is suboptimal, but we're used to that.

motlin (approved):
If you have the energy for it, I think we uncovered one or two issues worth reporting upstream. Either way, I think this looks like a good way to get compiling with all versions of Java again.

serialization-tests/src/test/java/org/eclipse/collections/impl/map/ordered/mutable/UnmodifiableMutableOrderedMapSerializationTest.java
stuart-marks: Hi all, JDK Sequenced Collections guy here. First, I hope I'm not causing you folks too much trouble. :-) @donraab brought this to my attention. Thanks Don!

Regarding this serialized form change, this was a mistake; there was no intent to change the serialized form. I've filed bug https://bugs.openjdk.org/browse/JDK-8309882 to restore to previous serialized form. It will likely be fixed in the next build of JDK 21.

That said, adding (and removing) fields from an object's serialized form is a compatible change; the behavior of "extra" and "missing" fields is well-defined in the serialization specification. We add or remove fields from time to time _without_ changing the object's serialVersionUID. (Changing the serialVersionUID is incompatible and is almost impossible for clients to cope with.) That might mean this test is overly brittle, depending on what it's trying to do.

motlin: Thanks for everything @stuart-marks! In the tests of the serialized forms of our wrapper classes, we're often wrapping JDK types. I agree with you that our tests are overly brittle. These tests have been quietly asserting that the serialized form doesn't change for empty ArrayLists, LinkedLists, HashSets, HashMaps, LinkedHashMaps, etc. for years. This is the first time I remember seeing a change to the serialized form, and I see that the addition of `transient` will bring back the old serialized form. I bet the serialized forms of these classes will never change. It might be helpful to add tests like these to the JDK! But if not, you can rely on us to ask about it when the form changes :)

While I have your attention, I have an unrelated question. Would it make sense to check for `RandomAccess` in List.getLast()?

```java
    default E getLast() {
        if (this.isEmpty()) {
            throw new NoSuchElementException();
        } else if (this instanceof RandomAccess) {
            return this.get(this.size() - 1);
        } else {
            return this.reversed().iterator().next();
        }
    }
```

stuart-marks: Hi @motlin! Haven't talked in a while. Yeah, the serialized forms of collections haven't changed much if at all over time. Other classes like `BigInteger` have changed some though. And their internal representations have changed too, so there are several layers of compatibility code that handle mapping of different serialized forms to the current internal representation.

And yes, I'm surprised that nothing in the JDK tests caught this change. I know we've added checks to prevent some obvious mistakes (such as forgetting to add a `serialVersionUID` declaration) but apparently not accidental changes to serialized forms. I'll have to check into this.

On the `getLast()` default implementation... the JDK's `LinkedList` implementation -- which is the most common `List` implementation I'm aware of that doesn't implement `RandomAccess` -- will traverse from the end if the requested index is closer to the end. Thus access to `size() - 1` will still be pretty much constant time. I suppose there might be some sequential list out there that doesn't know its size and that has to traverse from the front to find a particular index, but that would seem pretty unusual. I'm hard pressed to think of cases where doing a `RandomAccess` check will make a significant difference. If you're aware of any examples though, by all means let me know.

PR #1463, opened 2023-06-12
remove prompt from documentation code blocks

Bananeweizen:
People are used to copying the complete block from the documentation, and then the prompt will lead to syntax errors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,7 +39,7 @@ The Eclipse Collections build requires below as dependencies.
 The Eclipse Collections build performs code generation to create primitive collections. Run the full build once before opening your IDE.
 
 ```bash
-$ mvn install -DskipTests=true
+mvn install -DskipTests=true
 ```
 
 Now you can open the project in your IDE and it won't complain about missing files. You'll be able to use the IDE to perform incremental builds and run tests. You should rarely need to run the maven build, except when:
@@ -68,7 +68,7 @@ Coding Style
 Eclipse Collections follows a coding style that is similar to [Google's Style Guide for Java][style-guide], but with curly braces on their own lines. Many aspects of the style guide are enforced by CheckStyle, but not all, so please take care.
 
 ```bash
-$ mvn clean install checkstyle:check --projects "!scala-unit-tests,!jmh-scala-tests,!jmh-tests" -DskipTests=true
+mvn clean install checkstyle:check --projects "!scala-unit-tests,!jmh-scala-tests,!jmh-tests" -DskipTests=true
 ```
 
 Avoid changing whitespace on lines that are unrelated to your pull request. This helps preserve the accuracy of the git blame view, and makes code reviews easier.


donraab (approved):
LGTM. Thanks for the contribution @Bananeweizen. Please rebase and I will merge.

Bananeweizen:
@donraab If you want to ensure a more nice git history, there are 2 options you might want to enable:
* enable rebase merging of pull requests. that way rebases just happen on merge instead of repeatedly doing manual rebase beforehand: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-rebasing-for-pull-requests (I would strongly recommend this, because it leads to a nice to easy to use linear history of the main branch, without anyone doing manual work. I use rebase merges in all projects I manage privately or at work, and rebasing has led to maybe 2 red main branch builds after thousands of PRs handled that way.)
* enable the "update branch" button for all pull requests to trigger a manual rebase from the web UI: https://github.blog/changelog/2022-02-03-more-ways-to-keep-your-pull-request-branch-up-to-date/

Manual approaches just don't scale IMO. E.g. I currently have 4 open PRs starting from the same commit. Repeatedly merging only one and then rebasing all remaining is just wasted work.

mohrezaei:
@Bananeweizen and @motlin 
Can someone who's used the github rebase PR feature explain how that works with CI?
What happens in this simple case:
1) PR A is created. It passes CI and it's green.
2) PR's B, C, ... are merged. They do not have a git/textual conflict with A, but they have a logic conflict that completely breaks A.
3) github's rebase feature is used to merge A into master.
Question: Does the CI on master break? If the answer is yes, why do we bother running CI on PR's at all?

Bananeweizen:
@mohrezaei Yes, the master build is broken in that case.

Still, in my experience from many other projects it's a neglectable problem. The master build breaks way more often by _other_ changes that came into the CI system in an uncontrolled way. E.g. some dependencies or build steps not being versioned at all leading to uncontrolled updates. The maven build of this project uses 2 plugins without version numbers, so that's a real danger. Or Github actions not being pinned to their commit SHA, which means that hostile people could exchange externally defined github actions (even with a version number specified in the workflows of this project). Or external hosts updating their SSH certificates, leading to the cipher suite of the CI system no longer being able to connect to that host for some https based download. I've seen all of those and many more in real life, and semantical changes being introduced by rebase has been really low on that list.

mohrezaei:
Agree:
The update branch button sounds interesting. Sounds like it would fit nicely into our workflow.

Disagree:

I don't recall the last time the master build on EC was broken. Our PRs running github actions are not gated simply because nothing that runs in CI can affect real assets. A "hostile" PR is welcome to change all the .yml action files and the end result will be that we won't merge them, because we review the changes (and if it looks deliberate, we'll report them to github and get their account banned). We don't consider .yml files sacred: they change like any other file: PR + review + green CI => merge.

> The maven build of this project uses 2 plugins without version numbers, so that's a real danger.

Real danger that master breaks, because between the time a PR went green and it was merged, the versions changed in such an incompatible way that master goes red? If that ever happened, we have a simple trail that shows the versions that produced green and they could be pinned at that time, but really, that's a very far fetched scenario.

The point of having a guaranteed green CI before merge is one of responsibility. If a PR is auto-rebased and breaks master, who is supposed to fix that? The person who created the PR and may have not logged into Github for months? The person who did the merge? The entire team? 

With this workflow, I personally tend to avoid micro-commits. For example, I would've combined at least all 4 maven PR's into one, if not all 5 of your PRs. It's a lot less work for everyone.

motlin:
I want to enable the update branch feature. This feature allows maintainers
to update the branch with a rebase, which reruns checks. It makes it very
easy for maintainers to know that the branch is up to date, because the
button to update the branch doesn’t appear or is disabled. And of course
it’s easier for authors.

We need help from Eclipse folks to change the project settings.

donraab:
@motlin @mohrezaei @nikhilnanivadekar @prathasirisha Take a look at one of the approved PRs. It looks like I was able to enable the setting that enables the Update Branch button on a PR that needs to be rebased.

Bananeweizen:
> If a PR is auto-rebased and breaks master, who is supposed to fix that? The person who created the PR and may have not logged into Github for months? The person who did the merge? The entire team?

In that case I would always just revert to the last known good state, no matter how important the change was that broke it. It can be re-submitted after taking more time to rework it. Fixing a broken build by adding more stuff is generally not a good idea, IMO.

motlin:
We’re talking about problems caused by one of the options we’re not going
to use. The flow will be something like:

- a maintainer looks at a pull request
- they see that the checks are green and the update branch button is
showing. This is good enough to begin a review
- the maintainer clicks update branch (with rebase)
- the checks start to rerun
- while they are running, the maintainer goes through the changes to decide
whether to merge
- if the maintainer is happy, but the checks are still running, they may
have to wait a bit
- then they click merge

Now the history is clean and there’s no chance of our branches being red.
This approach scales until a project becomes so popular that pull requests
come in at a rate faster than the build takes to run.

PR #1465, opened 2023-06-12
Use correct minimum Maven version

Bananeweizen:
The Maven plugin versions specified require a higher Maven core version. This can be checked via `mvn versions:display-plugin-updates`.


motlin (approved):
Thank you! Our tests run with 3.6.3, so it's hard to tell what is the real minimum required version.

PR #1468, opened 2023-06-17
Clean up BND instructions

Bananeweizen:
* Remove jcni version (it's been removed in 2017 from the code base).
* Remove api version (it's never been used since moving the code to eclipse).
* Set a fixed Bundle-Developer in generated manifests to avoid bnd-maven-plugin complaining about all the developers in the parent pom.xml not having a mandatory ID.
* Set the Bundle-Description header in BND. Otherwise the pom.xml description is taken for that header, and unfortunately the pom.xml description tag contains whitespace from the XML formatting and indenting.
* Use the versionmask macro instead of the version macro. version is an alias for versionmask that is for compatibility only and should not be used.

If you are wondering what the above means:
grafik
grafik

### DIFF
--- a/eclipse-collections-api/bnd.bnd
+++ b/eclipse-collections-api/bnd.bnd
@@ -1,6 +1,9 @@
-package-version=${version;===;${Bundle-Version}}
+package-version=${versionmask;===;${Bundle-Version}}
 
 Automatic-Module-Name: org.eclipse.collections.api
 Bundle-SymbolicName: org.eclipse.collections.api
+# explicitly set bundle-developers to avoid bnd-maven-plugin raising warnings for missing ids in the developers section of pom.xml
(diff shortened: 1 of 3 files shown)


donraab:
@Bananeweizen Is this PR ok to merge now as is with the changes @fipro78 made in the other osgi PR? I'm ok with going with the Bundle-Developers approach.

donraab:
@Bananeweizen @fipro78 Can we close this PR now?

fipro78:
I can only add from my side that the `package-version` changes are already applied. The `Bundle-Developer` and `Bundle-Description` headers are still missing. The p2-repository does not exist anymore. @Bananeweizen needs to answer whether he wants to update the PR or if it should be simply dropped.

PR #1469, opened 2023-06-18
Upgrade Maven wrapper

Bananeweizen:
The maven wrapper is meanwhile an official Maven project and should be used with the maven groupid. Also silence the download transfer progress on all Maven invocations.

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -37,11 +37,11 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
       - name: Maven Install
-        run: mvn clean install -DskipTests=true
+        run: mvn clean install -DskipTests=true --no-transfer-progress
         env:
           MAVEN_OPTS: "-Xmx1g"
       - name: Mutation Test
-        run: mvn eu.stamp-project:pitmp-maven-plugin:run -DwithHistory -DtimeoutConstant=8000
+        run: mvn eu.stamp-project:pitmp-maven-plugin:run -DwithHistory -DtimeoutConstant=8000 --no-transfer-progress
         env:
           MAVEN_OPTS: "-Xmx1g"
       - name: Archive pitest reports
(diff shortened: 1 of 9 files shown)


.github/workflows/mutation.yml
```diff
@@ -37,11 +37,11 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
       - name: Maven Install
-        run: mvn clean install -DskipTests=true
+        run: mvn clean install -DskipTests=true --no-transfer-progress
         env:
           MAVEN_OPTS: "-Xmx1g"
       - name: Mutation Test
-        run: mvn eu.stamp-project:pitmp-maven-plugin:run -DwithHistory -DtimeoutConstant=8000
+        run: mvn eu.stamp-project:pitmp-maven-plugin:run -DwithHistory -DtimeoutConstant=8000 --no-transfer-progress
```
motlin, line 44: Shouldn't these calls to maven use `./mvnw`? I realize that the calls were already that way, but since you added the flag `--no-transfer-progress` this no relies on the floating version of maven that ships with github action runners.

.github/workflows/performance.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Git Checkout
         uses: actions/checkout@v3
       - name: Set Maven Wrapper
-        run: mvn -N io.takari:maven:0.7.7:wrapper -Dmaven=3.6.3
+        run: mvn -N wrapper:wrapper -Dmaven=3.6.3 --no-transfer-progress
```
motlin, line 22: Now I realize all the calls are like this. `--no-transfer-progress` was added in maven 3.6.1. And here we're specifying 3.6.3. We must have written these lines around the time that maven 3.6.1 was new and github runners didn't have access to that version yet.

motlin (approved):

donraab:
@motlin Is this ready to merge?

motlin:
@donraab yes, there are a few remaining calls to `mvn` that I'd like to replace with `mvnw` but that can be done separately and was an older problem.

PR #1476, opened 2023-06-30
Simplify Iterate forEach and sort by calling Java 8 default methods.

donraab:


eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/Iterate.java
```diff
@@ -124,14 +123,6 @@ else if (iterable instanceof BatchIterable)
...
-        {
-            ArrayListIterate.forEach((ArrayList<T>) iterable, procedure);
-        }
-        else if (iterable instanceof List)
-        {
-            ListIterate.forEach((List<T>) iterable, procedure);
-        }
         else if (iterable != null)
         {
             iterable.forEach(procedure);
```
motlin, line 128: The call to the default `Iterable.forEach()` was already here on line 128. So I'd describe the change as removing the specific cases for ArrayList and List. I think that's fine but I'm curious... didn't you add these cases previously, as an optimization?

mohrezaei: benchmarks?

donraab: No new benchmarks. I wasn't removing the instanceof checks to optimize the code. Just to simplify it.

mohrezaei: Umm, I meant have we run benchmarks to make sure this doesn't cause a performance regression?

donraab: No, I haven't. This is a fair point. Let me see what I can do.

donraab: @motlin @mohrezaei I finally got around to writing some benchmarks. There seems to be no regression for `ArrayList`. `Iterate` is barely faster than calling `ArrayListIterate` directly for a list of 100K elements with the change.

More importantly, I believe this change makes using `Iterate.forEach()` with `CopyOnWriteArrayList` and `Collections.synchronizedList()` safer and more desirable. We generally favor internal iterators over external, and now that the JDK has internal iterators, I believe it is more desirable to depend on them.

`JDKListIterationTest`
```
Benchmark                                                 Mode  Cnt     Score    Error  Units
JDKListIterationTest.arrayList_arrayListIterate_forEach  thrpt   20  1340.941 ± 19.529  ops/s
JDKListIterationTest.arrayList_iterate_forEach           thrpt   20  1363.489 ± 15.845  ops/s
JDKListIterationTest.cowal_iterate_forEach               thrpt   20  1342.872 ± 26.216  ops/s
JDKListIterationTest.cowal_listIterate_forEach           thrpt   20  1351.343 ± 31.598  ops/s
JDKListIterationTest.synchArrayList_iterate_forEach      thrpt   20  1324.700 ± 12.945  ops/s
JDKListIterationTest.synchArrayList_listIterate_forEach  thrpt   20  1338.609 ±  3.802  ops/s

```

mohrezaei (approved):

PR #1478, opened 2023-07-03
Forward deprecated methods to their new replacement

Bananeweizen:
* main code, tests are not affected
* only methods with a documented replacement

There is no functional change, as long as all deprecation replacements have been documented and delegated correctly before.


donraab (changes requested):
Hi @Bananeweizen, this PR looks good to me with one change I'd like to request. I'd like to have the StringPredicates changes extracted into a separate PR. All of the other changes here are related to primitive injectInto calls. Thanks!

donraab (approved):
Hi @Bananeweizen, thank you for this contribution. Looking again at this PR after a while, and I think your changes to `StringPredicates` is in the same spirit as the other changes. Please disregard my earlier comment about needing a separate PR for those changes. I am approving this.

PR #1479, opened 2023-07-16
Java 11 upgrade

prathasirisha:
I left my comments on some of the changes to add clarity. 

Created this PR to start discussions on the changes, but we should merge them only after we cut a milestone release (as suggested by @nikhilnanivadekar )

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,9 @@
   ~ All rights reserved. This program and the accompanying materials
   ~ are made available under the terms of the Eclipse Public License v1.0
   ~ and Eclipse Distribution License v. 1.0 which accompany this distribution.
-  ~ The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+  ~ The Eclipse Public License is available at https://www.eclipse.org/legal/epl-v10.html
   ~ and the Eclipse Distribution License is available at
-  ~ http://www.eclipse.org/org/documents/edl-v10.php.
+  ~ https://www.eclipse.org/org/documents/edl-v10.php.
   -->
 [![][maven img]][maven]
 [![][release img]][release]
@@ -22,9 +22,9 @@
 [![][actions checkstyle img]][actions checkstyle]
 [![][actions javadoc img]][actions javadoc]
 
-<a href="https://www.eclipse.org/collections/"><img src="https://github.com/eclipse/eclipse-collections/blob/master/artwork/eclipse-collections-logo.png" height="50%" width="50%"></a>
+<a href="https://eclipse.dev/collections/"><img src="https://github.com/eclipse/eclipse-collections/blob/master/artwork/eclipse-collections-logo.png" height="50%" width="50%"></a>
 
-#### [English](https://www.eclipse.org/collections/) | [中文](https://www.eclipse.org/collections/cn/index.html) | [Deutsch](https://www.eclipse.org/collections/de/index.html) | [Español](https://www.eclipse.org/collections/es/index.html) | [Ελληνικά](https://www.eclipse.org/collections/el/index.html) | [Français](https://www.eclipse.org/collections/fr/index.html) | [日本語](https://www.eclipse.org/collections/ja/index.html) | [Norsk (bokmål)](https://www.eclipse.org/collections/no-nb/index.html) | [Português-Brasil](https://www.eclipse.org/collections/pt-br/index.html) | [Русский](https://www.eclipse.org/collections/ru/index.html) | [हिंदी](https://www.eclipse.org/collections/hi/index.html)
+#### [English](https://eclipse.dev/collections/) | [Deutsch](https://eclipse.dev/collections/de/index.html) | [Ελληνικά](https://eclipse.dev/collections/el/index.html) | [Español](https://eclipse.dev/collections/es/index.html) | [中文](https://eclipse.dev/collections/cn/index.html) | [Français](https://eclipse.dev/collections/fr/index.html) | [日本語](https://eclipse.dev/collections/ja/index.html) | [Norsk (bokmål)](https://eclipse.dev/collections/no-nb/index.html) | [Português-Brasil](https://eclipse.dev/collections/pt-br/index.html) | [Русский](https://eclipse.dev/collections/ru/index.html) | [हिंदी](https://eclipse.dev/collections/hi/index.html) | [Srpski (latinica)](https://eclipse.dev/collections/sr-rs-latn/index.html)
 Eclipse Collections is a comprehensive collections library for Java. The library enables productivity and performance by delivering an expressive and efficient set of APIs and types. The iteration protocol was inspired by the Smalltalk collection framework, and the collections are compatible with the Java Collection Framework types.
 
 Eclipse Collections is compatible with Java 8+. Eclipse Collections is a part of the OpenJDK [Quality Outreach](https://wiki.openjdk.java.net/display/quality/Quality+Outreach) program, and it is validated for different versions of the OpenJDK.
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -557,19 +557,19 @@
             <plugin>
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <configuration>
-                    <source>${maven.compiler.source}</source>
+                    <source>1.8</source>
```
prathasirisha, line 560: I could not find a better solution after spending some time on this. I am open to alternate solutions if there are any. Without this change, the builds fail with 
`Exit code: 2 - Loading source files for package org.eclipse.collections.codegenerator.model...
error: No source files for package org.eclipse.collections.codegenerator.model`

I tried a few different solutions suggested [here](https://stackoverflow.com/questions/58836862/jdk-11-and-javadoc) and this is the only one that seems to work. I added a new upload artifact step in javadoc workflow to test that it actually generates javadoc with this change.

pom.xml
```diff
@@ -557,19 +557,19 @@
...
                     <links>
                         <link>https://junit.org/junit4/javadoc/latest</link>
-                        <link>https://docs.oracle.com/javase/8/docs/api/</link>
+                        <link>https://docs.oracle.com/en/java/javase/17/docs/api</link>
                     </links>
                     <destDir>${project.version}</destDir>
                     <doclint>html,syntax,accessibility</doclint>
                     <excludePackageNames>
-                        org.openjdk,org.eclipse.collections.impl.jmh,org.eclipse.collections.codegenerator,org.eclipse.collections.codegenerator.maven
+                        org.openjdk,org.eclipse.collections.impl.jmh,org.eclipse.collections.codegenerator.maven
```
prathasirisha, line 572: `excludePackageNames` doesn't seem to take as there is javadoc created for these excluded package names. I can revert this change, doesn't matter. I wanted to bring this to attention.

pom.xml
```diff
@@ -615,18 +615,19 @@
...
                 <configuration>
+                    <source>1.8</source>
                     <doctitle>Eclipse Collections - ${project.version}</doctitle>
                     <windowtitle>Eclipse Collections - ${project.version}</windowtitle>
                     <show>public</show>
                     <detectJavaApiLink>true</detectJavaApiLink>
                     <links>
                         <link>https://junit.org/junit4/javadoc/latest</link>
-                        <link>https://docs.oracle.com/javase/8/docs/api/</link>
+                        <link>https://docs.oracle.com/en/java/javase/17/docs/api</link>
```
prathasirisha, line 625: Reference: https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#links

missing-dependency/missing-dependency.iml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module org.jetbrains.idea.maven.project.MavenProjectsManager.isMavenModule="true" type="JAVA_MODULE" version="4">
-  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_8">
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_11">
```
prathasirisha, line 3: I made a change to this file, but I am not entirely sure what this module does. @motlin do we need this, it seems this was committed in 2018 with just this iml file. Happy to revert this change and take it up later.

.github/workflows/javadoc.yml
```diff
@@ -29,4 +29,14 @@ jobs:
         run: ./mvnw -Dsurefire.useFile=false -DskipTests -Denforcer.skip=true clean install javadoc:aggregate -pl '!unit-tests,!serialization-tests,!jcstress-tests,!junit-trait-runner,!unit-tests-java8,!test-coverage,!p2-repository,!org.eclipse.collections:org.eclipse.collections' --no-transfer-progress
         env:
           MAVEN_OPTS: "-Xmx1g"
+      - name: checks
+        run: |
+          ls -ltr ${{ github.workspace }}/target
+          ls -ltr ${{ github.workspace }}/target/site
+          ls -ltr ${{ github.workspace }}/target/site/apidocs
+      - name: Upload Javadoc
```
prathasirisha, line 37: I added this in case anyone wants to see the generated javadoc given this PR has related javadoc plugin changes. I will revert this section before the PR gets merged.

eclipse-collections-api/pom.xml
```diff
@@ -88,7 +88,7 @@
                     <windowtitle>Eclipse Collections API - ${project.version}</windowtitle>
                     <show>public</show>
                     <links>
-                        <link>https://docs.oracle.com/javase/8/docs/api/</link>
+                        <link>https://docs.oracle.com/en/java/javase/17/docs/api</link>
```
donraab, line 91: Why was this changed from 8 to 17 instead of 8 to 11? Was this because we are building Javadoc with 17?

donraab: > Why was this changed from 8 to 17 instead of 8 to 11? Was this because we are building Javadoc with 17?

Never mind @prathasirisha, I see you added a comment above referencing this change.

prathasirisha: I meant to reference Java 11 docs, looks like committed the changes before I could make that change. I will update them shortly.

donraab (approved):
This PR looks good to me. Thanks for contributing @prathasirisha! I would like to give time to hear any feedback and/or approvals from @nikhilnanivadekar and @motlin.

nikhilnanivadekar:
Will take a look tomorrow. Please do not merge till milestone is cut.

nikhilnanivadekar:
Milestone is published, will reflect in maven central as per the usual SLA. Will do a review later today/tomorrow.

prathasirisha:
> Milestone is published, will reflect in maven central as per the usual SLA. Will do a review later today/tomorrow.

Thanks @nikhilnanivadekar for cutting the milestone release.

.github/workflows/javadoc.yml
nikhilnanivadekar: if you have verified that the javadocs look ok, We should be fine 😄

missing-dependency/missing-dependency.iml
nikhilnanivadekar: this should ideally auto-update. But the change is ok.

pom.xml
nikhilnanivadekar: We are using `<version>3.4.1</version>` for maven javadoc plugin. Do we need to upgrade this? It does not smell right that we are linking the 1.8 API docs in Javadoc but building on 11.

pom.xml
nikhilnanivadekar: I think this has to do with the javadoc plugin version. Also the delimiter might have changed: https://maven.apache.org/plugins/maven-javadoc-plugin/examples/exclude-package-names.html

nikhilnanivadekar (changes requested):

pom.xml
prathasirisha: Delimiter has not changed. From the link you referenced: 

_If more than one package is to be excluded, each package name must be separated either with a colon (':'), a comma (',') or a semicolon (';')._

pom.xml
prathasirisha: I started out with 3.5.0, which is the latest version available but rolled back the change when I was running into build errors to eliminate any issues stemming from this plugin version upgrade. 

I can update this to 3.5.0 as it changes nothing for this PR and we will be on the latest version too.

prathasirisha: I agree it doesn't smell right that we are using 1.8 as the source. One option that I didnt explore is creating an issue [here](https://issues.apache.org/jira/projects/MJAVADOC/summary). I didnt see an issue with our exact problem, so that may be worth creating. I am open to other ideas as well.

nikhilnanivadekar:
You can raise an issue. Please mention everything that you have tried.

pom.xml
prathasirisha: Created an issue - https://issues.apache.org/jira/browse/MJAVADOC-764

nikhilnanivadekar:
@prathasirisha whats the latest on this? Do we want to pursue fixing the Javadoc or just merge? If you can verify that the Javadoc is fine, and spot check a few classes (EC & JDK) should we merge and proceed?

PR #1492, opened 2023-08-22
Added comments to identify the usage of the ' of() ' and ' with() ' functions

BrijeshPatra:
Please review and approve 
@sullis @gastaldi @valery1707 @brunoborges @Godin 
Thanks,
Brijesh Patra

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/MapIterate.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/MapIterate.java
@@ -16,6 +16,8 @@
  * making your code more concise and readable.
  
  * This utility class is especially useful when you want to return a MutableMap as a result of your operations. 
+  
+ * Explore the capabilities of MapIterate to streamline your data manipulation tasks when working with maps. It is a versatile tool in your Java development toolkit.
  */
 
 package org.eclipse.collections.impl.utility;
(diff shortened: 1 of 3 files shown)


brunoborges (approved):

brunoborges (approved):

BrijeshPatra:
@brunoborges Successfully modified the doc part so that developers can identify it easily. Added the comments about usage of methods.

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/MapIterate.java
```diff
@@ -6,6 +6,18 @@
  * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
  * and the Eclipse Distribution License is available at
  * http://www.eclipse.org/org/documents/edl-v10.php.
+ * 
+ * Utility class for iterating and performing operations on java.util.Map objects.
+
```
brunoborges, line 11: This block of documentation should go above the class type definition. Check the source code of java.lang.String for an example.

brunoborges (changes requested):

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/MapIterate.java
BrijeshPatra: Sure I will update it ASAP.
Thanks

PR #1511, opened 2023-09-11
Upgrade CheckStyle.

motlin:


checkstyle-configuration.xml
```diff
@@ -246,8 +246,6 @@
 
         <module name="SimplifyBooleanReturn" />
```
nikhilnanivadekar, line 248: Being lazy here, was this removed or something changed that we dislike?

nikhilnanivadekar (approved):

checkstyle-configuration.xml
motlin: I looked at the build failure on the dependabot review: #1510

It says `You have 677 Checkstyle violations.` and they are all `(design) FinalClass: Class KeySet should be declared as final.`

I don't know if the check was broken and they fixed it, or if they made it more strict, or what. I just disabled the check to move forward.

nikhilnanivadekar:
Makes sense, thanks!

PR #1515, opened 2023-09-22
Remove Java 19 and 20 builds and add Java 22 EA build.

donraab:


.github/workflows/javadoc.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        java-version: [11, 17]
+        java-version: [11, 17, 21]
```
prathasirisha, line 14: This feels excessive. We can just do 21 (latest) or 11 (since that is current java version)

.github/workflows/unit-no-p2.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        java-version: [17]
+        java-version: [17, 21]
```
prathasirisha, line 14: We can just do 21 here.

prathasirisha (changes requested):

.github/workflows/javadoc.yml
donraab: I guess let's go with 11 then.

.github/workflows/unit-no-p2.yml
donraab: If I get rid of 17 here it means we are no longer running unit tests against Java 17 (just 11 and 21). We need to run unit tests against all three of the latest LTS versions.

prathasirisha (approved):

PR #1521, opened 2023-10-14
Move test template to the correct directory - fixes #1520

### LINKED ISSUE #1520: Put AbstractImmutablePrimitiveSetTestCase template in the right folder
Currently the abstractImmutablePrimitiveSetTestCase template file sits under directory 'mutable'.

image

The file should logically sit under the 'immutable' template directory.

Desislav-Petrov:
Move test template file to the correct location


donraab (changes requested):
Thanks for reporting this problem and for the contribution @Desislav-Petrov. I looked at the file and it looks like moving the directory is not the only problem that needs to be addressed. The files being generated will be generated to the mutable package - `package org.eclipse.collections.impl.set.mutable.primitive;` and then this class is being included incorrectly in import statements in other files that should be corrected. For example in immutablePrimitiveHashSetTest.stg which imports `org.eclipse.collections.impl.set.mutable.primitive.AbstractImmutable<name>HashSetTestCase;`. I'd request that the additional fixes be applied so not only does the stg file sits in the correct directory, that the generated classes sit in the write packages. Thanks!

eclipse-collections-code-generator/src/main/resources/test/set/immutable/abstractImmutablePrimitiveSetTestCase.stg
```diff
@@ -66,12 +69,29 @@ public abstract class AbstractImmutable<name>HashSetTestCase extends AbstractImm
     {
         <name>ArrayList collisions = new <name>ArrayList();
         <name>HashSet set = new <name>HashSet();
-        for (<type> i = <(literal.(type))("32")>; collisions.size() \<= 10; i++)
+
+        Method m;
+        try
+        {
+            m = set.getClass().getDeclaredMethod("spreadAndMask", <type>.class);
```
Desislav-Petrov, line 76: This is a bit of a funny one because this method is accessible only form the mutable packages. After the move to immutable, this is no longer available. I thought of a few options to address that:

1) Expose to public interface but I doubt this is what we want to do
2) Pull all these helper methods into a util class that's written in Java and imported in all tests, accessible by both mutable & immutable packages
3) Do the same as 2 but thru the template generation
4) Use reflection which isn't great but seems like the least impactful change and is test code.. that's why I chose this one

Desislav-Petrov:
@donraab do you mind casting an eye whenever you get a chance please

eclipse-collections-code-generator/src/main/resources/test/set/immutable/abstractImmutablePrimitiveSetTestCase.stg
donraab: Thanks for the clarification @Desislav-Petrov and providing some alternatives. I'm looking at the source and will see if I can find any other alternatives.

donraab: Ok, taking a look at `AbstractIntSetTestCase` I can see there is a duplicate protected method named `generateCollisions1`. This is in the mutable package. 

Here's what I would recommend. Make the method `generateCollisions1` public, and change the generateCollisions implementation to call the static method on `AbstractIntSetTestCase`. After the change the method should look like this on the abstract immutable test case.

    protected static IntArrayList generateCollisions()
    {
        return AbstractIntSetTestCase.generateCollisions1();
    }

This creates a dependency between the immutable package on the immutable, but that dependency already existed. This is all test code anyway, so I would rather remove duplication than worry that a static test method is protected or public.

donraab (changes requested):
Hi @Desislav-Petrov, I made one suggestion on removing duplicate code for `generateCollisions` in the comments.

unit-tests/src/test/java/org/eclipse/collections/impl/set/immutable/primitive/AbstractImmutableByteHashSetTestCase.java
```diff
@@ -52,6 +58,40 @@ protected MutableSet<Byte> newObjectCollectionWith(Byte... elements)
         return UnifiedSet.newSetWith(elements);
     }
 
+    protected static ByteArrayList generateCollisions()
```
Desislav-Petrov, line 61: This one we can't easily avoid though as it's outside of the autogen classes & there's no generateCollisions1() for ByteArrayList - shall we leave like this?

eclipse-collections-code-generator/src/main/resources/test/set/immutable/abstractImmutablePrimitiveSetTestCase.stg
Desislav-Petrov: Ah - haven't thought about that - thx for pointing out, it does seem like a cleaner approach. Changes made to reflect it

eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutablePrimitiveMapKeySet.stg
```diff
@@ -2,7 +2,7 @@ import "primitiveImmutableKeySets.stg"
 
 skipBoolean() ::= "true"
 
-targetPath() ::= "org/eclipse/collections/impl/map/mutable/primitive"
+targetPath() ::= "org/eclipse/collections/impl/map/immutable/primitive"
```
donraab, line 5: This cannot be moved as it will break serialization.

eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutablePrimitivePrimitiveMapKeySet.stg
```diff
@@ -2,7 +2,7 @@ import "primitiveImmutableKeySets.stg"
 
 skipBoolean() ::= "true"
 
-targetPath() ::= "org/eclipse/collections/impl/map/mutable/primitive"
+targetPath() ::= "org/eclipse/collections/impl/map/immutable/primitive"
```
donraab, line 5: This cannot be moved as it will break serialization.

eclipse-collections-code-generator/src/main/resources/impl/map/immutable/primitiveImmutableKeySets.stg
```diff
@@ -6,7 +6,7 @@ import "primitiveLiteral.stg"
 primitiveKeySetBody(className, type, name, keyArray, increment, sameTwoPrimitiveMapKeySet, calledFrom) ::= <<
 <copyright()>
 
-package org.eclipse.collections.impl.map.mutable.primitive;
+package org.eclipse.collections.impl.map.immutable.primitive;
```
donraab, line 9: This cannot be moved as it will break serialization.

donraab (changes requested):
Hi @Desislav-Petrov, I can't recall if the impl/map generated files were in the PR previously, but they should not be moved as these classes are Serializable and this will break serialization. Please try and limit this PR to moving test templates only. I see that we have some immutable map types that wound up in mutable packages, and that is unfortunate, but would be a more difficult change. If this is not possible without these moves, then we should probably rescope or rethink the effort.

eclipse-collections-code-generator/src/main/resources/test/set/immutable/immutablePrimitivePrimitiveMapKeySetTest.stg
```diff
@@ -5,7 +5,7 @@ isTest() ::= "true"
 
 skipBoolean() ::= "true"
 
-targetPath() ::= "org/eclipse/collections/impl/map/mutable/primitive"
+targetPath() ::= "org/eclipse/collections/impl/map/immutable/primitive"
```
Desislav-Petrov, line 8: Should probably get these 3 back into the mutable package - incorrect from semantics perspective but this is where their main classes are so would be consistent

eclipse-collections-code-generator/src/main/resources/impl/map/immutable/primitiveImmutableKeySets.stg
Desislav-Petrov: reverted

eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutablePrimitivePrimitiveMapKeySet.stg
Desislav-Petrov: reverted

eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutablePrimitiveMapKeySet.stg
Desislav-Petrov: reverted

donraab:
Hi @Desislav-Petrov, one of the inspections builds is failing.

Desislav-Petrov:
> Hi @Desislav-Petrov, one of the inspections builds is failing.

Sry - fixed

eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveImmutableKeySets.stg
```diff
@@ -44,7 +44,7 @@ class <className> extends AbstractImmutable<name>Set implements Serializable
     private final boolean containsZeroKey;
     private final boolean containsOneKey;
 
-    <className>(<type>[] <keyArray>, int occupiedWithData, boolean containsZeroKey, boolean containsOneKey)
+    public <className>(<type>[] <keyArray>, int occupiedWithData, boolean containsZeroKey, boolean containsOneKey)
```
donraab, line 47: This should not need to be made public.

eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveImmutableKeySets.stg
```diff
@@ -30,7 +30,7 @@ import org.eclipse.collections.impl.set.mutable.primitive.<name>HashSet;
  *
  * @since 6.0.
  */
-class <className> extends AbstractImmutable<name>Set implements Serializable
+public class <className> extends AbstractImmutable<name>Set implements Serializable
```
donraab, line 33: This should not need to be made public.

donraab (changes requested):
The domain class should not need to be made public. Read my suggestion on where to move generateCollisions. I thought I suggested moving the method to the mutable test case and calling it from the immutable test cases. This change should not require changing any of the actual Set classes. Thanks!

eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveImmutableKeySets.stg
Desislav-Petrov: The suggested generateCollisions approach works for all cases that come from out generated templates but there's a special hierarchy for *ByteSetTestCase that this didn't account for.

I've reverted these publics to package visible again and I've introduced the CollisionsUtil class to overcome the limitations for the *ByteSetTestCase hierarchy.

donraab (commented):
Thanks @Desislav-Petrov for the updates. I think we're getting close but I first want to understand why ImmutableByteMapKeySetTest was deleted. Thanks!

Desislav-Petrov:
> Thanks @Desislav-Petrov for the updates. I think we're getting close but I first want to understand why ImmutableByteMapKeySetTest was deleted. Thanks!

Hi @donraab - the one I have deleted wasn't an autogenerated and it was incorrectly placed in the *mutable* package. After the template adjustments, we now have ImmutableByteMapKeySetTest in the *immutable* package that's auto generated. 

PR rebased as well.

donraab (approved):
Thanks for the explanation on the deleted test @Desislav-Petrov. I can see that the code between the deleted hand coded version and the template were very similar. I'm approving this now. Thanks for the contribution!

Issue #1453, opened 2023-04-12
The result types in sumOfPrimitiveProcedure.stg need to be widened to prevent overflow


donraab:
I discovered there is an overflow error using `sumOfInt` in the following code when the batch size exceeded 536

This code returned the correct sum:

```
long sum = Interval.oneTo(2_000_000)
        .toList()
        .asParallel(Executors.newSingleThreadExecutor(), 536)
        .sumOfInt(each -> each * 2);
```

This code returned the incorrect sum:

```
long sum = Interval.oneTo(2_000_000)
        .toList()
        .asParallel(Executors.newSingleThreadExecutor(), 537)
        .sumOfInt(each -> each * 2);

```
I tracked the problem down to `SumOfIntProcedure` returning `int` instead of `long` for the result. This Procedure must be used in the individual batches and if a batch gets large enough with large values being summed, overflow is occurring.

https://github.com/eclipse/eclipse-collections/blob/master/eclipse-collections-code-generator/src/main/resources/impl/block/procedure/sumOfPrimitiveProcedure.stg#L33

rzrobin213:
Hi, I would like to take up this issue. Could you assign it to me? Thank you!

donraab:
Thanks for volunteering @rzrobin213 ! I have assigned the issue to you.

TheJavaGuy:
@donraab File `eclipse-collections/src/main/java/org/eclipse/collections/impl/block/procedure/SumOfIntProcedure.java` seems not to be auto-generated. OTOH, file `eclipse-collections/target/generated-sources/java/org/eclipse/collections/impl/block/procedure/SumOfByteProcedure.java` says that it is generated from the `sumOfPrimitiveProcedure.stg`.

lightgrey7:
![git changed code](https://github.com/eclipse/eclipse-collections/assets/121189332/e0a03037-cb69-436e-84be-809c506816a8)
Hi! I have made some changes and optimized the code, please take a look.

lightgrey7:
hello!! I have made a pull request regarding this issue..please review it.
@donraab

PR #1531, opened 2023-12-01
#1453 issue, From int to LongAdder

lightgrey7:


donraab (changes requested):
Thanks for the contribution @lightgrey7 ! I think this change looks good. The extra benefit of being thread safe, while not strictly necessary here, I think is not bad if someone wants to use this type with a parallel Stream. The checkstyle build is failing, probably on the placement of the comments. If you can address those issues and rebase, I will approve and merge. Thanks!

PR #1534, opened 2023-12-11
Update README.md.

donraab:


README.md
```diff
@@ -184,6 +187,7 @@ We welcome contributions! We accept contributions via pull requests here in GitH
 [BlogUnparalleledDesignOfEclipseCollections]: https://medium.com/javarevisited/the-unparalleled-design-of-eclipse-collections-e4340b00f79f?source=friends_link&sk=629e6384171b18233a167a499b46408c
 [BlogIterateOverAnyIterableInJava]: https://medium.com/javarevisited/iterate-over-any-iterable-in-java-bec78eeeb452?source=friends_link&sk=7d460d1494cb76ce6bc9a5543785224a
 [BlogAsAMatterOfFactory]: https://medium.com/oracledevs/as-a-matter-of-factory-factories-matter-482d8adff094?source=friends_link&sk=96a4cd8fbc42e309c39a917449e6bff2
+[BlogJavaHasStreamsDoWeNeedCollections]: https://motlin.medium.com/java-has-streams-do-we-need-third-party-collections-dd12f473d105
```
motlin, line 190: 👍

motlin (approved):

README.md
```diff
@@ -71,15 +72,17 @@ Eclipse Collections is compatible with Java 8+. Eclipse Collections is a part of
...
-| 11.x.x |           |&check;| &check;    | &check;     |
+| EC     | JDK 5 - 7 | JDK 8   | JDK 9 - 10 | JDK 11 - 14 | JDK 15 - 21 |
+|--------|-----------|---------|------------|-------------|-------------|
+| 7.x.x  | &check;   | &check; |            |             |             |
+| 8.x.x  |           | &check; |            |             |             |
+| 9.x.x  |           | &check; | &check;    |             |             |
+| 10.x.x |           | &check; | &check;    |             |             |
+| 10.4.0 |           | &check; | &check;    | &check;     | &check;     |
+| 11.x.x |           | &check; | &check;    | &check;     | &check;     |
+| 12.x.x |           |         |            | &check;     | &check;     |
```
motlin, line 83: I can't exactly describe what's confusing me about this table, but the last two columns for JDK 11-14 and JDK 15-21 are identical, which makes them look like they can be collapsed. Looking at the old table, it's clear that 10.4.0 is the first version supporting Java 15+, but that doesn't show up in the new table. Maybe there's a mixxing check for 9.x.x and 10.x.x supporting JDK 11-14?

motlin (changes requested):
I already clicked approve but think you might want to edit the table, so I'm leaving this comment just to wipe out the approval. It's really up to you whether the table is ok as-is.

README.md
donraab: You're correct, I forgot to carry over some checkmarks when I split JDK 9 - 14 up into 9-10 and 11-14

motlin (approved):

Issue #581, opened 2018-07-24
IDE files cause continuous problems


mohrezaei:
The IDE files for Intellij IDEA are not stable. I have to constantly exclude them when committing. Every upgrade of the IDE creates new conflicts (new files, removed files, changed files, rearranged files, etc). On every rebase or other git action (amend, merge, etc), I have to stash them or things go wrong. 

Is a one time setup cost really worth the continuous torture?

bamapookie:
I would suggest adding the recommended ignore entries to your .gitignore file.  You can find them here: https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

Add that to .gitignore, then run `git rm --cached <filename>` for each of the ignored files.

bamapookie:
Also, if you use the .ignore plugin in IntelliJ, that will make the steps above much simpler.  You can simply add the template to .gitignore, then restart and it will suggest removing the tracked ignored files.

nikhilnanivadekar:
@bamapookie would you like to raise a PR to untrack the tracked files and update the .gitignore. Thanks for your pointers!

bamapookie:
If I can get to it today, I will.

bamapookie:
@nikhilnanivadekar Question:  Do you autoimport your project from maven or gradle?  If so, there are additional ignore entries in the template to uncomment.

mohrezaei:
Here is what git tells me is modified in my source tree:

        modified:   .idea/compiler.xml
        modified:   .idea/encodings.xml
        modified:   .idea/libraries/Maven__org_scala_lang_scala_library_2_11_7.xml
        modified:   .idea/misc.xml
        modified:   .idea/modules.xml
        modified:   eclipse-collections-parent.iml

I'm not sure about .idea/libraries/Maven...: does that match ```.idea/**/libraries```? None of the others would be fixed by the Jetbrain's default .gitignore. If we assume maven auto-import, that would fix modules.xml. That still leaves at least 4 files.

bamapookie:
@mohrezaei That matches. "**" matches 0 or more subdirectories.

mohrezaei:
@bamapookie ok, great. What about the other 4?

bamapookie:
@mohrezaei some config files should remain.  If you notice any churn in them (list entries being reordered, for instance), there is an IntelliJ bug to track and fix the reordering here: https://youtrack.jetbrains.com/issue/IDEA-91607.  Just comment with the file being reordered and give an example.

[...]

PR #1537, opened 2024-01-01
Replace @Test(expected=X) with assertThrows(X) in unit-tests-java8

josephks:
`@Test` does not have the "expected" field in junit-jupiter.  This is a small first step in moving to junit-jupiter.

### DIFF
--- a/unit-tests-java8/src/test/java/org/eclipse/collections/test/collection/mutable/MultiReaderMutableCollectionTestCase.java
+++ b/unit-tests-java8/src/test/java/org/eclipse/collections/test/collection/mutable/MultiReaderMutableCollectionTestCase.java
@@ -13,15 +13,17 @@ package org.eclipse.collections.test.collection.mutable;
 import org.eclipse.collections.impl.collection.mutable.AbstractMultiReaderMutableCollection;
 import org.junit.Test;
 
+import static org.junit.Assert.assertThrows;
+
 public interface MultiReaderMutableCollectionTestCase extends MutableCollectionTestCase
 {
     @Override
     <T> AbstractMultiReaderMutableCollection<T> newWith(T... elements);
 
-    @Test(expected = UnsupportedOperationException.class)
+    @Test
     default void Iterable_iterator_throws()
     {
-        this.newWith(3, 2, 1).iterator();
+        assertThrows(UnsupportedOperationException.class, () -> this.newWith(3, 2, 1).iterator());
     }
 
     @Test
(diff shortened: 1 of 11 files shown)


unit-tests-java8/src/test/java/org/eclipse/collections/test/stack/mutable/UnmodifiableMutableStackTestCase.java
```diff
@@ -24,18 +25,18 @@ default void Iterable_remove()
...
     }
 
     @Override
-    @Test(expected = UnsupportedOperationException.class)
+    @Test
     default void MutableStack_pop_throws()
     {
         MutableStack<Integer> mutableStack = this.newWith();
-        mutableStack.pop();
+        Assert.assertThrows(UnsupportedOperationException.class, () -> mutableStack.pop());
```
motlin, line 40: The other usages are static imports. I think these are the only usages with non-static imports. What's the distinction here?

unit-tests-java8/src/test/java/org/eclipse/collections/test/collection/mutable/MultiReaderMutableCollectionTestCase.java
```diff
@@ -18,10 +18,10 @@ public interface MultiReaderMutableCollectionTestCase extends MutableCollectionT
     @Override
     <T> AbstractMultiReaderMutableCollection<T> newWith(T... elements);
 
-    @Test(expected = UnsupportedOperationException.class)
+    @Test
     default void Iterable_iterator_throws()
     {
-        this.newWith(3, 2, 1).iterator();
+        org.junit.Assert.assertThrows(UnsupportedOperationException.class, () -> this.newWith(3, 2, 1).iterator());
```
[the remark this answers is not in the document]
josephks: I'm not sure how these inconsistencies crept in.  I originally wanted to make much more drastic changes in moving in the direction of junit-jupiter but issue #581 was driving me crazy and I ran into some things I didn't understand in the code base so I reverted a lot of my code changes to end up with this.  In the process I guess things like this crept in.  Fixing now.

unit-tests-java8/src/test/java/org/eclipse/collections/test/list/mutable/UnmodifiableMutableListTestCase.java
```diff
@@ -15,24 +15,25 @@
...
 
     @Override
-    @Test(expected = UnsupportedOperationException.class)
+    @Test
     default void MutableList_shuffleThis()
     {
-        this.newWith(5, 1, 4, 2, 3).shuffleThis();
-        this.newWith(5, 1, 4, 2, 3).shuffleThis(new Random(8));
+        assertThrows(UnsupportedOperationException.class, () -> this.newWith(5, 1, 4, 2, 3).shuffleThis());
+        assertThrows(UnsupportedOperationException.class, () -> this.newWith(5, 1, 4, 2, 3).shuffleThis(new Random(8)));
```
motlin, line 36: This is a perfect example of why @Test(expected) should never be used. Thank you for getting rid of these.

PR #1540, opened 2024-01-08
Implement RichIterable.reduceBy(). Fixes #134

### LINKED ISSUE #134: Implement RichIterable.reduceBy().
The existing method `RichIterable.aggregateBy(Function, Function0, Function2)` is conceptually a `groupBy()` followed by `injectInto()` on each group of values. This method would be a `groupBy` followed by `reduce()` on each group of values. Multiple variants are possible, but the first signature ought to be:

``` java
<K, T> MapIterable<K, T> reduceBy(
    Function<? super T, ? extends K> groupBy,
    Function2<? super T, ? super T, ? extends T> reduceFunction);
```

motlin:


donraab (changes requested):
@motlin This looks like a nice addition. I would recommend adding at since 12.0 tags in all of the interfaces. I also think there needs to be special handling for MultiReader classes, similar to the synchronized types.

motlin:
> @motlin This looks like a nice addition. I would recommend adding at since 12.0 tags in all of the interfaces. I also think there needs to be special handling for MultiReader classes, similar to the synchronized types.

I'm about to push a new change that updates since from `?` to `12.0` and overrides these methods in `AbstractMultiReaderMutableCollection`. However, I picked the places to place overrides by following the places where we have overridden the methods `aggregateBy`, `countBy`, `groupBy` etc. None of these methods are overridden in the MultiReader collections. I don't think they need to be for correctness, because they're all implemented by delegating to forEach() which does the right thing. It might be nice to override them for other reasons.

motlin:
@donraab I see your 👍  but you still have the request for changes so I want to make sure it's all good before I click Merge.

donraab (changes requested):
One small fix and then LGTM

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -2363,6 +2363,42 @@ default <K, V, R extends MutableMapIterable<K, V>> R aggregateBy(
...
+                groupBy,
+                reduceFunction,
+                Maps.mutable.empty());
+    }
+
+    /**
+     * Applies an aggregate function over the iterable grouping results into a map based on the specific groupBy function.
+     * Aggregate results are allowed to be immutable as they will be replaced in place in the map.
+     *
+     * @since ?
```
donraab, line 2387: Should be since 12.0 not since ?

donraab (approved):
LGTM

PR #1542, opened 2024-01-08
Override Map.merge() default method for correctness and efficiency. Partially addresses #500

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
```diff
@@ -2307,6 +2308,89 @@ private <P> V slowUpdateValueWith(
...
+                        {
+                            //noinspection ContinueStatementWithLabel
+                            continue outer;
+                        }
+
+                        if (newValue == null)
+                        {
+                            this.addToSize(-1);
+                        }
+                        return newValue;
```
motlin, line 2380: This is the trickiest part of the implementation and deserves a careful look. If the lambda returns null, the contract says to remove the existing entry for the key. Unlike the updateValue and getIfAbesent methods, the map can grow, stay the same, or shrink from this method call.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -569,6 +571,90 @@ private <P> V chainedUpdateValueWith(
...
+        {
+            this.table[index] = UnifiedMap.toSentinelIfNull(key);
+            this.table[index + 1] = value;
+            ++this.occupied;
+            return value;
+        }
+        if (cur != CHAINED_KEY && this.nonNullTableObjectEquals(cur, key))
+        {
+            V newValue = remappingFunction.apply((V) this.table[index + 1], value);
+            this.table[index + 1] = newValue;
```
mohrezaei, line 592: this doesn't look right. if `newValue` is null, we must clear the key slot.

mohrezaei (changes requested):

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
motlin: Good point, will update

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MutableMapIterableTestCase.java
```diff
@@ -277,4 +277,30 @@ default void MutableMapIterable_updateValue()
                 Collections.nCopies(1000, 2),
                 map4.values());
     }
+
+    @Test
+    default void Map_merge()
```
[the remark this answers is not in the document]
motlin: Will do

mohrezaei:
Just thinking this over:
- some collision tests (chained entries) would improve coverage for Unfied*
- if the result of calling the function is `oldValue` (i.e. `newValue == oldValue`) a lot of processing can be skipped.

mohrezaei:
additional things to do from the javadoc:

1)  If the remapping function itself throws an (unchecked) exception, the exception is rethrown, and the current mapping is left unchanged. 
2) In particular, all implementations of subinterface [ConcurrentMap](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ConcurrentMap.html) must document whether the remapping function is applied once atomically only if the value is not present. 

for (2), we have to document that all these are possible:
- function is called multiple times
- function is called one or more times, but the result is not used (because the old entry was concurrently removed)

Also I think the javadoc language for (2) is broken. "applied once atomically only if the value is not present. " is not sensible: if it's not present, it shouldn't have been applied, no?

for (1), we can't leave the map in an intermediate state. I don't see any such case, but a test or two would be good.

unit-tests/src/test/java/org/eclipse/collections/impl/map/mutable/UnifiedMapTest.java
```diff
@@ -866,6 +866,16 @@ public void trimToSize()
...
 
+    @Test
+    public void merge()
+    {
+        MutableMap<Integer, Integer> actualMap = this.mapWithCollisionsOfSize(8);
+        Integer merge = actualMap.merge(COLLISION_1, 1, (integer1, integer2) -> null);
+        Assert.assertNull(merge);
+        MutableMap<Integer, Integer> expectedMap = MORE_COLLISIONS.subList(1, 8).toMap(k -> k, v -> v);
+        Assert.assertEquals(expectedMap, actualMap);
+    }
```
motlin, line 877: I stepped through this test in a debugger and it hits the line `this.overwriteWithLastElementFromChain(chain, index, i)`

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
motlin: Fixed this, and stepped through the tests in the debugger to make sure the new lines are hit in tests.

PR #1567, opened 2024-03-26
Update JMH Benchmarks and library dependencies.

donraab:
I removed all of the `extends AbstractJMHTestRunner` code as IntelliJ has supported running JMH Benchmarks in the IDE for quite a while now. I added annotations to each of the benchmark classes for the number of forks, warmups, and iterations.

### DIFF
--- a/jmh-scala-tests/pom.xml
+++ b/jmh-scala-tests/pom.xml
@@ -43,9 +43,9 @@
         </dependency>
 
         <dependency>
-            <groupId>commons-lang</groupId>
-            <artifactId>commons-lang</artifactId>
-            <version>2.6</version>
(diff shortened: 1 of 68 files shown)


motlin:
> I added annotations to each of the benchmark classes for the number of forks, warmups, and iterations.

Where do these numbers come from? What are the defaults? Why add them at all?

donraab:
@motlin I left `AbstractJMHTestRunner` alone so these numbers would still be visible. If you are good with the changes, I will remove the class and update the PR.

donraab:
@motlin Are there any changes you would suggest for this PR?

motlin (approved):

Issue #1564, opened 2024-03-22
Interval types do not support reverse ranges for certain helper methods


donraab:
A few `Interval` helper methods are missing support for reverse ranges, which are ranges where the `from` value is greater than the `to` value, which would require the `by` value to be negative.

Classes: `Interval`, `IntInterval`, `LongInterval`

The following examples throw exceptions today:

1. `zeroTo`

```
Interval.zeroTo(-10);

// java.lang.IllegalArgumentException: Step by is incorrect for the range
```

This code should result in:  `Interval.fromToBy(0, -10, -1)`

2. `to`

```
Interval.from(10).to(-10);

// java.lang.IllegalArgumentException: Step by is incorrect for the range
```

This code should result in:  `Interval.fromToBy(10, -10, -1)`

3. `oneTo`

```
Interval.oneTo(-10); 

// java.lang.IllegalArgumentException: Only positive ranges allowed using oneToBy
```

This exception looks like it was intentional but I think if we fix the other two, this one should be changed to be consistent. This code should result in: `Interval.fromToBy(1, -10, -1)`

Failing tests should be written to cover all missing scenarios first. I was not exhaustive in my search, so there may be other helper methods that do not support reverse ranges.

airtyon:
hi @donraab,
may I give this one a try?

donraab:
Hi @airtyon, yes, thank you! I will assign it to you.

PR #1570, opened 2024-04-03
reverse interval ranges

airtyon:
Closes: #1564 

Hi @donraab,

In this PR, I tried to keep the changes as much consistent as possible.

Upon consideration, I think there might be a different way to approach this.

In my opinion we might consider allowing the direction of the interval to be determined by difference of `from - to`.

1. We allow `step` to have different values and sign
1.a. Pro: Not invasive
1.b. Con: The API doesn't look nice. i.e. fromToBy(0, 10, -1)

2. We allow `step` to only be positive
2.a. Pro: This would reflect more a "real life" situation, where having negative steps is not allowed. i.e. a car is moving in some direction with this many km/h. if the car reverses, the direction changes but the speed is still positive.
2.b. Con: Somehow invasive.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/Interval.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/Interval.java
@@ -100,7 +100,8 @@ public final class Interval
      */
     public Interval to(int newTo)
     {
-        return Interval.fromToBy(this.from, newTo, this.step);
+        int adjustedStep = IntervalUtils.calculateAdjustedStep(this.from, newTo, this.step);
+        return Interval.fromToBy(this.from, newTo, adjustedStep);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


donraab (approved):
LGTM. Thanks for the contribution @airtyon !

donraab:
Hi @airtyon,

I like your suggestions (especially number one), but I would be hesitant to make any changes to `Interval` because it is very old and has a lot of usages. I will think a bit on your first suggestion. I can't think of any reason not to do it off hand, but need to stare at the code a bit more. I'd also like to add some more tests for the three `Interval` classes first. For small Intervals I think we should be testing more than size alone.

Thanks!

> 
> In my opinion we might consider allowing the direction of the interval to be determined by difference of `from - to`.
> 
> 1. We allow `step` to have different values and sign
>    1.a. Pro: Not invasive
>    1.b. Con: The API doesn't look nice. i.e. fromToBy(0, 10, -1)
> 2. We allow `step` to only be positive
>    2.a. Pro: This would reflect more a "real life" situation, where having negative steps is not allowed. i.e. a car is moving in some direction with this many km/h. if the car reverses, the direction changes but the speed is still positive.
>    2.b. Con: Somehow invasive.

PR #1571, opened 2024-04-08
Add withOccurrences and withoutOccurrences to MutableBag.

donraab:


eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MultiReaderBag.java
```diff
@@ -41,6 +41,20 @@ default MultiReaderBag<T> without(T element)
         return this;
     }
 
+    @Override
+    default MultiReaderBag<T> withOccurrences(T element, int occurrences)
+    {
+        this.addOccurrences(element, occurrences);
```
nikhilnanivadekar, line 47: Just to confirm, we use this pattern for Multi reader collections correct? The return this is why I am asking

unit-tests/src/test/java/org/eclipse/collections/impl/bag/mutable/MutableBagTestCase.java
```diff
@@ -278,12 +278,31 @@ public void addOccurrences()
         MutableBagTestCase.assertBagsEqual(HashBag.newBagWith("1", "2", "2", "2", "2", "2", "2", "3"), bag);
     }
 
+    @Test
```
nikhilnanivadekar, line 281: Do we need to assert the instance is the same as well?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/bag/MultiReaderBag.java
donraab: Yes, the `addOccurrences` method will take a lock, so I didn't introduce the lock in this override.

nikhilnanivadekar (approved):

unit-tests/src/test/java/org/eclipse/collections/impl/bag/mutable/MutableBagTestCase.java
donraab: Yes, that is a good thing to assert as well. I can add it.

Issue #466, opened 2018-02-16
Update license from deprecated EPL 1.0 to EPL 2.0


andreikoval:
As per https://www.eclipse.org/legal/epl-2.0/faq.php#h.60mjudroo8e5

Bananeweizen:
👍 
If you intend to get on a simultaneous release train in the future, this will be a requirement.

donraab:
@waynebeaton Can a project be dual-licensed under EPL 2.0 and EDL 1.0? EPL 2.0 with Secondary License Clause will not work for us. I did not see this particular option addressed in the FAQ on licensing. https://www.eclipse.org/legal/epl-2.0/faq.php#epl-general

waynebeaton:
Updating to the EPL-2.0 would result in the same sort of dual licensing that you're doing with the EPL-1.0 now, just with the newer version of the license.

> EPL 2.0 with Secondary License Clause will not work for us. 

Dual licensing (which is what you're doing) is different from secondary licensing. Secondary licensing only applies in cases where you want to link EPL content with GPL-2.0+ content.

> _“Secondary License” means either the GNU General Public License, Version 2.0, or any later versions of that license, including any exceptions or additional permissions as identified by the initial Contributor._ 

Dual licensing with the EDL (a BSD-3-Clause licence) makes this sort of linking work without requiring the secondary licence.

donraab:
> Dual licensing with the EDL (a BSD-3-Clause licence) makes this sort of linking work without requiring the secondary licence.

I'll take this as a yes. What do we need to do to update to dual licensed under EPL 2.0 and EDL 1.0?

waynebeaton:
I should have led with yes.

Update your file headers and license files, and send a note to the EMO asking them to update their records.

FWIW, the EPL has a provision that grants you permission to update to a later version of the licence without needing to get approval from anybody. It's good form, however, to make sure that the project committers are aware of what you're doing and get a chance to ask questions.

donraab:
> It's good form, however, to make sure that the project committers are aware of what you're doing and get a chance to ask questions.

@nikhilnanivadekar @motlin @mohrezaei @prathasirisha @itohiro73

EC Committers, any questions or thoughts please add. This will be a somewhat cumbersome upgrade since we would need to update all file headers.

waynebeaton:
I notice that at least a few of your file headers do not have an `SPDX-License-Identifier`; this would be a good opportunity to add them.

[...]

PR #1572, opened 2024-04-12
Add NOTICE.md file.

donraab:


waynebeaton:
LGTM.

Please address #466

donraab:
> LGTM.
> 
> Please address #466

Thanks @waynebeaton ! Can a project be dual-licensed under EPL 2.0 and EDL 1.0? EPL 2.0 with Secondary License Clause will not work for us.

motlin:
I'm curious to learn more. I saw the conversation where you explained that this project is dual licensed. Is the NOTICE.md file meant to reduce that confusion in the future? Does it follow some template?

waynebeaton:
The current copyright and licence headers already describe the project as dual licensed. By including this in the NOTICE file, it just makes that more explicit. Notice files tend to also list things like the licences of third party dependencies required by the binary (so that adopters can have easy access to this information), but -- other than JUnit -- I'm pretty sure that Eclipse Collections doesn't have any.

We do provide a template for NOTICE files in the [handbook](https://www.eclipse.org/projects/handbook/#legaldoc-notice).

The inclusion of `SPDX-License-Identifier: EPL-1.0 OR BSD-3 Clause` is the best disambiguator.

donraab:
> We do provide a template for NOTICE files in the [handbook](https://www.eclipse.org/projects/handbook/#legaldoc-notice).

That's what I used. Thank you!

donraab:
> Is the NOTICE.md file meant to reduce that confusion in the future?

We do not currently have SPDX License Identifiers in any of our file headers (that I am aware of) so I wanted to have at least one file where we started doing that, assuming that bot scanners might be looking for this now.

NOTICE.md
```diff
@@ -0,0 +1,33 @@
...
+All content is the property of the respective authors or their employers. For more information regarding authorship of content, please consult the listed source code repository logs.
+
+## Declared Project Licenses
+
+This project is made available under the terms of the Eclipse Public License (EPL) v. 1.0 or Eclipse Distribution License (EDL) v. 1.0 which are available at the following locations.
+
+* EPL 1.0: https://www.eclipse.org/legal/epl-v10.html
+* EDL 1.0: https://www.eclipse.org/org/documents/edl-v10.html
+
+SPDX-License-Identifier: EPL-1.0 OR BSD-3 Clause
```
pan3793, line 23: a question for education: why say BSD-3 Clause instead of EDL-1.0?

donraab: > a question for education: why say BSD-3 Clause instead of EDL-1.0?

I had the same question. I found the answer in the EDL 1.0 documents link.

https://www.eclipse.org/org/documents/edl-v10.php

donraab: @waynebeaton Now that I look at it the SPDX License Identifier in the list from SPDX site, the identifier on the EDL 1.0 page is missing a dash. It should be `BSD-3-Clause`. If this is correct, then it should be fixed on the EDL 1.0 page as well.

waynebeaton: > a question for education: why say BSD-3 Clause instead of EDL-1.0?

It's a great question. I did ask the SPDX team to add the EDL to their list and their response was (correctly) that the EDL is really just the `BSD-3-Clause`, which is a template for a licence. Basically everybody who uses the `BSD-3-Clause` modifies it and having every single one of them have a separate SPDX code doesn't actually make anything easier.

SPDX has a means of specifying licences that aren't on their list. We could use `LicenseRef-EDL-1.0` as the SPDX code, but that would just add confusion as we'd have to explain what it means. We don't have this problem when we use `BSD-3-Clause`.

> the identifier on the EDL 1.0 page is missing a dash

Nice catch. There should never be spaces in the SPDX code. I'll fix that.

motlin (approved):

PR #1575, opened 2024-05-11
Add allSatisfyKeyValue method to Object/Primitive Maps to optimize HashBag equals method.

donraab:
I wrote a JMH Benchmark to compare the performance of the equals methods in HashBag, HashMultiset (Guava) and HashMap (Java). I was surprised to find that HashBag equals was slower than HashMap. I added a new method named shortCircuitKeyValue to Object/Primitive Maps and based the equals method in AbstractHashBag on using this method as a manual way of implementing allSatisfyKeyValue. I had originally added any/all/noneSatisfyKeyValue as well, but backed out those additions as a potential later change.

The following are the results of the JMH tests before and after changes.

Original Code

```
Benchmark                      Mode  Cnt      Score      Error   Units
BagEqualsTest.bagEquals       thrpt   20  19770.401 ±  139.757  ops/ms
BagEqualsTest.mapEquals       thrpt   20  23120.905 ±  246.497  ops/ms
BagEqualsTest.multisetEquals  thrpt   20  12867.562 ± 1087.486  ops/ms
```

Using `AbstractHashBag` `instanceof` and direct items map `equals`

```
Benchmark                      Mode  Cnt      Score     Error   Units
BagEqualsTest.bagEquals       thrpt   20  19328.068 ± 158.372  ops/ms
BagEqualsTest.mapEquals       thrpt   20  22618.070 ± 362.793  ops/ms
BagEqualsTest.multisetEquals  thrpt   20  13990.621 ±  29.081  ops/ms
```

Using `allSatisfyKeyValue`

```
Benchmark                      Mode  Cnt      Score     Error   Units
BagEqualsTest.bagEquals       thrpt   20  24719.542 ±  17.943  ops/ms
BagEqualsTest.mapEquals       thrpt   20  23109.328 ± 235.381  ops/ms
BagEqualsTest.multisetEquals  thrpt   20  14542.701 ±  66.522  ops/ms
```
Additionally, I refactored the value based any/all/noneSatisfy methods on the Object/Primitive Maps to use shortCircuitKeyValue. I tried using Object/Primitive Map equals to implement the HashBag equals, but found that was even slower. I updated the equals method slightly to remove some extra index lookups and multiple conversions. This sped the code up by around 10% when I measured it, but it was still slower than the shortCircuitKeyValue solution. 

Update: I rolled back the `shortCircuitKeyValue` implementation and renamed to `allSatisfyKeyValue`.

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/map/objectPrimitiveMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/map/objectPrimitiveMap.stg
@@ -95,6 +95,21 @@ public interface Object<name>Map\<K> extends <name>Iterable
     <if(!primitive.booleanPrimitive)><(flipUniqueValues.(name))(name)><endif>
 
     /**
+     * @since 12.0
+     */
+    default boolean allSatisfyKeyValue(Object<name>Predicate\<K> predicate)
+    {
+        for (Object<name>Pair\<K> pair : this.keyValuesView())
+        {
+            if (!predicate.accept(pair.getOne(), pair.getTwo()))
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    /**
      * Implements the injectInto pattern with each key/value pair of the map.
      * @param value to be injected into the map
      * @param function to apply to the injected value and key/value pairs
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/AbstractHashBag.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/AbstractHashBag.java
@@ -67,8 +67,7 @@ public abstract class AbstractHashBag<T> extends AbstractMutableBag<T>
         {
             return false;
         }
-
-        return this.items.keyValuesView().allSatisfy(each -> bag.occurrencesOf(each.getOne()) == each.getTwo());
+        return this.items.allSatisfyKeyValue((key, count) -> bag.occurrencesOf(key) == count);
     }
 
     @Override
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutableObjectPrimitiveHashMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutableObjectPrimitiveHashMap.stg
@@ -298,6 +298,15 @@ final class ImmutableObject<name>HashMap\<K> extends AbstractImmutableObject<nam
         this.delegate.forEachKeyValue(object<name>Procedure);
     }
 
+    /**
+     * @since 12.0
+     */
+    @Override
+    public boolean allSatisfyKeyValue(Object<name>Predicate\<K> predicate)
+    {
+        return this.delegate.allSatisfyKeyValue(predicate);
+    }
+
     @Override
     public ImmutableObject<name>Map\<K> select(Object<name>Predicate\<? super K> object<name>Predicate)
     {
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutableObjectPrimitiveSingletonMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/immutable/immutableObjectPrimitiveSingletonMap.stg
@@ -1,4 +1,4 @@
-import "copyright.stg"
+import "copyrightAndOthers.stg"
 import "primitiveEquals.stg"
 import "primitiveHashCode.stg"
 import "primitiveLiteral.stg"
@@ -12,7 +12,7 @@ class(primitive) ::= <<
 >>
 
 body(type, name) ::= <<
-<copyright()>
+<copyrightAndOthers()>
 
 package org.eclipse.collections.impl.map.immutable.primitive;
 
@@ -326,6 +326,19 @@ final class ImmutableObject<name>SingletonMap\<K> extends AbstractImmutableObjec
     public void forEachKeyValue(Object<name>Procedure\<? super K> object<name>Procedure)
     {
         object<name>Procedure.value(this.key1, this.value1);
+    }
+
+    /**
+     * @since 12.0
+     */
+    @Override
+    public boolean allSatisfyKeyValue(Object<name>Predicate\<K> predicate)
+    {
+        if (!predicate.accept(this.key1, this.value1))
+        {
+            return false;
+        }
+        return true;
     }
 
     @Override
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/objectPrimitiveHashMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/objectPrimitiveHashMap.stg
@@ -216,9 +216,14 @@ public class Object<name>HashMap\<K> implements MutableObject<name>Map\<K>, Exte
 
         for (int i = 0; i \< this.keys.length; i++)
         {
-            if (isNonSentinel(this.keys[i]) && (!other.containsKey(this.toNonSentinel(this.keys[i])) || <(notEquals.(type))("this.values[i]", "other.getOrThrow(this.toNonSentinel(this.keys[i]))")>))
+            Object key = this.keys[i];
+            if (isNonSentinel(key))
             {
-                return false;
+                K nonSentinel = this.toNonSentinel(key);
+                if (!other.containsKey(nonSentinel) || <(notEquals.(type))("this.values[i]", "other.getOrThrow(nonSentinel)")>)
+                {
+                    return false;
+                }
             }
         }
         return true;
@@ -863,6 +868,26 @@ public class Object<name>HashMap\<K> implements MutableObject<name>Map\<K>, Exte
             }
         }
         return count;
+    }
+
+    /**
+     * @since 12.0
+     */
+    @Override
+    public boolean allSatisfyKeyValue(Object<name>Predicate\<K> predicate)
+    {
+        for (int i = 0; i \< this.keys.length; i++)
+        {
+            Object key = this.keys[i];
+            if (isNonSentinel(key))
+            {
+                if (!predicate.accept(this.toNonSentinel(key), this.values[i]))
+                {
+                    return false;
+                }
+            }
+        }
+        return true;
     }
 
     @Override
--- a/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/objectPrimitiveHashMapWithHashingStrategy.stg
+++ b/eclipse-collections-code-generator/src/main/resources/impl/map/mutable/objectPrimitiveHashMapWithHashingStrategy.stg
@@ -788,6 +788,26 @@ public class Object<name>HashMapWithHashingStrategy\<K> implements MutableObject
         }
     }
 
+    /**
+     * @since 12.0
+     */
+    @Override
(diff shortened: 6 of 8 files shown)


eclipse-collections-code-generator/src/main/resources/api/map/objectPrimitiveMap.stg
```diff
@@ -94,6 +94,31 @@ public interface Object<name>Map\<K> extends <name>Iterable
...
+     *
+     * anySatisfy  = this.shortCircuitKeyValue(predicate, true, true, false);
+     * allSatisfy  = this.shortCircuitKeyValue(predicate, false, false, true);
+     * noneSatisfy = this.shortCircuitKeyValue(predicate, true, false, true);
+     *
+     * @since 12.0
+     */
+    default boolean shortCircuitKeyValue(
+            Object<name>Predicate\<K> predicate,
+            boolean shortCircuitCondition,
```
mohrezaei, line 108: at first blush, this is a rather odd function signature. the inclusion of `shortCircuitCondition` goes against the grain of having a `Predicate` -- I don't think there is anywhere else where we compare the output of a predicate with a parameter. Normally, we just expect the predicate to return true for whatever condition the function requires, which necessitates the use of `Predicates.not` on occasion.

I imagine removing this parameter would slow down `allSatisfy`, but it would be worth measuring that.

eclipse-collections-code-generator/src/main/resources/api/map/objectPrimitiveMap.stg
```diff
@@ -94,6 +94,31 @@ public interface Object<name>Map\<K> extends <name>Iterable
...
+            boolean shortCircuitReturn,
```
mohrezaei, line 109: Can you imagine a case where these last two parameters are not the opposite of each other? They could be eliminated entirely (default to `true, false`):
```java
anySatisfy  = this.shortCircuitKeyValue(predicate);
allSatisfy  = !this.shortCircuitKeyValue(Predicates.not(predicate));
noneSatisfy = !this.shortCircuitKeyValue(predicate);
```

mohrezaei (changes requested):

eclipse-collections-code-generator/src/main/resources/api/map/objectPrimitiveMap.stg
mohrezaei: looking at the above, it strikes me that this new function is just `anySatisfy` :smile:

motlin: This came out pretty similar to UnifiedMap.shortCircuit() and UnifiedSet.shortCircuit() but with different parameter names. Was that on purpose? Any advice that applies here (like not needing two redundant parameters) probably also applies there.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/AbstractHashBag.java
```diff
@@ -67,8 +67,7 @@ public boolean equals(Object other)
         {
             return false;
         }
-
-        return this.items.keyValuesView().allSatisfy(each -> bag.occurrencesOf(each.getOne()) == each.getTwo());
+        return this.items.shortCircuitKeyValue((k, v) -> bag.occurrencesOf(k) == v, false, false, true);
```
motlin, line 70: Did you performance test an implementation like this?

```java
return this.items.keysView().allSatisfy(each -> bag.occurrencesOf(each) == this.items.get(each));
```

I could see it being slower because the extra `this.items.get` might add more time than creating the Pair wrappers wastes, but I'm curious.

eclipse-collections-code-generator/src/main/resources/api/map/objectPrimitiveMap.stg
```diff
@@ -94,6 +94,31 @@ public interface Object<name>Map\<K> extends <name>Iterable
...
+    /**
+     * This method can be used to implement anySatisfy, allSatisfy, and noneSatisfy.
+     *
+     * anySatisfy  = this.shortCircuitKeyValue(predicate, true, true, false);
+     * allSatisfy  = this.shortCircuitKeyValue(predicate, false, false, true);
+     * noneSatisfy = this.shortCircuitKeyValue(predicate, true, false, true);
+     *
+     * @since 12.0
+     */
+    default boolean shortCircuitKeyValue(
```
motlin, line 106: I think it's odd to have this method as `public` and as `default`.

`public` because it's an implementation detail for performance.

`default` because the implementation at this level creates wrapper objects and isn't there for performance.

UnifiedMap/UnifiedSet's shortCircuit() methods are private. But they take values, not key-value pairs.

I've had this idea for a MapView for a long time that I never filed as an issue. It might make sense to add at this point. I'll show what it would look like in AbstractHashBag if it existed in another comment.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/mutable/AbstractHashBag.java
motlin: If we implemented lazy map views, the implementation here would become:

```java
return this.items.asLazyMap().allSatisfy((key, value) -> bag.occurrencesOf(key) == value);
```

motlin: Did you performance test an implementation like this?

```java

        if (bag instanceof AbstractHashBag)
        {
            AbstractHashBag<?> that = (AbstractHashBag<?>) other;
            return this.items.equals(that.items);
        }

        return bag.equals(this);
```

donraab: > Did you performance test an implementation like this?
> 
> ```java
> return this.items.keysView().allSatisfy(each -> bag.occurrencesOf(each) == this.items.get(each));
> ```
> 
> I could see it being slower because the extra `this.items.get` might add more time than creating the Pair wrappers wastes, but I'm curious.

Yes, it was significantly slower.

donraab: > Did you performance test an implementation like this?
> 
> ```java
>         if (bag instanceof AbstractHashBag)
>         {
>             AbstractHashBag<?> that = (AbstractHashBag<?>) other;
>             return this.items.equals(that.items);
>         }
> 
>         return bag.equals(this);
> ```

Yes, this was slower than the current implementation, most likely because it has to do `containsKey` and they `getOrThrow`.

eclipse-collections-code-generator/src/main/resources/api/map/objectPrimitiveMap.stg
donraab: I probably should have marked this PR as draft. I originally had all three implementations on the api (`any`, `all`, `none`) delegating to the shortCircuit method which I borrowed from `UnifiedMap`. I removed all of them from the PR to make it easier to review. The only one that is needed right now is allSatisfy, which I could add back. I thought it would be better to have a separate PR for all three methods with tests, and only including what was necessary for the performance improvement in this one. I think I'll just go with `allSatisfyKeyValue` on the API and get rid of the shortCircuit method. We can always add `any` and `none` later.

eclipse-collections-code-generator/src/main/resources/api/map/objectPrimitiveMap.stg
donraab: > This came out pretty similar to UnifiedMap.shortCircuit() and UnifiedSet.shortCircuit() but with different parameter names. Was that on purpose? Any advice that applies here (like not needing two redundant parameters) probably also applies there.

I fiddled with the parameter names, so yes it was on purpose.

eclipse-collections-code-generator/src/main/resources/api/map/objectPrimitiveMap.stg
donraab: Agreed. I didn't like having this method on the API. I went back and forth with having any, all, none and whether to make this private on the interface, which works but causes a lot more generated code to review.

In the end, for the initial PR, I wanted to only show that there was a performance improvement with the minimal code necessary so that a change here was worthwhile.

donraab:
@mohrezaei @motlin I'm going to make a bunch of changes and submit this as `allSatisfyKeyValue` and get rid of the short circuit method. I will rerun the benchmarks for the previous version and the map equals variety and post the results here so it is clearer that this change is worthwhile.

donraab:
@mohrezaei @motlin These are the benchmarks for comparison.

Original Code

```
Benchmark                      Mode  Cnt      Score      Error   Units
BagEqualsTest.bagEquals       thrpt   20  19770.401 ±  139.757  ops/ms
BagEqualsTest.mapEquals       thrpt   20  23120.905 ±  246.497  ops/ms
BagEqualsTest.multisetEquals  thrpt   20  12867.562 ± 1087.486  ops/ms
```

Using `AbstractHashBag` `instanceof` and direct items map `equals`

```
Benchmark                      Mode  Cnt      Score     Error   Units
BagEqualsTest.bagEquals       thrpt   20  19328.068 ± 158.372  ops/ms
BagEqualsTest.mapEquals       thrpt   20  22618.070 ± 362.793  ops/ms
BagEqualsTest.multisetEquals  thrpt   20  13990.621 ±  29.081  ops/ms
```

Using `allSatisfyKeyValue`

```
Benchmark                      Mode  Cnt      Score     Error   Units
BagEqualsTest.bagEquals       thrpt   20  24719.542 ±  17.943  ops/ms
BagEqualsTest.mapEquals       thrpt   20  23109.328 ± 235.381  ops/ms
BagEqualsTest.multisetEquals  thrpt   20  14542.701 ±  66.522  ops/ms
```

I'll be submitting an update to the PR in the next hour with the more simplified changes. 

Thank you for all the feedback and comments so far!

donraab:
@mohrezaei @motlin If these changes look good and you both approve, I will test out the same changes to primitive Bags and primitive/primitive Maps. The performance there looks worse than Object Bags, but hard to tell if the benefit will be comparable or better until I try it.

Currently on IntBag:

```
Benchmark                    Mode  Cnt      Score     Error   Units
IntBagEqualsTest.bagEquals  thrpt   20  16872.887 ±  46.903  ops/ms
IntBagEqualsTest.mapEquals  thrpt   20  15838.773 ± 103.729  ops/ms
```

mohrezaei (approved):

donraab:
Thanks @mohrezaei ! @motlin let me know if you are good with the changes as well and I will merge.

I finished the testing/benchmarking the change for primitive Bags and primitive/primitive Maps and the results are surprising.

Before changes:

```
Benchmark                    Mode  Cnt      Score     Error   Units
IntBagEqualsTest.bagEquals  thrpt   20  16872.887 ±  46.903  ops/ms
IntBagEqualsTest.mapEquals  thrpt   20  15838.773 ± 103.729  ops/ms

```
After allSatisfyKeyValue:

```
Benchmark                    Mode  Cnt      Score     Error   Units
IntBagEqualsTest.bagEquals  thrpt   20  51903.988 ± 237.046  ops/ms
IntBagEqualsTest.mapEquals  thrpt   20  15964.766 ± 172.302  ops/ms
```

PR #1578, opened 2024-05-15
Junit5 Migration Part 1 - #1576

Desislav-Petrov:
Part of the Junit5 migration preparation.

- Moves imports to static to reduce change size for future PRs
- Fixes a bug in an iterator test explained in [here](https://github.com/eclipse/eclipse-collections/pull/1577)
- Removes any junit4 syntax like @Test(expected = ) as it's not supported in junit5

PR #1579, opened 2024-05-17
Replace usages of @Test(expected) with Assert.assertThrows().

motlin:
Relates to #1578 and #1576. It turns out that just moving away from JUnit 4 syntax for expected exceptions is a fairly large change on its own. @Desislav-Petrov please take a look.


eclipse-collections-code-generator/src/main/resources/test/map/mutable/unmodifiablePrimitiveObjectMapKeySetTest.stg
```diff
@@ -42,6 +48,19 @@ public class Unmodifiable<name>ObjectMapKeySetTest extends Abstract<name>SetTest
...
+    public void <type>Iterator_throws_non_empty_collection()
+    {
+        Mutable<name>Collection collection = this.classUnderTest();
+        <name>Iterator iterator = collection.<type>Iterator();
+        while (iterator.hasNext())
+        {
+            iterator.next();
+        }
+        assertThrows(NoSuchElementException.class, () -> iterator.next());
+    }
```
motlin, line 62: These overrides don't make sense - I'll remove them. The test is for non-empty-collection, and this overrides it to only make assertions about empty collections.

eclipse-collections-code-generator/src/main/resources/test/list/immutable/abstractImmutablePrimitiveListTestCase.stg
```diff
@@ -80,16 +80,16 @@ public abstract class AbstractImmutable<name>ListTestCase extends AbstractImmuta
         }
     }
 
-    @Test(expected = IndexOutOfBoundsException.class)
+    @Test
     public void get_throws_index_greater_than_size()
     {
-        this.classUnderTest().get(this.classUnderTest().size());
+        Assert.assertThrows(IndexOutOfBoundsException.class, () -> this.classUnderTest().get(this.classUnderTest().size()));
```
donraab, line 86: Do you not want to make this a static import since you are changing this anyway?

donraab (changes requested):
One suggested change inline in regards to missing static import.

eclipse-collections-code-generator/src/main/resources/test/list/immutable/abstractImmutablePrimitiveListTestCase.stg
motlin: Oh sure, it was like that in the original but I'll change it.

eclipse-collections-code-generator/src/main/resources/test/collection/mutable/abstractMutablePrimitiveCollectionTestCase.stg
```diff
@@ -398,23 +397,6 @@ public abstract class AbstractMutable<name>CollectionTestCase extends Abstract<n
...
-        collection.add(<(literal.(type))("2")>);
-        collection.add(<(literal.(type))("3")>);
-        <name>Iterator iterator = collection.<type>Iterator();
-        while (iterator.hasNext())
-        {
-            iterator.next();
-        }
-        iterator.next();
-    }
-
```
motlin, line 417: this test is almost identical to the superclass, but it uses add() instead of a varargs creation method. Calling add() is the reason it doesn't work in unmodifiable tests which results in even more overrides.

donraab (approved):
LGTM

PR #1580, opened 2024-05-18
Replace usages of @Test(expected) with Assert.assertThrows().

motlin:
Relates to #1579 which still had one extra test override left.


donraab (approved):
I'm approving even though the text of the PR doesn't really match the work which is removing an unnecessary method I assume.

motlin:
I had meant to fold this into the previous commit but missed my chance, so I gave it the same commit message.

PR #1591, opened 2024-05-21
Specify the version of maven-wrapper-plugin.

motlin:
We're already running this plugin as part of continuous integration so the usual thing would be to lock down the version. Maven 4.x together with the enforcer plugin will require this, and I'm running maven 4 locally.


donraab (approved):
LGTM. Rebase and merge when you're ready.

PR #1599, opened 2024-05-22
Sync Checkstyle configuration with upstream examples.

motlin:
Our Checkstyle configuration borrows a lot of ideas from the configuration that the Checkstyle project runs on its own code.

https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle-checks.xml

They've cleaned up some configuration and turned on more rules that pass on our project, so this PR resyncs our configuration with "upstream"


donraab (approved):
LGTM

PR #1600, opened 2024-05-23
Fix generics in map factories.

motlin:


donraab (approved):
LGTM. Another good catch on some discrepancies. Thanks!

PR #1601, opened 2024-05-23
Change math function factories to use Math.*Exact methods.

motlin:


donraab (approved):
LGTM

mohrezaei:
This is a significant change to behavior. What is the reason this is being done? Nobody is going around replacing all the `+` signs in their code to `Math.addExact`. Shouldn't there be variant functions for these, just as there are in core java? Has the performance difference been measured? What about usecases that require wrapping behavior (e.g. simple hashing)?

unit-tests-java8/src/test/java/org/eclipse/collections/test/bag/mutable/sorted/MutableBagIterableTestCase.java
```diff
@@ -37,6 +38,10 @@ default void MutableBagIterable_addOccurrences()
         assertThrows(
                 IllegalArgumentException.class,
                 () -> mutableBag.addOccurrences(4, -1));
+        assertEquals(Bags.immutable.with(1, 1, 1, 2, 2, 3, 3, 3, 4, 4, 4, 4), mutableBag);
+
+        Assert.assertEquals(-2147483647, mutableBag.addOccurrences(3, Integer.MAX_VALUE - 1));
+        Assert.assertEquals(-2147483638, mutableBag.size());
```
motlin: Here I added assertions to show that addOccurrences and size already can overflows.

unit-tests-java8/src/test/java/org/eclipse/collections/test/bag/mutable/sorted/MutableBagIterableTestCase.java
```diff
@@ -37,6 +38,17 @@ default void MutableBagIterable_addOccurrences()
         assertThrows(
                 IllegalArgumentException.class,
                 () -> mutableBag.addOccurrences(4, -1));
+        assertEquals(Bags.immutable.with(1, 1, 1, 2, 2, 3, 3, 3, 4, 4, 4, 4), mutableBag);
+        assertEquals(Bags.immutable.with(1, 1, 1, 2, 2, 3, 3, 3, 4, 4, 4, 4), mutableBag);
+
+        assertThrows(
+                ArithmeticException.class,
+                () -> mutableBag.addOccurrences(3, Integer.MAX_VALUE - 1));
+        assertEquals(Bags.immutable.with(1, 1, 1, 2, 2, 3, 3, 3, 4, 4, 4, 4), mutableBag);
```
motlin, line 47: With the changes here, we can no longer overflow the number of occurrences for a single key.

unit-tests-java8/src/test/java/org/eclipse/collections/test/bag/mutable/sorted/MutableBagIterableTestCase.java
```diff
@@ -37,6 +38,17 @@ default void MutableBagIterable_addOccurrences()
...
+
+        mutableBag.addOccurrences(5, Integer.MAX_VALUE - 10);
+        mutableBag.addOccurrences(6, Integer.MAX_VALUE - 10);
+        Assert.assertEquals(-10, mutableBag.size());
```
motlin, line 51: But we can still overflow size with big numbers of occurrences at several keys.

motlin (commented):
Added tests and context to show what I'm up to here. I was looking at integer overflows in bags.

unit-tests-java8/src/test/java/org/eclipse/collections/test/bag/mutable/sorted/MutableBagIterableTestCase.java
mohrezaei: `MutableBag.addOccurances` does not specify in its javadoc that it will throw `ArithmeticException`. And that probably applies to everywhere else those functions are involved.

unit-tests-java8/src/test/java/org/eclipse/collections/test/bag/mutable/sorted/MutableBagIterableTestCase.java
mohrezaei: Java collection's `size` function has been known to have integer limits for a very long time. And for almost all applications, it doesn't matter. If some application needs a specialty collection with larger sizes, they should pick a suitable implementation, no?

unit-tests-java8/src/test/java/org/eclipse/collections/test/bag/mutable/sorted/MutableBagIterableTestCase.java
motlin: `strictAddOccurrences` is fine, then the next question would be whether to call it internally. For example, in HashBag.collect():

```java
    @Override
    public <V, R extends Collection<V>> R collect(Function<? super T, ? extends V> function, R target)
    {
        if (target instanceof MutableBagIterable<?>)
        {
            MutableBagIterable<V> targetBag = (MutableBagIterable<V>) target;

            this.forEachWithOccurrences((each, occurrences) -> targetBag.addOccurrences(function.valueOf(each), occurrences));
        }
        else
        {
            // ...
        }
        return target;
    }
```

Or add strictCollect()? Or just stop here?

mohrezaei: An alternative approach is to have a `StrictMutableBag` wrapper (like synchronized). Then the question of "where to stop" goes away.

motlin: > It's hard to figure out where to stop, as I don't understand why it was started in the first place. Is there a real world use case that triggered this?

So, I don't remember. I wrote the first version of this commit 6 years ago. I was talking to @donraab about having so much work in progress that I had almost given up on contributing and decided to try to land some of the smaller changes.

I think that the overflow has always bothered me and so I've always avoided the `addOccurrences` method. At my work, there are no usages of Bag in the company. In my side projects I've used `addOccurrences` [exactly once](https://github.com/motlin/klass/blob/edbd1efa8f68f91d80c91f88d877bb7adf736ef4/klass-reladomo/klass-reladomo-sample-data/src/main/java/cool/klass/reladomo/sample/data/AbstractDataTypePropertyVisitor.java#L196), and it was with the occurrences hard-coded to `1`. I just used the method so that I could increment and get the new count in a single method call.

> An alternative approach is to have a StrictMutableBag wrapper (like synchronized). Then the question of "where to stop" goes away.

> It occurred to me that implementing that wrapper would be easier if MutableBag has a strictAddOccurances method, so maybe a combination of the two works best.

I like these ideas and will try them out. I'll move this code review to a Draft, I don't plan on landing it in a form anything like it is now.

donraab: Thanks for jumping in on this @mohrezaei. I like your idea of having a `StrictMutableBag` if this functionality is desired. Another idea I have been pondering for a while is having a `LargeBag` which uses `long` instead of `int` for the counts. I haven't needed it, so haven't bothered. The backing Map is still limited by `int` for it's size. Java `Stream` counts into `Map<K, Long>`, which allows for the larger counts, while still suffering from the `int` backing size.

PR #1602, opened 2024-05-24
Fix return types of aggregateBy overrides.

motlin:


donraab (approved):
LGTM. It will be good to get these minor discrepancies sorted out. Thanks!

PR #1605, opened 2024-05-26
Add dependency on junit-jupiter-api.

motlin:
@Desislav-Petrov another small step on the way to JUnit 5.

### DIFF
--- a/acceptance-tests/pom.xml
+++ b/acceptance-tests/pom.xml
@@ -67,6 +67,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
             <scope>test</scope>
@@ -115,6 +121,9 @@
                                 </ignoredUnusedDeclaredDependency>
                                 <ignoredUnusedDeclaredDependency>
                                     org.slf4j:slf4j-simple:jar
+                                </ignoredUnusedDeclaredDependency>
+                                <ignoredUnusedDeclaredDependency>
+                                    org.junit.jupiter:junit-jupiter-api:jar
                                 </ignoredUnusedDeclaredDependency>
                             </ignoredUnusedDeclaredDependencies>
                         </configuration>
--- a/checkstyle-configuration.xml
+++ b/checkstyle-configuration.xml
@@ -392,8 +392,15 @@
         <module name="CovariantEquals" />
 
         <module name="AvoidStaticImport">
-            <property name="excludes"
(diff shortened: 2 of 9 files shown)


acceptance-tests/pom.xml
```diff
@@ -116,6 +122,9 @@
                                 <ignoredUnusedDeclaredDependency>
                                     org.slf4j:slf4j-simple:jar
                                 </ignoredUnusedDeclaredDependency>
+                                <ignoredUnusedDeclaredDependency>
+                                    org.junit.jupiter:junit-jupiter-api:jar
+                                </ignoredUnusedDeclaredDependency>
```
motlin, line 127: Temporarily allowing this dependency to be unused.

checkstyle-configuration.xml
```diff
@@ -392,8 +392,15 @@
...
-                value="org.junit.Assert.*,org.junit.Assume.*,org.hamcrest.CoreMatchers.*,org.hamcrest.Matchers.*,org.hamcrest.MatcherAssert.*,org.eclipse.collections.impl.test.Verify.*,org.eclipse.collections.impl.factory.Iterables.*,org.eclipse.collections.test.IterableTestCase.*" />
+            <property name="excludes" value="org.hamcrest.MatcherAssert.*" />
+            <property name="excludes" value="org.hamcrest.Matchers.*" />
+            <property name="excludes" value="org.junit.Assert.*" />
+            <property name="excludes" value="org.junit.Assume.*" />
+            <property name="excludes" value="org.junit.jupiter.api.Assertions.*" />
+            <property name="excludes" value="org.assertj.core.api.Assertions.*" />
+            <property name="excludes" value="org.eclipse.collections.impl.test.Verify.*" />
+            <property name="excludes" value="org.eclipse.collections.impl.factory.Iterables.*" />
+            <property name="excludes" value="org.eclipse.collections.test.IterableTestCase.*" />
```
motlin, line 403: Using the "new" syntax with one exclusion per line, and allowing static imports from JUnit 5 and AssertJ. I'm not sure yet if we'll use AssertJ, but if we do, static imports are ok for its assertThat method.

eclipse-collections-testutils/pom.xml
```diff
@@ -56,6 +56,12 @@
             <scope>compile</scope>
         </dependency>
 
+        <dependency>
```
Desislav-Petrov, line 59: Is there any reason why you aren't moving the junit4 to to jupiter-vintage so that the dependency on junit4 is removed altogether? Doing this required no code changes afair from my PR.

motlin: The reason that declaring a dependency on junit-vintage-engine works for you is that it depends on junit 4 and pulls it in as a transitive dependency.

https://mvnrepository.com/artifact/org.junit.vintage/junit-vintage-engine/5.10.2

This engine is used to launch JUnit 4 code, and should be used as a runtime dependency of the surefire plugin.

https://github.com/eclipse/eclipse-collections/blob/master/pom.xml#L480-L496

donraab (approved):
LGTM

PR #1606, opened 2024-05-26
Fix IntelliJ code style settings.

motlin:


.idea/codeStyles/Project.xml
```diff
@@ -179,6 +179,8 @@
       <option name="FOR_BRACE_FORCE" value="3" />
     </codeStyleSettings>
     <codeStyleSettings language="XML">
+      <option name="LINE_COMMENT_AT_FIRST_COLUMN" value="false" />
```
donraab, line 182: What is the effect of these two settings?

motlin: When you press the keyboard shortcut to comment stuff out, do the comment markers go at the current indent level or in column 0? For some reason the default is column 0 in both Java and XML. We already reconfigured it for Java years ago, but for some reason never configured it for XML. I noticed while working on the maven profiles.

donraab (approved):
LGTM

PR #1608, opened 2024-05-28
Set *.stg to linguist-linguist=StringTemplate in .gitattributes.

motlin:
This project's StringTemplate files are rendering in GitHub as plain text. Looking in the [Linguist supported languages](https://github.com/github-linguist/linguist/blob/master/lib/linguist/languages.yml), I see that [StringTemplate is present](https://github.com/github-linguist/linguist/blob/master/lib/linguist/languages.yml#L6867) but is set to match `*.st` files and not `*.stg` files. Following the [GitHub docs](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github) I set *.stg to linguist-linguist=StringTemplate in .gitattributes.

Before:

Screenshot 2024-05-28 at 9 57 24 AM

After:

Screenshot 2024-05-28 at 10 01 37 AM

PR #1609, opened 2024-05-28
Set .idea to linguist-generated=false in .gitattributes so that it shows up in code review diffs.

motlin:
In code reviews like #1606, the diffs are collapsed by default because files under .idea/ are considered generated sources (screenshot below).

Similar to the changes in #1608, I followed [GitHub's documentation](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github) on how to configure .gitattributes to control which files are considered generated.

Before:

Screenshot 2024-05-28 at 9 54 35 AM

After:

the diff will simply appear

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -25,3 +25,5 @@
 *.class         binary
 *.jar           binary
 *.zip           binary
+
+/.idea/** linguist-generated=false


donraab (approved):
LGTM

PR #1613, opened 2024-06-01
Consistently use `./mvnw --color=always` from GitHub workflows.

motlin:
Consistently run `mvn --non-recursive wrapper:wrapper -Dmaven=3.9.6` without extra flags to make sure we're using a known version, and then run `./mvnw --color=always` afterwards.


donraab (approved):
LGTM

PR #1616, opened 2024-06-02
Fix NullPointerException in IterableTestCase.assertEquals

motlin:
This one is a bit odd because it's an obvious bug fix inside test utility that I'm contributing without new tests. I ran into this while writing some Map tests that I'm about to put up in a separate PR, so I know that the fix helps.


donraab (approved):
LGTM

PR #1617, opened 2024-06-03
Fix inconsistencies in unmodifiable maps.

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
```diff
@@ -268,11 +268,7 @@ public V forcePut(K key, V value)
     @Override
     public void putAll(Map<? extends K, ? extends V> map)
     {
-        Set<? extends Map.Entry<? extends K, ? extends V>> entries = map.entrySet();
-        for (Map.Entry<? extends K, ? extends V> entry : entries)
-        {
-            this.put(entry.getKey(), entry.getValue());
-        }
+        map.forEach(this::put);
```
motlin, line 271: This change fixes assertions in HashBiMapNoIteratorTest, which now inherits more tests/coverage from MapTestCase.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/UnmodifiableBiMap.java
```diff
@@ -1133,25 +1133,48 @@ public void clear()
     @Override
     public V getIfAbsentPut(K key, Function0<? extends V> function)
     {
-        throw new UnsupportedOperationException("Cannot call getIfAbsentPut() on " + this.getClass().getSimpleName());
+        V result = this.get(key);
+        if (this.isAbsent(result, key))
+        {
+            throw new UnsupportedOperationException("Cannot mutate " + this.getClass().getSimpleName());
+        }
+        return result;
```
motlin, line 1141: Implementations like these are debatable, but are done consistently elsewhere like in UnmodifiableMutableMap. The trait-based testing approach is applying identical assertions to all the unmodifiable map wrappers and exposing differences like these.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/UnmodifiableBiMap.java
```diff
@@ -1219,4 +1242,52 @@ public MutableBiMap<K, V> withoutAllKeys(Iterable<? extends K> keys)
...
+    @Override
+    public void replaceAll(BiFunction<? super K, ? super V, ? extends V> function)
+    {
+        throw new UnsupportedOperationException("Cannot call replaceAll() on " + this.getClass().getSimpleName());
+    }
+
+    @Override
+    public V putIfAbsent(K key, V value)
+    {
+        throw new UnsupportedOperationException("Cannot call putIfAbsent() on " + this.getClass().getSimpleName());
```
motlin, line 1255: Note that other composite methods like putIfAbsent and the compute* methods all consistently throw, regardless of if the map would be mutated. These overrides make sure that the exception is consistently UnsupportedOperationException and that any lambdas passed into the compute* methods are never invoked.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/sorted/immutable/AbstractImmutableSortedMap.java
```diff
@@ -139,6 +140,60 @@ public void clear()
...
+    {
+        throw new UnsupportedOperationException("Cannot call putIfAbsent() on " + this.getClass().getSimpleName());
+    }
+
+    @Override
+    public void replaceAll(BiFunction<? super K, ? super V, ? extends V> function)
+    {
+        throw new UnsupportedOperationException("Cannot call replaceAll() on " + this.getClass().getSimpleName());
+    }
+
```
motlin, line 196: We still rely on abstract classes a lot, rather than an interface hierarchy with default methods, leading to lots of duplication. Cleaning that up is a bigger task.

eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/MapIterate.java
```diff
@@ -770,7 +770,7 @@ public static <K, V> void forEachKeyValue(Map<K, V> map, Procedure2<? super K, ?
     {
         if (map == null)
         {
-            throw new IllegalArgumentException("Cannot perform a forEachKeyValue on null");
+            throw new NullPointerException("Cannot perform a forEachKeyValue on null");
```
motlin, line 773: Most implementations of putAll() already throw NPE when the argument is null, but a few delegate to this method and throw IllegalArgumentException.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/MapIterableTestCase.java
```diff
@@ -34,7 +35,7 @@
 import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 
-public interface MapIterableTestCase extends RichIterableWithDuplicatesTestCase
+public interface MapIterableTestCase extends RichIterableWithDuplicatesTestCase, MapTestCase
```
motlin, line 38: This is a tricky detail. Every MapIterable in Eclipse Collections implements Map, even the immutable ones. That means we ought to be running tests against methods like Map.put() for every MapIterable. This is the core of the change that led to discovering so many inconsistencies between Mutable, Immutable, and Unmodifiable maps.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/UnmodifiableMapTestCase.java
```diff
@@ -0,0 +1,315 @@
...
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.fail;
+
+public interface UnmodifiableMapTestCase
+        extends MapTestCase
+{
+    @Test
+    default void Map_put()
+    {
+        Map<Integer, String> map = (Map<Integer, String>) this.newWithKeysValues(3, "Three", 2, "Two", 1, "One");
```
motlin, line 30: newWithKeysValues() returns Object in these tests, to allow for Map and MapIterable to have the same assertions. The result of the method call is cast to Map.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MapTestCase.java
```diff
@@ -20,15 +20,18 @@
...
 
 public interface MapTestCase
 {
-    <T> Map<Object, T> newWith(T... elements);
+    // Returns Object to allow subclasses to return either Map or MapIterable, and be cast to Map either way
+    <T> Object newWith(T... elements);
 
-    <K, V> Map<K, V> newWithKeysValues(Object... elements);
+    // Returns Object to allow subclasses to return either Map or MapIterable, and be cast to Map either way
+    <K, V> Object newWithKeysValues(Object... elements);
```
motlin, line 34: Here's where we return Object to allow testing of Map and MapIterable together. The generics are funny here. Deleting them leads to the overriding methods not being recognized as overrides.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/UnmodifiableHashMapTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public <T> Map<Object, T> newWith(T... elements)
+    {
+        Random random = new Random(CURRENT_TIME_MILLIS);
+        Map<Object, T> result = new HashMap<>();
+        for (T each : elements)
+        {
+            assertNull(result.put(random.nextDouble(), each));
+        }
+        return Collections.unmodifiableMap(result);
```
motlin, line 40: This is an important test to ensure consistency across implementations. This runs the same assertions against Collections.unmodifiableMap(new HashMap<>())

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/UnmodifiableMutableMapTest.java
```diff
@@ -62,133 +55,9 @@ public <K, V> MutableMap<K, V> newWithKeysValues(Object... elements)
...
-        assertThrows(UnsupportedOperationException.class, () -> map.merge(3, "4", (v1, v2) -> {
-            fail("Expected lambda not to be called on unmodifiable map");
-            return null;
-        }));
-        assertThrows(UnsupportedOperationException.class, () -> map.merge(4, "4", (v1, v2) -> {
-            fail("Expected lambda not to be called on unmodifiable map");
-            return null;
-        }));
-        Assert.assertEquals(this.newWithKeysValues(1, "1", 2, "2", 3, "3"), map);
-    }
```
motlin, line 193: These assertions are pulled up and now apply to all MutableMapIterables, not just MutableMaps. This led to finding some inconsistencies in BiMaps.

PR #1620, opened 2024-06-06
Fix performance problem in MutableList.subList() and implement similar optimization as ArrayList.subList()

yoa1226:


motlin:
This looks great, thanks. I'm wondering why additional tests are marked not applicable.

yoa1226:
@motlin hi, motlin, Thank you for your reply. The reason why additional tests are marked not applicable is that some methods  in `MutableList`  are not implemented In it's subclass. list below :

1.  `UnmodifiableMutableList` , ` MultiReaderFastList` and ` FastListNoIterator` don't support `listIterator` and `addAll`.
2.  `CompositeFastList` don't support `addAll(int index, Collection<? extends E> c)`.

So, when unit tests about `addAll` and `listIterator` are ignored in above class.

motlin (commented):
I'll share a commit with fixes to the checkstyle build as well as the ideas mentioned above.

motlin:
Fixed in #1621

PR #1632, opened 2024-06-15
Implement Stack.getLast()

Saurabh-Daware:


Saurabh-Daware:
Requesting review @motlin. Thanks!

unit-tests-java8/src/test/java/org/eclipse/collections/test/stack/StackIterableTestCase.java
```diff
@@ -85,7 +86,7 @@ default void InternalIterable_forEachWith()
     @Test
     default void RichIterable_getFirst_and_getLast()
     {
-        assertThrows(UnsupportedOperationException.class, this.newWith()::getLast);
+        assertThrows(EmptyStackException.class, this.newWith()::getLast);
     }
```
motlin, line 91: This override is no longer necessary. We can just delete the whole method RichIterable_getFirst_and_getLast here. Afterwards, we'll inherit the assertions for the superclass:

```java
    @Test
    default void RichIterable_getFirst_and_getLast()
    {
        RichIterable<Integer> iterable = this.newWith(3, 2, 1);
        assertNotEquals(iterable.getFirst(), iterable.getLast());
    }
```

unit-tests-java8/src/test/java/org/eclipse/collections/test/stack/StackIterableTestCase.java
```diff
@@ -99,36 +100,36 @@ default void RichIterable_getFirst_empty_null()
     @Test
     default void RichIterable_getLast_empty_null()
     {
-        assertThrows(UnsupportedOperationException.class, this.newWith()::getLast);
+        assertThrows(EmptyStackException.class, () -> this.newWith().getLast());
```
motlin, line 103: This is fine for now. @donraab what do you think of changing the behavior to match other collections? pop() would still throw, but getFirst() and getLast() would return null.

unit-tests-java8/src/test/java/org/eclipse/collections/test/stack/StackIterableTestCase.java
```diff
@@ -99,36 +100,36 @@ default void RichIterable_getFirst_empty_null()
...
     }
 
     @Override
     @Test
     default void RichIterable_getLast()
     {
-        assertThrows(UnsupportedOperationException.class, this.newWith(3, 3, 3, 2, 2, 1)::getLast);
+        assertEquals(1, this.newWith(3, 3, 3, 2, 2, 1).getLast());
```
motlin, line 110: Also delegate to the superclass here and in the next few methods.

```java
OrderedIterableWithDuplicatesTestCase.super.RichIterable_getLast();
```

motlin (changes requested):

unit-tests-java8/src/test/java/org/eclipse/collections/test/stack/StackIterableTestCase.java
Saurabh-Daware: `RichIterable` & `OrderedIterable` expect null when an element doesn't exists but in Stack we have EmptyStack exception.

Saurabh-Daware: If we were to return null on getLast(), we need to return null on getFirst() too.

unit-tests-java8/src/test/java/org/eclipse/collections/test/stack/StackIterableTestCase.java
donraab: I'm kind of conflicted with having `getLast()` for Stack. Stack is both LIFO and FILO. So what should getFirst and getLast mean in the context of Stack? I see that the implementation calls delegate.getFirst() for ArrayStack. I think getFirst and getLast should probably behave the same as peek in terms of throwing or not throwing.

motlin: I think overrides in tests are a hint that the interface hierarchy has a problem. We have getFirst and getLast on RichIterable and they return null on empty RichIterables unless:

- if you call getFirst() on an empty stack, you get an exception indicating it's an empty stack
- if you call getLast() on an empty stack, either the same thing or an UnsupportedOperationException because you're only supposed to look at the front of a stack

Even on other collection types, when getFirst() returns null, we don't know from the returned value alone whether the collection is empty or has a null at the front. I tend to think we should deprecate these methods across the board. They're already deprecated on RichIterable, but overridden to be not deprecated on OrderedIterable. We're talking here about a case where they behave inconsistently on an OrderedIterable.

motlin (approved):
Looks great, thank you very much for the contribution!

PR #1634, opened 2024-06-16
Remove expected from tests

Desislav-Petrov:
@motlin  - 2 more

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/ImmutableSingletonListTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/ImmutableSingletonListTest.java
@@ -14,6 +14,7 @@ import org.eclipse.collections.api.list.ImmutableList;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 
 public class ImmutableSingletonListTest extends AbstractImmutableListTestCase
 {
@@ -28,7 +29,7 @@ public class ImmutableSingletonListTest extends AbstractImmutableListTestCase
     public void min_null_throws()
     {
         // Collections with one element should not throw to emulate the JDK Collections behavior
-        super.min_null_throws();
+        assertDoesNotThrow(() -> this.classUnderTestWithNull().min(Integer::compareTo));
     }
 
     @Test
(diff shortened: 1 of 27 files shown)


unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/ImmutableSingletonListTest.java
```diff
@@ -28,31 +29,31 @@ protected ImmutableList<Integer> classUnderTest()
     public void min_null_throws()
     {
         // Collections with one element should not throw to emulate the JDK Collections behavior
-        super.min_null_throws();
+        assertDoesNotThrow(() -> this.classUnderTestWithNull().min(Integer::compareTo));
```
Desislav-Petrov, line 32: This one is interesting.. 
1) Junit4 doesn't have assertDoesNotThrow equivalent so w're using junit5 apis which isn't great but not a big ideal imo - i'm about to flip the imports soon
2) This test doesn't quite do what it says implemented in this way. The implementation is consistent with the comment but not with the test method name. Happy to just move to **// Not applicable**

unit-tests/src/test/java/org/eclipse/collections/impl/set/fixed/SingletonSetTest.java
```diff
@@ -558,30 +559,30 @@ public void newEmpty()
     public void min_null_throws()
     {
         // Collections with one element should not throw to emulate the JDK Collections behavior
-        super.min_null_throws();
```
Desislav-Petrov, line 561: Similar comment here..

unit-tests/src/test/java/org/eclipse/collections/impl/set/fixed/AbstractMemoryEfficientMutableSetTestCase.java
```diff
@@ -210,14 +210,15 @@ public void iterationWithIterator()
...
+    @Test
     public void iteratorWillGetUpsetIfYouPushItTooFar()
     {
         Iterator<String> iterator = this.classUnderTest().iterator();
-        for (int i = 0; i < this.classUnderTest().size() + 1; i++)
+        for (int i = 0; i < this.classUnderTest().size(); i++)
         {
             iterator.next();
         }
+        assertThrows(NoSuchElementException.class, iterator::next);
```
motlin, line 221: nice, much more precise

motlin (approved):

unit-tests/src/test/java/org/eclipse/collections/impl/list/immutable/ImmutableSingletonListTest.java
Desislav-Petrov: The only benefit that I see is that you explicitly declare your expectation - test isn't meant to throw an exception. Ideally every test should have some type of assertion so that one never wonders if this test was left incomplete for some reason.

Having said that, no strong opinion and I'm happy to remove the assertions if you like.

motlin (approved):
Thanks again!

PR #1636, opened 2024-06-22
Sync IntelliJ project files.

motlin:


motlin:
This doesn't fix the IntelliJ project file problem once and for all, but it deletes a bunch of stuff I suspected was unnecessary and doesn't cause the TeamCity build to fail. It should reduce but not eliminate project file churn. I'll keep working on this after this lands.

mohrezaei (approved):

Issue #1640, opened 2024-07-04
Optimize ImmutableArrayList.takeWhile() and dropWhile() for small lists.


motlin:
ImmutableArrayList.takeWhile() and dropWhile() always return ImmutableArrayList. Since these methods don't delegate to a factory, they can't return ImmutableEmptyList for example.

prithviraj2609:
Hello @motlin,

I have tried resolving this issue and created a PR for the same.
Would you review and merge my PR if this issue is still open
Here is my PR: https://github.com/eclipse/eclipse-collections/pull/1644

PR #1641, opened 2024-07-04
#1640 Issue: added condition for empty list

Sudhansh702:
This PR addresses the issue where takeWhile() and dropWhile() methods always return ImmutableArrayList. 

Now, these methods return ImmutableEmptyList when applicable, improving the performance and correctness of the code.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayList.java
@@ -605,6 +605,9 @@ final class ImmutableArrayList<T>
     {
         int endIndex = this.detectNotIndex(predicate);
         T[] result = (T[]) new Object[endIndex];
+
+        if(endIndex <= 0)return new ImmutableEmptyList<>();
+        
         System.arraycopy(this.items, 0, result, 0, endIndex);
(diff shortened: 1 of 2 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableArrayList.java
```diff
@@ -605,6 +605,9 @@ public ImmutableList<T> takeWhile(Predicate<? super T> predicate)
     {
         int endIndex = this.detectNotIndex(predicate);
         T[] result = (T[]) new Object[endIndex];
+
+        if(endIndex <= 0)return new ImmutableEmptyList<>();
+        
         System.arraycopy(this.items, 0, result, 0, endIndex);
         return new ImmutableArrayList<>(result);
```
motlin, line 612:
```suggestion
        int startIndex = this.detectNotIndex(predicate);
        int resultSize = this.size() - startIndex;
        T[] result = (T[]) new Object[resultSize];
        System.arraycopy(this.items, startIndex, result, 0, resultSize);
        return Lists.immutable.with(result);
```

motlin (changes requested):
Please remove .vscode/ and fix the implementation of takeWhile(). The implementation of dropWhile() looks ok.

PR #1643, opened 2024-07-07
Remove expected - final PR

Desislav-Petrov:
@motlin final one - will start moving imports after

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/bag/immutable/ImmutableEmptyBagTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/bag/immutable/ImmutableEmptyBagTest.java
@@ -346,45 +346,45 @@ public class ImmutableEmptyBagTest extends ImmutableBagTestCase
     }
 
     @Override
-    @Test(expected = NoSuchElementException.class)
+    @Test
     public void min()
     {
-        this.newBag().min(String::compareTo);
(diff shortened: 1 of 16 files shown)


unit-tests/src/test/java/org/eclipse/collections/impl/bag/immutable/ImmutableEmptyBagTest.java
```diff
@@ -576,7 +574,6 @@ public void groupByUniqueKey()
     @Test
     public void groupByUniqueKey_throws()
     {
-        super.groupByUniqueKey_throws();
```
motlin, line 579: I'm curious about these removals of the calls to super.

motlin (commented):
Looks good overall. One minor question which I can probably figure out on my own when I'm back at my main computer.

unit-tests/src/test/java/org/eclipse/collections/impl/bag/immutable/ImmutableEmptyBagTest.java
Desislav-Petrov: The super class call didn't do an explicit assertion for an exception due to the @Override which leads to the old (expected = xx.class) being ignored.

Now that we have moved to the junit5 style, we have explicit assertion that fails the test in the super.() call.

PR #1656, opened 2024-07-29
Moving serialization-tests to junit5

Desislav-Petrov:
@motlin i suspect you will get shocked when you see the number of files but 99.95% are import updates so using the intellij trick should hopefully make it less painful :))


motlin (approved):
I found an even better command for performing this kind of review.

```
❯ git show --ignore-matching-lines='^import .*;$' upstream/pr/1656
```

In this filtered view, there's one JUnit 4 removal in a pom.xml, and three places where we swap Before to BeforeEach. LGTM!

PR #1659, opened 2024-08-02
Prepare unit-tests-java8 for migration

Desislav-Petrov:
This PR is a preparation for the **unit-tests-java8** module for migration to Junit5. 

- Moves all junit4 assertions to static imports
- Renames a few helper methods from **assertXXX** to **assertIterables(Equal/NotEqual)** as it makes the refactoring easier

cc @motlin


motlin (commented):
Let's change the commit message to something that reflects the content, like Prepare unit-tests-java8 for migration.

Given the number of times that IterableTestCase.assertEquals() appears in code, I'd like it to have the best possible name. What do you think about assertIterablesEqual?

Desislav-Petrov:
@motlin - thx for the feedback, addressed

motlin (approved):

PR #1661, opened 2024-08-04
Optimize forEach method on Map views to not delegate to an iterator.

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -2384,17 +2391,63 @@ protected void copyKeys(Object[] result)
...
+            {
+                return false;
+            }
+
+            for (int i = 0; i < UnifiedMap.this.table.length; i += 2)
+            {
+                Object cur = UnifiedMap.this.table[i];
+                if (cur == CHAINED_KEY)
                 {
-                    return this.containsAll(other);
```
mohrezaei, line 2394: This doesn't look like a strict optimization to me. `containsAll` assumes `other` has a reasonably efficient iterator -- most collections do. The new implementation assumes `other` has a very efficient (`O(1)`) `contains`, which is not always true, for example with tree based implementations. If `other.contains` is `O(log N)`, then this method went from `O(N)` to `O(N log N)`.

mohrezaei:
In general, I'd suggest adding performance tests before optimizing code.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
motlin: It's a good point. I was making these methods more consistent with the regular equals() method, while trying to make test assertions more consistent. It's been this way since we open sourced.

```java
    @Override
    public boolean equals(Object object)
    {
        if (this == object)
        {
            return true;
        }

        if (!(object instanceof Map))
        {
            return false;
        }

        Map<?, ?> other = (Map<?, ?>) object;
        if (this.size() != other.size())
        {
            return false;
        }

        for (int i = 0; i < this.table.length; i += 2)
        {
            Object cur = this.table[i];
            if (cur == CHAINED_KEY)
            {
                if (!this.chainedEquals((Object[]) this.table[i + 1], other))
                {
                    return false;
                }
            }
            else if (cur != null)
            {
                K key = this.nonSentinel(cur);
                V value = (V) this.table[i + 1];
                Object otherValue = other.get(key);
                if (!UnifiedMap.nullSafeEquals(otherValue, value) || (value == null && otherValue == null && !other.containsKey(key)))
                {
                    return false;
                }
            }
        }

        return true;
    }
```

motlin:
> In general, I'd suggest adding performance tests before optimizing code.

The main motivation is to get the tests in https://github.com/eclipse/eclipse-collections/pull/1662 to pass.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
mohrezaei: `Map.equals` is bit more special, because iterating `other` may produce a lot of garbage via `Entry` object creation.
I've always felt that there was something missing in collections specifications in Java: a way to examine performance behavior at runtime and pick appropriate algorithms based on that. I guess this could be annotations or marker interfaces like `OhOneContains`, etc and maybe interface specific things like `Map.GarbageFreeEntry`...

mohrezaei:
> > In general, I'd suggest adding performance tests before optimizing code.
> 
> The main motivation is to get the tests in #1662 to pass.

I don't understand. Is there a bug in the original implementation?

motlin:
> > > In general, I'd suggest adding performance tests before optimizing code.
> > 
> > 
> > The main motivation is to get the tests in #1662 to pass.
> 
> I don't understand. Is there a bug in the original implementation?

It's not a bug, because the contract of equals doesn't say much about the implementation. Applying these strict assertions in an abstract test that applies to all of our collections shows inconsistencies:

- equals implementations that iterate over this or other
- equals and forEach methods that delegate to iterator
- forEach methods that create Entry

There's another that I found but didn't have time to address yet, is whether the map supports this pattern.

```
MutableMapIterable<String, Integer> map = this.newWithKeysValues("3", 3, "2", 2, "1", 1);
map.entrySet().forEach(each->each.setValue(each.getValue() + 1));
assertIterablesEqual(this.newWithKeysValues("3", 4, "2", 3, "1", 2), map);
```

These inconsistencies don't feel intentional, because each affects a different subset of maps and their keySet/entrySet views. WDYT?

mohrezaei:
I'm not sure I'd label these as inconsistencies:

>  - equals implementations that iterate over this or other

Sometimes it's better to iterate this, sometimes other. See above.

>  - equals and forEach methods that delegate to iterator

forEach delegating to iterator can likely be optimized in our collections, so that's a good check. Same for equals delegating to this.iterator, but equals calling other.iterator is ok.

>  - forEach methods that create Entry

Isn't this sometimes required by the interface, like Map.entrySet().forEach?

mohrezaei:
I don't understand the Map Iterable example. Does the assert fail and why?

> inconsistencies don't feel intentional, because each affects

affect how?

motlin:
> I don't understand the Map Iterable example. Does the assert fail and why?
> 
> > inconsistencies don't feel intentional, because each affects
> 
> affect how?

We have a bunch of tests like https://github.com/eclipse/eclipse-collections/blob/master/unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/UnifiedMapNoIteratorTest.java that assert that iterators are rarely used for anything. The tests below fail on UnifiedMapNoIteratorTest, UnifiedMapWithHashingStrategyNoIteratorTest, and HashBiMapNoIteratorTest. When I started to write this all out, I was thinking there were lots of inconsistencies but it seems like at least all hash tables are consistent with each other.

motlin:
I scaled this change back to no longer change any equals() methods, and only touch the forEach() methods on map views. I think this is a lot more straight-forward since it's just one kind of forEach method delegating to another.

mohrezaei (approved):

PR #1662, opened 2024-08-04
Add more assertions to IterableTestCase.assertIterablesEqual().

motlin:


motlin:
@donraab here's another little review if you have time to take a look

unit-tests-java8/src/test/java/org/eclipse/collections/test/IterableTestCase.java
```diff
@@ -143,9 +146,23 @@ static void assertIterablesEqual(Object o1, Object o2)
                 }
             }
         }
-        else if (o1 instanceof SortedMap<?, ?> || o2 instanceof SortedMap<?, ?>)
```
donraab, line 146: Why did you remove the `SortedMap` test?

motlin: Good catch, I'll restore this. I had replaced SortedMap with Map, which led me down a tangent. It caused test to fail because our hash tables implement equals() by delegating to iterator. I changed in that https://github.com/eclipse/eclipse-collections/pull/1661 and then eventually backed away from it. In that shuffle, I deleted the assertion here. Verify.assertMapsEqual() also fails here. And we've already called assertEquals before this line. So it's all subject to change but I might as well revert this one part.

motlin: @donraab reverted

donraab (approved):
LGTM

PR #1663, opened 2024-08-05
Migrate unit-tests-java8 to Junit5

Desislav-Petrov:
Migrate **junit-trait-runner** & **unit-tests-java8** to Junit5.

- Remove junit4 dependencies from pom.xmls
- Remove the junit-trainer runner as it's no longer needed
- Uplift all imports in the **unit-tests-java8** project

cc @motlin - two more in the bag


junit-trait-runner/src/main/java/org/eclipse/collections/impl/test/junit/Junit8TraitRunnerExtension.java
```diff
@@ -14,48 +14,49 @@
...
 
-public class Java8TestClass extends TestClass
+/**
+ * The class implements a custom Junit5 extension to facilitate the execution of trait-based tests
+ * that are implemented using default interface methods which Junit doesn't respect.
+ *
+ * The code manually appends the correct annotations to the test methods in the concrete classes if they are found
+ * anywhere in the interface hierarchy.
+ */
+public class Junit8TraitRunnerExtension implements BeforeEachCallback
```
motlin, line 34: Very cool. Are you sure it's still necessary?

Years ago I started a conversation about moving this into JUnit 4 which didn't go anywhere but I was under the impression that it might be the default behavior in JUnit 5.

https://github.com/junit-team/junit4/pull/760

Desislav-Petrov: Ah! You're right - you don't need it anymore so we can clean up the whole module. I will uplift soon and resubmit, thx for the catch.

Desislav-Petrov:
@motlin updated

motlin (approved):

PR #1669, opened 2024-08-17
Prepare test templates for migration

Desislav-Petrov:
This PR prepares the test templates for Junit5 Migration by moving all assertions to static imports.

cc @motlin


motlin (approved):
wow

Issue #1666, opened 2024-08-09
SortedSet.html#intersect does not always return elements from this


duponter:
## Context
According to the JavaDocs `SortedSet.html#intersect` should always return members from **_this_**:

https://eclipse.dev/collections/javadoc/11.1.0/org/eclipse/collections/api/set/sorted/ImmutableSortedSet.html#intersect(org.eclipse.collections.api.set.SetIterable)
> Description copied from interface: [SortedSetIterable](https://eclipse.dev/collections/javadoc/11.1.0/org/eclipse/collections/api/set/sorted/SortedSetIterable.html#intersect(org.eclipse.collections.api.set.SetIterable))
> Returns the set of all objects that are members of both this and set. The intersection of [1, 2, 3] and [2, 3, 4] is the set [2, 3]. The output will contain instances from this, not set.
 
## Issue
However, testing reveals that elements of the smallest of the 2 sets are returned instead. Not only contradicts this the API documentation, but this is counterintuitive as well.
In addition, when replacing `intersect` with the `select(Set::contains)` combination, the same functionality works as expected.

## Reproduction
Following test suite contains 4 test cases:
* SetA.intersect(SetB) where SetA > SetB
* SetA.intersect(SetB) where SetA < SetB
* SetA.select(SetB::contains) where SetA > SetB
* SetA.select(SetB::contains) where SetA < SetB

Technologies used:
* JDK 21
* Eclipse Collections 11.1
* JUnit 5.10
* AssertJ 3.25

```java
class SortedSetIntersectTest {

    @Nested
    class SelectContains {
        @Test
        void selectContainsReturnsElementsFromTheInvokingSetWhenFirstSetIsLargerThanSecond() {
            ImmutableList<Person> persons = Person.generate(20);
            ImmutableSortedSet<Person> first = SortedSets.immutable.withAll(Person.NAME_ONLY_COMPARATOR, persons.take(15));
            ImmutableSortedSet<Person> second = SortedSets.immutable.withAll(Person.NAME_ONLY_COMPARATOR, persons.drop(10).collect(Person::changeAge));

            System.out.printf("Set 1 with %d elements ...%n", first.size());
            first.forEachWithIndex((p, i) -> System.out.printf("%02d: %s%n", i, p));
            System.out.printf("%n...vs Set 2 with %d elements%n", second.size());
            second.forEachWithIndex((p, i) -> System.out.printf("%02d: %s%n", i, p));

[...]

PR #1672, opened 2024-08-18
Fix incorrect Javadoc for SortedSetIterable.intersect(),SetIterable.intersect() .

lzcyx:
For issue #1666

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/set/SetIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/set/SetIterable.java
@@ -42,7 +42,11 @@ public interface SetIterable<T> extends RichIterable<T>
 
     /**
      * Returns the set of all objects that are members of both {@code this} and {@code set}. The intersection of
-     * [1, 2, 3] and [2, 3, 4] is the set [2, 3]. The output will contain instances from {@code this}, not {@code set}.
+     * [1, 2, 3] and [2, 3, 4] is the set [2, 3]. The output will contain instances from the smaller of the two sets
+     * ({@code this} or {@code set}) based on the size. If the sizes are equal, the output will contain instances from
+     * {@code this}.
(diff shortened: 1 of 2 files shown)


motlin:
Rather than change the Javadoc, I think we should fix the behavior.

mohrezaei:
Just as point of reference, the difference in performance can be drastic if the set sizes are very different. E.g. if one set has 1M elements and the other has 10, changing the code will make this example 100,000x slower.

eclipse-collections-code-generator/src/main/resources/api/set/immutablePrimitiveSet.stg
```diff
@@ -61,7 +61,11 @@ public interface Immutable<name>Set extends Immutable<name>Collection, <name>Set
 
     /**
      * Returns the set of all objects that are members of both {@code this} and {@code set}. The intersection of
-     * [1, 2, 3] and [2, 3, 4] is the set [2, 3].
+     * [1, 2, 3] and [2, 3, 4] is the set [2, 3]. The output will contain instances from the smaller of the two sets
```
mohrezaei, line 64: primitives only have one sense of equality. They are value types, so there is no "memory reference" distinction. Ditto for the other 2 primitive stg files.

lzcyx: I apologize for the oversight in my previous commit. I have now reverted these 3 primitive stg files.

lzcyx:
Hi @mohrezaei ,
I wanted to check in on this PR. I've made the requested changes and would appreciate it if you could review them when you have a chance.

Thank you for your time and guidance!

mohrezaei:
LGTM, but I was hoping to get a second opinion from @motlin or @donraab

motlin (approved):

motlin:
LGTM too, I'll merge it.

0x78f1935:
I wouldn't trust this dude @lzcyx . Nasty cookie stealer. Wannabe hacker lol

PR #1674, opened 2024-08-19
Fix MapIterate.forEachKeyValue() to throw NullPointerException on null Map.

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/MapIterate.java
```diff
@@ -770,7 +770,7 @@ public static <K, V> void forEachKeyValue(Map<K, V> map, Procedure2<? super K, ?
     {
         if (map == null)
         {
-            throw new IllegalArgumentException("Cannot perform a forEachKeyValue on null");
+            throw new NullPointerException("Cannot perform a forEachKeyValue on null");
         }
```
donraab, line 774: This who if statement could be replaced with `Object.requireNonNull()` since you have changed the exception type thrown to NullPointerException.

donraab (approved):
LGTM

PR #1676, opened 2024-08-20
Sync .idea/compiler.xml.

motlin:


.idea/compiler.xml
```diff
@@ -7,12 +7,12 @@
         <sourceTestOutputDir name="target/generated-test-sources/test-annotations" />
         <outputRelativeToContentRoot value="true" />
         <module name="eclipse-collections-api" />
+        <module name="jcstress-tests" />
```
donraab, line 10: The modules are the same, just in a different order. Does order matter here?

motlin: Order doesn't matter. I was seeing churn in this file. Does it happen for you?

donraab: How do we know this change won't cause churn for someone else? Do you know why the order is being changed on your machine?

motlin: I don't know. I assume it's related to the version of IntelliJ. I'd like to know what you see on your machine. Does this file churn? Does this commit remove the diffs and show your file as unchanged again?

donraab: Yes, it churns, but differently than what you are committing.

mohrezaei: It's weird to see you going back and forth on something that's been obvious for 2 decades...
https://github.com/eclipse/eclipse-collections/issues/581

motlin: I deleted compiler.xml and the Inspections builds still pass. So I committed that and .gitignored it.

donraab (approved):
LGTM

PR #1679, opened 2024-08-21
Delete four flaky tests: Collectors2AdditionalTest.sumBy*Parallel().

motlin:
These four tests are flaky. They fail a small percentage of the time and I usually rerun them and move on. I finally looked at the code and they are clearly not thread-safe. They all run some form of `.parallelStream().collect(Collectors2.sumBy*(), ...)`.

The `Collectors2.sumBy*()` methods all delegate directly to a non-thread-safe map, so this is a pretty clear case of non-thread-safe code. I have no opinion on whether the Collectors ought to be thread-safe; if so let's file an issue to fix them separately. For now, I just want to remove the tests since they waste time.

There are serial versions of the same tests, so deleting the tests does not decrease coverage.

cc @donraab

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2AdditionalTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2AdditionalTest.java
@@ -150,14 +150,6 @@ public class Collectors2AdditionalTest
     }
 
     @Test
-    public void sumByIntParallel()
-    {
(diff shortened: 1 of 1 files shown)


donraab:
@motlin Makes sense to file an issue.

donraab (approved):
LGTM

motlin:
> @motlin Makes sense to file an issue.

Raised #1680

PR #1687, opened 2024-08-31
Add Spotless maven profiles.

motlin:
When working locally, I break the Checkstyle build frequently and manually fix it up. Probably the most frequently violated rule is unused imports. This feels silly, there are lots of CLI tools that can sort and minimize imports.

In other projects, at work and now in side projects, I've had good results with [Spotless](https://github.com/diffplug/spotless). It's a linter runner that unifies the configuration of other linters. For example, it allows configuring consistent on/off annotations to disable formatters on sections of code, even if the formatters were written by different authors. It also allows chaining linters For example, it can be configured to format code with an opinionated formatter like prettier, and then sort imports differently afterwards.

In this PR, I'm adding a few maven profiles that run Spotless if they are activated, but I'm not activating them in any GitHub Workflows. For now, I just want to be able to run them manually (to auto fix Java imports) and for you all to take a look and see if you like the tool.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,6 +135,8 @@ Arrays, Maps or Strings. The iteration protocol was inspired by the Smalltalk co
         <maven.compiler.meminitial>2048m</maven.compiler.meminitial>
         <maven.compiler.maxmem>2048m</maven.compiler.maxmem>
 
+        <spotless.check.skip>true</spotless.check.skip>
+
         <argLine></argLine>
 
         <project.previous.version>11.0.0</project.previous.version>
@@ -513,6 +515,17 @@ Arrays, Maps or Strings. The iteration protocol was inspired by the Smalltalk co
                 <version>3.3.1</version>
             </plugin>
 
+            <plugin>
+                <groupId>com.diffplug.spotless</groupId>
+                <artifactId>spotless-maven-plugin</artifactId>
+                <version>2.43.0</version>
+                <configuration>
+                    <upToDateChecking>
+                        <enabled>true</enabled>
+                    </upToDateChecking>
+                </configuration>
+            </plugin>
+
         </plugins>
 
     </build>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -832,6 +846,224 @@ Arrays, Maps or Strings. The iteration protocol was inspired by the Smalltalk co
...
+                <plugins>
+                    <plugin>
+                        <groupId>com.diffplug.spotless</groupId>
+                        <artifactId>spotless-maven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <goals>
+                                    <goal>check</goal>
+                                </goals>
+                                <phase>validate</phase>
```
motlin, line 867: The check goal is meant to be used in CI. It doesn't format code. It fails if the code isn't already formatted.

pom.xml
```diff
@@ -832,6 +846,224 @@ Arrays, Maps or Strings. The iteration protocol was inspired by the Smalltalk co
...
+                <plugins>
+                    <plugin>
+                        <groupId>com.diffplug.spotless</groupId>
+                        <artifactId>spotless-maven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <goals>
+                                    <goal>apply</goal>
+                                </goals>
+                                <phase>verify</phase>
```
motlin, line 892: The apply goal is meant to be run locally, to auto-format code. By itself, it doesn't do anything. We need to activate at least two profiles together to perform some auto-formatting. For example, `mvn spotless:apply --activate-profiles 'spotless-apply,spotless-java-sort-imports'` would fix up imports locally. A similar command could run from CI, just swapping in the word check for apply.

pom.xml
```diff
@@ -832,6 +846,224 @@ Arrays, Maps or Strings. The iteration protocol was inspired by the Smalltalk co
...
+                                    <off>@formatter:off</off>
+                                    <on>@formatter:on</on>
+                                </toggleOffOn>
+
+                                <importOrder>
+                                    <!-- use an empty string for all the imports not specified explicitly, '|' to join group without blank line, and '\#` prefix for static imports. -->
+                                    <order>java,javax,,\#com.diffplug,\#java|\#javax,\#</order>
+                                </importOrder>
+
+                                <removeUnusedImports />
```
motlin, line 920: Sort imports, then remove unused imports.

prathasirisha (approved):
LGTM, will give the tool a try locally.

PR #1688, opened 2024-08-31
Format yaml files using Spotless.

motlin:
Another example based on #1687, to show off the kind of thing that Spotless can do, this time in a language other than Java.

This was created with the command:

`mvn spotless:apply --activate-profiles 'spotless-apply,spotless-yaml'`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,10 @@
 version: 2
 updates:
-  - package-ecosystem: "maven"
-    directory: "/"
+  - package-ecosystem: 'maven'
+    directory: '/'
     schedule:
-      interval: "weekly"
-  - package-ecosystem: "github-actions"
-    directory: "/"
+      interval: 'weekly'
+  - package-ecosystem: 'github-actions'
+    directory: '/'
     schedule:
-      interval: "weekly"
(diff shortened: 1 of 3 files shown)


donraab (commented):
@motlin This looks to be replacing single quotes with double quotes. Are double quotes preferred in YAML?

motlin:
> @motlin This looks to be replacing single quotes with double quotes. Are double quotes preferred in YAML?

Honestly, I'm not sure. I just noticed we are inconsistent and it's one of the things that's easy to demo with Spotless. Searching for this question:

> Single quotes (''):
> 
> - Preserve the literal value of all characters inside
> - Don't allow escaping of characters
> - To include a single quote in the string, you need to double it
> 
> Double quotes (""):
> 
> - Allow for escaped characters using backslashes
> - Enable the use of special characters like \n for newline
> 
> General recommendations:
> 
> Use single quotes as the default for most strings, especially if they don't contain special characters or escapes.

I can switch to single!

motlin:
> @motlin This looks to be replacing single quotes with double quotes. Are double quotes preferred in YAML?

Switched to single quotes.

donraab (approved):
LGTM

PR #1690, opened 2024-09-01
Remove redundant tests from unit-tests that live in unit-tests-java8.

motlin:


donraab:
@motlin I have mixed feelings about this PR. On the one hand, if the tests are duplicated, then I have no issue. On the other hand, I have no way of verifying in this PR that the tests are duplicated, as they are sitting in two different places.

The thing I would feel more comfortable with is merging everything into one test suite, and removing duplication then.

motlin:
> @motlin I have mixed feelings about this PR. On the one hand, if the tests are duplicated, then I have no issue. On the other hand, I have no way of verifying in this PR that the tests are duplicated, as they are sitting in two different places.
> 
> The thing I would feel more comfortable with is merging everything into one test suite, and removing duplication then.

I put it up together with https://github.com/eclipse/eclipse-collections/pull/1684. One adds the assertions, the other removes the redundant assertions.

donraab (approved):
LGTM

PR #1691, opened 2024-09-01

motlin:
Another example based on #1687.

This was created with the command:

`mvn spotless:apply --activate-profiles 'spotless-apply,spotless-java-sort-imports'`

I was using this command to fix up the imports in my own changes, but it found a few older examples of files with unsorted imports, and files that don't end with a newline.


donraab (approved):
LGTM

PR #1693, opened 2024-09-01
Add job forbid-merge-commits to .github/workflows/pull-request.yml.

motlin:
We recently had a ["backwards" merge commit](https://github.com/eclipse/eclipse-collections/commit/cb4f64274e905492f0c70f9981a6097c718fad97) land on the master branch.

It's visible in the git history as the white arrow here.

Screenshot 2024-09-01 at 9 35 17 AM

I wrote and open sourced [a GitHub Action](https://github.com/marketplace/actions/forbid-merge-commits) to prevent this sort of thing. Its README has more details about why this is a problem and how folks get into this situation. When it happens, the error message is pretty helpful about how to fix up git history.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -298,3 +298,9 @@ jobs:
           MAVEN_OPTS: '-Xmx2g'
       - name: 'Clean Maven cache'
         run: rm -rf ~/.m2/repository/org/eclipse/collections
+
+  forbid-merge-commits:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run Forbid Merge Commits Action
+        uses: motlin/forbid-merge-commits-action@main


.github/workflows/pull-request.yml
```diff
@@ -298,3 +298,9 @@ jobs:
           MAVEN_OPTS: "-Xmx2g"
       - name: 'Clean Maven cache'
         run: rm -rf ~/.m2/repository/org/eclipse/collections
+
+  forbid-merge-commits:
```
prathasirisha, line 302: I am not entirely sure if it will immediately be evident to the contributor what their next steps are if this action fails - is it possible to include a section regarding the rebase steps in the contributing guidelines?

motlin: When the check fails, it leaves instructions on how to fix the problem. I can't demonstrate it on this repository until it lands, but I could demonstrate it in another repository if that would be helpful.

PR #1697, opened 2024-09-02
Add missing @Test annotations and fix broken assertion.

motlin:
I just learned that a missing `@Test` annotation on an `@Override` method causes the test to not be run, probably only in IntelliJ. I'm not sure if this is related to the JUnit 5 refactoring, but I'm looping in @Desislav-Petrov just in case.

I added a bunch of missing annotations. Along the way, I found one that wasn't an Override, and it was broken. I fixed its assertion.


donraab (approved):
LGTM

PR #1703, opened 2024-09-15
Add more files to .idea/.gitignore.

motlin:


.idea/.gitignore
```diff
@@ -1,2 +1,3 @@
-/workspace.xml
```
donraab, line 1: Why remove it and then add it back?

motlin: Just ran `:sort` on the file

donraab (approved):
LGTM

PR #1704, opened 2024-09-15
Stop throwing mangled exceptions.

motlin:
If you hide spaces, this looks like a smaller change. 99 additions and 1,413 deletions.

### DIFF
--- a/eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
+++ b/eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
@@ -98,115 +98,9 @@ public final class Verify extends Assert
         throw new AssertionError("Suppress default constructor for noninstantiability");
     }
 
-    /**
-     * Mangles the stack trace of {@link AssertionError} so that it looks like it's been thrown from the line that
-     * called to a custom assertion.
-     * <p>
-     * This method behaves identically to {@link #throwMangledException(AssertionError, int)} and is provided
-     * for convenience for assert methods that only want to pop two stack frames. The only time that you would want to
-     * call the other {@link #throwMangledException(AssertionError, int)} method is if you have a custom assert
-     * that calls another custom assert i.e. the source line calling the custom asserts is more than two stack frames
-     * away
-     *
-     * @param e The exception to mangle.
-     * @see #throwMangledException(AssertionError, int)
-     */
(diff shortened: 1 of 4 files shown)


donraab:
@motlin I don't think I have a problem with this, but curious why the change is being made now, as I believe this has been like this for a very long time?

motlin:
> @motlin I don't think I have a problem with this, but curious why the change is being made now, as I believe this has been like this for a very long time?

Do you know the history of why we have it? It has always felt unintuitive to me. It interferes subtly with tools like IntelliJ's analyze trace, and Datadog's error tracking and test tracking.

donraab:
@motlin I can't recall why this was done.

motlin:
> @motlin I can't recall why this was done.

From the behavior, it feels like it was designed to change stack traces of assertion failures so that the Verify class is on top. Someone probably thought that was more clear.

Usually the top of the stack trace is in your own test code, on the line that made the assertion. I think that the behavior is worse; I want to see the line in my own code.

Even if it were clearly better, it is unusual. JUnit's classes like Assert don't do it. I think it's better to test the way everyone else does.

donraab (commented):
@motlin Would it be better to leave the behavior alone and just deprecate `Verify`? Eventually it needs to stop extending `Assert`, right?

donraab (approved):
LGTM. I have no issues with you simplifying this if it helps move it forward for the JUnit upgrade.

motlin:
> @motlin Should we evenutally deprecate `Verify` in favor of `Assertions`? Eventually `Verify` needs to stop extending `Assert` as well, right?

`Verify` has some useful things in it, that are collection-specific assertions and failure messages. Examples:

- When comparing ordered iterables, it will show the index of the first failure
- When asserting maps are equal, it will also assert that the key and entry sets are equal

`Verify` is based on JUnit 4 primitives like `Assert` and has a lot of redundant stuff in it.

I like the idea of separate assertion libraries. JUnit 4 supports assertThat with Hamcrest matchers. Unfortunately Hamcrest is no longer maintained.

I've started to use [assertj](https://github.com/assertj/assertj) in other projects. It has [add-on modules](https://assertj.github.io/doc/) including a [Guava module](https://assertj.github.io/doc/#assertj-guava), for assertions for its Multimaps and Multisets. My ideal would be we build assertj-eclipse-collections and once stable contribute it to the upstream project.

PR #1705, opened 2024-09-15
Add test coverage for MutableBagIterable.

motlin:


unit-tests-java8/src/test/java/org/eclipse/collections/test/bag/mutable/sorted/MutableBagIterableTestCase.java
```diff
@@ -38,6 +38,10 @@ default void MutableBagIterable_addOccurrences()
         assertThrows(
                 IllegalArgumentException.class,
                 () -> mutableBag.addOccurrences(4, -1));
+        assertEquals(Bags.immutable.with(1, 1, 1, 2, 2, 3, 3, 3, 4, 4, 4, 4), mutableBag);
+
+        assertEquals(-2147483647, mutableBag.addOccurrences(3, Integer.MAX_VALUE - 1));
```
donraab, line 43: This looks like a bug.

donraab: I'm going to approve the PR, because while this look like a bug because it doesn't fail fast, documents current behavior.

donraab (approved):
LGTM. The overflow of `addOccurrences` and `size` documents current expected behavior. We should decide if this is an issue that should be addressed.

unit-tests-java8/src/test/java/org/eclipse/collections/test/bag/mutable/sorted/MutableBagIterableTestCase.java
motlin: This is related to #1601 where my intention was to fix a bug, but it wasn't covered in tests. I will leave that PR closed because that's not the approach we'll go with the fix. Now the bug is at least visible.

PR #1718, opened 2024-10-14
Fixes #1717 - overflow issues in LongInterval

### LINKED ISSUE #1717: Some LongInterval methods produce overflow for large interval values
The implementations of `each*`, `injectInto*`, and `chunk` methods appear to use `int` values which results in overflow when long interval values are outside of the valid range for `int`

vmzakharov:


unit-tests/src/test/java/org/eclipse/collections/impl/list/primitive/LongIntervalTest.java
```diff
@@ -251,6 +251,66 @@ public void each()
...
+                        0, -1_000_000_000L, -2_000_000_000L, -3_000_000_000L, -4_000_000_000L,
+                        -5_000_000_000L, -6_000_000_000L, -7_000_000_000L, -8_000_000_000L, -9_000_000_000L),
+                values2);
+    }
+
+    @Test
+    public void forEachWithIndexForLargeValues()
+    {
+        var values1 = LongLists.mutable.empty();
+        var indices1 = LongLists.mutable.empty();
```
donraab, line 282: It would be clearer if the indices1 were a `MutableIntList`.

There is also another method you changed named `forEachWithLongIndex` which would make perfect sense for this test.

donraab (changes requested):

unit-tests/src/test/java/org/eclipse/collections/impl/list/primitive/LongIntervalTest.java
vmzakharov: `forEachWithLongIndex` is not used and deprecated for removal (that was the change 😀) so I would rather not used it. Will change indices to `MutableIntList`.

donraab (approved):
LGTM. Thanks for the fixes!

PR #1721, opened 2024-10-19
Migrate test templates to junit5

Desislav-Petrov:
1. Migrates junit4 to junit5 for all test templates
2. Uplifts the assertX syntax as the signature between juni4 and 5 isn't compatible (assertion description now always goes at the end of the assertion method)

@motlin i'm back :-)


motlin (approved):

motlin:
Welcome back @Desislav-Petrov and thanks again!

PR #1724, opened 2024-10-25
Fixed flaky test in ParallelIterableTestCase#sumOfFloatConsistentRounding

IamLRBA:
## Changes

- **Increased the Delta (Tolerance)**: I increased the delta value in the assertEquals call to allow for small floating-point precision differences. In the first case, the delta was currently set to 1.0e-15.

 Since the difference in your test is in the range of 1.0e-16, we could increase the delta slightly, for example to 1.0e-13. This should be able to fix the flaky test caused by slight floating-point rounding differences.


IamLRBA:
cc
@motlin

IamLRBA:
> The change here is to increase the epsilon / tolerance of the rounding error, as you mentioned. Based on the name of the test, it seems to be designed to ensure the rounding error is not high, so I'm not sure this is a valid change to make. I don't remember being involved in the implementation of the code under test or the test, so I'm going to loop in @donraab for opinions here.

Alright!

PR #1729, opened 2024-12-09
Migrating Verify to Junit5

Desislav-Petrov:


eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
```diff
@@ -2242,59 +2240,29 @@ public static void assertEqualsAndHashCode(String itemNames, Object objectA, Obj
...
-        assertEquals("Expected " + itemNames + " to be equal.", objectA, objectB);
-        assertEquals("Expected " + itemNames + " to be equal.", objectB, objectA);
+        assertEquals(objectA, objectA, "Expected " + itemNames + " to be equal.");
+        assertEquals(objectB, objectB, "Expected " + itemNames + " to be equal.");
+        assertEquals(objectA, objectB, "Expected " + itemNames + " to be equal.");
+        assertEquals(objectB, objectA, "Expected " + itemNames + " to be equal.");
         assertEquals(
-                "Expected " + itemNames + " to have the same hashCode().",
                 objectA.hashCode(),
-                objectB.hashCode());
```
Desislav-Petrov, line 2259: Newer versions of java don't like this - so time to clean up

eclipse-collections-code-generator/src/main/resources/test/set/mutable/boxedMutableHashSetTest.stg
```diff
@@ -101,9 +101,9 @@ public class BoxedMutable<name>SetTest
     public void remove()
     {
         BoxedMutable<name>Set set = this.classUnderTest();
-        Verify.assertTrue(set.remove(<(literal.(type))("1")>));
-        Verify.assertFalse(set.remove(<(literal.(type))("1")>));
-        Verify.assertFalse(set.remove("abc"));
+        assertTrue(set.remove(<(literal.(type))("1")>));
```
Desislav-Petrov, line 104: No point to do via Verify

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
```diff
@@ -65,28 +64,27 @@
...
+ * Adds useful additional "assert" methods.
  * You can import this class instead of Assert, and use it thus, e.g.:
  * <pre>
  *     Verify.assertEquals("fred", name);  // from original Assert class
  *     Verify.assertContains("fred", nameList);  // from new extensions
  *     Verify.assertBefore("fred", "jim", orderedNamesList);  // from new extensions
  * </pre>
  */
-public final class Verify extends Assert
+public final class Verify
```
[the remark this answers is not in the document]
Desislav-Petrov: Fair point - extending Assertions, however, the only thing that needed to change as a result of this are a few places. They had plain Java assertions used but I'd say Verify should only be used for the special cases it caters for, and all other generic cases should go straight to Junit5 Assertions.xxx ?

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
```diff
@@ -120,15 +118,15 @@ public static void assertEmpty(String iterableName, Iterable<?> actualIterable)
 
         if (Iterate.notEmpty(actualIterable))
         {
-            fail(iterableName + " should be empty; actual size:<" + Iterate.sizeOf(actualIterable) + '>');
+            Assertions.fail(iterableName + " should be empty; actual size:<" + Iterate.sizeOf(actualIterable) + '>');
```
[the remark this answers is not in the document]
Desislav-Petrov: There was a fail() method in Verify that was used in two places in other tests. Given that the standard Assertions class had a method with the same signature, I've decomed the one in Verify.

eclipse-collections-testutils/pom.xml
```diff
@@ -49,17 +49,17 @@
...
         </dependency>
 
         <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-api</artifactId>
-            <scope>test</scope>
+            <groupId>org.opentest4j</groupId>
+            <artifactId>opentest4j</artifactId>
+            <version>1.3.0</version>
+            <scope>compile</scope>
```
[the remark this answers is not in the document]
Desislav-Petrov: VerifyTest class has this - that's what Junit5 throws now.

performance-tests/pom.xml
```diff
@@ -64,6 +64,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
```
Desislav-Petrov, line 67: No longer comes transitively via the **testutils** module so had to be explicitly added.

Desislav-Petrov:
sry @motlin wasn't previously fully ready for review that's why I left it as a draft but should be good now - thanks for looking at it.

eclipse-collections-testutils/pom.xml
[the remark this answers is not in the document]
donraab: Not sure I understand. You're saying this is an explicit dependency of JUnit5?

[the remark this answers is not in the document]
donraab: @Desislav-Petrov, if this is a transitive dependency of JUnit5, but we don't reference any classes in any of the code, then why do we need to include it as an explicit dependency?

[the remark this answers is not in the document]
donraab: Ok, I expanded VerifyTest... this is the import:

`import org.opentest4j.AssertionFailedError;`

This is in a test only, not included in the test-utils jar, so why is it not a dependency in the unit tests pom instead?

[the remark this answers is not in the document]
Desislav-Petrov: So opentest4j comes transitively from junit5 and the enforcer plugin complains we're using a dependency that's not been explicitly declared:
image

Given the build does pass, it might be more sensible to suppress the plugin for this specific library in that module?

eclipse-collections-testutils/pom.xml
```diff
@@ -49,17 +49,17 @@
         </dependency>
 
         <!-- External dependencies -->
-
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
             <scope>compile</scope>
```
motlin, line 55:
```suggestion
```

motlin (changes requested):
Made minor suggestions to fix the dependency. I tested the changes locally.

Desislav-Petrov:
Thanks for the suggestion @motlin - implemented. Could you please take another look?

acceptance-tests/src/test/java/org/eclipse/collections/impl/forkjoin/FJIterateAcceptanceTest.java
```diff
@@ -60,6 +60,8 @@
 import org.junit.Before;
 import org.junit.Test;
 
+import static org.junit.Assert.assertThrows;
```
motlin, line 63: This is a JUnit 4 import. It looks like the sort of thing we did in other modules but I'm wondering if it's intentional that it's part of the same PR. It means this is not the last one, but includes prep for one more.

eclipse-collections-code-generator/src/main/resources/test/set/mutable/boxedMutableHashSetTest.stg
motlin: There might be a lot of examples like this. It's nice to clean them up and maybe we can come back and do another PR that gets more of them. IntelliJ can show them with a warning, because it's accessing a static method via a subclass. However, we have to temporarily configure IntelliJ to treat generated code as non-generated, or the IDE will suppress most warnings.

eclipse-collections-testutils/pom.xml
```diff
@@ -49,16 +49,15 @@
...
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
         </dependency>
 
         <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-api</artifactId>
+            <groupId>org.opentest4j</groupId>
+            <artifactId>opentest4j</artifactId>
+            <version>1.3.0</version>
```
motlin, line 60: not crucial but consider moving the version number to the `dependencyManagement` section of the root pom.xml where we define the version of junit-jupiter-api.

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/AbstractListTestCase.java
```diff
@@ -317,21 +317,7 @@ public void asUnmodifiable()
...
-        }
-        catch (Exception e)
-        {
-            // Suppress if a Java 9 specific exception related to reflection is thrown.
-            if (!e.getClass().getCanonicalName().equals("java.lang.reflect.InaccessibleObjectException"))
-            {
-                throw e;
-            }
-        }
+      // do nothing
```
motlin, line 320: Why is this test hollowed out?

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
```diff
@@ -65,28 +64,28 @@
...
+ * Adds useful additional "assert" methods.
  * You can import this class instead of Assert, and use it thus, e.g.:
  * <pre>
  *     Verify.assertEquals("fred", name);  // from original Assert class
  *     Verify.assertContains("fred", nameList);  // from new extensions
  *     Verify.assertBefore("fred", "jim", orderedNamesList);  // from new extensions
  * </pre>
  */
-public final class Verify extends Assert
+public final class Verify extends Assertions
```
motlin, line 88: I'd like to get input from @donraab. This is a source breaking change, only for users of Verify, in order to move from JUnit 4 to JUnit 5. I think it makes sense to do. I'd like to think through together if we've done everything to make the source breaking changes as small as possible.

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
```diff
@@ -98,11 +97,6 @@ private Verify()
         throw new AssertionError("Suppress default constructor for noninstantiability");
     }
 
-    public static void fail(String message, Throwable cause)
-    {
-        throw new AssertionError(message, cause);
-    }
```
motlin, line 104: I think it makes sense to keep methods like this to preserve source compatibility. We have a few options for the implementation though.

- Delegate to `Assert.fail()` as long as the arguments are compatible. If we go this route, we should mark the method Deprecated and write Javadoc saying that it's inline-able.
- Continue to just throw, but use an Error type from JUnit 5 if the parameters are compatible, or our own subclass of the JUnit 5 error type.
- Unchanged. I think JUnit 5 will work by catching JUnit 4 Error types.

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
```diff
@@ -240,14 +234,6 @@ public static void assertIterableEmpty(String iterableName, Iterable<?> iterable
         }
     }
 
-    /**
-     * Assert that the given object is an instanceof expectedClassType.
-     */
-    public static void assertInstanceOf(Class<?> expectedClassType, Object actualObject)
-    {
-        Verify.assertInstanceOf(actualObject.getClass().getName(), expectedClassType, actualObject);
-    }
```
motlin, line 249: Why delete this? You didn't delete the other override of the same method so it's not like it's unsupported.

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
motlin: But this is unrelated to the version of JUnit, isn't it? Where did you see that error? Locally, using newer versions of Java? In our GitHub workflows/actions?

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
```diff
@@ -2245,48 +2229,18 @@ public static void assertEqualsAndHashCode(String itemNames, Object objectA, Obj
...
-    @Deprecated
-    public static void assertShallowClone(Cloneable object)
-    {
-        Verify.assertShallowClone("object", object);
-    }
-
-    /**
-     * @deprecated since 8.2.0 as will not work with Java 9
-     */
-    @Deprecated
```
motlin, line 2274: This says it won't work with Java 9, but our builds pass?

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
donraab: @motlin I'm ok with this in theory. Hard to tell what issues this might run into in practice. What might be better, although would cause more work and potentially source breaks would be to break the inheritance relationship altogether. This means Verify can't be used as a means to automatically route calls to Assert or Assertions, but also means there won't be a similar issue when JUnit 6 happens.

motlin: I dread the amount of work it would be, but I'd like to "override" every single method, delegate from Verify to Assert or Assertions, and mark the new overrides as deprecated and inlineable.

Desislav-Petrov: I take no follow ups for this PR based on this conversation?

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
Desislav-Petrov: The essentially doesn't execute for Java 9+ because the exception Java throws is caught and ignored in the test:
image

eclipse-collections-testutils/pom.xml
Desislav-Petrov: Done

eclipse-collections-code-generator/src/main/resources/test/set/mutable/boxedMutableHashSetTest.stg
Desislav-Petrov: This was necessary when I removed the `extends Assertions` as it was breaking. Now that I have restored the Assertion, we can revert this and deal with in a separate PR.

Desislav-Petrov:
@motlin thanks for questioning some of the stuff I have done, it got me to dig a bit deeper and as a result I've restructured the PR. 

To cut a long story short, I didn't realise we run maven by default with *--add-opens java.base/java.lang=ALL-UNNAMED*. Believe all comments should be addressed now.

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
Desislav-Petrov: Reverted - it wasn't necessary at the end of the day. Same for the others.

acceptance-tests/src/test/java/org/eclipse/collections/impl/forkjoin/FJIterateAcceptanceTest.java
Desislav-Petrov: Reverted - no longer needed after your proposed pom dependency changes, just forgot to update.

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
Desislav-Petrov: Realised it was locally and due to not running with **--add-opens java.base/java.lang=ALL-UNNAMED**.

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
motlin: Looks like fail() is still removed. Does this clash with the generics in the super type?

```java
	public static <V> V fail(String message, Throwable cause) {
		AssertionUtils.fail(message, cause);
		return null; // appeasing the compiler: this line will never be executed.
	}
```

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
motlin: Same here, this method is still deleted. Does it clash with the generics in the super type?

```java
	@API(status = STABLE, since = "5.10")
	public static <T> T assertInstanceOf(Class<T> expectedType, Object actualValue) {
		return AssertInstanceOf.assertInstanceOf(expectedType, actualValue);
	}
```

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
```diff
@@ -1054,14 +1041,14 @@ public static void assertListsEqual(String listName, List<?> expectedList, List<
...
-        assertEquals(listName + " size", expectedList.size(), actualList.size());
+        assertEquals(expectedList.size(), actualList.size(), listName + " size");
         for (int index = 0; index < actualList.size(); index++)
         {
             Object eachExpected = expectedList.get(index);
             Object eachActual = actualList.get(index);
             if (!Objects.equals(eachExpected, eachActual))
             {
-                junit.framework.Assert.failNotEquals(listName + " first differed at element [" + index + "];", eachExpected, eachActual);
+                Assertions.assertEquals(eachExpected, eachActual, listName + " first differed at element [" + index + "];");
```
motlin, line 1051: Nitpick: this is a static import and doesn't need to be qualified.

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/ExceptionThrownTest.java
```diff
@@ -44,12 +42,12 @@ public void callableException()
     {
         try
         {
-            Verify.assertThrows(NullPointerException.class, new EmptyCallable());
+            assertThrows(NullPointerException.class, EmptyCallable::new);
```
motlin, line 45: This looks like it was intended to test `Verify.assertThrows()` and even though it doesn't make much sense anymore, because the method was removed, keeping the test the way it was asserts source backwards compatibility. Is it a linter violation to keep it? If it works, I think we should keep `Verify` here.

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/VerifyTest.java
```diff
@@ -68,70 +68,70 @@ public void assertThrowsWithCause()
         Verify.assertThrowsWithCause(RuntimeException.class, NullPointerException.class, (Callable<Void>) () -> {
             throw new RuntimeException(new NullPointerException());
         });
-        Verify.assertError(AssertionError.class, () -> Verify.assertThrowsWithCause(RuntimeException.class, NullPointerException.class, (Callable<Void>) () -> null));
```
motlin, line 71: I didn't remember this so I looked it up. AssertionFailedError extends AssertionError, so on the surface, it looks like this assertion might pass. But `Verify.assertError()` delegates to assertSame between the error class files, so it's a very strict assertion. This change makes sense to me.

motlin (commented):
Two nitpicks which we might want to address first but this is basically ready to land at this point.

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
donraab: @Desislav-Petrov  I think @motlin was agreeing on the breaking the inheritance model, but implementing every static method on `Assert` in `Verify`, and simply delegating the calls to the same static methods on `Assert`. I don't know how many methods this would mean implementing, but the implementations should be straightforward delegations.

The problem I have with this is that these are static methods, not instance methods. The fact that Java delegates from a subclass to a superclass for a static method is bizarre. This is not part of the public API of the class. I feel like this is more of a cheap trick, that unfortunately we leveraged for `Verify` to use it as a stand in for `Assert`. What is interesting is that static methods on classes behave differently than static methods on interfaces. For instance, `MutableList` extends `List`, but we are unable to say `MutableList.of()`, thankfully.

There are 55 public static methods on `Assert` that would have to be replicated on `Verify`.

The benefit I see from this change @Desislav-Petrov is that we don't extend the problem of folks using Verify as an alternative to Assertions. If there is not a method on `Verify`, they should use `Assertions`. I'd prefer if folks used `Assertions` more than `Verify`. `Verify` should only be used for `Collection` specific assertions. That was what its original purpose was (I think).

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
Desislav-Petrov: Yes - removed because of the clash.

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
Desislav-Petrov: Yes - removed because of the clash.

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/ExceptionThrownTest.java
Desislav-Petrov: done - cast is needed though because the type between Verify.assertThrows and the Assertions.throws doesn't match and the method call becomes ambiguous.

Desislav-Petrov:
Thanks for the review - should be addressed now @motlin

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
motlin: I'm curious how this manifested. Was it a compiler error?

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/ExceptionThrownTest.java
```diff
@@ -30,12 +28,12 @@ public void specificRuntimeException()
     {
         try
         {
-            assertThrows(NullPointerException.class, new EmptyRunnable());
+            assertThrows(NullPointerException.class, EmptyRunnable::new);
```
motlin, line 31: This implies we have a source breaking change?

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/ExceptionThrownTest.java
```diff
@@ -44,12 +42,12 @@ public void callableException()
     {
         try
         {
-            Verify.assertThrows(NullPointerException.class, new EmptyCallable());
+            Verify.assertThrows(NullPointerException.class, (Callable<?>) EmptyCallable::new);
```
motlin, line 45: Is this a Callable that returns Callable?

unit-tests-java8/src/test/java/org/eclipse/collections/test/FixedSizeIterableTestCase.java
```diff
@@ -14,7 +14,7 @@
 
 import org.junit.jupiter.api.Test;
 
-import static org.eclipse.collections.impl.test.Verify.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
motlin, line 17: Does the code compile if you don't swap this import?

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
Desislav-Petrov: [2334,23] error: name clash: assertInstanceOf(Class<?>,Object) in Verify and <T>assertInstanceOf(Class<T>,Object) in Assertions have the same erasure, yet neither hides the other
  where T is a type-variable:
    T extends Object declared in method <T>assertInstanceOf(Class<T>,Object)


Compilation error - yes.

unit-tests-java8/src/test/java/org/eclipse/collections/test/FixedSizeIterableTestCase.java
Desislav-Petrov: Revertable now - updating.

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/ExceptionThrownTest.java
Desislav-Petrov: ` private static final class EmptyCallable implements Callable<Void> ` - callable that returns nothing. Funny enough - I can switch to new EmptyCallable() and it detects the type correctly, but not via the ::new expression.

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/ExceptionThrownTest.java
Desislav-Petrov: This call used to be based on the Junit4 assertThrows method, now it's based on Junit5. The signature of the method changes from Junit4 to 5 so as a result of the migration we had to change that.

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
donraab: @motlin I had a chat today with @Desislav-Petrov and we agreed on the approach which would be taken up in a separate PR. 

Assert has 55 methods. Assertions has 24 methods in common with Assert. Assert has 32 methods not on Assertions. Assertions has another 267 methods that are new.

Extending Assertions would be a galactic explosion to the public methods on Verify.

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/ExceptionThrownTest.java
motlin: The expressions don't have the same type, so it makes sense. Roughly:

- `new EmptyCallable()` is an `EmptyCallable`, or `Callable<Void>`
- `EmptyCallable::new` is a `Callable<EmptyCallable>`, or `Callable<Callable<Void>>`

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/ExceptionThrownTest.java
motlin: This has the same "problem" as the code below. It's no longer a runnable that does nothing, it's now a runnable that creates a runnable. This isn't really a problem on its own, but it no longer proves that Verify type signatures are source compatible with the previous version.

eclipse-collections-testutils/src/main/java/org/eclipse/collections/impl/test/Verify.java
motlin: I think I follow the options that you discussed, but could you summarize your conclusion?

donraab: `Verify` should extend nothing. Implement all of the methods from `Assert` on `Verify`, and deprecate all of them, forwarding to appropriate `Assertions` alternatives.

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/ExceptionThrownTest.java
Desislav-Petrov: So.. we can override assertThrows in Verify to preserve the signature?

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/ExceptionThrownTest.java
Desislav-Petrov: What's your advice then?

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/ExceptionThrownTest.java
motlin: I think this change can be reverted. If there's a compiler error, we can probably fix it by adding generics here `Verify.<here>assertThrows(`

Desislav-Petrov: Indeed - it's been reverted.

PR #1731, opened 2025-01-06
Add a test of Verify derived from JUnit's test for Assert.

motlin:
The idea is that this new test won't change in #1729 as long as the changes there are source compatible. cc @Desislav-Petrov


eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/AssertTest.java
motlin, line 1: See https://github.com/junit-team/junit4/blob/main/src/test/java/org/junit/tests/assertion/AssertionTest.java

eclipse-collections-testutils/src/test/java/org/eclipse/collections/impl/test/AssertTest.java
```diff
@@ -0,0 +1,1221 @@
...
+import org.junit.internal.ArrayComparisonFailure;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.is;
+
+/**
+ * Based on the test org.junit.tests.assertion.AssertionTest in JUnit 4, replacing all usages of {@link Assert} with {@link Verify}. Ensures that {@code Verify} remains source compatible with {@code Assert}.
+ */
+@SuppressWarnings("StaticMethodReferencedViaSubclass")
```
motlin, line 19: This suppression stops the IDE from warning that these methods should be called through Assert.

Desislav-Petrov:
@motlin I'm thinking we should probably merge this after https://github.com/eclipse-collections/eclipse-collections/pulls, alongside the Verify uplift?

motlin:
> @motlin I'm thinking we should probably merge this after [eclipse-collections/eclipse-collections/pulls](https://github.com/eclipse-collections/eclipse-collections/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen), alongside the Verify uplift?

I tried cherry picking the JUnit 5 changes on top and they don't compile, because methods like Verify.assertArrayEquals no longer have a string as their first parameter (it moved to the last parameter).

Ideally this should land first, so we can see every source breaking change.

motlin:
@donraab I think that https://github.com/eclipse-collections/eclipse-collections/pull/1729 is basically ready to go but that this one ought to land first. WDYT?

donraab (commented):
@motlin Is the plan to add methods from `Assert` to `Verify` once we switch over to JUnit 5 and remove the dependency on `Verify` extending `Assert`? My preference would be to NOT add functioning JUnit 4 `Assert` methods to `Verify` to provide source compatibility. I would be ok with adding deprecated methods to `Verify` that throw `UnsupportedOperationException` for all current `Assert` methods. The deprecated message would then indicate the appropriate method to use on `Assertions`. We should be making a statement with the release of EC 12.x that all the modules requires Java 11+ and JUnit 5 is required for test-utils. What do you think?

motlin:
@donraab I have no implementation plan for Verify. I just want to increase test coverage of the API. When we change it, whether we break source compatibility or preserve it, I want it to be visible in tests.

donraab:
> @donraab I have no implementation plan for Verify. I just want to increase test coverage of the API. When we change it, whether we break source compatibility or preserve it, I want it to be visible in tests.

Ok, makes sense.

donraab (approved):
LGTM. There is a failing JavaDoc build, but once that is fixed looks good to merge.

motlin:
> LGTM. There is a failing JavaDoc build, but once that is fixed looks good to merge.

The Javadoc failure looks like a bug in Java 11. The other Javadoc build using the EA build continues to pass.

motlin:
I tried a bunch of things to get the Javadoc 11 build to pass. I eventually gave up and added a Javadoc 17 build which passed on the first try.

Desislav-Petrov:
@motlin - are u still planning to merge that?

PR #1734, opened 2025-02-19
Refactor rehash() method to improve performance by replacing equals()… Fix resolved #issues/1728

bazzoneto:
… with == for object comparison

Refactor rehash() method to improve performance by replacing equals() with == for object comparison

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
@@ -334,54 +334,54 @@ public class UnifiedSet<T>
     }
 
     @Override
-    protected void rehash(int newCapacity)
-    {
-        int oldLength = this.table.length;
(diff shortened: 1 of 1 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/set/mutable/UnifiedSet.java
```diff
@@ -334,54 +334,54 @@ private boolean chainedAdd(T key, int index)
...
+    int oldLength = this.table.length;
+    Object[] old = this.table;
+    this.allocate(newCapacity);
+    this.occupied = 0;
 
-        for (int i = 0; i < oldLength; i++)
+    for (int i = 0; i < oldLength; i++)
+    {
+        Object oldKey = old[i];
+        if (oldKey instanceof ChainedBucket)
```
motlin, line 347: There are some space-only changes here that ought to be reverted.

motlin (changes requested):
It looks like the only changes are formatting changes.

PR #1745, opened 2025-05-22
Add default toArray implementations to LazyIterable.

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/AbstractBag.java
```diff
@@ -82,6 +82,18 @@ public abstract class AbstractBag<T>
         extends AbstractRichIterable<T>
         implements Collection<T>, Bag<T>
 {
+    @Override
+    public Object[] toArray()
```
donraab, line 86: Are these overrides necessary?

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/immutable/AbstractImmutableCollection.java
```diff
@@ -53,6 +53,18 @@ public abstract class AbstractImmutableCollection<T>
 {
     protected abstract MutableCollection<T> newMutable(int size);
 
+    @Override
+    public Object[] toArray()
```
donraab, line 57: Are these overrides necessary?

donraab (commented):
Mostly LGTM. Just a couple questions about overrides that call super and do nothing else.

eclipse-collections/src/main/java/org/eclipse/collections/impl/collection/immutable/AbstractImmutableCollection.java
motlin: Same here:

```
[ERROR]   class AbstractImmutableCollection<T#2> inherits abstract and default for toArray(E#2[]) from types RichIterable and Collection
[ERROR]   where T#1,E#1,T#2,E#2 are type-variables:
```

eclipse-collections/src/main/java/org/eclipse/collections/impl/bag/AbstractBag.java
donraab: ok, thanks!

donraab (approved):
LGTM

PR #1767, opened 2025-06-16
Simplify p2 repository generation

fipro78:
This PR simplifies the build of a p2 update site.

It removes the explicit target definition and configures it as pom-first. This way the dependencies are configured via the p2-feature/pom.xml and there is no need for an additional target file for this task.


nikhilnanivadekar:
@fipro78 when I do the release, it’s from a tag. And the tag should have updated versions. See here: https://github.com/eclipse-collections/eclipse-collections/releases/tag/12.0.0.M20.JDK11

Listing down the changes required because I want to make sure I understand the changes, so that I can automate the pom updates to ensure things work in a repeatable way with least manual updates.

Update all the pom files to have the version number and delete all target files. There are some nested pom files as well, do I need to update those?

fipro78:
There are no versions in the nested pom files. You only need the update in case you change the version in the parent pom.

nikhilnanivadekar:
That’s where I am confused, if you look at the code as of the tag (link above), the parent pom is updated. So, I wonder what is missing? Is this all because of the presence of the target files? Happy to cut another milestone to test it out.

fipro78:
But I don't see any version update in the p2-feature and sub-folders there. At least not in the commit.

nikhilnanivadekar:
I meant `{project.version}` would be inherited, so ideally it should work. But I see your point. I’ll merge this PR and try to build from master to test it out. If it works, I’ll update the build to ensure these pom files are also updated as a part of the build. Thanks for your patience and help

fipro78:
The p2-feature projects are not part of the whole build reactor. Maybe that is why the update was not performed automatically. The reason is that the p2-feature build is a tycho build that would interfere with your normal build. At least that was my impression back then. Not sure if this could be fixed with the updated variant without the target files now.

nikhilnanivadekar:
Good point. I’ll update the poms via the plugin with file paths. That should work around the reactor issue

nikhilnanivadekar (approved):

nikhilnanivadekar:
@fipro78 do I need to change the version in the below files at the lines of code I have tagged?

1. https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-feature/pom.xml#L15
2. https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-feature/org.eclipse.collections.repository/pom.xml#L18
3. https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-repository/pom.xml#L9
4. https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-repository/org.eclipse.collections/pom.xml#L9
5. https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-repository/org.eclipse.collections/pom.xml#L21
6. https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-repository/org.eclipse.collections/pom.xml#L26

I will need to create a separate replacer plugin step which is fine, just want to make sure I cover all the necessary files.

fipro78:
@nikhilnanivadekar 

1. yes
2. yes
3. no
4. no
5. no
6. no

you missed https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-feature/org.eclipse.collections.feature/pom.xml#L18 in your list

The folder p2-repository is actually obsolete now. The whole p2 repository build is now in p2-feature. So p2-repository can be removed. I just didn't do it in case anything is not working as intended and we need to revert back to the old p2 build variant.

fipro78:
@nikhilnanivadekar 
wait, the version of the parent pom in p2-feature is questionable. The version of the p2 feature is actually derived from there. And the version you used for the milestone is actually invalid. At least the Tycho build fails because it does not follow the typical pattern major-minor-bugfix(-SNAPSHOT). Not sure if this is only an issue in Tycho or general.

nikhilnanivadekar:
Sorry, I don’t follow. I ran a build on master after the PR was merged. The p2 artifacts should be published as mentioned here: https://github.com/eclipse-collections/eclipse-collections/issues/1741#issuecomment-2978205241

So, what needs to be fixed with versions?

fipro78:
Sorry, my fault. Actually you don't need to change the version if you are fine with the following result:

image

Tycho derives the version of the p2 feature from the p2-feature/pom.xml and replaces the -SNAPSHOT with the build timestamp. So you can only see from the update site URL that you are consuming a milestone. But the milestone version can not be used, as it is invalid in the OSGi context. At least Tycho will not accept such a version, as it can not derive the proper version with a build timestamp.

From my point of view that should be ok. So only the versions that I changed with this PR need to be adjusted. And of course once you create a new version 12.1.0 or so, you need to update the parent pom.xml versions that I mentioned in the previous post.

nikhilnanivadekar:
I don’t follow this: _ But the milestone version can not be used, as it is invalid in the OSGi context. At least Tycho will not accept such a version, as it can not derive the proper version with a build timestamp._

how do we fix it? And what’s the right thing to do? Sorry for my incessant questions because I am not an OSGi expert/user.

fipro78:
don't worry, it's ok that you ask questions. :)

In the p2-feature/pom.xml file we have the version **12.0.0-SNAPSHOT**. Tycho replaces **-SNAPSHOT**  with the build timestamp, so the version of the features deployed in the p2 update site are **12.0.0.v20240611-0433**. IIRC that would be the timestamp of the last commit included as we use the `tycho-sourceref-jgit` plugin to create the version qualifier out of it.

Now your milestone version is **12.0.0.M20.JDK11**. And if I read https://semver.org/ correctly, it is a valid semantic version. But Tycho does not understand it and fails with the following error message if the parent version is set this way:

```
org.apache.maven.InternalErrorException: Internal error: java.lang.IllegalArgumentException: Format "format(n[.n=0;[.n=0;[-S]]])" was unable to parse 12.0.0.M20.JDK11
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:109)
    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:903)
    at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:280)
    at org.apache.maven.cli.MavenCli.main (MavenCli.java:203)
    at jdk.internal.reflect.DirectMethodHandleAccessor.invoke (DirectMethodHandleAccessor.java:103)
    at java.lang.reflect.Method.invoke (Method.java:580)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced (Launcher.java:255)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launch (Launcher.java:201)
    at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode (Launcher.java:361)
    at org.codehaus.plexus.classworlds.launcher.Launcher.main (Launcher.java:314)
Caused by: java.lang.IllegalArgumentException: Format "format(n[.n=0;[.n=0;[-S]]])" was unable to parse 12.0.0.M20.JDK11
    at org.eclipse.equinox.internal.p2.metadata.VersionFormat.parse (VersionFormat.java:307)
    at org.eclipse.equinox.internal.p2.metadata.VersionParser.parse (VersionParser.java:167)
    at org.eclipse.equinox.p2.metadata.Version.create (Version.java:96)
```

So probably my statement was incorrect and it is neither a general issue nor an issue in OSGi. But it looks like to be an issue in p2 as it does not understand the version format in this way.

What to do with this? Well there are options:
1. Ask @laeubi if there is something you could do here. To be honest, I am getting out of ideas here, because I never used such version formats. And although I tried to get around this, I was not able to find a solution to deal with it.
2. Live with it. As I said, the version is only used for the p2 feature, which is kind of a group of the Eclipse Collections API and IMPL bundles. The bundle versions are resolved as **Bundle-Version: 12.0.0.M20JDK11** which looks good to me. Although Tycho is not able to resolve the sources, which are then missing in the p2 repository

```
[WARNING] The following referenced plugins have missing sources
        org.eclipse.collections.api_12.0.0.M20JDK11
        org.eclipse.collections.impl_12.0.0.M20JDK11
```

which is obvious, because the bundle version is modified by concatenating the last two fragments to a single one. This issue should be gone with the release once the two fragments in the version (milestone + JDK) are gone.

If you don't want the date as qualifier of the feature, you could also use a property to force it to a specific value

```xml
<forceContextQualifier>M20JDK11</forceContextQualifier>
```

but if this is really making it better, I am not sure.

Short:
The issue is that the p2 version does not understand the format you are using. But using the -SNAPSHOT will result in replacing it with a timestamp. Just for the feature. 

But maybe @laeubi has an idea on this.

nikhilnanivadekar:
Got it, so to confirm only these need to be updated:

1. https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-feature/org.eclipse.collections.repository/pom.xml#L18
2. https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-feature/org.eclipse.collections.feature/pom.xml#L18
3. https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-feature/pom.xml#L46
4. https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-feature/pom.xml#L51

The selective update using the replacer plugin might be a bit tricky, but I will see what I can do 😄

fipro78:
1 and 2:
yes, but it needs to stay with a -SNAPSHOT to avoid the build error.
And additionally there is https://github.com/eclipse-collections/eclipse-collections/blob/14a7d7ff2cb14bd4198cf6949a68ffbd7fda4f02/p2-feature/org.eclipse.collections.repository/pom.xml#L18

Not to be confused with the p2-repository project one level higher.

2 and 4:
yes, there you need to use the version to consume (e.g. 12.0.0.M20.JDK11). Of course you could add a property, then it could be only a single place you need to take care of.

laeubi:
I don't think one can do much about this, one would expect to use a valid P2 Version for P2 managed items (like a p2 feature or update site). Even though there are many version schemes not all are compatible (e.g. maven <-> OSGi <-> p2 <-> ...) or leading to unexpected results (e.g. `3.9.0-alpha` will be included in a range `[3,3.9)` by maven).

One might argue that Tycho should replace the version here with something cleaned up (like it does in the manifest), but I think its a quite uncommon case, still it might be worth to provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue.

nikhilnanivadekar:
@fipro78 lets try this a bit differently because I am a tad confused 😄 

can you please give which pom files need to change with which version please mention which ones need `-SNAPSHOT` and which ones need a version in maven central? Sorry for asking this, between the `SNAPSHOT` and concrete versions, I am totally going to make a mistake

fipro78:
Seems I am also getting confused in the meanwhile. Ok let's make it concrete:

https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-feature/pom.xml#L15
https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-feature/org.eclipse.collections.feature/pom.xml#L18
https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-feature/org.eclipse.collections.repository/pom.xml#L18

Here use the -SNAPSHOT version needed for the p2 version formatting. Basically the p2 build parent pom version. You can stick here with the current 12.0.0-SNAPSHOT, as 12.0.0 is the upcoming release. Once that is done and you do the update for the next release (bugfix or minor) you need to update the version part once.

https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-feature/pom.xml#L46
https://github.com/eclipse-collections/eclipse-collections/blob/master/p2-feature/pom.xml#L51

Here you need to set the version that you want to consume and integrate into the p2 repository. Simple Maven dependency. This version needs to be changed everytime you do a release. So everytime a new milestone or release version is published and you want to provide the matching p2 update site, the update needs to be done. 
I suppose these two are the only places you need to cover with your replacer, as the -SNAPSHOT stuff does not change that often. And you could reduce that to a single place that needs to be changed, if you configure the Maven version as a property that you use in the two places (tiny improvement).

Does that make it clearer now?

nikhilnanivadekar:
This is perfect and clarifies my understanding!

nikhilnanivadekar:
I tried a bit to automate the replacements however, I did run in to some unforeseen issues. Since these steps are repeatable, I am going to update these instructions in the release steps. This way I can at least take care of the release and then work on automating the replacements. Thanks a ton for your help!

PR #1780, opened 2025-07-31
 Fix Map.merge to handle null values without invoking remapping function

MishaDemianenko:
This PR updates Map.merge to correctly handle entries with null values by bypassing the remapping function and directly storing the provided value.
The change was introduced in PR #1542 and violates the optional Map.merge contract, deviating from the previous behavior of maps affected by mentioned PR.


mohrezaei:
PR #1542 changes more than just UnifiedMap. Please fix the others.

MishaDemianenko:
@mohrezaei updated. I have updated PR name and description as well since now its a bit more than just UnifiedMap patch

motlin:
Thank you very much @MishaDemianenko! If you have the energy for it, there are other methods that we ought to implement and that I could use help with, like Map.compute(). They all require this much care to make sure we get the contract right.

PR #1781, opened 2025-08-07
Fix updateValue method ordering in primitiveObjectHashMap template. Fixes #1517.

### LINKED ISSUE #1517: Unexpected behaviour in updateValue() with empty key
This issue exits in Eclipse Collections framework. Unexpected behaviour in, for instance, IntObjectHashMap.updateValue() with empty key. Apparently, the issue will be reproduced also in any PrimitiveObjectHashMap and in updateValueWith() method.

Steps to reproduce
1. Create new `IntObjectHashMap map = new IntObjectHashMap()`;.
2. Update value for empty key: `map.updateValue(0, () -> null, (existingValue) -> { if (map.containsKey(0)) throw new RuntimeException("key 0 does not exits at the moment"); return new Object();});`
3. Exception is thrown

Currently, map has flag `containsZeroKey` before the value for zero key was evaluated. It leads to true result of `map.containsKey(0)` call. That's quite unexpected and it's different from behaviour for non empty key. `map.containsKey(0)` returning false inside function expected before value for zero key was inserted to the map.

I assume, the issue will be reproduced also in any PrimitiveObjectHashMap and in `updateValueWith()` method.
Just replacing `this.sentinelValues.containsZeroKey = true;` after `this.sentinelValues.zeroValue = function.valueOf(factory.value());` is enough for fixing this

motlin:


motlin:
@GauriKokate6 @romanvmarkov here's a fix for #1517

eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
```diff
@@ -2454,17 +2454,18 @@ public class <name>ObjectHashMap\<V> implements Mutable<name>ObjectMap\<V>, Exte
             if (this.sentinelValues == null)
             {
                 this.sentinelValues = new SentinelValues\<>();
```
mohrezaei: in this case, it's better to invoke the function before mutating `this.sentinelValues`. Something like:
```
SentinelValues s = new SentinelValues();
s.zeroValue = function.valueOf(factory.value());
s.containsZeroKey = true;
this.sentinelValues = s;
```

and similarly below

mohrezaei (changes requested):

mohrezaei:
I was thinking through the logic here and it occurred to me that while the original issue was weird (the test case in the issue doesn't make sense), the test case in this PR does make sense. It effectively expects a new invariant, which can be summarized as:

**_If a collection mutation-function is invoked with a function-as-parameter that eventually throws, the collection is left in the same state as prior to the mutation-function invocation._**

If this is the behavior we want/expect, we should add the hundreds of missing test cases. Test cases should cover the special values (e.g. sentinels) and assert on contents and other state, e.g. size.

Out of curiosity, I checked primitive-primitive map, and that one serendipitously avoids this particular problem on this particular function. Checking all the others would be quite a big effort.

mohrezaei:
Functions that have been verified for the new invariant should also be documented as such.

@motlin please add test cases for non-sentinel values. Assert size in all test cases. Document this invariant for this function.

motlin:
Thanks for the thorough review and the feedback @mohrezaei. I addressed most of it. As you pointed out, there are probably many places with missing test coverage, and I've stuck to just updateValue and updateValueWith for now.

eclipse-collections-code-generator/src/main/resources/impl/map/mutable/primitiveObjectHashMap.stg
```diff
@@ -2453,37 +2453,41 @@ public class <name>ObjectHashMap\<V> implements Mutable<name>ObjectMap\<V>, Exte
         {
             if (this.sentinelValues == null)
             {
-                this.sentinelValues = new SentinelValues\<>();
-                this.sentinelValues.containsZeroKey = true;
-                this.sentinelValues.zeroValue = function.valueOf(factory.value());
+                SentinelValues\<V> sentinelValues = new SentinelValues\<>();
+                sentinelValues.zeroValue = function.valueOf(factory.value());
+                sentinelValues.containsZeroKey = true;
+                this.sentinelValues = sentinelValues;
```
motlin, line 2459: Rather than move the lambda call up, I chose to just assign the field at the end.

eclipse-collections-code-generator/src/main/resources/test/map/mutable/primitiveObjectHashMapTest.stg
```diff
@@ -586,6 +586,114 @@ public class <name>ObjectHashMapTest extends AbstractMutable<name>ObjectMapTestC
...
+        RuntimeException functionException = new RuntimeException("Function exception");
+
+        RuntimeException actualException1 = assertThrows(
+                RuntimeException.class, () ->
+                        map.updateValue(
+                                key,
+                                () -> { throw factoryException; },
+                                v -> { throw functionException; }));
+
+        assertSame(factoryException, actualException1);
```
motlin, line 640: I settled on a test where all lambdas throw, we catch what is thrown, and assert that what we caught is the exact instance we expect to be thrown.

eclipse-collections-code-generator/src/main/resources/test/map/mutable/primitiveObjectHashMapTest.stg
```diff
@@ -586,6 +586,114 @@ public class <name>ObjectHashMapTest extends AbstractMutable<name>ObjectMapTestC
...
+        assertTrue(!map.containsKey(key));
+        assertNull(map.get(key));
+        assertEquals(expectedSize, map.size());
```
motlin, line 643: belt-and-suspenders assertions that we don't contain the key, get returns null, and the size hasn't changed.

eclipse-collections-code-generator/src/main/resources/test/map/mutable/primitiveObjectHashMapTest.stg
```diff
@@ -586,6 +586,114 @@ public class <name>ObjectHashMapTest extends AbstractMutable<name>ObjectMapTestC
...
+
+        map.put(key, "temp");
+        RuntimeException actualException2 = assertThrows(
+                RuntimeException.class, () ->
+                        map.updateValue(
+                                key,
+                                () -> { throw factoryException; },
+                                v -> { throw functionException; }));
+
+        assertSame(functionException, actualException2);
```
motlin, line 653: Here the key is present, so we expect that the function gets called (and throws) and the factory is not called.

mohrezaei (approved):

mohrezaei:
>  I've stuck to just updateValue and updateValueWith

Yeah, sorry, I didn't mean to imply those changes should go into this PR 😄 
While the total amount of work is large, it's trivial to do it piecemeal.

PR #1782, opened 2025-08-09
Improve rehash() performance by eliminating unnecessary equals() calls

### LINKED ISSUE #1728: '`rehash()'

znight1020:
This PR resolves issue #1728 by optimizing the `UnifiedSet.rehash()` method. The change introduces a specialized `rehashAdd()` helper method that avoids the redundant `equals()` checks performed by the general-purpose `add()` method during the rehashing process.

### Summary of Changes

- A new private method `rehashAdd(T key)` has been added to `UnifiedSet`. This method is designed specifically for the rehash operation and omits the unnecessary `equals()` checks, as all elements being rehashed are guaranteed to be unique.
- The `rehash(int newCapacity)` method has been updated to use `rehashAdd()` instead of `add()`.

### Performance Benchmark

To validate the performance impact of this change, I conducted a series of JMH benchmarks. The benchmark was specifically designed to measure the effect of this optimization under varying degrees of hash collision, which was the core concern of the original issue.

- **Test Setup:** 1,000,000 objects were added to a `UnifiedSet`.
- **Variable:** The hash collision rate was controlled by adjusting the `hashCode()` implementation (`id % collisionFactor`).
- **Benchmark Protocol:** Each scenario was run with 3 warmup iterations, followed by 10 measurement iterations. The scores in the table below represent the average of these 10 measurement iterations.
- **Benchmark Code:** The full benchmark setup and code can be found in this repository for complete reproducibility: https://github.com/znight1020/ec-test

### Benchmark Results

The results clearly demonstrate that the performance improvement is directly proportional to the hash collision rate. The optimization provides a significant gain in scenarios with high hash collisions.

*(All scores are in milliseconds per operation - lower is better)*

| Collision Rate | `hashCode()` | Before (ms/op) | After (ms/op) | **Improvement** |
| :--- | :--- | ---:| ---:| :---:|
| **High** | `id % 100` | 68,772.8 | 47,852.3 | **30.4%** |
| **Medium** | `id % 500` | 9,587.8 | 9,079.8 | **5.3%** |
| **Low** | `id % 1000` | 4,911.2 | 4,713.1 | **4.0%** |

### Analysis

The data confirms that under conditions of severe hash collision (i.e., when `collisionFactor` is low), the redundant `equals()` calls in `rehash()` become a significant source of performance degradation. The implemented `rehashAdd()` method successfully mitigates this issue, delivering a performance improvement of approximately 30% in this worst-case scenario.

Fixes #1728


mohrezaei:
LGTM

mohrezaei:
A better benchmark would be for an object that has a complex `equals()` method, with normal collision expectations.

znight1020:
Thank you for the feedback! @mohrezaei

 I completely agree that seeing the results for a more "normal" use case is important.

My initial benchmark was focused on the worst-case scenario to clearly demonstrate the effect of the optimization.

Would it be helpful if I add another benchmark using an object with a more complex `equals()` and a well-distributed `hashCode()` (to simulate "normal collision expectations") and then update the PR with both sets of results?

mohrezaei:
> Would it be helpful if I add another benchmark 

It would be interesting, but not necessary. The worst equals methods we've seen (really, an anti-pattern) are those that allocate large strings inside `equals()`.

PR #1783, opened 2025-08-15
Add exception safety tests for getIfAbsentPut methods.

motlin:


mohrezaei:
I suggest documenting the function's behavior.
Also, I assume `UnifiedMapWithHashingStrategy` is covered in here somehow (it's a bit hard to tell with the abstractions)?

motlin:
I added the Javadoc:

> This method guarantees that if the factory function throws an exception, the map will be left in the same state as prior to the method invocation.

I hope that someday soon, we'll know that this is true of all methods, and it won't be worth mentioning in the docs any more.

Yes, `UnifiedMapWithHashingStrategy` is also covered by tests.

mohrezaei:
> I hope that someday soon, we'll know that this is true of all methods, and it won't be worth mentioning in the docs any more.

I think this guarantee needs to be documented somewhere, possibly not on a per-method basis, but somewhere central?

mohrezaei (approved):

PR #1786, opened 2025-08-21
Fix updateValue exception safety in UnifiedMap and UnifiedMapWithHashingStrategy.

motlin:
I have a branch where I've added a bunch of tests and most of them pass, but it turns out that updateValue() and updateValueWith() actually have the bug so I figured I'd fix it first.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
@@ -424,8 +424,8 @@ public class UnifiedMap<K, V> extends AbstractMutableMap<K, V>
         Object cur = this.table[index];
         if (cur == null)
         {
-            this.table[index] = UnifiedMap.toSentinelIfNull(key);
             V result = function.valueOf(factory.value());
+            this.table[index] = UnifiedMap.toSentinelIfNull(key);
             this.table[index + 1] = result;
             ++this.occupied;
             return result;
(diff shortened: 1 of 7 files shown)


mohrezaei:
> have the bug

Not a bug. Unimplemented feature perhaps?
I wonder if the JDK functions have this guarantee, or even all function-taking mutating-functions across other libs??

mohrezaei:
Please add a note to the docs

mohrezaei:
> JDK functions have this guarantee

Functions such as this:
```java
public void replaceAll(BiFunction<? super K, ? super V, ? extends V> function)
```
do not have such a guarantee (for more or less obvious reasons), so it's certainly not universal.

motlin:
Lots of methods document their behavior in Javadoc in terms of an implementation. Map.replaceAll() says:

```java
@implSpec
     * <p>The default implementation is equivalent to, for this {@code map}:
     * <pre> {@code
     * for (Map.Entry<K, V> entry : map.entrySet())
     *     entry.setValue(function.apply(entry.getKey(), entry.getValue()));
     * }</pre>
     *
     * <p>The default implementation makes no guarantees about synchronization
     * or atomicity properties of this method. Any implementation providing
     * atomicity guarantees must override this method and document its
     * concurrency properties.
```

I see that Map.merge() does document the property we want.

```java
     * If the remapping function itself throws an (unchecked) exception, the
     * exception is rethrown, and the current mapping is left unchanged.
```

motlin:
I accidentally already added the Javadoc in the previous PR. I was working on all this at the same time and didn't split up my commits properly. Sorry about that, but it does mean this PR is ready to land.

https://github.com/eclipse-collections/eclipse-collections/blob/3b3c55a47406ca11ece6c077b96c9564f73dfc53/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java#L145

PR #1787, opened 2025-08-23
Add exception safety tests.

motlin:
I added tests for four more methods. I didn't add Javadoc for the Map methods because the contacts in the JDK already say that the methods are safe from mutation in the face of exceptions.

### DIFF
--- a/unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MapTestCase.java
+++ b/unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MapTestCase.java
@@ -295,9 +295,13 @@ public interface MapTestCase
         assertIterablesEqual(this.newWithKeysValues(1, "1", 2, "2Two", 4, "4", 5, "5"), map);
 
         // existing key, remapping function throws exception
-        assertThrows(IllegalArgumentException.class, () -> map.merge(4, "Four", (v1, v2) -> {
-            throw new IllegalArgumentException();
+        RuntimeException exception = new RuntimeException("Test exception");
+        RuntimeException actualException1 = assertThrows(RuntimeException.class, () -> map.merge(4, "Four", (v1, v2) -> {
+            assertEquals("4", v1);
+            assertEquals("Four", v2);
+            throw exception;
         }));
+        assertSame(exception, actualException1);
         assertIterablesEqual(this.newWithKeysValues(1, "1", 2, "2Two", 4, "4", 5, "5"), map);
 
         // existing key with null value, remapping function is not called and new value is used
--- a/unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MapTestCase.java
+++ b/unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MapTestCase.java
@@ -21,9 +21,11 @@ import org.eclipse.collections.api.set.MutableSet;
 import org.eclipse.collections.impl.tuple.ImmutableEntry;
 import org.junit.jupiter.api.Test;
 
+import static org.eclipse.collections.impl.test.Verify.assertSize;
 import static org.eclipse.collections.test.IterableTestCase.assertIterablesEqual;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertSame;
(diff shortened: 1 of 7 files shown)


unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MapTestCase.java
```diff
@@ -310,6 +337,183 @@ default void Map_merge()
...
+            assertNotNull(k);
+            assertNotNull(v);
+            return v + k;
+        });
+        assertIterablesEqual(this.newWithKeysValues(1, "11", 2, "22", 3, "33"), map);
+
+        if (this.supportsNullValues())
+        {
+            Map<Integer, String> map2 = this.newWithKeysValues(1, "1", 2, "2");
+            map2.replaceAll((k, v) -> null);
```
mohrezaei, line 455: `supportsNullValues` requires a `containsKey` check

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MapTestCase.java
```diff
@@ -310,6 +337,183 @@ default void Map_merge()
...
+        RuntimeException actualException = assertThrows(RuntimeException.class, () -> map3.replaceAll((k, v) -> {
+            if (k.equals(2))
+            {
+                throw exception;
+            }
+            return v + "Modified";
+        }));
+        assertSame(exception, actualException);
+        assertNotNull(map3);
+        assertTrue(map3.containsKey(1) || map3.containsKey(2) || map3.containsKey(3));
```
mohrezaei: I think you meant "&&", right?
Also assert size.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MutableMapIterableTestCase.java
```diff
@@ -160,6 +160,37 @@ default void MutableMapIterable_removeIf()
...
+                () -> map6.removeIf((key, value) ->
+                {
+                    if (key == 3)
+                    {
+                        throw predicateException;
+                    }
+                    return key % 2 == 0;
+                }));
+        assertSame(predicateException, actualException);
+        assertIterablesEqual(this.newWithKeysValues(1, "One", 3, "Three", 4, "Four"), map6);
```
mohrezaei: I don't understand how this could be guaranteed. iteration has no order.

motlin: assertIterablesEqual won't assert the same order for two non-sorted maps. But it's enhanced over just assertEquals because it runs stuff like

```java
        assertEquals(o1, o1);
        assertEquals(o2, o2);
        assertEquals(o1, o2);
        assertEquals(o2, o1);
        assertEquals(o1.hashCode(), o2.hashCode());
```

mohrezaei: sorry, that's not what I meant. I meant, when `removeIf` is executing, there is no guarantee that it'll see "2" before failing on "3", for that matter, not starting with "4". The fact that the resulting map is 1/3/4 is pure coincidence, no?

motlin:
I think I addressed all the feedback, deleted some redundant assertions, and found some places where behavior ought to change. It's all in https://github.com/eclipse-collections/eclipse-collections/compare/d4567ab1b1c75dc1caca951deffdd67eaa879a68..ca5b5c9f2ac5e117bc29be39e48979cf747900cc

I chose to keep all the changes in tests for now, so I left TODOs for the potential behavior changes.

mohrezaei:
The `Unmodifiable` ones are concerning. Do those methods allow modification?
If so, I think it's best to fix those (here or in a PR before this -- I would do it here).

motlin:
> The `Unmodifiable` ones are concerning. Do those methods allow modification? If so, I think it's best to fix those (here or in a PR before this -- I would do it here).

I'm pretty confident that no Unmodifiable wrappers allow any modifications in any cases. However, when we forget to override a method to throw Unsupported, then we get exceptions late, when a mutation is attempted. For methods like removeIf, compute, computeIfPresent, etc. we may not get any exception if no mutation is attempted.

I have noticed this behavior before, and tried to fix it in PRs like #1617. I got some pushback on that PR, and if I recall it was just because of its size. I've been trying to write the smallest PRs possible since then.

One thing which we just need to decide/confirm is that any mutating method that _might_ mutate ought to throw Unsupported immediately. That would be easy to implement and test, and the only remaining challenge for me is the size of the PRs and number of places that need to change.

PR #1791, opened 2025-09-01
Make updateValue exception-safe in primitive maps.

motlin:


nikhilnanivadekar (approved):
Good catch!

nikhilnanivadekar:
Nit: before you merge can you add a bit more description in the PR. It helps while compiling release notes

motlin:
Thanks @nikhilnanivadekar, added a bit of detail. When we get up to release notes, we can collapse all the PRs that address #1517 into a single item, about all primitive maps.

PR #1808, opened 2025-11-11
Improve generated JavaDoc in mutablePrimitiveObjectMap.stg

pochopsp:
Fixes https://github.com/eclipse-collections/eclipse-collections/issues/1802

@motlin also agreed this improvement should be added.

### DIFF
--- a/eclipse-collections-code-generator/src/main/resources/api/map/mutablePrimitiveObjectMap.stg
+++ b/eclipse-collections-code-generator/src/main/resources/api/map/mutablePrimitiveObjectMap.stg
@@ -62,15 +62,17 @@ public interface Mutable<name>ObjectMap\<V> extends <name>ObjectMap\<V>, Mutable
     void putAll(<name>ObjectMap\<? extends V> map);
 
     /**
-     * Removes the mapping associated with the key, if one exists, from the map.
-     * @param key the key to remove
+     * Remove an entry from the map at the specified {@code key}.
+     *
+     * @return The value removed from entry at key, or null if not found.
      * @see #remove(<type>)
      */
     V removeKey(<type> key);
 
     /**
-     * Removes the mapping associated with the key, if one exists, from the map.
-     * @param key the key to remove
+     * Remove an entry from the map at the specified {@code key}.
+     *
+     * @return The value removed from entry at key, or null if not found.
      * @see #removeKey(<type>)
      */
     V remove(<type> key);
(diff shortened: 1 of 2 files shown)


nikhilnanivadekar (approved):

pochopsp:
Hi @nikhilnanivadekar I force pushed the commit again since I had a failing checkstyle (there was a tab instead of spaces in one line)

eclipse-collections-code-generator/src/main/resources/api/map/mutablePrimitiveObjectMap.stg
```diff
@@ -62,15 +62,17 @@ public interface Mutable<name>ObjectMap\<V> extends <name>ObjectMap\<V>, Mutable
     void putAll(<name>ObjectMap\<? extends V> map);
 
     /**
-     * Removes the mapping associated with the key, if one exists, from the map.
-     * @param key the key to remove
+     * Remove an entry from the map at the specified {@code key}.
```
pzygielo, line 65: 😕 

This introduces inconsistency. For example compare doc for `put` above:
- it says: _Introduce**s**_, here it's changed from _Remove**s**_ to _Remove_
- it has `@param` for parameters, which is gone here
- `@return` - there it starts from lowercase, as they usually do in Java API, here it's sentence with fullstop.

_Remove an entry_ it's something like a command to do a thing, not a description what it does.

pzygielo: _How to Write Doc Comments for the Javadoc Tool_:
>  The description is in 3rd person declarative rather than 2nd person imperative.
> Gets the label. (preferred)
> Get the label. (avoid)

pochopsp: Hi @pzygielo thanks for your comment. I think you're actually right and we can make this even better.

On the inconsistency, I agree when talking about this file, but my new javadoc was just copied character by character from `MutableMapIterable` (I also mentioned that in my [issue](https://github.com/eclipse-collections/eclipse-collections/issues/1802)):
https://github.com/eclipse-collections/eclipse-collections/blob/master/eclipse-collections-api/src/main/java/org/eclipse/collections/api/map/MutableMapIterable.java#L71

so I would say it is actually consistent with other parts of the code base.

Perhaps `MutableMapIterable` javadocs are also improvable and we should fix them as well ?

PR #1816, opened 2025-11-25
Implement Binary Search Algorithm in Java

ArsiyaTasleem90250:
This PR adds an implementation of the binary search algorithm in Java. 
The `search` method takes a sorted integer array `nums` and a `target` value, 
and returns the index of the target if found, or -1 if it is not present in the array.

Key features:
- Efficient O(log n) time complexity.
- Handles edge cases like empty arrays.
- Returns -1 when the target is not found.

This implementation can be used as a utility method for searching in sorted arrays.

### DIFF
--- a/search.java
+++ b/search.java
@@ -1,0 +1,21 @@
+class Solution {
+    public int search(int[] nums, int target) {
+        int i = 0;
+        int j = nums.length - 1;
+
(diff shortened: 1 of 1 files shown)


mohrezaei:
Please stop creating PRs unrelated to this repo.

PR #1818, opened 2025-11-29
Refactor RichIterable test cases to test unique elements first, then duplicates.

motlin:
The conceptual change here is to pull up assertions from RichIterableUniqueTestCase to RichIterableTestCase. I thought this might reduce lines of code - it turns out it doesn't, the total number of lines is approximately unchanged. However, the test coverage improves and these tests caught the problem fixed in #1817, that we could create multiple copies of what was supposed to be the empty singleton. This PR will fail until #1817 lands, but I wanted to have both PRs open at the same time so they can be viewed together. This PR is big but low risk since it's just moving test assertions. The other PR is small, but fixes production code.

```
[ERROR] Failures: 
[ERROR]   ImmutableBagTest.RichIterable_selectInstancesOf expected: org.eclipse.collections.impl.bag.immutable.ImmutableEmptyBag@2419a5db<[]> but was: org.eclipse.collections.impl.bag.immutable.ImmutableArrayBag@52bb3dcd<[]>
```


motlin:
Rebased on top of https://github.com/eclipse-collections/eclipse-collections/pull/1817 and now the tests pass.

PR #1819, opened 2025-12-03
Fix SortedSetIterable getFirst/getLast to return null on empty sets.

motlin:
The sorted set implementations (TreeSortedSet, SortedSetAdapter, AbstractImmutableSortedSet) previously delegated getFirst() and getLast() to java.util.SortedSet's first() and last() methods, which throw NoSuchElementException when the set is empty. This violated the RichIterable/OrderedIterable contract which specifies that getFirst() and getLast() should return null for empty collections.

Changed to check isEmpty() before calling first()/last(), returning null for empty sets. Updated SortedSetIterableTestCase to remove the assertThrows tests that verified the incorrect exception behavior.

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/list/fixed/ArrayAdapterTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/list/fixed/ArrayAdapterTest.java
@@ -220,26 +220,6 @@ public class ArrayAdapterTest extends AbstractListTestCase
 
     @Override
     @Test
-    public void getFirst()
-    {
-        super.getFirst();
-
-        assertEquals(Integer.valueOf(1), this.newWith(1, 2, 3).getFirst());
-        assertNotEquals(Integer.valueOf(3), this.newWith(1, 2, 3).getFirst());
-    }
-
-    @Override
-    @Test
-    public void getLast()
-    {
-        super.getLast();
-
-        assertNotEquals(Integer.valueOf(1), this.newWith(1, 2, 3).getLast());
-        assertEquals(Integer.valueOf(3), this.newWith(1, 2, 3).getLast());
-    }
-
-    @Override
-    @Test
     public void isEmpty()
     {
         super.isEmpty();
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/FastListTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/FastListTest.java
@@ -421,24 +421,6 @@ public class FastListTest extends AbstractListTestCase
 
     @Override
     @Test
-    public void getFirst()
-    {
-        assertNull(FastList.<Integer>newList().getFirst());
-        assertEquals(Integer.valueOf(1), FastList.newListWith(1, 2, 3).getFirst());
-        assertNotEquals(Integer.valueOf(3), FastList.newListWith(1, 2, 3).getFirst());
-    }
-
-    @Override
-    @Test
-    public void getLast()
-    {
-        assertNull(FastList.<Integer>newList().getLast());
-        assertNotEquals(Integer.valueOf(1), FastList.newListWith(1, 2, 3).getLast());
-        assertEquals(Integer.valueOf(3), FastList.newListWith(1, 2, 3).getLast());
-    }
-
-    @Override
-    @Test
     public void isEmpty()
     {
         Verify.assertEmpty(FastList.<Integer>newList());
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/immutable/AbstractImmutableSortedSet.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/immutable/AbstractImmutableSortedSet.java
@@ -157,13 +157,13 @@ abstract class AbstractImmutableSortedSet<T> extends AbstractImmutableCollection
     @Override
     public T getFirst()
     {
-        return this.first();
+        return this.isEmpty() ? null : this.first();
     }
 
     @Override
     public T getLast()
     {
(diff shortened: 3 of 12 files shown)


unit-tests/src/test/java/org/eclipse/collections/impl/list/fixed/ArrayAdapterTest.java
```diff
@@ -218,26 +218,6 @@ public void collectIf()
                 new ArrayList<>()), "1", "2", "3");
     }
 
-    @Override
```
nikhilnanivadekar, line 221: Did these get pulled somewhere?

unit-tests/src/test/java/org/eclipse/collections/impl/list/mutable/FastListTest.java
```diff
@@ -419,24 +419,6 @@ public void testForEachIf()
         assertEquals(3, countProcedure.getCount());
     }
 
-    @Override
```
nikhilnanivadekar, line 422: Did these get pulled elsewhere?

nikhilnanivadekar (commented):
Is this really a _fix_ or a change in behavior? Wont it have trickle effects like for `getAny()`? I went through some of my code bases and it is written expecting an exception. So, do we really need to fix it?

motlin:
> Is this really a _fix_ or a change in behavior? Wont it have trickle effects like for `getAny()`? I went through some of my code bases and it is written expecting an exception. So, do we really need to fix it?

The Javadoc of RichIterable's getFirst/getLast says:

> If the iterable is empty, null is returned

The Javadoc of SortedSet's first/last says:

> `@throws NoSuchElementException if this set is empty`

The contract of RichIterable is _weird_ but it's documented, so I consider this a bug and a bugfix. I don't really remember the history of how we got here, but it looks like we implemented some of our own sorted sets assuming that getFirst/first are exactly the same and getLast/last are exactly the same and having one of each pair delegate to the other.

Your other question is about the deleted tests though. Those are overrides of tests that used to break because they test the contracts as written, and were overridden to assert the actual implementation. So by deleting the overrides, we now assert that getFirst/getLast return null / don't throw.

Any change in behavior can be a big deal, even if it's to implement a Javadoc contract or fix a bug, so I'm open to discussing this more and even not landing the change.

My recent work has been to pull up assertions from all over the unit-tests-java8 test suite into RichIterableTestCase and delete overrides, and this causes me to find all sorts of inconsistencies like this one.

motlin:
@donraab this is a behavior change, but one that makes the actual behavior match the interface contract. WDYT?

donraab:
@motlin @nikhilnanivadekar I have mixed feelings about this one. It's not great to have conflicts in expected behavior for methods, since we do try and provide consistency with JDK for compatibility. The methods `getFirst` and `getLast` are at least similar to `detect` which returns `null` if nothing is found. The proper way to code for the potential of an empty collection would be to check if the collection `isEmpty()`, not catch if an exception is thrown. We do have `detectOptional`, `minOptional`, `maxOptional`, so maybe it would make sense to have `getFirstOptional` and `getLastOptional` for symmetry.

Updated: I just looked. We already have `getFirstOptional` and `getLastOptional` on `OrderedIterable`.

If we changed the behavior of `getFirst`/`getLast` to be consistent with JDK, it would break anyone who explicitly is checking to see if either of these returns `null`, instead of checking `isEmpty()`. I'm starting to think maybe this would be preferable since `SequencedCollection` was added with more places where `getFirst` / `getLast` are available and we'll have potentially more confusion with folks getting used to these methods throwing. The proper solution is still the same IMO. Check for `isEmpty()`. Checking for `null` is weird. Catching an `Exception` is weird. The implementations of `getFirstOptional` and `getLastOptional` show that `isEmpty()` is the thing to check for.

```
default Optional<T> getFirstOptional()
{
    if (this.isEmpty())
    {
        return Optional.empty();
    }
    return Optional.of(this.getFirst());
}
```

I'm trying to remember the case where returning `null` of empty simplifies coding. If anything, it is probably only slightly less verbose than having to check `isEmpty()`.

motlin:
@donraab I'm not sure what you're suggesting. I just want to clarify how I got here.

I was working on the trait based tests, in unit-tests-java8, and trying to pull more assertions up to RichIterableTestCase.

I found an internal inconsistency where I could not make a single assertion for all of our collections - if I asserted null some failed and if I asserted an exception threw, others failed.

I saw the contract of RichIterable says:

```java
    /**
     * Returns the first element of an iterable. In the case of a List it is the element at the first index. In the
     * case of any other Collection, it is the first element that would be returned during an iteration. If the
     * iterable is empty, null is returned. If null is a valid element of the container, then a developer would need to
     * check to see if the iterable is empty to validate that a null result was not due to the container being empty.
```

And I made this PR to make our own collections compatible with our own contract. I didn't even realize that SequencedCollection clashes, but it does!

```java
    /**
     * Gets the first element of this collection.
     *
     * @implSpec
     * The implementation in this interface obtains an iterator of this collection, and
     * then it obtains an element by calling the iterator's {@code next} method. Any
     * {@code NoSuchElementException} thrown is propagated. Otherwise, it returns
     * the element.
     *
     * @return the retrieved element
     * @throws NoSuchElementException if this collection is empty
     */
    default E getFirst() {
        return this.iterator().next();
    }
```

So I see now we have a bigger problem on our hands, but I wasn't trying to touch that one.

motlin:
I want to revisit this one @donraab @nikhilnanivadekar. Our getFirst() method usually returns null on empty, but in rare cases it throws. The JDK added SequencedCollection and its getFirst() method in Java 21 which throws NoSuchElementException when empty. I want to make all implementations consistent. This PR makes all implementations return null, but I could go the other way and make them all throw NoSuchElementException. WDYT?

PR #1823, opened 2025-12-14
Fix ImmutableOrderedMapAdapter.entrySet() to throw UnsupportedOperationException on Entry.setValue().

motlin:
This fixes the most important bug uncovered in https://github.com/eclipse-collections/eclipse-collections/pull/1821

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/immutable/ImmutableOrderedMapAdapter.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/immutable/ImmutableOrderedMapAdapter.java
@@ -69,7 +69,7 @@ public class ImmutableOrderedMapAdapter<K, V>
 
     public ImmutableOrderedMapAdapter(Map<K, V> delegate)
     {
-        this.delegate = OrderedMapAdapter.adapt(new LinkedHashMap<>(delegate));
+        this.delegate = OrderedMapAdapter.adapt(Collections.unmodifiableMap(new LinkedHashMap<>(delegate)));
     }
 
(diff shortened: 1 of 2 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/immutable/ImmutableOrderedMapAdapter.java
```diff
@@ -69,7 +69,7 @@ public class ImmutableOrderedMapAdapter<K, V>
 
     public ImmutableOrderedMapAdapter(Map<K, V> delegate)
     {
-        this.delegate = OrderedMapAdapter.adapt(new LinkedHashMap<>(delegate));
+        this.delegate = OrderedMapAdapter.adapt(Collections.unmodifiableMap(new LinkedHashMap<>(delegate)));
```
nikhilnanivadekar, line 72: This will change behavior, correct? So we will need the next version to be a major version, to confirm.

motlin: It's a bug fix, making a class that is supposed to be an immutable type actually immutable, so I don't think this necessitates a major bump. That said, I have other changes queued up that make getFirst/getLast behave consistently (either return null or throw) which probably ought to be in a major release.

nikhilnanivadekar: I had forgotten, we are already queued up for a major version. So, we should be good.

PR #1829, opened 2025-12-15
Override merge() to throw UnsupportedOperationException immediately.

motlin:
instead of relying on the default Map.merge() which calls the lambda before failing on put(). This ensures consistent behavior with other mutating operations on immutable maps.


motlin:
> Good spotting of this. Curious how did you find it? It’s like a needle in the haystack

I'm breaking up my work into smaller pieces to make it easier to review and land. I have more work I haven't turned into PRs yet where I'm pulling up tests, or assertions, into RichIterableTestCase and MapTestCase. When I pull assertions up this high, they run on everything, and they don't always pass.

PR #1830, opened 2025-12-15
Fix ErrorProne rules ClassCanBeStatic, StringCaseLocaleUsage, and MissingOverride in tests.

motlin:
I've been experimenting with turning on ErrorProne, mostly because I needed the MissingOverride rule in tests to find the problems I've been fixing in my recent PRs. It caught a few more very minor problems and I figured I'd socialize the fixes and start to see if it's realistic and helpful to turn on this additional linter.

### DIFF
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/WordleTest.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/WordleTest.java
@@ -10,6 +10,8 @@
 
 package org.eclipse.collections.impl;
 
+import java.util.Locale;
+
 import org.eclipse.collections.api.bag.primitive.MutableCharBag;
 import org.eclipse.collections.api.factory.Lists;
 import org.eclipse.collections.api.factory.primitive.CharLists;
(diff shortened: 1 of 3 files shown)


nikhilnanivadekar:
May be I missed it, where did you turn on the rule? Inspection or checkstyle? Because we should have the corresponding changes to run it systematically as a part of builds too. Thoughts?

motlin:
> May be I missed it, where did you turn on the rule? Inspection or checkstyle? Because we should have the corresponding changes to run it systematically as a part of builds too. Thoughts?

This relates to the comment I just left on another PR: https://github.com/eclipse-collections/eclipse-collections/pull/1829#issuecomment-3668228699

When I pull up assertions into RichIterableTestCase and MapTestCase, I want to make sure that they aren't overridden by test methods, unless those tests also call super. I remembered that ErrorProne has the annotation `@OverridingMethodsMustInvokeSuper` which enforces exactly that, and I've been working on turning it on.

The thing is, I found lots of inconsistencies and tests that don't pass. I also found the tests that were missing `@Test` and so they weren't running at all. I can't land the ErrorProne build yet - it doesn't pass. But that's what I'm building toward with these PRs.

nikhilnanivadekar (approved):

PR #1831, opened 2025-12-16
Add inject into with index

### LINKED ISSUE #1203: Add injectIntoIndex to OrderedIterable
We have injectIntoIndex on primitive OrderedIterables (e.g. `OrderedIntIterable`), but not on the object equivalent `OrderedIterable`.

valbaca:
_This is my first open source commit and first commit to this repo. Please provide any advice._

This commit adds injectIntoWithIndex which is like injectInto with an
additional parameter for the index of the element.

This provides a default implementation for OrderedIterable<T> that
reuses forEachWithIndex.

Implementation note: I used an `AtomicReference<IV> ref` to have a "box" for the running value. I see that other code uses a similar pattern for primitives, like `  int[] index = {0};` but doing something like `IV[] box` becomes a bit of an mess with the mix of Type and generics and started giving warnings. Another option is to use a single-element List.

If you have another suggested approach, I'm happy to update. 

Closes #1203

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/ordered/OrderedIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/ordered/OrderedIterable.java
@@ -15,7 +15,6 @@ import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.block.function.Function;
@@ -390,12 +389,13 @@ public interface OrderedIterable<T> extends RichIterable<T>
      * index as the third parameter. This method is commonly called foldIndexed or sometimes reduce.
      *
      * @since 14.0
-     * @see {@link #injectInto(Object, Function2)}
+     * @see #injectInto(Object, Function2)
      * @param injectedValue first parameter of first evaluation
      * @param function function invoked with arguments: injectedValue or previous result, element, and index of element
      */
 
-    default <IV> IV injectIntoWithIndex(IV injectedValue, Function3<? super IV, ? super T, Integer, ? extends IV> function) {
+    default <IV> IV injectIntoWithIndex(IV injectedValue, Function3<? super IV, ? super T, Integer, ? extends IV> function)
+    {
         int[] index = {0};
         return this.injectInto(injectedValue, (prev, each) -> function.value(prev, each, index[0]++));
     }
--- a/unit-tests-java8/src/test/java/org/eclipse/collections/test/OrderedIterableTestCase.java
+++ b/unit-tests-java8/src/test/java/org/eclipse/collections/test/OrderedIterableTestCase.java
@@ -10,12 +10,8 @@
 
 package org.eclipse.collections.test;
 
-import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.Iterator;
-import java.util.List;
 import java.util.Optional;
-import java.util.Set;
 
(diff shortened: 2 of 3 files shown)


unit-tests-java8/src/test/java/org/eclipse/collections/test/OrderedIterableTestCase.java
```diff
@@ -203,4 +208,50 @@ default void OrderedIterable_zipWithIndex_target()
                 result);
         assertSame(target, result);
     }
+
+    @Test
+    default void OrderedIterable_injectIntoWithIndex()
```
motlin, line 209: OrderedIterable_collectWithIndex has an override in SortedNaturalOrderTestCase, and this probably should too.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/ordered/OrderedIterable.java
```diff
@@ -380,4 +382,19 @@ default <P, V> OrderedIterable<V> flatCollectWith(Function2<? super T, ? super P
...
+     * index as the third parameter. This method is commonly called foldIndexed or sometimes reduce.
+     *
+     * @since 14.0
+     */
+
+    default <IV> IV injectIntoWithIndex(IV injectedValue, Function3<? super IV, ? super T, Integer, ? extends IV> function) {
+        AtomicReference<IV> ref = new AtomicReference<>(injectedValue);
+        this.forEachWithIndex((value, index) -> ref.set(function.value(ref.get(), value, index)));
+        return ref.get();
+    }
```
motlin: The implementation of collectWithIndex is

```java
    default <V> OrderedIterable<V> collectWithIndex(ObjectIntToObjectFunction<? super T, ? extends V> function)
    {
        int[] index = {0};
        return this.collect(each -> function.valueOf(each, index[0]++));
    }
```

I understand why the AtomicReference is here, but it might be more consistent and performant to use         `int[] index = {0};`
and delegate to injectInto, rather than forEachWithIndex.

*WithIndex methods tend to be overridden in Bags, so that we can delegate to forEachWithOccurrences as a micro optimization. See AbstractMutableSortedBag#collectWithIndex as an example.

motlin (commented):
Looks like a great start! Thank you for the contribution. I left a few suggestions - let me know what you think.

valbaca:
thanks for the comments, I'll take another pass with the comments incorporated.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/ordered/OrderedIterable.java
valbaca: I've updated to use `int[] index` and delegate to injectInto

I didn't add the Bags override as the injectIntoWithIndex as I would need to benchmark to see if the occurrences for-loop would be an optimization.

unit-tests-java8/src/test/java/org/eclipse/collections/test/OrderedIterableTestCase.java
valbaca: updated the tests to be simpler and added an override in SortedNaturalOrderTestCase

.github/workflows/pull-request.yml
```diff
@@ -267,7 +267,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - uses: actions/cache@v4
+      - uses: actions/cache@v5
```
motlin, line 270: These changes aren't bad, but they are unrelated?

motlin:
This is getting hard to review because it has commits from upstream cherry picked into it. Could you run `git pull upstream master --rebase`?

PR #1832, opened 2025-12-17
Fix #1801: ConcurrentHashMap view spliterators should be CONCURRENT and not SIZED

### LINKED ISSUE #1801: ConcurrentHashMap#values().stream().toArray() is not thread-safe
**Bug Reproduction Unit Test**

	@Test
	void raceConditionReproductionTest() {
		ConcurrentMap<Integer, String> eclipseMap = new org.eclipse.collections.impl.map.mutable.ConcurrentHashMap<>();
		IntStream.range(0, 200000).boxed()
				.forEach(i -> eclipseMap.put(i, String.valueOf(i)));

		CompletableFuture.allOf(
				CompletableFuture.runAsync(
						() -> eclipseMap.values().stream().peek(s -> randomWait()).toArray(String[]::new)),
				CompletableFuture.runAsync(
						() -> IntStream.range(200000, 400000)
								.boxed()
								.peek(i -> randomWait())
								.forEach(i -> eclipseMap.put(i, String.valueOf(i))))
		).join();
	}

	private static void randomWait() {
		try {
			if (ThreadLocalRandom.current().nextInt(1000) == 0) {
				Thread.sleep(1);
			}
		} catch (InterruptedException e) {
			throw new RuntimeException(e);
		}
	}

The above test code adds new values to a ConcurrentHashMap while a Stream#toArray() operation is being performed on the map's values. 
Running the test on the latest version of Eclipse Collections (13.0.0) results in an exception like the following being thrown:
`java.util.concurrent.CompletionException: java.lang.IllegalStateException: Accept exceeded fixed size of 202893`

Note: Using `java.util.concurrent.ConcurrentHashMap` instead of the Eclipse one fixes the test, so the vanilla Java concurrent map does **not** have this issue.

**Bug Details**
Upon an initial investigation, the problem seems to be that the collection's spliterator supplies the size of the resulting array before the Stream pipeline processing, as if it is sure that the size of the collection won't change mid-iteration. And so the exception occurs when more elements then expected are appended to the array.

MiguelGarcia-E:
### What was changed
Updated the `spliterator()` implementations for `keySet()`, `values()`, and `entrySet()` views of
`ConcurrentHashMap` to ensure they correctly report `Spliterator.CONCURRENT` and do **not**
report `SIZED` or `SUBSIZED`.

This aligns the behavior with the concurrent nature of the map and prevents incorrect
assumptions when these views are used with streams.

### Why this change
The previous behavior could expose spliterators that appeared sized, even though the
underlying collection can change concurrently. This could lead to race conditions or
incorrect stream behavior, as reported in #1801.

### Tests
- Added unit test verifying that `keySet`, `values`, and `entrySet` spliterators:
  - have `Spliterator.CONCURRENT`
  - do not have `Spliterator.SIZED` or `Spliterator.SUBSIZED`
- Ran the full Maven test suite successfully.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
@@ -24,6 +24,8 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Set;
+import java.util.Spliterator;
+import java.util.Spliterators;
 import java.util.concurrent.Executor;
 import java.util.concurrent.FutureTask;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -1631,6 +1633,12 @@ public final class ConcurrentHashMap<K, V>
         }
 
         @Override
+        public Spliterator<K> spliterator()
+        {
+            return Spliterators.spliteratorUnknownSize(this.iterator(), Spliterator.CONCURRENT | Spliterator.DISTINCT | Spliterator.NONNULL);
+        }
+
+        @Override
         public int size()
         {
             return ConcurrentHashMap.this.size();
@@ -1661,6 +1669,12 @@ public final class ConcurrentHashMap<K, V>
         public Iterator<V> iterator()
         {
             return new ValueIterator();
+        }
+
+        @Override
+        public Spliterator<V> spliterator()
+        {
+            return Spliterators.spliteratorUnknownSize(this.iterator(), Spliterator.CONCURRENT | Spliterator.NONNULL);
         }
 
         @Override
@@ -1720,6 +1734,12 @@ public final class ConcurrentHashMap<K, V>
         public Iterator<Map.Entry<K, V>> iterator()
         {
             return new EntryIterator();
+        }
+
+        @Override
+        public Spliterator<Map.Entry<K, V>> spliterator()
+        {
+            return Spliterators.spliteratorUnknownSize(this.iterator(), Spliterator.CONCURRENT | Spliterator.DISTINCT | Spliterator.NONNULL);
         }
 
         @Override
--- a/RELEASE_NOTE_DRAFT.md
+++ b/RELEASE_NOTE_DRAFT.md
@@ -18,6 +18,14 @@ The Eclipse Collections team gives a huge thank you to everyone who participated
 * Upgrade bnd plugin to 7.1.0.
 * Publish p2 artifacts to maven central. Fixes #294 . Example [here](https://central.sonatype.com/artifact/org.eclipse.collections/p2-site)
 
+
+# Bug Fixes
+-----------------
+* Fixed ConcurrentHashMap keySet(), values(), and entrySet() spliterators to ensure they are CONCURRENT and not SIZED. Fixes #1801.
+* Fixed ConcurrentHashMapUnsafe keySet(), values(), and entrySet() spliterators to ensure they are CONCURRENT and not SIZED.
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -75,3 +75,4 @@ fabric.properties
 .project
 .settings/
 .classpath
+.idea/
```
mohrezaei: please revert

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
```diff
@@ -33,6 +33,9 @@
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
```
mohrezaei, line 35: Please make the corresponding changes to ConcurrentHashMapUnsafe

mohrezaei (changes requested):

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentHashMap.java
motlin: I just noticed the same thing and pushed the changes plus a few more improvements to https://github.com/eclipse-collections/eclipse-collections/commit/63f060040aee119bcd88c081ab5a0b331cf841d1

MiguelGarcia-E: Added the corresponding changes to ConcurrentHashMapUnsafe and updated release notes. Local build/tests succeeded.

mohrezaei: thanks. Please incorporate @motlin 's suggestions.

MiguelGarcia-E: Incorporated @motlin 's suggestions in ConcurrentHashMap (imports + DISTINCT/NONNULL flags). ConcurrentHashMapUnsafe already has the corresponding spliterator change. Thanks!

mohrezaei: `DISTINCT` and `NONNULL` seem to be missing from Unsafe.
He also added a `raceConditionReproductionTest`

MiguelGarcia-E: Updated ConcurrentHashMapUnsafe spliterator flags to include DISTINCT/NONNULL (matching ConcurrentHashMap) and added the raceConditionReproductionTest as suggested. Tests pass locally.

mohrezaei (approved):

MiguelGarcia-E:
Thanks for the review! Happy to address any follow-ups if needed.

mohrezaei:
@motlin any other suggestions? Feel free to merge otherwise.
@MiguelGarcia-E thanks for the contribution.

motlin:
I'd prefer to squash the commits down before merging. Doesn't have to be a single commit, but I'd prefer not to see reverts in the git history.

mohrezaei:
oh, sorry, I've been using a "Squash and merge" workflow for the past couple of years. But I agree, for eclipse collections, we should stick to manual squash for now.
@MiguelGarcia-E please squash your commits.

MiguelGarcia-E:
Squashed commits as requested (also removed incidental .idea and unrelated test file changes).

PR #1833, opened 2025-12-29
Fix #1790: make sumOfDoubleConsistentRounding deterministic and ULP-tolerant

### LINKED ISSUE #1790: Fix or disable flaky test sumOfDoubleConsistentRounding().
This test is flaky (and slow). After seeing it fail randomly a few times, I decided to record some evidence for anyone who wants to fix/reenable it. In a separate PR, I intend to disable the test.

I edited the test to use a fixed seed, and iterated until I found one that reliably causes the test to fail.

```java
    @Test
    public void sumOfDoubleConsistentRounding()
    {
        DoubleFunction<Integer> roundingSensitiveElementFunction = i -> (i <= 99995) ? 1.0e-18d : 1.0d;

        Random random = new Random(1303L);
        MutableList<Integer> list = Interval.oneTo(100_000).toList().shuffleThis(random);
        double baseline = this.getExpectedWith(list.toArray(new Integer[]{}))
                .sumOfDouble(roundingSensitiveElementFunction);

        for (Integer batchSize : BATCH_SIZES)
        {
            this.batchSize = batchSize;

            ParallelIterable<Integer> testCollection = this.newWith(list.toArray(new Integer[]{}));
            assertEquals(
                    baseline,
                    testCollection.sumOfDouble(roundingSensitiveElementFunction),
                    1.0e-15d,
                    "Batch size: " + this.batchSize);
        }
    }
```

This gets all the Parallel versions of the test to fail with batch size 10000.

- ImmutableListParallelListIterableTest.sumOfDoubleConsistentRounding()
- MultiReaderFastListParallelListIterableTest.sumOfDoubleConsistentRounding()
- ParallelCollectListIterableTest.sumOfDoubleConsistentRounding()
- ParallelListIterableTest.sumOfDoubleConsistentRounding()
- RandomAccessListAdapterParallelListIterableTest.sumOfDoubleConsistentRounding()
- UnmodifiableMutableListParallelListIterableTest.sumOfDoubleConsistentRounding()

With error messages like:

```
org.opentest4j.AssertionFailedError: Batch size: 10000 ==> 
Expected :5.0000000000000995
Actual   :5.000000000000101
```

MiguelGarcia-E:
### Description

This change fixes a flaky `sumOfDoubleConsistentRounding()` test caused by non-associative floating-point summation in parallel reductions.

The test now:
- Uses a deterministic shuffle with a fixed random seed for reproducibility.
- Replaces the fixed delta assertion with an ULP-based tolerance to account for rounding differences due to varying summation order across batch sizes.

Fixes #1790.

### DIFF
--- a/RELEASE_NOTE_DRAFT.md
+++ b/RELEASE_NOTE_DRAFT.md
@@ -21,8 +21,7 @@ The Eclipse Collections team gives a huge thank you to everyone who participated
 
 # Bug Fixes
 -----------------
-* Fixed flaky parallel sumOfDoubleConsistentRounding() test by using deterministic shuffle seed and ULP-based tolerance. Fixes #1790.
-
+* Fixed flaky sumOfDoubleConsistentRounding() test by using deterministic input ordering and validating the sum stays within the mathematically expected range for rounding-sensitive values across batch sizes. Fixes #1790.
 
 
 # Note
--- a/unit-tests/src/test/java/org/eclipse/collections/impl/lazy/parallel/ParallelIterableTestCase.java
+++ b/unit-tests/src/test/java/org/eclipse/collections/impl/lazy/parallel/ParallelIterableTestCase.java
@@ -912,29 +912,31 @@ public abstract class ParallelIterableTestCase
     @Test
     public void sumOfDoubleConsistentRounding()
     {
-        DoubleFunction<Integer> roundingSensitiveElementFunction = i -> (i <= 99995) ? 1.0e-18d : 1.0d;
+        DoubleFunction<Integer> roundingSensitiveElementFunction =
+                i -> (i <= 99995) ? 1.0e-18d : 1.0d;
 
-        // Static seed
         Random random = new Random(1303L);
         MutableList<Integer> list = Interval.oneTo(100_000).toList().shuffleThis(random);
-
-        double baseline = this.getExpectedWith(list.toArray(new Integer[]{}))
-                .sumOfDouble(roundingSensitiveElementFunction);
 
         for (Integer batchSize : BATCH_SIZES)
         {
             this.batchSize = batchSize;
 
             ParallelIterable<Integer> testCollection = this.newWith(list.toArray(new Integer[]{}));
+
+            long bigCount = testCollection.count(i -> i > 99995);
+            long totalCount = testCollection.count(each -> true); // <-- en lugar de size()
+
+            double min = (double) bigCount;
+            double max = (double) bigCount + ((double) (totalCount - bigCount)) * 1.0e-18d;
+
             double actual = testCollection.sumOfDouble(roundingSensitiveElementFunction);
 
-            double tolerance = 4.0d * Math.max(Math.ulp(baseline), Math.ulp(actual));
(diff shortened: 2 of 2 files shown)


motlin:
@donraab I'm curious for your opinion. I think that using a static seed might undermine the idea of the test, but on the other hand the test was never really correct, and it fails for me about 1% of the time and I think I'd merge a PR that deletes the test at this point.

mohrezaei:
>  I think that using a static seed might undermine the idea of the test

I tend to think along the same lines. Additionally, the solution with the magic number 4 * ulp is just a mirror of the test setup: it's adding 5 tiny numbers to 99995 big ones, and it can lose track of 4. So it just renders the test kinda moot.

> on the other hand the test was never really correct

I see that point too. The test itself is essentially designed to prove that adding in parallel is not deterministic, but you can get close. How you test this closeness is not all that interesting.

MiguelGarcia-E:
I agree with the concern. This test seems to assert deterministic behavior that we don’t really guarantee for parallel floating-point summation, so any tolerance is ultimately arbitrary.

Given the long-standing flakiness and unclear contract, I’d be fine with removing the test entirely if that’s the preferred direction. Happy to update this PR accordingly.

mohrezaei:
> I’d be fine with removing the test entirely

My feeling is that the test pretends to be fancy (before and even more after your change).
It's a sum of `99995*1 + 5 * 10^-18`. I can do that in my head. When the little bits get lost because of parallel stuff, it shouldn't be off by more than `5*10^-18`. So it can be drastically simplified by specifying the results as constants. In fact, the actual condition for the result should be:
```
99995.0 <= result <= 99995.0 + 5*10^-18
```
(it's asymmetric, not a +- delta)

MiguelGarcia-E:
> ```
> 99995.0 <= result <= 99995.0 + 5*10^-18
> ```

Good point. Instead of enforcing determinism (baseline/seed), I can simplify the test to assert the correct bounded result for this specific setup: 99995.0 <= result <= 99995.0 + 5e-18 across batch sizes.

mohrezaei:
It appears I misread the test setup. It's five big ones (1.0) and 99995 small ones (1e-18). The sum is therefore in the range 5.0 <= result <= 5.0 + 99995e-18.

MiguelGarcia-E:
> It appears I misread the test setup. It's five big ones (1.0) and 99995 small ones (1e-18). The sum is therefore in the range 5.0 <= result <= 5.0 + 99995e-18.

Good catch — you’re right: it’s 5 values of 1.0 and 99995 values of 1e-18, so the expected range is 5.0 <= result <= 5.0 + 99995e-18. The test now derives min/max from the collection counts (big vs small) and checks the result is within that range for each batch size, with a tiny ULP slack for the upper bound representation.

MiguelGarcia-E:
Hi @donraab @mohrezaei @motlin  — I’ve updated the test to use a bounded result assertion (no baseline/determinism), matching the corrected setup (5 × 1.0 + 99995 × 1e-18). 

All checks are green. Would appreciate a review when you have a chance.

PR #1834, opened 2025-12-29
Set up ErrorProne and fix violations.

motlin:


motlin:
@nikhilnanivadekar I put up the whole code review at once, so I can show you where this is all going, but I could also split it up.

- The first 17 commits fix ErrorProne rule violations that I agree with, one rule at a time. They could land together or individually. I think all the rules are good, but if any are controversial at all, I'd split this up, and it's not important to land all 17.
- The last 5 commits suppress violations for rules that I generally agree with, but we have reasons to break in tests, like following the contract of equals(). These could land separately, or together with turning on ErrorProne.
- The 18th commit turns on ErrorProne in the build, with these 17 rules plus other rules that are on by default and have no violations.

eclipse-collections/src/main/java/org/eclipse/collections/impl/tuple/AbstractImmutableEntry.java
```diff
@@ -87,36 +87,6 @@ public String toString()
...
-    {
-        private static final long serialVersionUID = 1L;
-
-        @Override
-        public V valueOf(Map.Entry<?, V> entry)
-        {
-            return entry.getValue();
-        }
-    }
-
```
motlin, line 119: No serialization tests break. Do I need to keep these?

PR #1839, opened 2026-01-17
Introduces a concurrent HashMap implementation with a custom hashing strategy

MarcMil:
Similar to UnifiedMapWithHashingStrategy and UnifiedMap, I have left the ConcurrentHashMap class untouched. This implementation is based on the ConcurrentHashMap class, which was modified to use the custom hashing strategy.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
@@ -1,0 +1,2483 @@
+/*
+ * Copyright (c) 2024 Goldman Sachs and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompany this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.impl.map.strategy.mutable;
+
+import java.io.Externalizable;
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.util.AbstractCollection;
+import java.util.AbstractSet;
+import java.util.Collection;
+import java.util.ConcurrentModificationException;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.Set;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.concurrent.Executor;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicIntegerArray;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.function.BiFunction;
+import java.util.function.Predicate;
+
+import org.eclipse.collections.api.block.HashingStrategy;
+import org.eclipse.collections.api.block.function.Function;
+import org.eclipse.collections.api.block.function.Function0;
+import org.eclipse.collections.api.block.function.Function2;
+import org.eclipse.collections.api.block.function.Function3;
+import org.eclipse.collections.api.block.procedure.Procedure;
+import org.eclipse.collections.api.block.procedure.Procedure2;
+import org.eclipse.collections.api.block.procedure.primitive.ObjectIntProcedure;
+import org.eclipse.collections.api.factory.Maps;
+import org.eclipse.collections.api.map.ConcurrentMutableMap;
+import org.eclipse.collections.api.map.ImmutableMap;
+import org.eclipse.collections.api.map.MapIterable;
+import org.eclipse.collections.api.map.MutableMap;
+import org.eclipse.collections.api.tuple.Pair;
+import org.eclipse.collections.impl.block.procedure.MapEntryToProcedure2;
+import org.eclipse.collections.impl.list.mutable.FastList;
+import org.eclipse.collections.impl.map.mutable.AbstractMutableMap;
+import org.eclipse.collections.impl.utility.Iterate;
+import org.eclipse.collections.impl.utility.MapIterate;
+import org.eclipse.collections.impl.utility.internal.IterableIterate;
+
+@SuppressWarnings({"rawtypes", "ObjectEquality"})
+public final class ConcurrentHashMapWithHashingStrategy<K, V>
+        extends AbstractMutableMap<K, V>
+        implements ConcurrentMutableMap<K, V>, Externalizable
+{
+    private static final long serialVersionUID = 1L;
+
+    private static final Object RESIZE_SENTINEL = new Object();
+    private static final int DEFAULT_INITIAL_CAPACITY = 16;
+
+    /**
+     * The maximum capacity, used if a higher value is implicitly specified
+     * by either of the constructors with arguments.
+     * MUST be a power of two <= 1<<30.
+     */
+    private static final int MAXIMUM_CAPACITY = 1 << 30;
+
+    private static final AtomicReferenceFieldUpdater<ConcurrentHashMapWithHashingStrategy, AtomicReferenceArray> TABLE_UPDATER = AtomicReferenceFieldUpdater.newUpdater(ConcurrentHashMapWithHashingStrategy.class, AtomicReferenceArray.class, "table");
+    private static final AtomicIntegerFieldUpdater<ConcurrentHashMapWithHashingStrategy> SIZE_UPDATER = AtomicIntegerFieldUpdater.newUpdater(ConcurrentHashMapWithHashingStrategy.class, "size");
+    private static final Object RESIZED = new Object();
+    private static final Object RESIZING = new Object();
+    private static final int PARTITIONED_SIZE_THRESHOLD = 4096; // chosen to keep size below 1% of the total size of the map
+    private static final int SIZE_BUCKETS = 7;
+
+    /**
+     * The table, resized as necessary. Length MUST Always be a power of two.
+     */
+    private volatile AtomicReferenceArray table;
+
+    private AtomicIntegerArray partitionedSize;
+
+    private HashingStrategy strategy;
+    @SuppressWarnings("UnusedDeclaration")
+    private volatile int size; // updated via atomic field updater
+
+    /**
+     * @deprecated No argument default constructor used for serialization. Instantiating an ConcurrentHashMapWithHashingStrategy with
+     * this constructor will have a null hashingStrategy and throw NullPointerException when used.
+     */
+    @Deprecated
+    public ConcurrentHashMapWithHashingStrategy()
+    {
+    }
+
+    public ConcurrentHashMapWithHashingStrategy(HashingStrategy<? super K> hashingStrategy)
+    {
+        this(hashingStrategy, DEFAULT_INITIAL_CAPACITY);
+    }
+
+    public ConcurrentHashMapWithHashingStrategy(HashingStrategy<? super K> hashingStrategy, int initialCapacity)
+    {
+        this.strategy = hashingStrategy;
+        if (initialCapacity < 0)
+        {
+            throw new IllegalArgumentException("Illegal Initial Capacity: " + initialCapacity);
+        }
+        if (initialCapacity > MAXIMUM_CAPACITY)
+        {
+            initialCapacity = MAXIMUM_CAPACITY;
+        }
+
+        int threshold = initialCapacity;
+        threshold += threshold >> 1; // threshold = length * 0.75
+
+        int capacity = 1;
+        while (capacity < threshold)
+        {
+            capacity <<= 1;
+        }
+        if (capacity >= PARTITIONED_SIZE_THRESHOLD)
+        {
+            this.partitionedSize = new AtomicIntegerArray(SIZE_BUCKETS * 16); // we want 7 extra slots and 64 bytes for each slot. int is 4 bytes, so 64 bytes is 16 ints.
+        }
+        this.table = new AtomicReferenceArray(capacity + 1);
+    }
+
+    public static <K, V> ConcurrentHashMapWithHashingStrategy<K, V> newMap(HashingStrategy<? super K> hashingStrategy)
+    {
+        return new ConcurrentHashMapWithHashingStrategy<>(hashingStrategy);
+    }
+
+    public static <K, V> ConcurrentHashMapWithHashingStrategy<K, V> newMap(HashingStrategy<? super K> hashingStrategy, int newSize)
+    {
+        return new ConcurrentHashMapWithHashingStrategy<>(hashingStrategy, newSize);
+    }
+
+    private static int indexFor(int h, int length)
+    {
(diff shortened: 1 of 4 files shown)


eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
```diff
@@ -0,0 +1,2437 @@
...
+                currentArray = resizeContainer.nextArray;
+            }
+        }
+        while (resizeContainer != null);
+        return false;
+    }
+
+    private boolean nullSafeEquals(Object v, Object value)
+    {
+        return v == value || v != null && strategy.equals(v, value);
```
mohrezaei: nullSafeEquals is only used on values. use plain `.equals`

mohrezaei (changes requested):

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
```diff
@@ -0,0 +1,2439 @@
...
+        public boolean equals(Object o)
+        {
+            if (!(o instanceof Map.Entry<?, ?>))
+            {
+                return false;
+            }
+            Map.Entry<K, V> e = (Map.Entry<K, V>) o;
+            K k1 = this.key;
+            Object k2 = e.getKey();
+            if (k1 == k2 || k1 != null && k1.equals(k2))
```
motlin: I'm not sure if this is intentional, because there's no way to write a symmetric equals method, but the hashingStrategy is missing here, and it ought to be here to be consistent with our other types that use a hashing strategy.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
```diff
@@ -0,0 +1,2439 @@
...
+                Object v2 = e.getValue();
+                return v1 == v2 || v1 != null && v1.equals(v2);
+            }
+            return false;
+        }
+
+        @Override
+        public int hashCode()
+        {
+            return (this.key == null ? 0 : this.key.hashCode()) ^ (this.value == null ? 0 : this.value.hashCode());
```
motlin: And the hashing strategy is missing here.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
```diff
@@ -0,0 +1,2439 @@
...
+            if (o == RESIZED || o == RESIZING)
+            {
+                throw new ConcurrentModificationException("can't compute hashcode while resizing!");
+            }
+            Entry<K, V> e = (Entry<K, V>) o;
+            while (e != null)
+            {
+                Object key = e.getKey();
+                Object value = e.getValue();
+                h += (key == null ? 0 : key.hashCode()) ^ (value == null ? 0 : value.hashCode());
```
motlin: Hashing strategy is missing here as well.

motlin (changes requested):

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
mohrezaei: I checked UnifiedMapWithHashingStrategy and the Entry object (ImmutableEntry) has the same behavior: the strategy is ignored.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
mohrezaei: ditto

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
mohrezaei: the hashcode for the entire map has no need to use the strategy

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
motlin: Oh wow, it's inconsistent in a way I wasn't expecting. entrySet() exposes ImmutableEntry which does not use hashing strategy. But entrySet().iterator() exposes WeakBoundEntry which does. And entrySet().contains() doesn't expose entries of course but uses the hashing strategy. I'm just skimming so I'm not 100% sure, but I'll see if I have time to add tests that prove this out.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
motlin: There's no way to implement this in a way that's symmetrical for non-hashing-strategy maps, but the property I want to see is that if two hashing-strategy-maps are equal and use the same hashing strategy, then map.hashCode() returns the same value for both.

mohrezaei: I checked UnifiedMapWithHashingStrategy and we do use strategy there, so let's keep it consistent and use it here as well.

motlin: I actually couldn't find any bugs yet in the existing implementation. I couldn't get an ImmutableEntry to be visible.

```java
        MutableMap<Integer, Integer> map = this.newMapWithConstantStrategy();
        map.put(3, 2);

        int expectedCombinedHashCode = 1 ^ 2; // sum of strategy.computeHashCode(key) ^ value.hashCode()

        assertEquals(expectedCombinedHashCode, map.entrySet().hashCode());
        assertEquals(expectedCombinedHashCode, map.hashCode());

        map.entrySet().forEach(entry -> assertEquals(expectedCombinedHashCode, entry.hashCode()));
        for (Map.Entry<Integer, Integer> entry : map.entrySet())
        {
            assertEquals(expectedCombinedHashCode, entry.hashCode());
        }

        assertTrue(map.entrySet().contains(new ImmutableEntry<>(3, 2)));
        assertTrue(map.entrySet().contains(new ImmutableEntry<>(4, 2)));
        assertFalse(map.entrySet().contains(new ImmutableEntry<>(3, 3)));
        assertFalse(map.entrySet().contains(new ImmutableEntry<>(4, 3)));
```

mohrezaei: (@motlin I think your last comment was meant for line 1893, not this -- continuing there)

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
mohrezaei: There are a few places UnifiedMapWithHashingStrategy is using ImmutableEntry in (`containsEntry`, `batchForEach`, `retainAll`), which seems inconsistent.
I think it's best if we:
1) Use the strategy in the Entry object here (Concurrent)
2) Fix UnifiedMapWithHashingStrategy to consistently use WeakBoundEntry

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategy.java
motlin: I don't really understand whyUnifiedMapWithHashingStrategy.EntrySet#getEntry returns ImmutableEntry. It's mostly only used to see if the result is non-null. But there's one case where they can escape, fromUnifiedMapWithHashingStrategy.EntrySet#batchForEach.

```java
        Set<Map.Entry<Integer, Integer>> entries = map.entrySet();
        if (entries instanceof BatchIterable<?> batchIterable)
        {
            batchIterable.batchForEach(entry -> assertEquals(expectedCombinedHashCode, entry.hashCode()), 0, 1);
        }
```

mohrezaei: Yeah, UMWHS has a couple of leaks (`containsEntry`, `batchForEach` and `retainAll` at least). Do you agree with the plan above?

motlin: Yes I put up #1841 to fix bugs and get test coverage in place. I'd like to land that first, so that we can run those tests on this new concurrent map.

motlin:
Now that #1841 landed, we can rebase this on it and add unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/strategy/ConcurrentHashMapWithHashingStrategyTest.java:

```java
public class ConcurrentHashMapWithHashingStrategyTest implements HashingStrategyMapTestCase
{
    private static final long CURRENT_TIME_MILLIS = System.currentTimeMillis();

    @Override
    public <T> MutableMap<Object, T> newWith(T... elements)
    {
        Random random = new Random(CURRENT_TIME_MILLIS);
        MutableMap<Object, T> result = ConcurrentHashMapWithHashingStrategy.newMap(new DefaultHashingStrategy<>());
        for (T each : elements)
        {
            assertNull(result.put(random.nextDouble(), each));
        }
        return result;
    }

    @Override
    public <K, V> MutableMap<K, V> newWithKeysValues(Object... elements)
    {
        if (elements.length % 2 != 0)
        {
            fail(String.valueOf(elements.length));
        }

        MutableMap<K, V> result = ConcurrentHashMapWithHashingStrategy.newMap(new DefaultHashingStrategy<>());
        for (int i = 0; i < elements.length; i += 2)
        {
            assertNull(result.put((K) elements[i], (V) elements[i + 1]));
        }
        return result;
    }

    @Override
    public MutableMap<Integer, Integer> newMapWithConstantStrategy()
    {
        return ConcurrentHashMapWithHashingStrategy.newMap(new ConstantHashingStrategy());
    }

    @Override
    public boolean supportsNullKeys()
    {
        return false;
    }

    @Override
    @Test
    public void Map_entrySet_setValue()
    {
        MutableMapIterable<String, Integer> map = this.newWithKeysValues("3", 3, "2", 2, "1", 1);
        map.entrySet().forEach(each -> assertThrows(UnsupportedOperationException.class, () -> each.setValue(each.getValue() + 1)));
        assertIterablesEqual(this.newWithKeysValues("3", 3, "2", 2, "1", 1), map);
    }

    @Override
    @Test
    public void MutableMapIterable_entrySet_setValue()
    {
        this.Map_entrySet_setValue();
    }

    /**
     * TODO: Implement a custom replaceAll() using the JDK's retry-loop approach with replace().
     * ConcurrentHashMapWithHashingStrategy's Entry.setValue() throws UnsupportedOperationException,
     * so the default replaceAll (which uses setValue internally) cannot work.
     */
    @Override
    @Test
    public void Map_replaceAll()
    {
        Map<Integer, String> map = this.newWithKeysValues(1, "1", 2, "2", 3, "3");
        assertThrows(UnsupportedOperationException.class, () -> map.replaceAll((k, v) -> v + k));
        assertIterablesEqual(this.newWithKeysValues(1, "1", 2, "2", 3, "3"), map);
    }
}
```

MarcMil:
One testcase is failing now, ```Object_equalsAndHashCode```:
```
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
	at org.eclipse.collections.test.map.mutable.strategy.HashingStrategyMapTestCase.Object_equalsAndHashCode(HashingStrategyMapTestCase.java:79)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
```

https://github.com/eclipse-collections/eclipse-collections/blob/0a404ed70aeff07809e3270f7da8c6a18a08f4ef/unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/strategy/HashingStrategyMapTestCase.java#L79

This fails because the entry is not symmetric, like @motlin already mentioned.
What do we do about that? Override this specific testcase in ```ConcurrentHashMapWithHashingStrategyTest```  and disable it?

Regarding the ECA: I've signed the ECA, but I am unable to link my github account to the eclipse account. Apparently, it seems that it is already linked to some account (although I do not know which one). I've written a mail to the Eclipse infrastructure team on Saturday, but haven't gotten a reply yet. I am not sure how I can proceed there?

motlin:
> This fails because the entry is not symmetric, like @motlin already mentioned. What do we do about that? Override this specific testcase in `ConcurrentHashMapWithHashingStrategyTest` and disable it?

I found that map.entrySet().hashCode() is the same as map.hashCode() for all existing Maps.

KeySet is missing hashCode()

```java
        @Override
        public int hashCode()
        {
            int hashCode = 0;
            for (K key : this)
            {
                hashCode += ConcurrentHashMapWithHashingStrategy.this.strategy.computeHashCode(key);
            }
            return hashCode;
        }
```

EntrySet.contains() has a bug. The last line should be:

```java
        @Override
        public boolean contains(Object o)
        {
            if (!(o instanceof Map.Entry<?, ?>))
            {
                return false;
            }
            Map.Entry<K, V> e = (Map.Entry<K, V>) o;
            Entry<K, V> candidate = ConcurrentHashMapWithHashingStrategy.this.getEntry(e.getKey());
            return candidate != null && Objects.equals(e.getValue(), candidate.getValue());
        }
```

EntrySet is missing a hashCode() implementation.

```java
        @Override
        public int hashCode()
        {
            return ConcurrentHashMapWithHashingStrategy.this.hashCode();
        }
```

MarcMil:
I've added the changes.

mohrezaei:
> KeySet is missing hashCode()
> EntrySet.contains() has a bug. The last line should be:
> EntrySet is missing a hashCode() implementation.

Do we have tests for these?

unit-tests/src/test/java/org/eclipse/collections/impl/map/strategy/mutable/ConcurrentHashMapWithHashingStrategyTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.eclipse.collections.impl.block.factory.HashingStrategies;
+import org.eclipse.collections.impl.map.mutable.ConcurrentHashMapTest;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class ConcurrentHashMapWithHashingStrategyTest extends ConcurrentHashMapTest
```
motlin, line 24: This would need to extend HashingStrategyMapTestCase. Please refer to https://github.com/eclipse-collections/eclipse-collections/pull/1839#issuecomment-3774071393

PR #1841, opened 2026-01-20
Fix UnifiedMapWithHashingStrategy.entrySet().batchForEach() to expose Entries that use the HashingStrategy.

motlin:
Discovered while code reviewing #1839

### DIFF
--- a/unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/strategy/HashingStrategyMapTestCase.java
+++ b/unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/strategy/HashingStrategyMapTestCase.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright (c) 2024 Goldman Sachs and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompany this distribution.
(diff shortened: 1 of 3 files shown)


unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/strategy/HashingStrategyMapTestCase.java
```diff
@@ -0,0 +1,134 @@
...
+
+        if (entries instanceof BatchIterable<?> batchIterable)
+        {
+            batchIterable.batchForEach(entry -> assertEquals(expectedCombinedHashCode, entry.hashCode()), 0, 1);
+        }
+
+        assertTrue(entries.contains(new ImmutableEntry<>(3, 2)));
+        assertTrue(entries.contains(new ImmutableEntry<>(4, 2)));
+        assertFalse(entries.contains(new ImmutableEntry<>(3, 3)));
+        assertFalse(entries.contains(new ImmutableEntry<>(4, 3)));
```
motlin, line 81: To be really thorough, we could also test mutating methods like removeAll and retainAll, but I read them and saw that they are all consistent.

mohrezaei: I checked and we don't have ImmutableEntry referenced anywhere anymore.

PR #1851, opened 2026-02-09
Add NavigableSet support to SortedSetIterable

### LINKED ISSUE #1746: Add support for NagivableSet like SortedSetIterable

tampix:
Resolves #1746.

## Motivation

`SortedSetIterable` lacks equivalents for `NavigableSet` methods (`floor`, `ceiling`, `lower`, `higher`, `descendingSet`), and `MutableSortedSet` implementations throw `UnsupportedOperationException` on `toReversed`. This PR adds full `NavigableSet` support to the sorted set hierarchy.

## Structure

`ImmutableTreeSet` is rewritten to delegate to an `ImmutableList<T>` of elements in logical comparator order. View operations (`descendingSet`, `subSet`, `headSet`, `tailSet`, `take`, `drop`) are O(1) and share the backing list via the `ReversedImmutableList` / `ImmutableSubList` infrastructure (EC 12.0). Wrapper depth is bounded to at most 2 regardless of how many view operations are chained.

Navigation methods (`lower`, `floor`, `ceiling`, `higher`) use `lowerBound`/`upperBound` binary-search helpers. `indexOf` uses comparator-based binary search (consistent with `TreeSet`; documented).

## `SortedSetAdapter` change

Caches the `NavigableSet` delegate at construction time (`private final NavigableSet<T> navigableDelegate`, null when the delegate is not a NavigableSet). All 11 NavigableSet methods delegate through `asNavigableSet()` which throws `UnsupportedOperationException` for non-NavigableSet delegates. Avoids repeated `instanceof` checks.

## API changes

New methods on `SortedSetIterable`:
- `lower(T)`, `floor(T)`, `ceiling(T)`, `higher(T)`
- `descendingSet()`, `descendingIterator()`
- `subSet(T, boolean, T, boolean)`, `headSet(T, boolean)`, `tailSet(T, boolean)`

Default implementations for `reverseForEach`, `reverseForEachWithIndex`, and `detectLastIndex`, delegating through `descendingSet()`.

`MutableSortedSet` now extends `NavigableSet<T>`.
`AbstractImmutableSortedSet` now implements `NavigableSet<T>`.
`ImmutableSortedSet` adds `castToNavigableSet()`.

## Implementation per class

| Class | Strategy |
|---|---|
| `TreeSortedSet` | Delegates to underlying `java.util.TreeSet` NavigableSet methods |
| `ImmutableTreeSet` | Binary search + structural sharing via EC `ImmutableList` view API |
| `ImmutableEmptySortedSet` | Returns `null` for navigation, `this`/empty for views |
| `SortedSetAdapter` | Caches `NavigableSet` cast at construction; throws `UOE` for non-NavigableSet delegates |
| `SynchronizedSortedSet` | Synchronized wrappers; views share the same lock object |
| `UnmodifiableSortedSet` | Read-only delegation; `pollFirst`/`pollLast` throw; views wrapped as unmodifiable |
| `AbstractImmutableSortedSet` | Implements `NavigableSet`; `pollFirst`/`pollLast` throw |

## Complexity

| Operation | Time | Space |
|---|---|---|
| `contains(e)` | O(log n) | O(1) |
| `lower/floor/ceiling/higher(e)` | O(log n) | O(1) |
| `descendingSet()` | O(1) | O(1) |
| `subSet/headSet/tailSet(...)` | O(log n) | O(1) |
| `first()` / `last()` | O(1) | O(1) |

## Test plan

- [x] `ImmutableTreeSetTest`: navigation methods, descendingSet, subSet/headSet/tailSet with inclusive flags, comparator-agnostic IAE test
- [x] `ImmutableEmptySortedSetTest`: navigation on empty returns null, views return empty
- [x] `TreeSortedSetTest`: navigation, pollFirst/pollLast, descendingSet, subSet with flags
- [x] `AbstractSortedSetTestCase`: fixed detectLastIndex, toReversed, reverseForEachWithIndex
- [x] Removed stale `UnsupportedOperationException` overrides from adapter/wrapper tests

Fixes: #1746

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableReversibleArrayList.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/list/immutable/ImmutableReversibleArrayList.java
@@ -74,11 +74,6 @@ public final class ImmutableReversibleArrayList<T>
         return new ImmutableReversibleArrayList<>(elements.clone());
     }
 
-    public static <E> ImmutableReversibleArrayList<E> copyOfRange(E[] array, int from, int to)
-    {
-        return new ImmutableReversibleArrayList<>(Arrays.copyOfRange(array, from, to));
-    }
-
     private int physicalIndex(int logicalIndex)
     {
         return this.ascending
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/immutable/ImmutableTreeSet.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/set/sorted/immutable/ImmutableTreeSet.java
@@ -28,6 +28,7 @@ import org.eclipse.collections.api.block.procedure.Procedure;
 import org.eclipse.collections.api.block.procedure.primitive.ObjectIntProcedure;
 import org.eclipse.collections.api.factory.SortedSets;
 import org.eclipse.collections.api.list.ImmutableList;
+import org.eclipse.collections.api.list.MutableList;
 import org.eclipse.collections.api.map.MapIterable;
 import org.eclipse.collections.api.multimap.sortedset.ImmutableSortedSetMultimap;
 import org.eclipse.collections.api.ordered.OrderedIterable;
@@ -44,6 +45,7 @@ import org.eclipse.collections.impl.lazy.parallel.set.sorted.FlatCollectSortedSe
 import org.eclipse.collections.impl.lazy.parallel.set.sorted.RootSortedSetBatch;
 import org.eclipse.collections.impl.lazy.parallel.set.sorted.SelectSortedSetBatch;
 import org.eclipse.collections.impl.lazy.parallel.set.sorted.SortedSetBatch;
+import org.eclipse.collections.impl.list.fixed.AbstractArrayAdapter;
 import org.eclipse.collections.impl.list.immutable.ImmutableReversibleArrayList;
 import org.eclipse.collections.impl.map.mutable.ConcurrentHashMap;
 import org.eclipse.collections.impl.utility.ArrayIterate;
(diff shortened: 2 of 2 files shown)


motlin:
ImmutableReversibleArrayList is interesting, but I was thinking something like overriding SequencedCollection.reversed() with a mutable view for a mutable type, in immutable view for an immutable type, etc. I'll look through in more detail shortly.

tampix:
@motlin
My understanding is that EC currently targets Java 17+, while `SequencedCollection` is Java 21+. I agree it would be cleaner as an API, but it still comes with non-trivial complexity for lists:

* You still need to handle `binarySearch` properly on a reversed view, which brings some complexity by itself.
* There's also the problem of composing `subList` with a reverse view. Piling wrappers upon wrappers isn't ideal: you're more or less forced to forgo direct iteration over the underlying array and fall back to `iterator()` / `get(index)`, which defeats much of the performance benefit.

The approach in `ImmutableReversibleArrayList` avoids both problems: reversed views and sublists are just different (`from`, `size`, `ascending`) windows over the same backing array, so every iteration method (`each`, `forEachWithIndex`, `toArray`, `binarySearch`) branches on ascending and loops directly over the physical range. No wrappers, no indirection, and `toReversed()` / `subList()` compose freely in O(1) without stacking delegation layers.

That said, I believe the current implementation would adapt easily to `SequencedCollection` if EC moves to Java 21+ in the future. The (`array`, `from`, `size`, `ascending`) representation is essentially what a `reversed()` view needs under the hood: it would just be a matter of wiring it to the `SequencedCollection` API rather than exposing it through EC-specific methods.

The one real issue I see with the current implementation is constructor management.

* `ImmutableReversibleArrayList` must ensure the input array is a defensive copy to guarantee immutability, hence package-private constructors and public factory methods only.
* `ImmutableTreeSet` has the same constraint.
  * Some cases are naturally clean (e.g. constructing from a `SortedSet`, which goes through `toArray`).
  * Others are uglier: the array is copied, then sorted and deduplicated in-place, then copied again to trim. The last copy could be skipped when `uniqueCount == length`, but at the cost of additional complexity and a somewhat brittle design (though documentable).

This would all be much simpler if there were a way (e.g. a wrapper type) to mark an array as "safe" (already a defensive copy) versus not. But absent that, it's the implementer's responsibility to know what they're doing.

P.S. I've pushed a fixup commit that eliminates the extra array copy during sort + dedup when there are no duplicates, by introducing a `SortedUniqueArray` helper that extends `AbstractArrayAdapter` and overrides `toArray()` to return the backing array directly. `ImmutableReversibleArrayList.newList()` then picks it up through `Iterate.toArray()` with no redundant copy. When there are duplicates, `Arrays.copyOfRange` trims as before. I've also removed the now useless `ImmutableReversibleArrayAdapter.copyOfRange`, which was sort of a leaky API.

motlin:
I'm still thinking about this and working on vaguely related PRs. First, more test coverage for sorted data structures:

- https://github.com/eclipse-collections/eclipse-collections/pull/1852
- https://github.com/eclipse-collections/eclipse-collections/pull/1853

And I'm taking a crack at implementing reversed() in a way that is compatible with both Java 17 and 21.

motlin:
I think I got it. Please check #1854

tampix:
@motlin I looked at your PR and it LGTM.

While it has similar characteristics to `java.util.ReverseOrderListView` (simple wrapper that composes with other wrappers such as `java.util.AbstractList.RandomAccessSubList)`, I'm not sure it fits well with the expected usage patterns of `NavigableSet`.

If you look at `java.util.TreeMap.NavigableSubMap` and its subclasses, they took an approach similar to the one I used in #1851 — avoiding composed wrappers in favor of keeping at most one level of indirection.

For `ImmutableTreeSet` specifically, I think this matters because it relies heavily on `Arrays.binarySearch` over the backing array. Stacking wrapper layers would force going through iterator() / get(index) instead, which could have a non-negligible performance impact for legitimate use cases like repeated navigation queries on large sorted sets. I'd argue that, `ImmutableTreeSet` (and `ImmutableTreeMap` down the road) warrants a dedicated optimized list view rather than a generic wrapper.

The alternative would be to drop the array-backed optimization entirely and have ImmutableTreeSet / ImmutableTreeMap simply wrap a TreeSet / TreeMap, but that would come at a significant memory overhead compared to the current approach.

That said, I'd be curious to hear your thoughts on these tradeoffs.

P.S. I've seen in your PR a comment about `reversed().add(Object)` losing a guarantee to be O(1) — seems expected according to the Java 21 spec, and isofunctional with their implementation.

motlin:
@tampix I am not too worried about the performance overhead of extra wrappers as long as it's a constant number of wrappers. This conversation reminded me that list.subList(...).subList(...) is optimized to return a single wrapper, not two. I haven't optimized the reversed view, but I could with code like:

```java
  @Override
  public MutableList<T> subList(int fromIndex, int toIndex)
  {
      return this.delegate.subList(this.size() - toIndex, this.size() - fromIndex).reversed();
  }
```

tampix:
@motlin Sorry, I was busy, so I didn't keep up to date since then.

I've seen you have merged https://github.com/eclipse-collections/eclipse-collections/pull/1854, plus lots of tests on subLists and navigable interfaces.

I can rework this PR to use these and focus on trying to minimize the amount of wrappers if you want :)

motlin:
> @motlin Sorry, I was busy, so I didn't keep up to date since then.
> 
> I've seen you have merged #1854, plus lots of tests on subLists and navigable interfaces.
> 
> I can rework this PR to use these and focus on trying to minimize the amount of wrappers if you want :)

sounds good to me!

PR #1854, opened 2026-02-10
Add covariant reversed() overrides with O(1) view implementations for lists and sorted sets.

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/list/mutable/ReversedRandomAccessMutableList.java
```diff
@@ -0,0 +1,124 @@
...
+    public void add(int index, T element)
+    {
+        Objects.checkIndex(index, this.size() + 1);
+        this.delegate.add(this.size() - index, element);
+    }
+
+    @Override
+    public boolean add(T element)
+    {
+        this.delegate.add(0, element);
```
mohrezaei, line 65: not exactly O(1)?

motlin: True! I don't think there's a better way though, is there? Should I remove the comment about O(1)?

mohrezaei: A linked list (appropriately implemented) or a deq would be O(1). Too bad we don't have either 😄

PR #1855, opened 2026-02-11
Replace deprecated boxed-type constructors with valueOf() in tests.

motlin:
The boxed constructors like new Integer(1) are marked deprecated for removal, and emit warnings during the build. Some are used for tests that care about reference equality but many aren't, and the ones that aren't are removed in this PR.


donraab (approved):
LGTM

PR #1861, opened 2026-02-18
Improve test coverage in LazyNoIteratorTestCase implementations.

motlin:


unit-tests-java8/src/test/java/org/eclipse/collections/test/lazy/CompositeIterableTestNoIteratorTest.java
```diff
@@ -27,4 +32,29 @@ public <T> LazyIterable<T> newWith(T... elements)
         Iterable<T>[] iterables = (Iterable<T>[]) noIteratorLists.toArray(new Iterable[]{});
         return CompositeIterable.with(iterables);
```
motlin, line 33: I never noticed this before but CompositeIterable is a bit different than all the others in that it has a private constructor. I can't create a subclass that always throws from iterator(). Instead, this test has the Composite wrap other iterables which don't support iterators. It's pretty close but requires me to override the three tests below now that they have such strict assertions in the parent test.

motlin: I just remembered I'm talking as if another PR has landed but I split it up. I'm about to make these tests more strict in f9cd1dec894971374496671a8650befe87779131

PR #1865, opened 2026-02-21
Update CONTRIBUTING.md and README_EXAMPLES.md to use Java 17.

donraab:


CONTRIBUTING.md
```diff
@@ -33,7 +33,7 @@ Build
 -----
 The Eclipse Collections build requires below as dependencies.
 
-- Java 11+
+- Java 17+
 - Maven 3.1.0+
```
pzygielo, line 37: Seems that even maven 3.8.9 won't do.

donraab: @pzygielo Thanks, I was wondering what minimum version of Maven would work with 17. I think I will be conservative and increase the recommended version to 3.9.6+ and get consensus from the other committers.

pzygielo: To clarify - maven 3.1.0 is too low for reasons not related to Java version. For example
https://github.com/eclipse-collections/eclipse-collections/blob/7070213d39c0015a6e483b10817da0d08aa77d8d/.mvn/maven.config#L3
is available only in much much later maven. And few more reasons for 3.8.9 to not work.

PR #1883, opened 2026-03-16
Provide efficient removeIf implementation for FastList (#147)

### LINKED ISSUE #147: java 8 convergence: provide efficient implementation of Java 8 default methods

sgworld123:
## Summary
Provides an efficient override of `removeIf(Predicate)` in `FastList` 
instead of falling back to the JDK default which uses an Iterator.

## Changes
- Overridden `removeIf` in `FastList` using direct array traversal
- Added tests in `FastListTest`

Closes #147


motlin:
The issue #147 covers more than this. I started to write up what else would need to be done but figured it would be easier to show. I'm closing this in favor of #1890

PR #1897, opened 2026-03-23
Support wrapping in UntouchableMutableList.asUnmodifiable() and asSynchronized().

motlin:


donraab (commented):
@motlin This PR doesn't make sense to me. An UntouchableMutableList can only exist in the context of a read or write lock. Why would it ever be useful to create a synchronized or unmodifiable version?

motlin:
> @motlin This PR doesn't make sense to me. An UntouchableMutableList can only exist in the context of a read or write lock. Why would it ever be useful to create a synchronized or unmodifiable version?

I hadn't really thought about that. My attention wound up here when I was working on the reversed list view, and I wasn't sure if we could support views on untouchable collections. I was looking at how we do all this work to track iterators and sublists, and how if we support sublists we could definitely support other views as well. I implemented the reversed view, but couldn't figure out why we left unmodifiable and synchronized unimplemented. I still think unmodifiable may be useful but synchronized probably isn't.

motlin:
@donraab I think it still makes sense to support unsupported specifically, in case we want to pass the list to another method. Maybe not synchronized though. What do you think?

PR #1900, opened 2026-03-25
use maven.compiler.release

TheJavaGuy:
Maven compiler plugin documentation strongly recommends to use `maven.compiler.release` instead of old `maven.compiler.source` and `maven.compiler.target`. This is documented at https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html and https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html.

This PR removes `maven.compiler.source` and `maven.compiler.target` from the `pom.xml` and uses `maven.compiler.release` instead. This also removed a few build warnings.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
@@ -818,7 +818,7 @@ abstract class AbstractMutableBiMap<K, V> extends AbstractBiMap<K, V> implements
                 K next = iterator.next();
                 if (!collection.contains(next))
                 {
-                    this.remove(next);
+                    iterator.remove();
                 }
             }
             return oldSize != AbstractMutableBiMap.this.size();


nikhilnanivadekar:
Will this cause issues for release builds? I am supportive of maven hygiene, however, the release process always throws a twist in things.

TheJavaGuy:
> Will this cause issues for release builds? I am supportive of maven hygiene, however, the release process always throws a twist in things.

From the logical perspective I don't see how it could cause issues; however I don't know enough of release process to answer the question. How the release builds differ from "normal" checks executed on every PR?

Maybe one of you can chip in here @motlin @donraab @mohrezaei?

donraab:
@nikhilnanivadekar I was working on adding the early access build for JDK 27, and it seems to require the release setting as @TheJavaGuy has it. If it causes an issue, we'll just have to revert or be redundant and have both.

TheJavaGuy:
GitHub closed the PR without warning when I tried to rebase which is definitely **not** what I wanted to do. I also don't have permission to reopen :(

TheJavaGuy:
I see now the same change already landed with commit `f30201c0ea098c953e38297705e2a4cca6781690` by @donraab so this PR is not needed anymore.

donraab:
Hi @TheJavaGuy, yes, IIRC I had to add this when I upgraded our early access build to JDK 27.

Issue #1798, opened 2025-09-30
Add support for Enum optimized collections


tampix:
### Problem Statement

Eclipse Collections currently lacks dedicated implementations for enum-based collections, while the JCF provides `EnumSet` and `EnumMap` as memory-efficient and performance-optimized containers for enums. This gap forces users to either:
1. Fall back to JCF types, losing the benefits of Eclipse Collections' rich API
2. Use general-purpose collections with unnecessary memory overhead and reduced performance
3. Always use `Sets.adapt` / `Maps.adapt` depite the lack of immutable variants

### Proposed Solution

Add enum-specific collection types following Eclipse Collections' mutable/immutable pattern:

#### Core types (high priority)
- `MutableEnumSet` / `ImmutableEnumSet`
- `MutableEnumMap` / `ImmutableEnumMap`

#### Extended types (medium to low priority)
- `MutableEnumBiMap` / `ImmutableEnumBiMap` -> use cases are limited
- Support for all Multimaps combinations -> very complex API wise

### Benefits

**For MutableEnumSet/MutableEnumMap:**
- Maintains API consistency with `SetIterable`/`MapIterable`
- Provides performance benefits of JCF's enum-optimized implementations
- Enables seamless integration with existing Eclipse Collections code

**For ImmutableEnumSet/ImmutableEnumMap:**
- Allows caching of declarative configurations with immutability guarantees
- Maintains API consistency across the library
- Supports functional programming patterns common in Eclipse Collections usage

**For MutableEnumBiMap/ImmutableEnumBiMap:**
- Completes the enum collection family for consistency
- Utility is more limited but valuable for specific use cases

**For enum-based multimaps:**
- Enables efficient enum-to-collection mappings
- See notes below regarding API complexity considerations

### Notes for Multimaps

Enum-based multimaps introduce significant API complexity due to combinatorial explosion of factory methods and type combinations:

```java
// Different factory entry points depending on key/value types:
var objToEnums = Multimaps.mutable.enums.of("", AnEnum.KEY);
var enumToObj = EnumMultimaps.mutable.list.of(AnEnum.KEY, "");
var enumToEnums = EnumMultimaps.mutable.enums.of(AnEnum.KEY, AnotherEnum.STUFF);

// Usage can become verbose:
MutableEnumEnumMultimap<AnEnum, AnotherEnum> enumToEnums = EnumSets.mutable
    .allOf(AnotherEnum.class)
[...]

PR #1918, opened 2026-04-04
Add EnumUniverse: ClassValue-based cache for enum constants

tampix:
## Summary

- Adds `EnumUniverse<E>`, a `ClassValue`-based cached accessor for the universe (array of all constants) of an enum type
- First building block for enum-optimized collections (`MutableEnumSet`, `MutableEnumMap`, etc.)
- `Class.getEnumConstants()` allocates a defensive copy on every call — `EnumUniverse` eliminates that overhead via a GC-aware, thread-safe `ClassValue` cache

### `eclipse-collections/.../impl/utility/internal/EnumUniverse.java` (new)

- `forClass(Class<E>)` — factory method with `ClassValue` cache lookup; rejects non-enum types and anonymous enum subclasses via `isEnum()` guard
- `fromOrdinal(int)` — returns the enum constant at the given ordinal
- `size()` — returns the number of constants in the enum type
- Backing array never exposed (encapsulation by design)
- `Objects.requireNonNull` in constructor for defense-in-depth against null constants array

### `unit-tests/.../impl/utility/internal/EnumUniverseTest.java` (new)

6 tests covering:
- `size()` for multi-constant, singleton, and body-override enums
- `fromOrdinal()` for all constants of a 3-value enum
- Out-of-bounds ordinal (negative and beyond range)
- `ClassValue` caching identity (`assertSame`)
- Rejection of anonymous enum subclasses (`WithBody.VALUE.getClass()`)
- Rejection of non-enum classes via raw type (`Object.class`)

### `eclipse-collections/.../impl/utility/internal/package-info.java` (updated)

- Added `EnumUniverse` to the package documentation listing

## Design decisions

| Decision | Rationale |
|----------|-----------|
| `ClassValue` over `ConcurrentHashMap` | GC-aware — entries evicted on class unload, handles OSGi and dynamic classloaders correctly |
| `asSubclass(Enum.class)` in `computeValue` | Clearer failure mode than raw cast; runtime check at cache population |
| `isEnum()` guard in `forClass` | Anonymous enum subclasses (`Foo.VALUE.getClass()`) have `isEnum() == false`; rejects them early with a clear message |
| `Objects.requireNonNull` in constructor | Defense-in-depth — structurally prevents a null-constants `EnumUniverse` from entering the cache |
| Minimal API (`size` + `fromOrdinal`) | Only what `MutableEnumSet`/`MutableEnumMap` need; no array exposure, no mutable collection wrapper |
| `forClass` naming over `of` | `of` is idiomatic in Eclipse Collections for creating instances with values; `forClass` communicates cache lookup semantics |
| `fromOrdinal` naming over `get`/`valueOf` | Explicit about the ordinal→constant mapping; avoids collision with `Enum.valueOf(String)` |

## Test plan

- [x] `mvn test -Dtest=EnumUniverseTest` — 6/6 pass
- [x] `mvn checkstyle:check` — 0 violations

Ref: #1798

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/EnumUniverse.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/utility/internal/EnumUniverse.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2026 Goldman Sachs and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompany this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.impl.utility.internal;
+
+import java.util.Objects;
+
+/**
+ * A cached, GC-aware accessor for the universe (array of all constants) of an enum type.
+ * <p>
+ * {@code Class.getEnumConstants()} allocates a defensive copy on every call. This class
+ * uses a {@link ClassValue}-based cache to eliminate that overhead for hot-path operations
+ * in enum-optimized collections ({@code MutableEnumSet}, {@code MutableEnumMap}, etc.).
+ * <p>
+ * The backing array is never exposed; all access goes through {@link #size()} and
+ * {@link #fromOrdinal(int)}.
+ *
+ * @since 14.0
+ */
+public final class EnumUniverse<E extends Enum<E>>
+{
+    @SuppressWarnings("rawtypes")
+    private static final ClassValue<EnumUniverse<?>> CACHE = new ClassValue<>()
+    {
+        @Override
+        protected EnumUniverse<?> computeValue(Class<?> type)
+        {
+            return new EnumUniverse<>(type.asSubclass(Enum.class).getEnumConstants());
+        }
+    };
+
+    private final E[] constants;
+
+    private EnumUniverse(E[] constants)
+    {
+        this.constants = Objects.requireNonNull(constants);
+    }
+
+    /**
+     * Returns the cached {@code EnumUniverse} for the given enum type.
+     *
+     * @param enumClass the enum class; must be a direct enum type, not an anonymous subclass
+     * @throws IllegalArgumentException if {@code enumClass} is not a direct enum type
+     */
+    @SuppressWarnings("unchecked")
+    public static <E extends Enum<E>> EnumUniverse<E> forClass(Class<E> enumClass)
+    {
+        if (!enumClass.isEnum())
+        {
+            throw new IllegalArgumentException(enumClass + " is not a direct enum type");
+        }
+        return (EnumUniverse<E>) CACHE.get(enumClass);
+    }
+
+    /**
+     * Returns the enum constant with the given ordinal.
+     */
+    public E fromOrdinal(int ordinal)
+    {
+        return this.constants[ordinal];
+    }
+
+    /**
+     * Returns the number of constants in the enum type.
+     */
+    public int size()
+    {
+        return this.constants.length;
+    }
+}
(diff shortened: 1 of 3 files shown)


motlin:
I've never looked into building an EnumMap before and don't fully understand why a utility like this is necessary. Is it because the JDK EnumMap uses JDK internals that we would avoid, to get access to the "key universe"?

tampix:
Yes. If you read the issue I referenced in the desc (https://github.com/eclipse-collections/eclipse-collections/issues/1798), JCF uses internal low level tricks to track the key universe of enums, relying on the underlying shared array to do it in-place.

Sadly, it's not something accessible to end users, as exposing that would be extremely unsafe. Things like end-users modifying the shared array and break enums beyond repair.

A naive approach can have terrible effects in term of performance and GC pressure when used unknowlingly, as the facilities offered to Java devs :

* Enum.values()
* Class.getEnumConstants()

Are both safe : meaning they both return a clone of the underlying shared array (so each call = allocation + System.arraycopy).

It is a good thing when used sporadically, but dangerous in the context of things like EnumSet, where you could want to allocate them on the fly in a tight loop, naively thinking it's just some arithmetics over a long bitfield.

I described the core problem and each approaches I studied in the issue to circumvent it.

motlin:
This makes sense to me, and I'm usually in favor of landing small PRs, but maybe this functionality would be best to land together with an enum map? At the moment it would be unused by the rest of the library.

tampix:
I'm fine with implementing at least EnumSet and/or EnumMap to go along with this PR :)

nikhilnanivadekar:
I am not sure if Eclipse Collections is the right place to have these Enum specific collections.

tampix:
@nikhilnanivadekar I gave the rationale in https://github.com/eclipse-collections/eclipse-collections/issues/1798

Currently, lots of config code has to rely on a mix of SetIterable/MapIterable + `asUnmodifiable` + proper usage of adapters.

Immutable enum maps are somewhat common in large app code, where you tend to cache some sort of state machines / strategy routers.

Plus, I'd argue that, as the library has dedicated primitives collections, Enum-optimized variants also have their place for similar reasons (mainly performance and memory).

Does that framing make sense to you?

nikhilnanivadekar:
It makes sense. However, this is such a niche use case, I am not able to rationalize the cost to add, maintain, and update these data structures. I might be missing something, hence I asked. I’ll be happy to change my stance.

tampix:
@nikhilnanivadekar That's fair, and I want to make sure I'm not over-selling a niche. Let me try to reframe what I think the value could be, and you can tell me if it still doesn't add up:

1. There might be a nice synergy with #1901. [JDK-6278287](https://bugs.openjdk.org/browse/JDK-6278287) has been open for ~20 years asking `EnumSet` to implement `SortedSet` : enums have a natural ordering via the ordinal, but the JDK never wired it up (probably for binary-compat reasons). Now that EC just got `NavigableSet` support, an enum-backed sorted/navigable set feels like it would slot in fairly naturally. I could be wrong about how much demand there is, but it does seem like something the JDK won't ship and EC is well-placed for.
2. Immutable variants might be where this earns its keep. `Collections.unmodifiableSet(EnumSet.of(...))` is just a wrapper. Enum sets commonly end up in static final fields modeling state machines or strategy routers, where a true immutable type with with/without/builders would be a much better fit — which is exactly what EC's Immutable* hierarchy provides elsewhere. This is probably the use case where the proposal would deliver the most concrete value.
3. Smaller wins on factory ergonomics. `EnumSet.complementOf` requires an existing `EnumSet`, and `EnumSet.copyOf(Collection)` doesn't handle empty collections cleanly => well-known gotchas that come up regularly. Minor papercuts on their own, but easy to smooth over here.
4. On the maintenance cost specifically. I think this is the part I should have led with: SetIterable already specifies difference, intersect, union, isSubsetOf, etc. So we wouldn't really be adding new API surface — just providing an implementation of an existing contract, optimized for the enum case. That feels like a less scary maintenance proposition than it might sound at first.

Would this work as a way forward? I could extend the PR to actually implement `MutableEnumSet` and `ImmutableEnumSet`, and then we'd have something concrete to evaluate real integration cost, real perf numbers, real API feel. If after seeing it you still think the value isn't there, I'd completely understand walking away from it. But I'd rather give you something tangible to react to than keep arguing in the abstract. Does that sound reasonable?

PR #1932, opened 2026-04-24
Fix retainAll in BiMap ValuesCollection to use iterator.remove().

motlin:
Very similar to #1929 but in a different view.

### DIFF
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
@@ -953,7 +953,7 @@ abstract class AbstractMutableBiMap<K, V> extends AbstractBiMap<K, V> implements
                 V next = iterator.next();
                 if (!collection.contains(next))
                 {
-                    this.remove(next);
+                    iterator.remove();
                 }
             }
             return oldSize != AbstractMutableBiMap.this.size();


eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
```diff
@@ -953,7 +953,7 @@ public boolean retainAll(Collection<?> collection)
                 V next = iterator.next();
                 if (!collection.contains(next))
                 {
-                    this.remove(next);
+                    iterator.remove();
```
Desislav-Petrov, line 956: I assume this is done to avoid throwing a `ConcurrentModificationException` ? If so, wondering if it's worth to add a test

Desislav-Petrov (approved):
Approve with a suggestion.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
motlin: No, because the old code navigates through the structure using the iterator, but removes without using it, it misses elements - whichever element shifts from the end of the collision chain into the slot where we are removing an element. I don't think I'm going to land this as is. I'll rewrite the removeAll/retainAll to iterate through the backing structure like forEachWithKeyValue would, and fix the bug while avoiding the iterator entirely. The test failures in #1939 exercise the bug fix here.

motlin: I wound up replacing this PR with #1937

PR #1937, opened 2026-04-30
Fix ValuesCollection.removeAll() and retainAll() in BiMap and UnifiedMap.

motlin:


mohrezaei:
What is this fixing? the new implementation is potentially much slower.

motlin:
Thanks @mohrezaei, these are all bug fixes. I'm still working on the test suite but I put it up in #1939. They are fairly simple tests of the views as if they were regular collections. In other words, they don't even assert that the view and the map are in sync. They're just asserting stuff like after removing all items, they are all removed. I was looking for a specific bug that is caught by the Guava suite that's much more of an edge case, and started getting more coverage in place, and stumbled upon this.

mohrezaei:
@motlin sorry, I think I misread the diff. This is only on `map.values().removeAll()`, etc. (not `map.removeAll()`). The specific fix here is that `values()` may have duplicates, all of which need to be removed.

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
```diff
@@ -812,15 +812,8 @@ public boolean addAll(Collection<? extends K> source)
...
-            Iterator<K> iterator = this.iterator();
-            while (iterator.hasNext())
-            {
-                K next = iterator.next();
-                if (!collection.contains(next))
-                {
-                    iterator.remove();
-                }
-            }
+            AbstractMutableBiMap.this.delegate.keySet().retainAll(collection);
```
mohrezaei: I don't understand why this change is necessary. Is something wrong with `bimap.keySet().iterator().remove()`?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/UnifiedMap.java
```diff
@@ -3229,31 +3229,79 @@ public boolean remove(Object o)
...
         public boolean removeAll(Collection<?> collection)
         {
-            // todo: this is N^2. if c is large, we should copy the values to a set.
-            boolean changed = false;
+            return this.removeByValueCondition(collection, true);
+        }
 
-            for (Object obj : collection)
+        @Override
+        public boolean retainAll(Collection<?> collection)
```
mohrezaei, line 3236: tests?

motlin: I'm working on the tests in #1939. I could land them together, but I also like to put them up as separate PRs so that we can see the failing tests.

```
[ERROR] Failures: 
[ERROR] org.eclipse.collections.test.bimap.mutable.HashBiMapInverseTest.Collection_retainAll
[ERROR]   Run 1: HashBiMapInverseTest.Collection_retainAll expected: <[]> but was: <[3]>
[INFO]   Run 2: PASS
[INFO] 
[ERROR] org.eclipse.collections.test.bimap.mutable.HashBiMapTest.Collection_retainAll
[ERROR]   Run 1: HashBiMapTest.Collection_retainAll expected: <[]> but was: <[1]>
[INFO]   Run 2: PASS
[INFO] 
[ERROR] org.eclipse.collections.test.map.mutable.UnifiedMapTest.Collection_remove_removeAll
[ERROR]   Run 1: UnifiedMapTest.Collection_remove_removeAll expected: <false> but was: <true>
[INFO]   Run 2: PASS
```

eclipse-collections/src/main/java/org/eclipse/collections/impl/bimap/mutable/AbstractMutableBiMap.java
motlin: There's no bug with consistent use of iterator.next() and iterator.remove() but the testing strategy I used in #1939 applied lots of existing tests to our views, and they include the tests that assert no delegation to iterator for a set of methods. It exposed the minor inconsistency that this is the only Map implementation in the library that uses iterator in its implementation of retainAll. I thought about overriding the one test to allow the inconsistency. Then I thought about this implementation and I figured it's a decent implementation anyway.

mohrezaei: > Then I thought about this implementation and I figured it's a decent implementation anyway.

I'm not sure this is true. For a large BiMap, worse case might be: bimap.retainAll(bimap.keySet())

mohrezaei (approved):

PR #1939, opened 2026-05-01
Add @Nested KeySet and Values view tests to all map test classes.

motlin:


unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/sorted/TreeMapTest.java
```diff
@@ -10,16 +10,33 @@
...
+import static org.eclipse.collections.test.IterableTestCase.assertIterablesEqual;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
 public class TreeMapTest implements NavigableMapTestCase
 {
+    private static final long CURRENT_TIME_MILLIS = System.currentTimeMillis();
```
Desislav-Petrov, line 29: Nit: This introduces some non-determinism in the tests, given how it's used it's unlikely for it to cause an issue but might still be worth to just capture a static value that doesn't change?

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/sorted/TreeSortedMapNoIteratorTest.java
```diff
@@ -10,20 +10,28 @@
...
+import org.eclipse.collections.test.map.MapValuesCollectionTestCase;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.fail;
 
 public class TreeSortedMapNoIteratorTest implements MutableSortedMapIterableTestCase, OrderedIterableNoIteratorTestCase
 {
+    private static final long CURRENT_TIME_MILLIS = System.currentTimeMillis();
```
Desislav-Petrov, line 33: Same comment as above, and I think the same pattern is used in a couple of other places below.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/FixedSizeMapTest.java
```diff
@@ -95,4 +100,57 @@ public <K, V> MutableMap<K, V> newWithKeysValues(Object... elements)
...
+        public final <T> Collection<T> newWith(T... elements)
+        {
+            return ((Map<Object, T>) FixedSizeMapTest.this.newWith(elements)).values();
+        }
+
+        @Test
+        @Override
+        public void Object_equalsAndHashCode()
+        {
+            IterableTestCase.assertIterablesNotEqual(this.newWith(4, 3, 2, 1), this.newWith(3, 2, 1));
```
Desislav-Petrov:
```suggestion
            IterableTestCase.assertIterablesNotEqual(this.newWith(4, 3, 2, 1), this.newWith(3, 2, 1));
IterableTestCase.assertIterablesNotEqual(this.newWith(3, 2, 1), this.newWith(3, 2, 0));            
```

Worth expanding with 1 more test case - same size but different value?

Desislav-Petrov (commented):
couple of nits but looks good

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/sorted/TreeMapTest.java
motlin: Not necessarily defending the choice, but this is consistent with patterns we've been using for a long time. One example:

```java
public class ImmutableHashBiMapInverseTest implements ImmutableBiMapTestCase
{
    private static final long CURRENT_TIME_MILLIS = System.currentTimeMillis();

    @Override
    public <T> ImmutableBiMap<Object, T> newWith(T... elements)
    {
        Random random = new Random(CURRENT_TIME_MILLIS);
        MutableBiMap<T, Object> result = new HashBiMap<>();
        for (T each : elements)
        {
            assertNull(result.put(each, random.nextDouble()));
        }
        return result.toImmutable().inverse();
    }
```

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/FixedSizeMapTest.java
motlin: I'm going to redo this section. The thing here is that the super implementation does something like assertEquals(this.newWith(3, 2, 1), this.newWith(3, 2, 1)) which doesn't pass for collection views with no implementation of equals. I'm thinking I will add one of those boolean methods which answers whether the collection supports equality, and then assert equal or not equal accordingly. It should get rid of the override entirely.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/sorted/TreeMapTest.java
Desislav-Petrov: Fair.

PR #1943, opened 2026-05-07
Add OrderedMaps factory in eclipse-collections-api.

motlin:


eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/MutableOrderedMapFactoryImpl.java
```diff
@@ -0,0 +1,148 @@
...
+
+@aQute.bnd.annotation.spi.ServiceProvider(MutableOrderedMapFactory.class)
+public class MutableOrderedMapFactoryImpl implements MutableOrderedMapFactory
+{
+    public static final MutableOrderedMapFactory INSTANCE = new MutableOrderedMapFactoryImpl();
+
+    @Override
+    public <K, V> MutableOrderedMap<K, V> empty()
+    {
+        return OrderedMapAdapter.adapt(new LinkedHashMap<>());
```
motlin, line 28: One reason I never did this before is that we don't have our own ordered map - yet. I have one ready but the PR is big so I'm splitting it up.

PR #1944, opened 2026-05-11
Eclipse Collections AI Policy

mohrezaei:


motlin:
Overall the policy is balanced and pragmatic. I don't love putting "Assisted-by: Claude:claude-3-opus" into version control but could live with it. I definitely wouldn't mind putting it into pull request descriptions.

vmzakharov:
The policy sounds sensible. Just one comment along possibly the similar lines of @motlin was talking about. If we say, "The submitting person must understand and be able to discuss, modify and take responsibility for the submitted code", do we also need to have a notice when "submitting non-trivial AI generated code"? The human making the submission is wholly responsible for it after all?

mohrezaei:
> do we also need to have a notice when "submitting non-trivial AI generated code"

There are two reasons this made sense to me:

1) The law on this stuff is on our side right now, but it may change (Don's worry), and we may have to account for everything in the codebase that's AI generated sometime in the future.
2) For another perspective, let's consider the question "Do we care about the author field for a commit that's 10 years old?" I'd say, not really, it's just a historical curiosity now. It'll be interesting to look back at these commits in 10 years time and see how they hold up.

AI_POLICY.md
```diff
@@ -0,0 +1,29 @@
...
+          (e.g., coccinelle, sparse, smatch, clang-tidy)
+
+        Basic development tools (git, Eclipse, Intellij, etc) should not be listed.
+
+        Example::
+
+          Assisted-by: Claude:claude-3-opus coccinelle sparse
+
+	- No AI specific files should be included in the commit.
+	- All other contribution guidelines must be followed, including copyright/license compatibility and notices, Eclipse ECA, etc.
```
Desislav-Petrov, line 28: Does it make sense to include a line to cover `Reasonably sized PRs that a human can review & understand` or do we think this is given irrespective of how the contribution was generated? I guess with AI generating a largely PR is a lot more likely

Desislav-Petrov (commented):
Looks sensible.

Desislav-Petrov:
> Overall the policy is balanced and pragmatic. I don't love putting "Assisted-by: Claude:claude-3-opus" into version control but could live with it. I definitely wouldn't mind putting it into pull request descriptions.

I wonder if we should reflect the AI-usage at the commit level, or in the code itself - mark sections, or append something to the file headers?

AI_POLICY.md
mohrezaei: Not sure that's AI specific, as I've seen humans break that quite often and the response is always the same: break it up. Sometimes, that's not even sensible, which leads to PRs languishing.

mohrezaei:
> I wonder if we should reflect the AI-usage at the commit level, or in the code itself - mark sections, or append something to the file headers?

We generally don't mark stuff in code; git + diff are better at that. 
We also don't put authors in headers (and AI is really the author role here).

We do use headers for copyright/license, but we've stayed away from being super particular, generally sticking with a single copyright + boiler plate license. We could revisit that (for example, most code contributed since the open sourcing is not really copyright Goldman Sachs), but I think that's a separate topic. In general, I think we're very much average in our handling of that. As an example, here is a tiny file in linux kernel tree, with 8 contributors:
https://github.com/torvalds/linux/blame/master/kernel/dma/remap.c
There is a single copyright: "Copyright (c) 2014 The Linux Foundation". The file is edited more recently and likely touched by folks not from the linux foundation. Not trying to make fun here, just saying it's not a particularly worth while effort. If push came to shove, git + diff would be good enough.

donraab:
This looks good to me. Thanks for submitting @mohrezaei. One thing I would request is that an update is added to this page (https://github.com/eclipse-collections/eclipse-collections/wiki/Working-with-GitHub) to include an example of a commit with an Assisted-by: trailer using `--trailer`. We may want to reference the example in the AI Policy doc as well so folks can find it easily.

There seems to be some convergence on this approach in the wild. https://allthingsopen.org/articles/open-source-ai-contributions-assisted-by-git-trailer-standard

Apache Groovy has added this: https://github.com/apache/groovy/blob/master/AGENTS.md

donraab (approved):
The policy looks good to me. I think we've given enough time for folks to comment. I request that the other committers submit a review as well.

I would like an update to this page (https://github.com/eclipse-collections/eclipse-collections/wiki/Working-with-GitHub) to include an example of a commit with an Assisted-by: trailer using `--trailer`. We should reference the example in the AI Policy doc as well so folks can find it easily.

mohrezaei:
> I would like an update to this page (https://github.com/eclipse-collections/eclipse-collections/wiki/Working-with-GitHub) to include an example of a commit with an Assisted-by: trailer using --trailer. We should reference the example in the AI Policy doc as well so folks can find it easily.

Edited the wiki page and added the reference

vmzakharov (approved):

donraab:
@motlin @nikhilnanivadekar @prathasirisha @Desislav-Petrov Please submit a review for this PR. Thanks!

Desislav-Petrov (approved):

motlin (approved):

nikhilnanivadekar (approved):

prathasirisha (approved):

PR #1947, opened 2026-05-20
Override Java 8 Map default methods for efficiency and correctness (#147, #500)

snowykte0426:
## Summary

Closes part of #500 (and relates to #147). Adds efficient, thread-safe overrides for the remaining Java 8 `Map` default methods that were missing from the eclipse-collections implementation.

### What was missing

The Java default implementations for `computeIfAbsent`, `computeIfPresent`, and `compute` perform 2–3 separate hash lookups and, for synchronized wrappers, acquire and release the lock multiple times per call — making them both inefficient and non-atomic.

### Changes

**`AbstractMutableMap`** — adds correct fallback implementations for all subclasses:
- `computeIfAbsent(K, Function<? super K, ? extends V>)`
- `computeIfPresent(K, BiFunction<? super K, ? super V, ? extends V>)`
- `compute(K, BiFunction<? super K, ? super V, ? extends V>)`

**`UnifiedMap`** — adds efficient single-lookup implementations following the established `merge`/`chainedMerge` pattern. Each public method is paired with a `chained*` private helper for collision chains, reducing hash lookups from 2–3 to 1:
- `computeIfAbsent` + `chainedComputeIfAbsent`
- `computeIfPresent` + `chainedComputeIfPresent`
- `compute` + `chainedCompute`

**`AbstractSynchronizedMapIterable`** — adds atomic synchronized wrappers (correctness fix). The Java default implementations are **not thread-safe** because they call `get()` and `put()` as two separate synchronized operations. These overrides hold the lock across the entire operation. `SynchronizedMutableMap`, `SynchronizedBiMap`, and `SynchronizedSortedMap` all inherit these fixes:
- `computeIfAbsent`, `computeIfPresent`, `compute`
- `replaceAll` (was missing — using the Java default was non-atomic)
- `putIfAbsent` (was missing — using the Java default was non-atomic)

### What is NOT in this PR (future work)

- `AbstractMutableBiMap` / `AbstractMutableSortedMap` — do not extend `AbstractMutableMap`; they fall back to Java defaults which are functionally correct through their overridden `put`/`remove`, but lack efficiency optimizations.
- `UnifiedMapWithHashingStrategy` — has its own `chainedMerge`; a follow-up could add `chainedCompute*` variants similarly.
- `MapAdapter` / `OrderedMapAdapter` — inherit through `AbstractMutableMap` or Java defaults; a follow-up could add explicit delegation to the underlying map's implementations.

### Test plan

- [ ] Existing `MapTestCase` interface tests (`Map_computeIfAbsent`, `Map_computeIfPresent`, `Map_compute`, `Map_replaceAll`, `Map_putIfAbsent`) automatically apply to all concrete implementations via interface inheritance (`UnifiedMapTest`, `SynchronizedMutableMapTest`, `UnmodifiableMutableMapTest`, `MapAdapterTest`, etc.)
- [ ] Run `mvn test -pl unit-tests-java8` — all tests pass (134/134 per class, 0 failures)
- [ ] Null-value edge cases verified: key mapped to `null` correctly triggers `computeIfAbsent` function call per Java 8 spec
- [ ] Exception propagation verified: function throwing `RuntimeException` leaves map unchanged

### DIFF
--- a/RELEASE_NOTE_DRAFT.md
+++ b/RELEASE_NOTE_DRAFT.md
@@ -1,5 +1,11 @@
 # Unrreleased
 
+## Bug Fixes
+* Fixed `SynchronizedMutableMap` (and `SynchronizedBiMap`, `SynchronizedSortedMap`) so that `computeIfAbsent`, `computeIfPresent`, `compute`, `replaceAll`, and `putIfAbsent` hold the lock across the entire operation. The Java default implementations were not atomic because they invoked `get()` and `put()` as two separate synchronized calls. ([#147](https://github.com/eclipse-collections/eclipse-collections/issues/147), [#500](https://github.com/eclipse-collections/eclipse-collections/issues/500))
+
+## Performance Improvements
+* `UnifiedMap` and `UnifiedMapWithHashingStrategy` now provide single-lookup overrides for `computeIfAbsent`, `computeIfPresent`, and `compute`, reducing hash lookups from 2–3 (Java default) to 1. ([#147](https://github.com/eclipse-collections/eclipse-collections/issues/147), [#500](https://github.com/eclipse-collections/eclipse-collections/issues/500))
+
 ## Documentation Changes
 * Improved Javadoc for 'Sets' factory class ([#782](https://github.com/eclipse-collections/eclipse-collections/issues/782))
 
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/AbstractSynchronizedMapIterable.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/AbstractSynchronizedMapIterable.java
@@ -319,6 +319,51 @@ public abstract class AbstractSynchronizedMapIterable<K, V>
     }
 
     @Override
+    public V computeIfAbsent(K key, java.util.function.Function<? super K, ? extends V> mappingFunction)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().computeIfAbsent(key, mappingFunction);
+        }
+    }
+
+    @Override
+    public V computeIfPresent(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().computeIfPresent(key, remappingFunction);
+        }
+    }
+
+    @Override
+    public V compute(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().compute(key, remappingFunction);
+        }
+    }
+
+    @Override
+    public void replaceAll(BiFunction<? super K, ? super V, ? extends V> function)
+    {
+        synchronized (this.lock)
+        {
+            this.getDelegate().replaceAll(function);
+        }
+    }
+
+    @Override
+    public V putIfAbsent(K key, V value)
+    {
+        synchronized (this.lock)
+        {
+            return this.getDelegate().putIfAbsent(key, value);
+        }
+    }
+
+    @Override
     public <VV> MutableMapIterable<VV, V> groupByUniqueKey(Function<? super V, ? extends VV> function)
     {
         synchronized (this.lock)
--- a/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentMutableHashMap.java
+++ b/eclipse-collections/src/main/java/org/eclipse/collections/impl/map/mutable/ConcurrentMutableHashMap.java
@@ -452,6 +452,24 @@ public final class ConcurrentMutableHashMap<K, V>
     }
 
     @Override
+    public V computeIfAbsent(K key, java.util.function.Function<? super K, ? extends V> mappingFunction)
+    {
+        return this.delegate.computeIfAbsent(key, mappingFunction);
+    }
+
+    @Override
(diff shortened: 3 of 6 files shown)


motlin:
Thanks for putting this together. Most of this I'd like to merge as-is.

One blocker though: the `AbstractMutableMap` fallback overrides cause a thread-safety regression in `ConcurrentMutableHashMap`. That's what the CI failure on `ConcurrentHashMapTestCase.computeIfPresent` is showing:

```
ConcurrentMutableHashMapTest>ConcurrentHashMapTestCase.computeIfPresent:184
expected: <[10, 10, 10, 10, 10, 10, 10, 10, 10, 10]>
but was:  <[10, 10, 10, 10, 10, 10, 10, 10, 9, 10]>
```

The test does 100 increments across 10 keys via `ParallelIterate.forEach`, expecting each bucket to reach 10. The `9` is a lost update.

`ConcurrentMutableHashMap` is a thin wrapper around `java.util.concurrent.ConcurrentMap`:

```java
public final class ConcurrentMutableHashMap<K, V>
        extends AbstractMutableMap<K, V>
        implements ConcurrentMutableMap<K, V>, Serializable
{
    private final ConcurrentMap<K, V> delegate;
    // ...
    @Override
    public boolean replace(K key, V oldValue, V newValue)
    {
        return this.delegate.replace(key, oldValue, newValue);
    }
}
```

It has no `compute*` overrides of its own and relies on inherited defaults.

Before this PR, method resolution for `computeIfPresent` was:

1. `ConcurrentMutableHashMap`: no override
2. `AbstractMutableMap` / `AbstractMutableMapIterable`: no override
3. Interface defaults: both [`Map.computeIfPresent`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Map.html#computeIfPresent(K,java.util.function.BiFunction)) and [`ConcurrentMap.computeIfPresent`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ConcurrentMap.html#computeIfPresent(K,java.util.function.BiFunction)) exist as defaults
4. Java picks the most specific subinterface, so `ConcurrentMap.computeIfPresent` wins. That implementation CAS-loops on [`replace(K, V, V)`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ConcurrentMap.html#replace(K,V,V)), which delegates to the atomic `j.u.c.ConcurrentMap.replace`.

After this PR, `AbstractMutableMap` adds class-level overrides:

```java
@Override
public V computeIfPresent(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction)
{
    Objects.requireNonNull(remappingFunction, "remappingFunction cannot be null");
    V oldValue = this.get(key);
    if (oldValue == null)
    {
        return null;
    }
    V newValue = remappingFunction.apply(key, oldValue);
    if (newValue == null)
    {
        this.remove(key);
    }
    else
    {
        this.put(key, newValue);
    }
    return newValue;
}
```

Class methods shadow interface defaults, so `ConcurrentMap`'s atomic default no longer applies.

A potential fix is to drop the `AbstractMutableMap` overrides. They're redundant with `Map`'s interface defaults anyway.

Happy to push the revert myself if that's easier, or you can.

snowykte0426:
@motlin  Thanks for catching this — you're absolutely right.

The root cause is exactly as you described: class-level overrides take priority over interface defaults regardless of interface specificity, so `AbstractMutableMap.computeIfPresent` shadowed `ConcurrentMap.computeIfPresent` and broke `ConcurrentMutableHashMap`'s atomic CAS-loop behavior. I've dropped all three `compute*` overrides from `AbstractMutableMap` in the follow-up commit (c47ef1d). The `Map` interface defaults that remain in effect for non-concurrent subclasses are identical to the removed code, so there's no behavior change there.

`UnifiedMap` still has its efficient single-lookup overrides and `AbstractSynchronizedMapIterable` retains the synchronized wrappers — those are unaffected.

motlin:
If we're fixing UnifiedMap, we ought to also fix UnifiedMapWithHashingStrategy.

snowykte0426:
@motlin Thanks for the review!

Added `computeIfAbsent`, `computeIfPresent`, and `compute` overrides to `UnifiedMapWithHashingStrategy` ([c7a7c1a](https://github.com/snowykte0426/eclipse-collections/commit/c7a7c1a0)) following the same chained-helper pattern as `UnifiedMap`. While at it, I also added explicit `this.delegate` delegation for the three compute methods in `ConcurrentMutableHashMap` ([3cd1ac7](https://github.com/snowykte0426/eclipse-collections/commit/3cd1ac79)) to keep it consistent with how `putIfAbsent` and `merge` are already handled there.

I also noticed there were no tests verifying that the `HashingStrategy` is actually respected during `compute*` operations — a strategy-equal key should find the existing entry rather than be treated as absent — so added three dedicated tests to `HashingStrategyMapTestCase` to cover that ([3d22c5b](https://github.com/snowykte0426/eclipse-collections/commit/3d22c5b6)).

Regarding the `Unit Test Early Access` failure ([job log](https://github.com/eclipse-collections/eclipse-collections/actions/runs/26234237579/job/77521751752)): the error is a compilation failure in `RichIterableTestCase.java`, unrelated to the files changed in this PR:

```
ERROR .../RichIterableTestCase.java:[120] error: code too large
```

I checked and the same job was already failing on upstream master before this PR was opened, so I don't think it's caused by our changes — but wanted to flag it in case it's something you're tracking.

motlin:

PR #1951, opened 2026-05-25
Introduce API-level PrimitiveTuples factory with SPI provider

vijayguhan10:
## Problem

Primitive tuple interfaces already live in the API module:

```text
org.eclipse.collections.api.tuple.primitive
```

However, the tuple factory currently resides in the implementation module:

```text
org.eclipse.collections.impl.tuple.primitive.PrimitiveTuples
```

As a result, users creating primitive tuples must import implementation classes:

```java
import org.eclipse.collections.api.tuple.primitive.IntIntPair;
import org.eclipse.collections.impl.tuple.primitive.PrimitiveTuples;

IntIntPair pair = PrimitiveTuples.pair(1, 2);
```

Although consumers interact with API interfaces, tuple creation itself requires importing from the implementation module.

---

## Before

Tuple creation was performed through the implementation factory:

```java
import org.eclipse.collections.impl.tuple.primitive.PrimitiveTuples;

IntIntPair pair = PrimitiveTuples.pair(1, 2);
```

Architecture:

```text
User Code
    ↓
Implementation PrimitiveTuples
    ↓
Generated Pair Implementations
```

---

## After

Tuple creation can be performed entirely through the API module:

```java
import org.eclipse.collections.api.factory.primitive.PrimitiveTuples;
import org.eclipse.collections.api.tuple.primitive.IntIntPair;

IntIntPair pair = PrimitiveTuples.pair(1, 2);
```

Architecture:

```text
User Code
    ↓
API PrimitiveTuples
    ↓
PrimitiveTupleFactoryProvider
    ↓
ServiceLoader
    ↓
PrimitiveTupleFactoryProviderImpl
    ↓
Implementation PrimitiveTuples
    ↓
Generated Pair Implementations
```

Consumers now interact entirely through API abstractions and no longer need to import implementation-side tuple factories.

The underlying tuple implementations and tuple creation logic remain unchanged.

---

## Backward Compatibility

This change adds a new API-level entry point without changing the existing implementation factory.

Existing code such as:

```java
import org.eclipse.collections.impl.tuple.primitive.PrimitiveTuples;

IntIntPair pair = PrimitiveTuples.pair(1, 2);
```

continues to work as before.

No existing APIs have been removed or modified, so current users are unaffected. This change simply provides an additional way to create primitive tuples through the API module while preserving the existing usage pattern.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,16 @@ jmh-scala-tests/.cache-main
 scala-unit-tests/.cache-tests
 test-coverage/**
 
+# VS Code
+.vscode/
+
+# Scala (Metals / Bloop)
+.metals/
+.bloop/
+.scala-build/
+.bsp/
+metals.sbt
+
 # https://raw.githubusercontent.com/github/gitignore/master/Global/JetBrains.gitignore
 
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and WebStorm
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,4 @@
-Contributor License
--------------------
+## Contributor License
 
 If this is your first time contributing to an Eclipse Foundation project, you'll need to sign the [Eclipse Contributor Agreement][ECA].
 
@@ -8,13 +7,11 @@ If this is your first time contributing to an Eclipse Foundation project, you'll
 - Read and [sign the ECA](https://dev.eclipse.org/site_login/myaccount.php#open_tab_cla)
 - Use the exact same email address for your Eclipse account and your commit author.
 
-Issues
-------
+## Issues
 
 Search the [issue tracker](https://github.com/eclipse-collections/eclipse-collections/issues) for a relevant issue or create a new one.
 
-Making changes
---------------
+## Making changes
 
 Fork the repository in GitHub and make changes in your fork.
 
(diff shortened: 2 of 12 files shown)


vijayguhan10:
Hi @donraab ,

I hope you're doing well.

I wanted to follow up on my PR, as it has been awaiting review for about 22 days now. I was wondering if there are any issues with the workflow, CI checks, or any additional changes required from my side.

If there is anything that needs to be updated or improved, please let me know. I'd be happy to make the necessary changes.

I'm excited to contribute to the Eclipse Collections project and would love to continue contributing more to the Java and Eclipse communities.

Thank you for your time and feedback.

Best regards,
Vijay

donraab:
Hi @vijayguhan10, thanks for submitting. I am currently on vacation, and will take a look at the full PR when I return, if one of the other committers doesn't get to it before me. Before I review the code I would ask if this PR was AI assisted, and if so, then please follow the guidance in the AI policy markdown file. When I read the problem statement there was a point that looked incorrect, and which I verified was incorrect by hand. The following code is not possible as a usage, as the constructor is package private.

`IntIntPair pair = new IntIntPairImpl(1, 2);`

Screenshot 2026-06-19 at 3 49 31 PM

Thanks!

vijayguhan10:
Hi @donraab,

Thanks for taking a look, and I hope you're enjoying your vacation.

Yes, I used AI assistance while working on this change, and I'll make sure to follow the guidance in the AI policy markdown file.

Also, thanks for catching the issue in the problem statement. I had mistakenly interpreted the `primitivePrimitivePairImpl.stg` template and incorrectly assumed that the constructor was public after seeing the `public class` declaration. You're right that direct instantiation is not part of the intended usage.

I've updated the problem statement accordingly.

The goal of this PR is simply to introduce an API-level `PrimitiveTuples` factory backed by an SPI provider. This allows consumers to create primitive tuples through the API module instead of importing the implementation-side factory directly, while preserving the existing tuple implementations and behavior internally.

Thanks again for pointing that out and for the helpful feedback!

PR #1954, opened 2026-05-29
Add OrderedHashMap, a memory-efficient insertion-ordered map.

motlin:


motlin:
@donraab this one is big in lines of code but is kind of conceptually simple. Do you want to take a look?

donraab:
@motlin Thanks, I will take a look. See if you can force push an update. The OpenJDK team has claimed the bug we found has been fixed, but not sure if they have released a new JDK 27 EA binary with the fix.

motlin:
@donraab all green

donraab:
@motlin I'm beginning my review of this PR now. I did not do any code reviews while I was on vacation.

donraab:
@motlin Since `OrderedHashMap` is meant to be a memory-efficient insertion-ordered map, I checked out your branch locally and wrote the following test which verifies it is more memory-efficient than `LinkedHashMap`. It would be good to have a similar test in the PR. But there is one problem. All of your tests are in `unit-tests-java8` module. Why were they put here instead of `unit-tests`? It would be good if we collapsed these two modules as a priority into a single module. We run the risk of losing the historical knowledge of why these two modules existed to begin with, and confusing contributors to the library.

```
@Test
public void orderedHashMapLinkedHashMap()
{
    MutableList<Integer> list = Interval.oneTo(100).toList();
    list.shuffleThis();
    OrderedHashMap<Integer, Integer> ordered = new OrderedHashMap<>();
    list.forEachWith(ordered::put, 0);
    LinkedHashMap<Integer, Integer> linked = new LinkedHashMap<>();
    list.forEachWith(linked::put, 0);
    assertEquals(ordered, linked);
    JolMemoryTestUtil.assertGraphMemoryEquals(3728L, 0L, ordered);
    JolMemoryTestUtil.assertGraphMemoryEquals(6720L, 0L, linked);
}
```

When I run `OrderedHashMapTest` locally in IntelliJ with coverage, it doesn't display the coverage of `OrderedHashMap`. The same works when I run `UnifiedMapTest` and look at the coverage of `UnifiedMap`. I'm trying now to figure out why the coverage is not working as expected, which I suspect has to do something with module setup.

donraab:
@motlin I was able to configure the `unit-tests-java8` module so coverage is reported in IntelliJ. 

Screenshot 2026-07-05 at 12 02 42 AM

There are no tests that cover the `KeyIterator.remove()` method. The anonymous inner implementations of `KeySet` and `EntrySet` and their Iterators need to be replaced with named classes as we have them in `UnifiedSet`.

donraab (changes requested):
@motlin I've made some comments on the PR about the location of unit tests, code coverage, and the anonymous inner keySet and entrySet which should be replaced by named protected inner classes as we have them in `UnifiedSet`. Test coverage is good, but should be as close to 100% as possible for a new data structure implementation. I also saw in the coverage that keySet `next()` did not have a failure case for not `hasNext()` being false. The clear() method on iterator() was also not tested.

motlin:
@donraab thank you very much for the thorough review. I'm addressing the missing coverage now. Not sure if you remember the history of unit-tests vs unit-tests-java8 but it's something I haven't written about in public before so I'm trying to write up a thorough description now.

donraab:
@motlin I remember the history vaguely, when we were still on Java 7 but wanted to start using Java 8 for tests in default methods in interfaces. It needs a better future. We have unit tests split across both modules, and when I run a test suite locally, it's usually the plain unit tests one. I had never run coverage before on unit-tests-java8 so it wasn't showing up when I tried.

donraab:
@motlin I also saw you went with open addressing but with a single array with alternating key/value pairs like `UnifiedMap`. You did't use any chaining though like `UnifiedMap`, and this made me wonder how this impacts the potential size of the Map. I was originally expecting to see separate key and value arrays like the primitive maps. I wanted to ask you what thought process went into the decision?

motlin:
> @motlin I also saw you went with open addressing but with a single array with alternating key/value pairs like `UnifiedMap`. You did't use any chaining though like `UnifiedMap`, and this made me wonder how this impacts the potential size of the Map. I was originally expecting to see separate key and value arrays like the primitive maps. I wanted to ask you what thought process went into the decision?

@donraab It sounds like you have another design in mind, and you're asking how I weighed the pros/cons between this design and the other. I had no second design in mind and I'm struggling to picture the design you're hinting at.

I wanted to build something that behaves like LinkedHashMap but with our fuller API, and ideally with no entry objects wasting space. I came up with the design you see here, which is similar to our object-primitive maps. If I tried to accomplish the same goals with a backing structure like UnifiedMap's, how would that work? I would need triple (not double) sized arrays, and that third slot would have to be some sort of next pointer. What would it be? Not an index, we don't have a single array to index into. I guess it would be the key of the next entry, but then this would require a hashtable lookup for basic iteration. And to support removal from the middle, I think we'd actually need a previous pointer as well. Is this what you're getting at or do you have another design in mind?

donraab:
@motlin Never mind. For some reason, I forgot we implemented same type primitive maps the same way we implemented UnifiedMap with a single array. So IntIntHashMap has a single array, and IntLongHashMap has two arrays.

> 
> @donraab It sounds like you have another design in mind, and you're asking how I weighed the pros/cons between this design and the other. I had no second design in mind and I'm struggling to picture the design you're hinting at.
> 
> I wanted to build something that behaves like LinkedHashMap but with our fuller API, and ideally with no entry objects wasting space. I came up with the design you see here, which is similar to our object-primitive maps. If I tried to accomplish the same goals with a backing structure like UnifiedMap's, how would that work? I would need triple (not double) sized arrays, and that third slot would have to be some sort of next pointer. What would it be? Not an index, we don't have a single array to index into. I guess it would be the key of the next entry, but then this would require a hashtable lookup for basic iteration. And to support removal from the middle, I think we'd actually need a previous pointer as well. Is this what you're getting at or do you have another design in mind?

motlin:
> @motlin I remember the history vaguely, when we were still on Java 7 but wanted to start using Java 8 for tests in default methods in interfaces. It needs a better future. We have unit tests split across both modules, and when I run a test suite locally, it's usually the plain unit tests one. I had never run coverage before on unit-tests-java8 so it wasn't showing up when I tried.

Thanks Don. I agree the module split is confusing, and the name `unit-tests-java8` is not helpful anymore.

The reason the tests are in `unit-tests-java8` is that this suite uses Java 8 default methods in test interfaces, but not for the usual backward compatibility reasons. We use default methods for multiple inheritance of test behavior.

That matters because the production interface type uses multiple inheritance of behavior. A `MutableList` is both a `MutableCollection` and a `ListIterable`. For the best coverage of all interface contracts, it's helpful for `MutableListTestCase` to inherit from both `MutableCollectionTestCase` and `ListIterableTestCase`. This structure gives great consistency, and even just now while working on test coverage for `OrderedHashMap`, I found a bug in an unrelated class, and I'll point it out on the PR.

From a high level, `unit-tests` has higher test coverage numbers. `unit-tests-java8` only achieves about 60% test coverage for our core types, and doesn't cover primitive collections at all, so its overall coverage is only around 12%. However, for anything it covers, it tends to be way more thorough.

One problem I had with `unit-tests` is that the tests are so small and so fast, that IntelliJ would spend more time instantiating and filtering all the tests than running them. `unit-tests-java8` tests tend to be chunkier because they gather all assertions possible into one place, delegate to the super interface(s) tests with the same names, and combine assertions for methods that are always declared together. An example of that last point is RichIterableTestCase#RichIterable_anySatisfy_allSatisfy_noneSatisfy, since those three methods are always declared and overridden together.

Some statistics:

| metric | `unit-tests` | `unit-tests-java8` |
|---|---:|---:|
| Java code LOC | 112k checked-in / 610k incl. generated | 18k |
| Test classes run | 2,433 | 117 |
| Test executions | 168,374 | 10,409 |
| Plain Surefire test time | 32.5s | 1.8s |
| Aggregate line coverage | 92.4% | 12.1% |

Breaking down the coverage numbers for `unit-tests-java8`, since the primitive packages dominate:

| scope | line coverage |
|---|---:|
| primitive packages | 3.0% |
| non-primitive packages | 47.5% |
| object lists | 58.6% |
| object sets | 49.5% |
| object maps | 65.6% |
| object bimaps | 61.3% |
| object multimaps | 12.6% |
| object bags | 63.8% |

I've been trying to make `unit-tests-java8` "the" test suite but it's challenging. First, as I expand coverage I find lots of inconsistencies and bugs that need to be fixed. #1937 is a good example of a real bug, but there have been so many minor inconsistencies in null handling, which type of exception is thrown when a method isn't supported, whether methods like getIfAbsentPut throw on the non-mutating path, whether default methods are overridden to be atomic, and more. It can be difficult to make our methods consistent without pulling a long thread, and even when I can do it, it can be challenging to get past code review. An example is #1819. In that example, I have no strong opinion on whether those methods should return null or throw, just that they should all be consistent, but we struggled to decide and land anything. As long as the inconsistency exists, a test like this cannot pass:

```java
    @Test
    default void RichIterable_getFirst_empty_null()
    {
        assertNull(this.newWith().getFirst());
    }
```

Nor can the opposite pass:

```java
    @Override
    @Test
    default void RichIterable_getFirst_empty_null()
    {
        assertThrows(NoSuchElementException.class, () -> this.newWith().getFirst());
    }
```

Of course, the fact that the `unit-tests-java8` exposes inconsistencies is also why I think it's valuable. This test suite drives much of my work.

Another challenges is that I expect it will be hard to delete anything from `unit-tests` since it will be hard to code review. We'd have to get comfortable that the tests are truly obsolete and don't contribute to coverage. Of the 112k lines that are checked in, I would estimate about 30k are fully redundant and could be deleted.

If this context is helpful, I could write it up and share it somewhere more permanent like a README in the relevant modules, or a blog. What do you think?

motlin:
> @motlin I have no other design in mind, just looking to understand the thoughts that went into this design approach.

We start with the linked list of Entry objects in LinkedHashMap. To store the same data without the overhead of the Entry wrappers and the before/after links, we store them in a dense Object[] called `entries`, in insertion order, alternating key/value. forEachKeyValue() is simple, just iterating over this array in order, skipping removed sentinels.

Next we need the actual hash table. The bucketing and probing is identical to our Object-Primitive maps. But here's the subtle thing. Even though the hash table uses the same approach as a our open addressing map with **Object keys**, the array is actually **int[]**. It's the array called `indices` and if the value is non-negative, it's an index into `entries`, the linked list. So we probe for the insertion-order-index, but whenever we need the actual key Object, we run `this.entries[idx << 1]` to get it.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedHashMap.java
```diff
@@ -0,0 +1,1747 @@
...
+                    {
+                        if (!this.lastReturned)
+                        {
+                            throw new IllegalStateException();
+                        }
+                        K key = OrderedHashMap.this.toNonSentinel(OrderedHashMap.this.entries[this.position - 2]);
+                        OrderedHashMap.this.removeKey(key);
+                        this.count--;
+                        this.lastReturned = false;
+                    }
```
motlin: @donraab you pointed out this was uncovered, and in fact it wasn't even implemented.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedHashMap.java
```diff
@@ -0,0 +1,1747 @@
...
+            {
+                OrderedHashMap.this.clear();
+            }
+        });
+    }
+
+    @Override
+    public MutableSet<Map.Entry<K, V>> entrySet()
+    {
+        return SetAdapter.adapt(new AbstractSet<>()
```
motlin: @donraab you pointed out that these should be named inner classes. Is that for serialization?

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedMapAdapter.java
```diff
@@ -227,7 +227,34 @@ public V removeKey(K key)
...
+                }
+
+                builder.append(key == OrderedMapAdapter.this ? "(this Map)" : key);
+                builder.append('=');
+                builder.append(value == OrderedMapAdapter.this ? "(this Map)" : value);
+            }
+        });
+
+        builder.append('}');
+        return builder.toString();
```
motlin: This is what I meant by finding unrelated bugs while working on test coverage. A new test I added for OrderedHashMap was causing a stack overflow here.

serialization-tests/src/test/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedHashMapSerializationTest.java
```diff
@@ -0,0 +1,39 @@
...
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.eclipse.collections.impl.map.ordered.mutable;
+
+import org.eclipse.collections.api.map.MutableOrderedMap;
+import org.eclipse.collections.impl.test.Verify;
+import org.junit.jupiter.api.Test;
+
+public class OrderedHashMapSerializationTest
```
motlin, line 17: The views are not serializable yet, but I could support that if desired. HashMap's views are not serializable.

unit-tests-java8/src/test/java/org/eclipse/collections/test/IterableTestCase.java
```diff
@@ -551,6 +551,9 @@ default void Iterable_remove()
             return;
         }
 
+        Iterator<Integer> freshIterator = this.newWith(3, 2, 1).iterator();
+        assertThrows(IllegalStateException.class, freshIterator::remove);
```
motlin, line 560: Here's the test coverage for calling remove() without calling next()

unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java
```diff
@@ -309,6 +309,43 @@ default void RichIterable_notEmpty()
         assertFalse(this.newWith().notEmpty());
     }
 
+    @Test
+    default void RichIterable_zip()
```
motlin, line 318: While looking for missing test coverage, I saw that zip() was somehow missing from RichIterableTestCase. This class has 90 concrete subclasses which are tests. By contrast, AbstractRichIterableTestCase#zip has 44 subclasses/tests.

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/MapTestCase.java
```diff
@@ -52,12 +55,79 @@ default boolean supportsNullValues()
...
+        if (this.allowsPut() && this.supportsNonComparableKeys())
+        {
+            Map<Object, Object> selfKey = this.newWithKeysValues();
+            selfKey.put(selfKey, "value");
+            assertEquals("{(this Map)=value}", selfKey.toString());
+
+            Map<Object, Object> selfValue = this.newWithKeysValues();
+            selfValue.put("key", selfValue);
+            assertEquals("{key=(this Map)}", selfValue.toString());
+        }
```
motlin: Here's the new test coverage that caught the unrelated bug.

mohrezaei:
> I also saw you went with open addressing but with a single array 

Single array, open addressed hashmap is text book. Used all over, including our primitive collections, Trove (if you still remember that), etc. They do ok, unless someone really tries to cause hash collisions. The latter is hard to do with primitives, trivial to do with objects.

Of course, this is not just an open addressed map:

>  we store them in a dense Object[] called entries, in insertion order

The map index array points into a (memory) heap where insertion order must be maintained. The trouble with this kind of heap is deletions. `reclaimTrailingTombstones` tries to handle the most trivial of cases, but IMHO it's not sufficient. It would be better to keep a separate remove count for the heap and compact it. Here is a simple test that grows the heap indefinitely, while the index array stays constant:

```java
map.put("A", 1);
map.put("B", 2);
for(int i=0;i<10000000;i++) {
    map.remove("A");
    map.put("A", 1);
    map.remove("B");
    map.put("B", 2);
}
```
(I haven't tested this, just doing this in my head. the above keys should be chosen so they don't collide for the test to work, I think.)

motlin:
> > I also saw you went with open addressing but with a single array
> 
> Single array, open addressed hashmap is text book. Used all over, including our primitive collections, Trove (if you still remember that), etc. They do ok, unless someone really tries to cause hash collisions. The latter is hard to do with primitives, trivial to do with objects.
> 
> Of course, this is not just an open addressed map:
> 
> > we store them in a dense Object[] called entries, in insertion order
> 
> The map index array points into a (memory) heap where insertion order must be maintained. The trouble with this kind of heap is deletions. `reclaimTrailingTombstones` tries to handle the most trivial of cases, but IMHO it's not sufficient. It would be better to keep a separate remove count for the heap and compact it. Here is a simple test that grows the heap indefinitely, while the index array stays constant:
> 
> ```java
> map.put("A", 1);
> map.put("B", 2);
> for(int i=0;i<10000000;i++) {
>     map.remove("A");
>     map.put("A", 1);
>     map.remove("B");
>     map.put("B", 2);
> }
> ```
> 
> (I haven't tested this, just doing this in my head. the above keys should be chosen so they don't collide for the test to work, I think.)

`addKeyValueAtSlot` checks:

```java
if ((this.entryCount << 1) >= this.entries.length)
{
    this.rehashAndGrow();
}
```

and `rehash()` compacts live entries.

mohrezaei:
> addKeyValueAtSlot checks:

that's good, but the dual accounting is a bit problematic, I think. Try this test:

```java
// fill the map, just before a resize (-1, because of the equality check. index array should be size 2^21.
// entries heap should be full -1
for(int i=0;i< (1 << 20) - 1; i++) {
    map.put(i, i);
}

// now remove and re-add just one entry. This wouldn't cause the index array to resize, but will cause the heap to resize
for(int i=0;i<10000000;i++) {
    map.remove(5000);
    map.put(5000, 5000);
}
```
something like that.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedHashMap.java
```diff
@@ -0,0 +1,1759 @@
...
+    };
+
+    // Sparse open-addressed hash table. Each slot holds an orderedKeyValues pair index, EMPTY, or REMOVED.
+    private int[] indicesHashTable;
+    // Counts live and REMOVED slots in indicesHashTable.
+    private int indicesHashTableOccupied;
+    // Dense interleaved key-value pairs in insertion order: [k0, v0, k1, v1, ...].
+    private Object[] orderedKeyValues;
+    // Counts live and interior REMOVED key-value pairs in orderedKeyValues.
+    private int orderedKeyValuesOccupied;
```
motlin, line 188: I renamed the main fields for clarity.

motlin:
I reworked the capacity and resizing math to address @mohrezaei's feedback.

The important change is that rehash capacity is now derived from live entries, not from the current backing array size.

The target capacity also leaves enough usable room after compaction for the A/B remove/re-add churn case to continue at the same backing sizes.

mohrezaei:
> I reworked the capacity and resizing math

Could you please document the sizing/resizing behavior, as it's fairly non-standard. It's also a bit difficult to work out from the code.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedHashMap.java
```diff
@@ -0,0 +1,1783 @@
...
+ *     <li>Removal-heavy workloads rehash at the same or a smaller capacity. Steady remove/re-add
+ *     churn periodically compacts tombstones away instead of growing the arrays unboundedly.</li>
+ * </ul>
+ * For example, a map with the default index table capacity of 8 has a usable capacity of 5. The first
+ * five insertions fit without rehashing. The sixth insertion triggers a rehash before it is stored:
+ * with 5 live entries, the new index table capacity is 16 (the smallest power of two at least
+ * {@code 5 * 3 = 15}), and the usable capacity becomes 10.
+ *
+ * @since 14.0
+ */
```
motlin, line 128: > > I reworked the capacity and resizing math
> 
> Could you please document the sizing/resizing behavior, as it's fairly non-standard. It's also a bit difficult to work out from the code.

I added Javadoc here @mohrezaei

unit-tests-java8/src/test/java/org/eclipse/collections/test/map/mutable/ordered/OrderedHashMapTest.java
```diff
@@ -0,0 +1,692 @@
...
+        assertEquals(10, getOrderedKeyValuesLength(map));
+        assertEquals(5, getIndicesHashTableOccupied(map));
+        assertEquals(5, getOrderedKeyValuesOccupied(map));
+
+        map.put(5, 5);
+
+        assertEquals(16, getIndicesHashTableLength(map));
+        assertEquals(20, getOrderedKeyValuesLength(map));
+        assertEquals(6, getIndicesHashTableOccupied(map));
+        assertEquals(6, getOrderedKeyValuesOccupied(map));
```
motlin, line 450: These assertions prove out the specific example written in the javadoc about how growth works. The numbers 10 and 20 refer to the double length entries array, meaning that they represent usable capacity of 5 and 10 entries respectively.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedHashMap.java
donraab: For Serialization and for consistency with all other implementations.

eclipse-collections/src/main/java/org/eclipse/collections/impl/map/ordered/mutable/OrderedHashMap.java
```diff
@@ -0,0 +1,1783 @@
...
+    }
+
+    @Override
+    public ImmutableOrderedMap<K, V> toImmutable()
+    {
+        return new ImmutableOrderedMapAdapter<>(this);
+    }
+
+    @Override
+    public MutableOrderedMap<K, V> asSynchronized()
```
donraab, line 1696: I checked out the latest and code coverage is improved, but this method is uncovered.

donraab (changes requested):
@motlin Updated locally and re-ran coverage. One missing test for asSynchronized. I couldn't find any JOL tests for the memory footprint comparison of OrderedHashMap and LinkedHashMap to explicitly cover the claim in the first sentence of the JavaDoc - "A memory-efficient, insertion-ordered map using a sparse index table with open addressing and dense orderedKeyValues storage". Testing memory at various sizes would be helpful to have as a comparison in Javadoc and maybe eventually our web site.

mohrezaei (approved):
I tested the resizing behavior and it works as described. It might be worth mentioning that repeated deletes will cause a rehash, even when the same keys are used.

motlin:
Thank you for the approval @mohrezaei. Rather than land it as-is though, I pushed up some changes to address some of @donraab's feedback.

- **Named inner classes:** `KeySetIterator`, `EntrySetIterator`, `WriteThroughEntry`. Entries from `entrySet().iterator()` now serialize; they previously threw `NotSerializableException`.
- **`asSynchronized()`:** covered by a test asserting the `UnsupportedOperationException`, same as `clone()`. Real implementation to follow separately.
- **JOL test:** `OrderedHashMapMemoryTest` in `unit-tests` compares overhead against `LinkedHashMap` for sizes 0 to 10,000. I originally saw that the empty case wasn't great, which leads to the last optimization.
- **Empty map:** optimized to allocate lazily; an empty map is 32 bytes vs 64 for `LinkedHashMap`.

One more idea which I'm going to put in a separate PR:

- **Narrow index table**: `byte[]`/`short[]`/`int[]` chosen by capacity, making small maps *much* smaller than LinkedHashMap.

mohrezaei:
> Rather than land it as-is

Sorry, I didn't mean you should merge without addressing other comments. I just meant my comments were addressed appropriately.

> JOL test

What was the result? Should we include that in the docs somewhere?

> Narrow index table

This makes sense, but you'd probably have to wrap the `int[]` field so the code is not horrendous. That'll have an extra memory footprint for non-empty (worst case: single entry) instance.

unit-tests/src/test/java/org/eclipse/collections/impl/JolMemoryTestUtil.java
```diff
@@ -51,4 +51,11 @@ public static void assertGraphMemoryEquals(long expected, long expectedCOH, Obje
         long memoryRequired = IS_COH_ENABLED ? expectedCOH : expected;
         assertEquals(memoryRequired, GraphLayout.parseInstance(instance).totalSize());
     }
+
+    public static void assertGraphMemoryExcludingEquals(long expected, long expectedCOH, Object instance, Object... excluded)
+    {
+        long memoryRequired = IS_COH_ENABLED ? expectedCOH : expected;
+        long excludedSize = excluded.length == 0 ? 0L : GraphLayout.parseInstance(excluded).totalSize();
```
donraab, line 58: Is excluding the var arg array correct here?

donraab:
@motlin Looks like there is a failure with the JOL memory test. I left a comment in the new method you added to the util class. I wonder if excluding the memory of the var arg array is related.

donraab (approved):
@motlin It looks like my suggested changes have been addressed. The failing memory test needs to be fixed before merging, and thank you for adding it! If the build passes and all of Moh's feedback is addressed I think this should be ready to merge.

donraab:
@motlin This should be unrelated to the failure, but COH is the default setting now in JDK 27.

PR #1956, opened 2026-06-15
Category annotations

aaccioly:
PoC using annotations to categorise `RichIterable` methods (under @donraab's guidance).

## Advantages complementary to foldable regions:

1. Improved Javadocs, displaying the category for each method: image
2. Similar to foldable regions, it enables users to navigate categories using the "Find Usages" feature of IDEs such as IntelliJ IDEA, but now at a global level.
3. Global "Category Usage" index in `package-info`, linking to automatically generated usage indexes per category across packages, classes, and interfaces. See it in action at:
   https://aaccioly-open-source.github.io/eclipse-collections/javadocs/org/eclipse/collections/api/annotation/category/package-summary.html
4. Indexes are now sorted and can be generated automatically by introspecting class annotations.

## Potential downsides:

* Additional maintenance overhead when adding new methods to the API:
  * Determine the appropriate category.
  * Update the index.
  * Add the method in the correct position within the corresponding folding region.

## Disclaimer:

This is an initial PR to evaluate the approach and get some feedback from maintainers.

@donraab and I have already classified several other interfaces behind the scenes while working on an AI skill  and a series of Open Rewrite rules to automatically classify methods. If maintainers decide to support this approach, the plan is to submit a series of small PRs, each categorising a limited number of interfaces/classes. This should make reviews more manageable and allow time for discussion around the proposed classifications.

In parallel, we intend to continue improving the AI classifier skill, which is planned to be released as a standalone open source project. This approach respects the Eclipse Collections AI policy by keeping AI-related assets outside the repository.

For this PR, I used Claude Opus to classify methods and validate them against the previous manual classifications. However, I did not use AI to generate any of the code itself.

This may change in future PRs, where the custom skill and Claude Opus may be used more explicitly to assist with modifications to API classes and interfaces, and as such, I'll add git commit trailers as required by Eclipse Collection AI's policy.

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
@@ -27,6 +27,15 @@ import java.util.stream.Collector;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
+import org.eclipse.collections.api.annotation.category.Aggregating;
+import org.eclipse.collections.api.annotation.category.Converting;
+import org.eclipse.collections.api.annotation.category.Counting;
+import org.eclipse.collections.api.annotation.category.Filtering;
+import org.eclipse.collections.api.annotation.category.Finding;
+import org.eclipse.collections.api.annotation.category.Grouping;
+import org.eclipse.collections.api.annotation.category.Iterating;
+import org.eclipse.collections.api.annotation.category.Testing;
+import org.eclipse.collections.api.annotation.category.Transforming;
 import org.eclipse.collections.api.bag.Bag;
 import org.eclipse.collections.api.bag.ImmutableBag;
 import org.eclipse.collections.api.bag.MutableBag;
@@ -110,8 +119,10 @@ import org.eclipse.collections.api.tuple.Pair;
  * </ul></li>
  * <li><b>Counting 🔢</b>
  * <ul><li>
- * {@link #count(Predicate)} , {@link #countBy(Function)} , {@link #countByEach(Function)},
- * {@link #countByWith(Function2, Object)}, {@link #countWith(Predicate2, Object)} , {@link #size()}
+ * {@link #count(Predicate)} , {@link #countBy(Function)} , {@link #countBy(Function, MutableBagIterable)} ,
+ * {@link #countByEach(Function)} , {@link #countByEach(Function, MutableBagIterable)} ,
+ * {@link #countByWith(Function2, Object)} , {@link #countByWith(Function2, Object, MutableBagIterable)} ,
+ * {@link #countWith(Predicate2, Object)} , {@link #size()}
  * </li></ul>
  * <li><b>Testing 🧪</b>
  * <ul><li>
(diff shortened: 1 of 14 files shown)


donraab:
@aaccioly Was RichIterable.java in the wrong directory? There is no diff, it looks like an added file.

aaccioly:
@donraab , apologies, PBKAC. Should be good now, can you run the pending workflows?

aaccioly:
@donraab, rebased and squashed into a single commit as requested.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/annotation/category/Mutating.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Marks a method as belonging to the <b>Mutating</b> category.
+ * Mutating methods modify the state or contents of a collection.
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+@Category(value = "Mutating", icon = "🧬")
+public @interface Mutating
```
Desislav-Petrov, line 27: What's the list of categories based on?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
```diff
@@ -192,2381 +222,2559 @@
 public interface RichIterable<T>
         extends InternalIterable<T>
 {
-    //region [Category: Iterating] 🔄
```
motlin, line 195: Could we keep the order we already had? So, iterating up top?

eclipse-collections-api/src/main/java/org/eclipse/collections/api/annotation/category/Mutating.java
aaccioly: Morning. They come from custom code folding regions in `RichIterable` + Eclipse Collections Categorically - Appendix A (if I'm not mistaken, from the tables for `MutableList` and `LazyIterable`).

This isn't meant to be set in stone though. New categories will likely emerge when classifying other classes, and maintainers may wish to further break down or combine some of these categories. This first PR is more about testing the waters than anything else.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/RichIterable.java
aaccioly: Can be done if the maintainers wish. The reasoning behind moving the methods around was to automate index creation and classification into custom folding regions. I wrote a small utility that takes a class, introspects its methods to discover categories, and then outputs a new version with the index and custom folding regions ordered lexicographically by category + method name. This is meant to evolve into OpenRewrite recipes that can easily be used by maintainers in the future, and potentially even be enforced in the pipeline.

Having said that, I have no problem keeping the existing order. We have several options for `RichIterable`: 

1. Revert to the previous custom region and method order, but keep the index sorted lexicographically.
2. Revert to the previous custom region and method order, and make the index match that order.
3. Keep everything sorted lexicographically as in this PR. 
4. Come up with a new deterministic order (e.g. each category would have an order field, where `Iterable` is `0`; sort categories by that order, then sort the methods lexicographically within each custom folding region).

Let me know your thoughts and what would be your preference.

donraab: @aaccioly I was thinking the same as @motlin. At least for this PR, it will make it much easier to review `RichIterable` since the diff is not helpful at all as it is. I think option 2 to be as close to the original as possible. This will make it easier for us to get your `Category` annotations approved and merged, and then we can figure out at a later date if we want to change the order of `RichIterable`. I'm ok with the order, as that is the order I put it in by hand. You can then move to `MutableList`, `MutableSet`, etc. and we can see if the diff gets easier once the methods there are grouped. They will need to be moved anyway because the methods are not already grouped like RichIterable is today in Custom regions.

aaccioly: @motlin, @donraab, done. The methods are back in their original order. The index ended up being a mix of strategies 1 and 2.

I kept the categories in their original order, but the links were already mostly (but not always) in lexicographical order . Since the output of my tooling also came out in lexicographical order, it was easier (and resulted in a smaller diff) to keep it that way.

So, in this PR, `RichIterable`'s index is treated as follows:

1. Categories where the `RichIterable` index was already complete were left unchanged (i.e. `Iterating 🔄` and `Testing 🧪`) and therefore kept in their original, "mostly lexicographical, with some exceptions to group related methods" order.
2. All other categories were sorted in lexicographical order.
3. Unlike in the previous version, I didn't remove existing non-local methods from the index, e.g. links to `InternalIterable#forEachWith`, `InternalIterable#forEachWithIndex`, `Iterable#iterable`, and `Iterable#spliterator`.

Let me know what you think. If you're both happy with this version, on my side, feel free to merge it.

donraab: Thanks @aaccioly! I'm hoping to have a look at this and if all good will approve/merge this in the next week or two.

donraab (approved):
LGTM! Thanks for the contribution @aaccioly ! The changes made it much easier to review.

aaccioly:
Thanks @donraab :). Should I rebase this branch or maintainers  got it from here? Also, you mentioned `MutableSet`, `MutableList`, and I also have `LazyIterable` and `InternalIterable` done in a branch. I assume separate PRs make it easier to review right? Should I open all 4 PRs at once or should I do it one at a time?

donraab:
> Thanks @donraab :). Should I rebase this branch or maintainers got it from here? Also, you mentioned `MutableSet`, `MutableList`, and I also have `LazyIterable` and `InternalIterable` done in a branch. I assume separate PRs make it easier to review right? Should I open all 4 PRs at once or should I do it one at a time?

Yes, please rebase @aaccioly. Agree on the separate PRs as well. I was thinking it would be easier to review again, if you take a similar approach to what you did with `RichIterable`. Add the category annotations first, and you can update the JavaDoc as that is new and won't need much work to diff. The custom code folding region moves might be useful then in a separate set of PRs. What do you think?

aaccioly:
@donraab  rebase done. On my end feel free to merge.
I can do the separate PRs no problem, so 8 PRS them? First 4 with annotations + javadoc index per file, then another 4 with custom folding regions? 
Should I create the first 4 PRs (with annotations + javadoc index) at once or would you prefer to do one PR / Review cycle at a time?

Cheers,

donraab:
> @donraab rebase done. On my end feel free to merge. I can do the separate PRs no problem, so 8 PRS them? First 4 with annotations + javadoc index per file, then another 4 with custom folding regions? Should I create the first 4 PRs (with annotations + javadoc index) at once or would you prefer to do one PR / Review cycle at a time?
> 
> Cheers

Merged. Thank you for your patience on this work. I love that we can find usages of the categories in the IDE as well. 

Let's start with one and see how it goes. I'm hoping the initial JavaDoc and annotation updates can go much faster than this first one.

PR #1960, opened 2026-07-09
Fix Map.toString() for self-referential entries.

motlin:


Desislav-Petrov (approved):
lgtm

motlin:
Thank you for the review @Desislav-Petrov! Merging now.

PR #1964, opened 2026-08-01
Add collectEachOccurrences; fix Bag collect stateful semantics (#420)

### LINKED ISSUE #420: Optimize primitive collect methods for primitive Bags when collecting into a primitive Bag.

arimu1:
Fixes #420

### Problem
Closed PR #1131 optimized bag `collect` by applying the function once per distinct element and using `addOccurrences`. That is incorrect for stateful functions (see #420 discussion).

### Changes
- **`collect` / `collectWith` / `collectIf` / `flatCollect` / primitive `collect*`** on bags now apply the function **once per occurrence**.
- **New** `collectEachOccurrences` / `collectWithEachOccurrences` and primitive `collectEachOccurrences*` methods apply the function **once per distinct element** and multiply occurrences (use `addOccurrences` when the target is a bag). Intended for pure functions.
- **`countBy` / `countByWith`** on bags use `collectEachOccurrences` so pure-key counting stays optimized.
- Primitive bag templates + `BooleanHashBag` updated accordingly.
- Tests for object and primitive bags (call counts + bag targets).
- `RELEASE_NOTE_DRAFT.md` updated.

### Avoids #1131 mistakes
Does **not** put the once-per-distinct optimization on `collect` itself.

### Tested
- `HashBagTest`, `SynchronizedBagTest`, `MultiReaderHashBagTest`, `TreeBagTest`, `ImmutableHashBagTest`, `ImmutableArrayBagTest`, `IntHashBagCollectEachOccurrencesTest` (JDK 21)
- checkstyle clean on `eclipse-collections-api` and `eclipse-collections`


motlin:
If I ignore the names we already have, and think of new names for the behavior that iterates over groups and over each item separately, I would expect `collectWithOccurrences` to have the optimized behavior (delegating to `forEachWithOccurrences`) and `collectEach` to have the non-optimized behavior (delegating to `forEach`). And I'm looking now and I see `collectWithOccurrences` exists already and does have the optimized behavior. The potential problem with it is that it takes the 2-arg lambda and it may sometimes be convenient to take a 1-arg lambda and not have to ignore the second parameter.

For backwards compatibility, I would prefer not to change the behavior of `Bag.collect()` which has existed in this form for years.

Putting it together, my suggestion would be to add a new method, `Bag.collectEach()` which has the form that delegates to `forEach`.

```java
public <V, R extends Collection<V>> R collectEach(Function<? super T, ? extends V> function, R target)
{
    this.forEach(each -> target.add(function.valueOf(each)));
    return target;
}
```

@donraab @mohrezaei WDYT?

PR #1968, opened 2026-09-06
Add category annotations to MutableList

aaccioly:
Next PR of the method classification effort (previous PR #1956).
As agreed with @donraab , delaying custom folding regions to a separate PR.

Assisted-by: Copilot: Claude Sonnet 4.6 (Classification) MAI-Code-1.1-Flash (Coding)

### DIFF
--- a/eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
+++ b/eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
@@ -15,6 +15,13 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Random;
 
+import org.eclipse.collections.api.annotation.category.Converting;
+import org.eclipse.collections.api.annotation.category.Filtering;
+import org.eclipse.collections.api.annotation.category.Finding;
+import org.eclipse.collections.api.annotation.category.Grouping;
+import org.eclipse.collections.api.annotation.category.Iterating;
+import org.eclipse.collections.api.annotation.category.Mutating;
+import org.eclipse.collections.api.annotation.category.Transforming;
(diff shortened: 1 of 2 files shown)


eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
```diff
@@ -296,19 +388,22 @@ default <P, V> MutableList<V> flatCollectWith(Function2<? super T, ? super P, ?
...
     @Override
+    @Filtering
     <V> MutableList<T> distinctBy(Function<? super T, ? extends V> function);
 
     /**
      * Sorts the internal data structure of this list and returns the list itself as a convenience.
      *
      * @since 10.0 - Added default implementation.
      */
+    @Mutating
```
donraab, line 406: @aaccioly I assume we can apply multiple category annotations to a method. I'm wondering whether having a Mutating and Sorting category would make sense. I won't hold up this PR because of this, but just curious.

donraab (approved):
This LGTM. Thanks for the contribution @aaccioly ! I left one comment that won't hold up this PR, but something for us to think about. Multiple categories for a single method actually make sense in certain cases.

eclipse-collections-api/src/main/java/org/eclipse/collections/api/list/MutableList.java
aaccioly: Yes we can. It works at the Javadoc level too, i.e. both `@Documented` tags appear in the method Javadoc, and `--use` adds the method to the global indexes of both categories.

We just need to agree on what we are going to do about code folding regions (e.g. use the first annotation as the code folding region) and the class-specific indexes (e.g. should the same method link appear twice, or just under the "main" category?).

